PR #83, opened 2020-07-29
docker: add protoc to docker image

jamaljsr:
This PR fixes the docker build which was previously broken by #81. 

`protoc` needs to be installed in the image for the app to compile the proto files during the build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN apk add --no-cache --update alpine-sdk \
     bash \
     binutils \
     tar \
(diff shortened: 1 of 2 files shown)


truesilver92:
I think I ran into this as a problem trying to build tag v0.1.0-alpha

Did this change get in before the tag?

jamaljsr:
Yes, this was merged after `v0.1.0-alpha` was tagged. This PR is the only change currently in master, outside of docs/CI. So its safe to build your image from master.

PR #103, opened 2020-08-21
add sorting of the channel and history lists

### LINKED ISSUE #86: Manual sorting of the channel/history lists
Users should have the ability to manually sort the channel and history lists. 

LiT currently sorts channels by the status of the channel balances, based on the `Channel Balance Mode` configured in the settings. Occasionally, you may want to sort by the other columns in the list.

The Loop history list is currently sorted by the swap's UpdatedDate descending. This also should allow manually changing the sort order.

jamaljsr:
Closed #86 

This PR adds the ability to manually sort the channel and history lists by clicking on the column headers.

### Steps to Test
1. Go to the Terminal page
1. Click on the column headers (click twice on the same header to change the direction)
1. Confirm the channels are sorting in the correct order
1. Go to the Loop History page
1. Clicking on the column headers
1. Confirm the loops are sorting in the correct order

### DIFF
--- a/app/src/__tests__/components/history/HistoryPage.spec.tsx
+++ b/app/src/__tests__/components/history/HistoryPage.spec.tsx
@@ -8,8 +8,9 @@ import HistoryPage from 'components/history/HistoryPage';
 describe('HistoryPage', () => {
   let store: Store;
 
-  beforeEach(() => {
+  beforeEach(async () => {
     store = createStore();
+    await store.swapStore.fetchSwaps();
   });
 
   const render = () => {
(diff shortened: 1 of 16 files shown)


guggero (approved):
LGTM :100: 
I really like the idea of the "reset sorting" icon on the main page. Any reason not to add it to the loop history page as well?

jamaljsr:
> Any reason not to add it to the loop history page as well?

The only reason I added the "reset sorting" icon is because the channel list is initially sorted by the Balance Mode defined in Settings and this column has no header to click. If the user sorts manually, they need a way to get back to the original Balance Mode sorting. I initially added a header to this column, but I didn't like the way it looked. So I added this icon instead.

The Loop history list is sorted by Update Date by default, which does have a column header. So the user can easily revert to this sorting method by clicking on the header if they choose to do so.

Hope that makes sense.

guggero:
Ah, makes sense, thanks!

PR #104, opened 2020-08-22
Update internal loop daemon to v0.8.0-beta

jamaljsr:
Updating the version of loopd used internally to v0.8.0. I also needed to update the client app as there were a few changes made to the proto file.


guggero (commented):
See comment about `go.mod` in #105. I think the two PRs need to be merged because the version updates need to happen in tandem with lnd, lndclient and faraday.

jamaljsr:
Closing in favor of #108

PR #108, opened 2020-08-24
Update internal LND and loopd versions

jamaljsr:
Updating the versions of the internal daemons:

LND v0.11.0-beta
Loop v0.8.0-beta

I also needed to update the client app as there were a few changes made to the proto file.


guggero (approved):
LGTM :1st_place_medal:

PR #109, opened 2020-08-24
loop: increase max swap routing fee

jamaljsr:
Currently, `MaxSwapRoutingFee` is set to 2% of the `SwapFee` instead of 2% of the swap `Amount`. This can cause Loop Out off-chain payments to fail more often than they should because the max fee is set very low.


guggero (approved):
LGTM, nice catch!

PR #112, opened 2020-09-05
loop: add options for conf target and loop out address

### LINKED ISSUE #106: Allow customization of swap parameters when performing a Loop
Currently in Terminal, the user is only allowed to specify the amount and channels to use for the Loop. There are a few more parameters which can be specified to customize how the swap will be performed by `loopd` and well as the backend server. We should surface the parameters below in the UI to allow the user to modify them as they see fit for their use.

### Loop In
 - `conf_target` - the target number of blocks the on-chain htlc broadcast by the swap client should confirm within

### Loop Out
- `addr` - the optional address that the looped out funds should be sent to, if let blank the funds will go to lnd's wallet
- `conf_target` - the number of blocks from the swap initiation height that the on-chain HTLC should be swept within

jamaljsr:
Closes #106 

Added additional options to the loop processes to allow users to specify the confirmation target for Loop In & Out, as well as the destination address for Loop Out.

image


app/src/i18n/locales/en-US.json
```diff
@@ -51,6 +51,12 @@
   "cmps.loop.swap.SwapConfigStep.heading": "{{type}} Amount",
   "cmps.loop.swap.SwapConfigStep.loopInDesc": "Loop In moves funds onto the Lightning Network from an on-chain wallet to provide outbound capacity to send funds over Lightning. Use the slider to select the amount to Loop In or Loop Out of the Lightning Network",
   "cmps.loop.swap.SwapConfigStep.loopOutDesc": "Loop Out moves funds off of the Lightning Network into an on-chain address while providing inbound capacity to receive funds on Lightning. Use the slider to select the amount to Loop In or Loop Out of the Lightning Network",
+  "cmps.loop.swap.SwapConfigStep.confTargetLabel": "Confirmation Target",
+  "cmps.loop.swap.SwapConfigStep.confTargetHint": "number of blocks (ex: 6)",
+  "cmps.loop.swap.SwapConfigStep.destAddrLabel": "Destination Address",
+  "cmps.loop.swap.SwapConfigStep.destAddrHint": "p2wkh address",
```
guggero, line 57: Loop allows any kind of address, even legacy ones. Maybe we should just write `segwit address` to make it a bit less specific but still incentivize people to use native segwit addresses?

guggero (approved):
Great addition to the UI! I like how you can show/hide the advanced options.

One thing I'm not sure about: Should we add a tooltip to the destination address that if this is left blank, the funds just go to `lnd`'s wallet? Otherwise a user might be confused whether or not this field must be filled (happened to me when I first tried out loop).

app/src/i18n/locales/en-US.json
jamaljsr: Good point! I wasn't sure what to put as a placeholder there. `segwit address` makes more sense so I'll switch to that.

I also agree about the tooltip. I'll add one for conf target as well since it likely isn't obvious for new users, what this does.

Thanks for the feedback.

jamaljsr:
PR updated:
- changed `p2wkh address` to `segwit address`
- added tooltips to the input labels

Issue #85, opened 2020-08-03
Add support for connecting to external nodes


jamaljsr:
Currently, LiT can only use the internal LND, Loop & Faraday nodes. It would be beneficial for some users who already have operational nodes to be able to connect to them. This would support a mix of internal & external nodes based on the user's preference and existing infrastructure.

PR #116, opened 2020-09-09
Allow remote lnd connection

### LINKED ISSUE #120: Display useful information after running litd
We should provide users with more useful information after starting `litd` for the first time.

- Node status
- Network (regtest, testnet, mainnet)
- Node alias
- LND version
- Web interface URL

carbon (14)

guggero:
Preview version only, needs much cleanup and manual testing. Also the whole config validation part is mostly missing.

Depends on:
 - ~https://github.com/lightningnetwork/lnd/pull/4464~
 - ~https://github.com/lightninglabs/loop/pull/281~
 - ~https://github.com/lightninglabs/faraday/pull/77~

Fixes #120.

Fixes the `lnd` part of https://github.com/lightninglabs/lightning-terminal/issues/85. Allowing `loop` and `faraday` to run externally as well should be a much smaller diff and will be done in a follow-up PR.

TODO:
- [x] Validate full configuration for both modes, assert no invalid combinations can be set
- [x] Make sure configuration files can be used for both modes with no or very limited required command line flags
- [x] Fully document all command line options and provide examples for the following possibilities:
  + `integrated` mode with command line flags only
  + `integrated` mode with config file only
  + `remote` mode with command line only
  + `remote` mode with config file only
  + Example `lncli`, `loop` and `frcli` commands for both modes


jamaljsr:
Here's my initial feedback from trying to switch my testnet node to remote mode.

1. For `--lit-dir`, I think we should display the default dir `/root/.lightning-terminal` in the help output, similar to:
   `--lnd.lnddir=  The base directory that contains lnd's data, logs, configuration file, etc. (default: /root/.lnd)`
   Personal nit: I'd prefer the default dir name to be `.lit` over the verbose `.lightning-terminal` :)
1. Is the config file still going to be loaded from `~/.lnd/lnd.config`? I have my LND node running on the same machine which already uses this file for it's configuration. If `litd` is also going to use this same file for its config (ex: httpslisten, lnd-mode), then it is not possible to run them both using config files on the same machine. I will have to use the command line for one of them. This is less than ideal IMO as I prefer to use conf files over the command line, so i can just run `lnd` and `litd` without any arguments.
1. I tried to run `litd --lnd.configfile=~/.lit/lit.conf`, but it seem like this ignores `lnd-mode=remote` set in that file. It is trying to connect to a bitcoin node, which leads me to believe it is attempting to load LND internally.
   ```
   $ litd --lnd.configfile=~/.lit/lit.conf
   loadConfig: either bitcoin.active or litecoin.active must be set to 1 (true)
   $ nano ~/.lit/lit.conf ## added lnd.bitcoin.active=1
   $ litd --lnd.configfile=~/.lit/lit.conf
   loadConfig: either --bitcoin.mainnet, or bitcoin.testnet, bitcoin.simnet, or bitcoin.regtest must be specified
   $ nano ~/.lit/lit.conf ## added lnd.bitcoin.testnet=1
   $ litd --lnd.configfile=~/.lit/lit.conf
   Attempting automatic RPC configuration to btcd
   unable to load RPC credentials for btcd: unable to extract RPC credentials: open /root/.btcd/btcd.conf: no such file or directory, cannot start w/o RPC connection
   $ cat ~/.lit/lit.conf 
   httpslisten=0.0.0.0:443
   uipassword=...
   letsencrypt=1
   letsencrypthost=lit.merchant.com
   lnd-mode=remote
   lit-dir=~/.lit/
 
   lnd.bitcoin.active=1
   lnd.bitcoin.testnet=1
   ```
1. I gave up trying to use the config file and switched to the command line only. Received this error and I'm not sure how to resolve it. I tried adding `--lnd.bitcoin.testnet` to the command but that didn't fix it. Maybe the network this is referring to has nothing to do with mainnet/testnet/regtest.
   ```
   $ litd --httpslisten=0.0.0.0:443 --uipassword=... --letsencrypt --letsencrypthost=lit.merchant.com --lnd-mode=remote
   2020-09-09 15:02:38.132 [ERR] LITD: Could not start subservers: error validating network: unknown network
   panic: runtime error: invalid memory address or nil pointer dereference
   [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xc3d1d9]
 
   goroutine 1 [running]:
   google.golang.org/grpc.(*ClientConn).Close(0x0, 0x0, 0x0)
           /root/go/pkg/mod/google.golang.org/grpc@v1.28.0/clientconn.go:971 +0x59
   github.com/lightninglabs/lightning-terminal.(*rpcProxy).Stop(0xc0000ce410, 0x2, 0x24)
           /root/dev/shushtar/rpc_proxy.go:140 +0x45
   github.com/lightninglabs/lightning-terminal.(*LightningTerminal).shutdown(0xc0001ea720, 0xa4504e, 0xc00000eb60)
           /root/dev/shushtar/terminal.go:435 +0x229
   github.com/lightninglabs/lightning-terminal.(*LightningTerminal).Run.func2(0xc0001ea720)
           /root/dev/shushtar/terminal.go:185 +0x2f
   github.com/lightninglabs/lightning-terminal.(*LightningTerminal).Run(0xc0001ea720, 0x1bdd200, 0xc000187020)
           /root/dev/shushtar/terminal.go:201 +0x754
   main.main()
           /root/dev/shushtar/cmd/litd/main.go:13 +0x103
   ```

I am blocked from further testing due to the last bullet point.

jamaljsr:
Updated to https://github.com/lightninglabs/lightning-terminal/commit/7415e8788529be74d3e9e9710e4bf5ee29e5d7c5 and now I can start it up successfully. 👍 

NotiAlso making a note here about our offline discussion to change the `--lnd-remote.*` flags to be `--remote.lnd.*` instead. Also to place `--lit-*` flags under `--remote.*` as well to group all of the remote node configuration under one section. This should make it easier for new users to discover the flags that need to be set to connect to a remote LND node.

guggero:
Thanks a lot for your suggestions, @jamaljsr! I pushed a new version with the adjusted flags. I also started with splitting up the configuration examples into two separate docs.
I'll continue working on this soon, there's still a lot to do. Added a TODO list to the PR body.

jamaljsr:
Cool. I'm running the latest (https://github.com/lightninglabs/lightning-terminal/commit/c0e6b1f2c5b1606bbbca61a3953f96e4bdb9777c) on my testnet node with no issues. The updated flags are definitely an improvement. 👍 

Let me know when you are ready for more testing & review.

guggero:
Taking this PR out of WIP now. Config validation is now complete and the docs have been completed with example commands.

GiovanniBordiga:
Hey there!
I think I found a small misspelling in the guide [Configuring LiT with integrated lnd node
](https://github.com/lightninglabs/lightning-terminal/blob/grub-v2/doc/config-lnd-integrated.md#upgrade-existing-nodes)
In the `loop` part saying:

> For `loop`:
> 
> - if you use an `loop.conf` file for configurations, copy the parameters into
>   the `lit.conf` and add the `loop.` prefix to each of the configuration
>   parameters. Also remove any section headers or replace them with comments.
> 
>   Before: (in `loop.conf`)
> 
>   ```text
>   [Application Options]
>   loopoutmaxparts=5
>   ```
> 
>   After: (in `lnd.conf`)
> 
>   ```text
>   # Loop
>   loop.loopoutmaxparts=5
>   ```
>

I think the "After" part was meant to refer to `lit.conf` and not `lnd.conf`, i.e.

> For `loop`:
> 
> - if you use an `loop.conf` file for configurations, copy the parameters into
>   the `lit.conf` and add the `loop.` prefix to each of the configuration
>   parameters. Also remove any section headers or replace them with comments.
> 
>   Before: (in `loop.conf`)
> 
>   ```text
>   [Application Options]
>   loopoutmaxparts=5
>   ```
> 
>   After: (in `lit.conf`)
> 
>   ```text
>   # Loop
>   loop.loopoutmaxparts=5
>   ```
> 

Hope this helps! :+1:

guggero:
Thanks, @GiovanniBordiga, I fixed the wrong file name mentioned.

config.go
```diff
@@ -128,7 +386,58 @@ func loadLndConfig(preCfg *Config) (*lnd.Config, error) {
...
+	}
+
+	r.LitTLSCertPath = lncfg.CleanAndExpandPath(r.LitTLSCertPath)
+	r.LitTLSKeyPath = lncfg.CleanAndExpandPath(r.LitTLSKeyPath)
+	r.LitLogDir = lncfg.CleanAndExpandPath(r.LitLogDir)
+
+	// Make sure the parent directories of our certificate files exist. We
+	// don't need to do the same for the log dir as the log rotator will do
+	// just that.
+	if err := makeDirectories(filepath.Base(r.LitTLSCertPath)); err != nil {
```
Jawshua, line 420: This should be `filepath.Dir` instead of `filepath.Base`. This is currently effectively running `mkdir tls.cert` and `mkdir tls.key`, which left me with quite the head scratcher when running `litd` from the same directory as my certs 😄

guggero: Oh, thanks for reporting this :grinning: Fixed it (in the PR, new RC will probably take a few days).

config.go
```diff
@@ -224,28 +537,76 @@ func buildTLSConfigForHttp2(config *Config) (*tls.Config, error) {
 			GetCertificate: manager.GetCertificate,
 		}
 
+	case config.LndMode == ModeRemote:
+		tlsCertPath := config.Remote.LitTLSCertPath
+		tlsKeyPath := config.Remote.LitTLSKeyPath
+
+		err := cert.GenCertPair(
```
Jawshua, line 544: This appears to be generating a self-signed cert every time on startup. I'm trying to use my own cert, but it gets overwritten on launch.

config.go
Jawshua: Excellent. Many thanks!

config.go
guggero: Yes, you are correct. Fixed this as well. Thank you for all the testing!

justinpobrien:
Assume default values for:

[ ] TLS cert path
[ ] Macaroon
[ ] Host
[ ] Assume network is `mainnet`

README.md
```diff
@@ -40,6 +40,7 @@ archives as well.
 
 | LiT              | LND          | Loop        | Faraday      |
 | ---------------- | ------------ | ----------- | ------------ |
+/ **v0.2.0-alpha** | v0.11.1-beta | v0.9.0-beta | v0.2.1-alpha |
```
jamaljsr, line 43: Replace `/` with `|`

doc/config-lnd-integrated.md
```diff
@@ -0,0 +1,333 @@
+# Configuring LiT with integrated lnd node
+
+The "integrated" mode means that `lnd` is started within the same process as
+`litd`. Once the integrated `lnd` has been unlocked, `litd` then goes ahead and
+starts `faraday` and `loop` and connects them to the integrated `lnd` node.
+As a final step the UI server is then also started.
```
jamaljsr, line 6: The UI server is now started prior to `lnd` being unlocked

doc/config-lnd-remote.md
```diff
@@ -0,0 +1,190 @@
+# Configuring LiT with remote lnd node
+
+The "remote" mode means that `lnd` is started as a standalone process, possibly
+on another host, and `litd` connects to it.
+Once the connection to the remote `lnd` node has been established, `litd` then
+goes ahead and starts `faraday` and `loop` and connects them to that `lnd` node
+as well. As a final step the UI server is also started.
```
jamaljsr, line 7: The UI server is now started prior to lnd being unlocked

config.go
```diff
@@ -77,16 +237,94 @@ func loadLndConfig(preCfg *Config) (*lnd.Config, error) {
...
+	}
+
+	// With the validated config obtained, we now know that the root logging
+	// system of lnd is initialized and we can hook up our own loggers now.
+	SetupLoggers(cfg.Lnd.LogWriter)
+
+	// Validate the lightning-terminal config options.
+	if cfg.LetsEncryptDir == "" {
+		cfg.LetsEncryptDir = filepath.Join(
+			cfg.Lnd.LndDir, defaultLetsEncryptDir,
```
jamaljsr, line 254: On my node, the `letsencrypt` dir is being placed in `~/.lnd/` instead of `~/.lit/`. Should this line join `cfg.LitDir` instead of `cfg.Lnd.LndDir`?

jamaljsr (commented):
I did another round of testing on my testnet node using `remote` and `integrated` modes along with LetsEncrypt. Just found one bug and a couple changes to the docs. Otherwise it all works as expected.

config.go
guggero: Yes, definitely. Fixed.

guggero:
Added a log output to address #120. Example from local regtest:
```
$ litd --uipassword=regtest3 --remote.lnd.network=regtest --loop.server.host=[::]:11009

2020-10-07 15:21:17.198 [INF] LITD: Dialing lnd gRPC server at localhost:10009
2020-10-07 15:21:17.333 [INF] LITD: Listening for http_tls on: 127.0.0.1:8443
----------------------------------------------------------
 Lightning Terminal (LiT) by Lightning Labs               
                                                          
 Operating mode      remote                                   
 Node status         locked                                   
 Alias               ???? (node is locked)                                   
 Version             0.11.1-beta.rc3                                   
 Web interface       https://127.0.0.1:8443                                   
----------------------------------------------------------
2020-10-07 15:21:17.356 [INF] LNDC: Creating lnd connection to localhost:10009
```

Unfortunately I wasn't able to suppress the rest of the logs as that would require a code change in the `lnd` codebase. We can address that at a later time I think.

This example also shows that the remote node connection is now the default and requires minimal flags to be set on the command line or in the config file.

pfergi42:
Hello guggero, just curious, why do you include the server host setting in your original post? Is it the same as setting the server hostname to localhost?

--loop.server.host=[::]:11009

I'm not familiar with the "[::]" part.

Thanks!

guggero:
The `[::]` is the IPv6 notation for the `0.0.0.0` equivalent in IPv4. This just means "any address" which normally is used when specifying a listening address where you want to be reachable from the network. So when connecting somewhere, this would be incorrect. It should be [::1] which is the equivalent of `127.0.0.1`. But it seems like it's internally still resolved to the loopback address which is why it works anyway.

jamaljsr:
For `remote` mode, it looks like `litd` requires all of the `lnd` macaroons, not just the `admin.macaroon`. I suggest renaming the `remote.lnd.macaroonpath` flag to be `remote.lnd.macaroondir` and updating the description to make it clear that ll macaroons are required in the specified dir.

guggero:
> For `remote` mode, it looks like `litd` requires all of the `lnd` macaroons, not just the `admin.macaroon`. I suggest renaming the `remote.lnd.macaroonpath` flag to be `remote.lnd.macaroondir` and updating the description which currently states "The full path to the macaroon file to use for the remote connection". It should be more like "The full path to the directory containing all lnd macaroons".

Makes sense, changed the flag as suggested.

go.mod
```diff
@@ -6,11 +6,11 @@ require (
...
 	github.com/improbable-eng/grpc-web v0.12.0
 	github.com/jessevdk/go-flags v1.4.0
-	github.com/lightninglabs/faraday v0.2.0-alpha.0.20200901065614-d72fdaeff888
+	github.com/lightninglabs/faraday v0.2.1-alpha.0.20200918072230-7ec3f67ed2f0
 	github.com/lightninglabs/lndclient v0.11.0-0
-	github.com/lightninglabs/loop v0.8.1-beta.0.20200831113239-d4d9aa102455
-	github.com/lightningnetwork/lnd v0.11.0-beta
-	github.com/lightningnetwork/lnd/cert v1.0.2
+	github.com/lightninglabs/loop v0.9.0-beta.0.20200918101140-4bb59f87bf95
+	github.com/lightningnetwork/lnd v0.11.1-beta.rc3
```
halseth, line 13: update to non-RC version?

rpc_proxy.go
```diff
@@ -0,0 +1,398 @@
...
+	var opts []grpc.DialOption
+	tlsConfig, err := credentials.NewClientTLSFromFile(tlsCertPath, "")
+	if err != nil {
+		return nil, fmt.Errorf("could not read lnd TLS cert %s: %v",
+			tlsCertPath, err)
+	}
+
+	opts = append(
+		opts,
+		grpc.WithCodec(grpcProxy.Codec()), // nolint
```
halseth, line 356: why do we need `WithCodec`?

config.go
```diff
@@ -29,45 +32,160 @@ const (
 
 	uiPasswordMinLength = 8
 
-	defaultLndMacaroon = "admin.macaroon"
+	ModeIntegrated = "integrated"
+	ModeRemote     = "remote"
+
+	defaultLndMode = ModeRemote
```
halseth, line 38: should this be integrated?

config.go
```diff
@@ -149,15 +378,36 @@ func loadLndConfig(preCfg *Config) (*lnd.Config, error) {
...
+	if err := makeDirectories(litDir); err != nil {
 		return nil, err
 	}
 
-	// With the validated config obtained, we now know that the root logging
-	// system of lnd is initialized and we can hook up our own loggers now.
-	SetupLoggers(cleanCfg.LogWriter)
+	switch cfg.LndMode {
+	// In case we are running lnd in-process, let's make sure its
+	// configuration is fully valid. This also sets up the main logger to
```
halseth, line 388: sentence unfinished

config.go
```diff
@@ -166,7 +416,71 @@ func loadLndConfig(preCfg *Config) (*lnd.Config, error) {
...
+	if err := makeDirectories(filepath.Dir(r.LitTLSCertPath)); err != nil {
+		return err
+	}
+	if err := makeDirectories(filepath.Dir(r.LitTLSKeyPath)); err != nil {
+		return err
+	}
+
+	logWriter := build.NewRotatingLogWriter()
+	cfg.Lnd.LogWriter = logWriter
+	err := logWriter.InitLogRotator(
```
halseth, line 472: shouldn't this also be set up in integrated mode?

halseth (approved):
Excellent work on remote node support, tested and works flawlessly 🙌  Only a few questions and clarifications, otherwise this LGTM.

go.mod
guggero: I tried. But `go mod` would change it back because `loop` and `faraday` also use the RC version. And I'd rather avoid adding an overwrite clause as that can have side effects in the versions that are actually resolved in the end.

rpc_proxy.go
guggero: According to the `grpcProxy` godoc, this is crucial for the gRPC web -> gRPC native proxy to work. Added a comment to clarify.

config.go
guggero: No, we decided that the default mode should be remote as most users might already be running an `lnd` node.

config.go
guggero: In integrated mode, we call `lnd.ValidateConfig()` which creates the default log writer. Added a comment to clarify this here.

jamaljsr (approved):
Tested https://github.com/lightninglabs/lightning-terminal/commit/da194d0fe6a56202b38f20aee48d2b8b502a310c on mainnet in remote mode with letsencrypt, and also regtest in integrated mode. Default settings work like a charm. Great job on this feature Oli. LGTM 🥇

config.go
halseth: Ok, then I think maybe the config description must be updated, as it says `integrated` is the default!

PR #130, opened 2020-10-13
Update dependencies, set loopd user agent to "litd"

guggero:
Updates the dependencies of `loopd` and `faraday` and sets the Agent Name part of `loopd`'s user agent string to `litd` so we can see LiT usage on the server side.


jamaljsr (approved):
tACK 🥇

go.mod
```diff
@@ -25,4 +25,6 @@ require (
 	gopkg.in/macaroon.v2 v2.1.0
 )
 
+replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.11.1-beta
```
halseth, line 28: is this because of the implicit loop dependency?

guggero: Yeah, both `loop` and `faraday` reference the `-beta.rc3` version. So without the replace directive, go mod changes it to that as well. Even though if semver was implemented correctly, the `-beta` should have precedence over `-beta.rc3`...

halseth (approved):

PR #141, opened 2020-11-12
Loop user agent initiator

guggero:
Results in this user agent string being sent from LiT:

`litd/v0.11.1-beta/commit=v0.11.1-beta,initiator=lit-ui`


alexbosworth (approved):

jamaljsr (approved):
tACK 🥇 

Thanks for updating the frontend. I don't know how to confirm the `lit-ui` value is stored on the backend but I can see it being sent in the Loop Out request correctly.

PR #151, opened 2020-11-17
Update golang to 1.15.5 for security update

stefanlsd:
https://groups.google.com/g/golang-announce/c/NpBGTTmKzpM/m/fLguyiM2CAAJ

Geth has made this change a high priority. Unsure if LND is affected. I have built LIT docker with 1.15.5 and it seems to run ok.  (havent tested lnd, pool, loop) - running to a remote lnd.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13-alpine as builder
+FROM golang:1.15.5-alpine as builder
 
 # Copy in the local repository to build from.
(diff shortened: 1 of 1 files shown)


guggero (approved):
Thank you for the update, LGTM!
We'd normally not put an explicit minor version into the `Dockerfile` to allow it to pick the latest version. But I think in this case it's a good thing to do, to make sure an older version isn't chosen from the cache.

In the meantime, I'm going to re-build all LiT release binaries with `go1.15.5`.

PR #153, opened 2020-11-23
docker+GitHub: add production Dockerfile and build automatically on tag push

guggero:
We add a new production `Dockerfile` and a GitHub workflow that is triggered whenever a new version tag is
pushed. It will trigger a docker image build for that version and automatically push it to the specified repo.

For this to work, the GitHub repository needs the following two secrets to be configured:

```
DOCKER_USERNAME: <the user to log into Docker Hub>
DOCKER_API_KEY: <an API key generated by Docker Hub for the above user>
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,13 @@ go-install:
 	@$(call print, "Installing lightning-terminal.")
 	$(GOINSTALL) -tags="$(LND_RELEASE_TAGS)" -ldflags "$(LDFLAGS)" $(PKG)/cmd/litd
 
+go-install-cli:
+	@$(call print, "Installing all CLI binaries.")
+	$(GOINSTALL) -trimpath -tags="$(LND_RELEASE_TAGS)" -ldflags "$(LDFLAGS)" github.com/lightningnetwork/lnd/cmd/lncli
+	$(GOINSTALL) -trimpath -ldflags "$(LDFLAGS)" github.com/lightninglabs/loop/cmd/loop
+	$(GOINSTALL) -trimpath github.com/lightninglabs/faraday/cmd/frcli
+	$(GOINSTALL) -trimpath github.com/lightninglabs/pool/cmd/pool
+
 app-build: yarn-install
 	@$(call print, "Building production app.")
 	cd app; yarn build
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,45 @@
+name: Docker image build
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  DOCKER_REPO: lightninglabs
+  DOCKER_IMAGE: lightning-terminal
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -21,18 +28,21 @@ RUN apk add --no-cache --update alpine-sdk \
...
-&&  make install \
-&& go install -v -trimpath github.com/lightningnetwork/lnd/cmd/lncli \
-&& go install -v -trimpath github.com/lightninglabs/faraday/cmd/frcli \
-&& go install -v -trimpath github.com/lightninglabs/loop/cmd/loop \
-&& go install -v -trimpath github.com/lightninglabs/pool/cmd/pool
+&& git clone https://github.com/lightninglabs/lightning-terminal /go/src/github.com/lightninglabs/lightning-terminal \
+&& cd /go/src/github.com/lightninglabs/lightning-terminal \
+&& git checkout $checkout \
+&& make install \
+&& make go-install-cli \
```
jamaljsr, line 45: Remove the trailing `\` as it's causing the docker build to fail

jamaljsr (changes requested):
Tested building both Dockerfiles locally. Ran into an issue with the `Dockerfile`.

Would it be better to have the default `Dockerfile` use the local files instead of cloning from github at a specific version? 
1. If someone pulls the repo, I would assume that they want the docker image to use the local source files versioned at the commit they pulled. If we force them to specify `-f Dockerfile.dev`, it adds a bit of friction. 
1. The `Dockerfile` that is used for the GitHub Action build and push to DockerHub is something no dev is likely to use, so I think it would be better to have the friction there in the automated workflow, since you just set it once and forget it.
1. The `checkout` build arg is only going to work with future tags. When I use the current `v0.3.1-alpha` tag or prior ones, the docker build fails with the error:
    ```
    make: *** No rule to make target 'go-install-cli'.  Stop.
    ```

Dockerfile
guggero: Argh, always these last minute changes that get me. Sorry, fixed.

guggero:
Thank you very much for testing and the feedback you provided!

> If someone pulls the repo, I would assume that they want the docker image to use the local source files versioned at the commit they pulled. If we force them to specify -f Dockerfile.dev, it adds a bit of friction.
> The Dockerfile that is used for the GitHub Action build and push to DockerHub is something no dev is likely to use, so I think it would be better to have the friction there in the automated workflow, since you just set it once and forget it.

We discussed this offline. The goal is to align the naming of the `Dockerfile`/`dev.Dockerfile` with how they're now named in `lnd`. I added a section to the `compile.md` that mentions the difference between the two.

> The checkout build arg is only going to work with future tags. When I use the current v0.3.1-alpha tag or prior ones, the docker build fails with the error: `make: *** No rule to make target 'go-install-cli'.  Stop.`

Yes, that's an unfortunate side effect of pulling the source from git. It will indeed not be possible to build older versions than the upcoming `v0.3.2-alpha`. But I think that is okay as that will also be the first version with an official Lightning Labs provided image.
To nevertheless test the production build, you can run `docker build -t prod-litd --build-arg checkout=prod-docker-file` to try with the current branch of this PR (which is pushed to upstream).

jamaljsr (approved):
LGTM 👍 

Thanks for adding the docs. This will certainly be helpful.

Makefile
```diff
@@ -112,6 +112,13 @@ go-install:
 	@$(call print, "Installing lightning-terminal.")
 	$(GOINSTALL) -tags="$(LND_RELEASE_TAGS)" -ldflags "$(LDFLAGS)" $(PKG)/cmd/litd
 
+go-install-cli:
+	@$(call print, "Installing all CLI binaries.")
```
Roasbeef, line 116: 👍

Roasbeef (approved):
LGTM 💈

Repo secrets are now present as well!

PR #158, opened 2020-12-09
docker: refactor build process to support building ARM images

guggero:
This PR prepares the build process to allow building ARM based docker images.
To achieve that, we need to remove the `*.proto` to JavaScript compilation step from the build process itself and require PRs that update the protos to update the code instead.

The resulting docker image(s) can be inspected here: https://hub.docker.com/layers/guggero/lightning-terminal/v0.0.0-docker-arm5/images/sha256-68baf47fc333b998c86fd07c325e4e1827ea9b0f60d9843395144ca9e7a9b701?context=explore


Makefile
```diff
@@ -192,6 +192,14 @@ list:
 		grep -v Makefile | \
 		sort
 
+protos:
+	@$(call print, "Compiling protos.")
+	cd ./app; yarn protos
+
+protos-check: protos
```
jamaljsr, line 201: Nice validation check! 👍

jamaljsr (commented):
Overall, I'm in favor of this approach. It'll remove the need for users to install `protoc` to compile from source, which has been an issue for some already.  Good stuff 👍

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Instruct the GitHub linguist tool to interpret the generated JS code as just
+# that, generated. This will make sure the JS (and TS types for that matter)
+# will not show up in the repository's language statistics.
+
+app/src/types/generated* linguist-generated
```
jamaljsr, line 5: 🙌

Thanks for finding this hidden gem.

jamaljsr (approved):
tACK 👍 

Tested both Dockerfile builds successfully.

.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Instruct the GitHub linguist tool to interpret the generated JS code as just
```
Roasbeef, line 1: Solid move, now I think only those doing development need to make a RAM reservation to compile `protoc` themselves...

Roasbeef (approved):
LGTM 🕌

PR #159, opened 2020-12-11
Update protos build script

jamaljsr:
As a follow up to #158 which removed the dependency on `protoc` at the machine level to compile from source, this PR makes a few updates:

- update the compile.md doc removing the `protoc` dependency
- update `yarn protos` script to download protos from github instead of needing to do this manually
- update the compile.md with instructions to compile the protos whenever we update the `lnd` and `loop` versions
- update the proto files to match the `lnd` and `loop` versions currently included with Terminal
- fix generated code being included in GitHub language stats

### DIFF
--- a/app/scripts/build-protos.js
+++ b/app/scripts/build-protos.js
@@ -1,10 +1,21 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const util = require('util');
+const https = require('https');
 const exec = util.promisify(require('child_process').exec);
 const { promises: fs } = require('fs');
 const { join, sep } = require('path');
 const { platform } = require('os');
 const appPath = join(__dirname, '..');
+
+/** Specify the versions of LND and Loop protos to download */
+const LND_VERSION = 'v0.11.1-beta';
+const LOOP_VERSION = 'v0.11.2-beta';
+
+/** mapping of proto files to the github url to download each from */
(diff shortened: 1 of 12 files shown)


app/scripts/build-protos.js
```diff
@@ -1,11 +1,22 @@
...
+const https = require('https');
 const exec = util.promisify(require('child_process').exec);
 const { promises: fs } = require('fs');
 const { join, sep } = require('path');
 const { platform } = require('os');
 const appPath = join(__dirname, '..');
 
+/** Specify the versions of LND and Loop protos to download */
+const LND_VERSION = 'v0.11.1-beta';
+const LOOP_VERSION = 'v0.11.2-beta';
```
guggero, line 12: Great idea checking that we have the correct version of the `*.proto` themselves checked in.
We _could_ even go a step further and parse this version out of the `go.mod` in the root directory. That way we'd notice if we update the dependency that's built into the unified binary without updating the protos.
Can be done it a separate PR though.

guggero (approved):
LGTM, nice cleanup :100:

app/scripts/build-protos.js
jamaljsr: Great point! I didn't think to use the `go.mod` for the versions but that makes total sense.

PR #161, opened 2020-12-14
Prepare for LiT v0.3.3-alpha

guggero:
With this PR we prepare to release a minor version of LiT that contains the following changes:
- Only build docker images for 64bit ARM, remove the 32bit target to be in sync with the `lnd` images.
- Update Loop to `v0.11.2-beta`.
- Extract `lnd` and `poold` versions from `go.mod` when compiling proto files.

### DIFF
--- a/app/scripts/build-protos.js
+++ b/app/scripts/build-protos.js
@@ -7,14 +7,34 @@ const { join, sep } = require('path');
 const { platform } = require('os');
 const appPath = join(__dirname, '..');
 
-/** Specify the versions of LND and Loop protos to download */
-const LND_VERSION = 'v0.11.1-beta';
-const LOOP_VERSION = 'v0.11.2-beta';
+/**
+ * Specify the pattern under which the project's version can be found in the
+ * root directory's go.mod file.
+ */
+const LND_VERSION_PATTERN = /^\tgithub\.com\/lightningnetwork\/lnd (v[\d.]+-beta)/ms;
+const LOOP_VERSION_PATTERN = /^\tgithub\.com\/lightninglabs\/loop (v[\d.]+-beta)/ms;
 
 /** mapping of proto files to the github url to download each from */
-const protoSources = {
-  lnd: `lightningnetwork/lnd/${LND_VERSION}/lnrpc/rpc.proto`,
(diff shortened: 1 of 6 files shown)


app/scripts/build-protos.js
```diff
@@ -7,14 +7,34 @@ const { join, sep } = require('path');
...
+ * root directory's go.mod file.
+ */
+const LND_VERSION_PATTERN = /^\tgithub\.com\/lightningnetwork\/lnd (v[\d.]+-beta)/ms;
+const LOOP_VERSION_PATTERN = /^\tgithub\.com\/lightninglabs\/loop (v[\d.]+-beta)/ms;
 
 /** mapping of proto files to the github url to download each from */
-const protoSources = {
-  lnd: `lightningnetwork/lnd/${LND_VERSION}/lnrpc/rpc.proto`,
-  loop: `lightninglabs/loop/${LOOP_VERSION}/looprpc/client.proto`,
+const protoSources = async () => {
```
jamaljsr, line 18: Much better implementation! Good stuff 🥇

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/lightninglabs/aperture v0.1.3-beta
 	github.com/lightninglabs/faraday v0.2.2-alpha
 	github.com/lightninglabs/lndclient v0.11.0-3
-	github.com/lightninglabs/loop v0.11.1-beta
+	github.com/lightninglabs/loop v0.11.2-beta
```
jamaljsr, line 13: Crap! I though we were on `v0.11.2` of `loop` already.

jamaljsr (approved):
LGTM 👍 

My only suggestion would be to update the [compile.md](https://github.com/lightninglabs/lightning-terminal/blob/master/doc/compile.md#compiling-grpc-proto-files) to remove step 2 since it's no longer necessary.

go.mod
guggero: No problem, gives us a good reason to do a minor release of LiT which then also will trigger the ARM docker image builds for the first time.

guggero:
> My only suggestion would be to update the compile.md to remove step 2 since it's no longer necessary.

Good idea, removed.

Issue #162, opened 2020-12-15
Allow disabling SSL


lukechilds:
We want to package LiT as an Umbrel app and to do this each app registers a port which the user can access over the local network or via a Tor hidden service.

You can read more about our app design here for context: https://github.com/getumbrel/umbrel/blob/master/apps/README.md

It would be great if we could disable SSL on LiT, it doesn't provide any extra security in the context of Umbrel and is a UX nightmare for users.

All remote connections are handled by exposing the app as a Tor hidden service which handles encryption too. Using SSL on top of Tor is redundant.

For local connections we can't generate CA signed certs, so we'd need to use self signed certs. This is a UX nightmare for users to enable, they get all sorts of scary browser warnings that make it look like it's not working.

It's also full of security footguns, if users don't correctly pin the certificate (most have no idea how to do this) then SSL is providing no security, despite the fact it appears to be working.

It would be great if LiT supported a config parameter to allow serving the web UI over plain HTTP and not requiring SSL. Would you be interested in implementing that?

jamaljsr:
Yes, I think we all would be interested in seeing LiT packaged with Umbrel. 

I'll do an investigation to see what needs to be done to accomplish this. My immediate worry is that we currently use `grpc-web` on the client with `grpc-webproxy` + `http2` on the backend. I believe that most, if not all, browsers that support `http2` require TLS. I'm sure there is a way to work around this limitation, it just probably isn't as simple as changing a config option. I imagine that this will be a bit more involved to get working.

I'll follow up as soon as I can, once I gather more info and do some testing.

lukechilds:
Awesome, thanks @jamaljsr!

Yeah I hadn't considered HTTP2 issues. Specter are facing similar issues due to camera APIs used for scanning QR codes not being available without HTTPS.

Let me know if there's anything we can do to help from our end.

jamaljsr:
Good news. I've tested locally and confirmed that the `grpc-webproxy` will work with a relatively small amount of changes to the backend. The only thing we have to figure out now is what config flags to expose. It is possible to run both `https` and `http` servers simultaneously, or we could only allow one at a time.

I suppose that you want to run the `http` server only, not both. So a simple `--insecure-http=true` flag would probably work for your specific use case. Is that correct?

lukechilds:
Amazing!

> So a simple --insecure-http=true flag would probably work for your specific use case. Is that correct?

That would be perfect, yes!

[...]

PR #163, opened 2020-12-17
add insecure HTTP only listener option

guggero:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/162.

For uses of LiT through Tor connections that add their own layer of
transport security, it doesn't make sense to add TLS on top. We
therefore allow running an additional HTTP only listener.

@justinpobrien requesting review from you as well because this contains a potential security footgun if used improperly. Does the flag name `--insecure-httplisten` convey the danger of enabling this sufficiently?

### DIFF
--- a/config.go
+++ b/config.go
@@ -112,6 +112,7 @@ var (
 // daemon's short name.
 type Config struct {
 	HTTPSListen    string `long:"httpslisten" description:"The host:port to listen for incoming HTTP/2 connections on for the web UI only."`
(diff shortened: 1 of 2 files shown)


jamaljsr (approved):
tACK 🔥

Man! I wasn't expecting a PR so quickly. Thanks so much @guggero

justinpobrien (approved):
Yes conveys danger sufficiently.

Issue #155, opened 2020-12-01
litd panic after poold RPC request while lnd is locked


jamaljsr:
This is a bit of an edge case today but may become a common problem in the future when the Pool UI is used by more people.

Steps to reproduce:
- run `litd` in integrated mode
- unlock `lnd` with `lncli unlock`
- open the UI in a browser and login
- restart `litd` so that `lnd` is now locked
- trigger a `loopd` gRPC request from the UI (click on the Loop button)
- observe an error notification is displayed stating:
   ```
   loop is not yet ready for requests, lnd possibly still starting or syncing
   ```
- trigger a `poold` gRPC request from the UI
- observe an error notification is displayed stating:
   ```
   Response closed without headers
   ```
- switch to the terminal where `litd` is running and see that the process has ended by a panic error

Logs from startup to panic:
```
2020-11-30 23:58:05.101 [INF] LTND: Version: 0.11.1-beta commit=lightning-terminal-v0.3.0-alpha-98-g88f0ac0ad27002bf7f30bf74c8e8c691877f1839, build=production, logging=default
2020-11-30 23:58:05.101 [INF] LTND: Active chain: Bitcoin (network=testnet)
2020-11-30 23:58:05.101 [INF] LTND: Opening the main database, this might take a few minutes...
2020-11-30 23:58:05.101 [INF] LTND: Opening bbolt database, sync_freelist=false
2020-11-30 23:58:06.213 [INF] CHDB: Checking for schema update: latest_version=17, db_version=17
2020-11-30 23:58:06.214 [INF] LTND: Database now open (time_to_open=1.112434896s)!
2020-11-30 23:58:06.225 [INF] RPCS: password RPC server listening on 0.0.0.0:10009
2020-11-30 23:58:06.225 [INF] LITD: Dialing lnd gRPC server at 127.0.0.1:10009
2020-11-30 23:58:06.225 [INF] RPCS: password gRPC proxy started at 127.0.0.1:8080
2020-11-30 23:58:06.226 [INF] LTND: Waiting for wallet encryption password. Use `lncli create` to create a wallet, `lncli unlock` to unlock an existing wallet, or `lncli changepassword` to change the password of an existing wallet and unlock it.
2020-11-30 23:58:06.351 [INF] LITD: Setting up Let's Encrypt for server <snipped>
2020-11-30 23:58:06.351 [INF] LITD: Setting up Let's Encrypt with cache dir /root/.lit/letsencrypt
[...]

PR #172, opened 2021-01-18
Add lnd v0.12.0-beta compatibility, allow single macaroon

guggero:
Depends on https://github.com/lightninglabs/pool/pull/202 and the final release of `lnd v0.12.0-beta`.

Adds compile time and run time compatibility with `lnd v0.12.0-beta`.
Also pulls in the latest version of `lndclient` which allows us to cleanly block on startup until `lnd` is unlocked.
We can also use a single macaroon now instead of needing to specify all subserver macaroons. Either the `admin.macaroon` is used if `--remote.lnd.macaroondir` is specified. Or a custom one (must contain all required permissions!) can be specified with `--remote.lnd.custommacaroonpath`.

Fixes https://github.com/lightninglabs/lightning-terminal/issues/155.


README.md
```diff
@@ -77,6 +78,7 @@ Lightning Terminal is backwards compatible with `lnd` back to version v0.11.1-be
 
 | LiT              | LND          | Loop        | Faraday      | Pool          |
 | ---------------- | ------------ | ----------- | ------------ |---------------|
+| **v0.3.5-alpha** | v0.12.0-beta | v0.11.2-beta | v0.2.3-alpha | v0.4.2-alpha |
```
jamaljsr, line 81: Why v0.3.5? AFAIK we want the next release to be v0.4.0 and include the pool ui.

README.md
```diff
@@ -66,6 +66,7 @@ Lightning Terminal is backwards compatible with `lnd` back to version v0.11.1-be
 
 | LiT              | LND          |
 | ---------------- | ------------ |
+| **v0.3.5-alpha** | v0.11.1-beta |
```
jamaljsr, line 69: Shouldn't this be v0.12.0-beta like below?

config.go
```diff
@@ -172,6 +173,13 @@ type RemoteDaemonConfig struct {
 	// MacaroonDir is the directory that contains all the macaroon files
 	// required for the remote connection.
 	MacaroonDir string `long:"macaroondir" description:"The directory containing all lnd macaroons to use for the remote connection."`
+	
+	// CustomMacaroonPath is the path to the custom macaroon that should be
+	// used instead of needing to specify the macaroon directory that
+	// contains all of lnd's macaroons. The custom macaroon MUST have all
+	// permissions that all the subservers use, otherwise permission errors
+	// will occur.
+	CustomMacaroonPath string `long:"custommacaroonpath" description:"The path to the custom macaroon to use. Cannot be specified at the same time as macaroondir. The custom macaroon must contain ALL permissions required for all subservers to work, otherwise permission errors will occur."`
```
jamaljsr, line 182: I like that we can now just specify the admin macaroon file instead of the full dir. I wonder if we should just deprecate `macaroondir` and change `custommacaroonpath` to `macaroonpath` instead of keeping both flags permanently. I believe that having `custommacaroonpath` adds a bit of confusion. My hunch is that everyone should just use this flag pointed at the admin macaroon. Maybe I'm misunderstanding the intention here though.

jamaljsr (commented):
I haven't tested the branch yet, but had a couple comments on the approach taken.

README.md
guggero: I wasn't sure if the UI would be ready together with lnd v0.12. But in that case it makes sense to update to `v0.4.0` in the PR that adds the UI. Going to remove this diff then.

README.md
guggero: In theory we should still be backward compatible down to `v0.11.1-beta`, at least for remote connections.

jamaljsr: Doh! I should've probably read the sentence above that wasn't visible in the diff. Ok, that makes sense.

README.md
jamaljsr: Cool 👍

config.go
guggero: Great point. I've added a deprecation comment to the dir flag and renamed the new one to `macaroonpath`.

jamaljsr: Awesome! 🥇 

Not sure if you want to do it in this PR, but we should also update the docs & examples to use `macaroonpath` as well.

jamaljsr:
Your latest rebase includes an updated `lnd` proto which breaks the CI build. The new proto requires an update to the frontend Typescript types. I think it's fine to keep the old proto in this PR as I will update it separately.

guggero:
The new RPC compile check prevents us from using an old version. Perhaps we need a way to override that.
Or I can update the testdata, should be straightforward.

The check will fail until we reference the final version of lnd anyway since it tries to download it from an URL without the rc tag.

jamaljsr:
I agree that the RPC compile check will fail, but the build is also failing on the "typescript compile" step because the frontend code is not updated for v0.12 protos. This also breaks my local docker build because the frontend won't build successfully. 

Can you either remove the proto update or update the `sampleData.ts` file to fix the frontend build?

config.go
guggero: Yes, I did update the docs. Or did I overlook anything?

guggero:
I fixed the `sampleData.ts` with a temporary commit. Will remove that once we've tagged the final `lnd v0.12.0-beta` version.

config.go
jamaljsr: Oh, my bad. I missed that on my end. Looks good 👍

jamaljsr (approved):
tACK LGTM 👍

proto/lnd.proto
```diff
@@ -3351,12 +3518,31 @@ message MacaroonPermission {
...
+message ListMacaroonIDsResponse {
+    // The list of root key IDs that are in use.
+    repeated uint64 root_key_ids = 1;
+}
+
+message DeleteMacaroonIDRequest {
+    // The root key ID to be removed.
+    uint64 root_key_id = 1;
+}
+message DeleteMacaroonIDResponse {
```
carlaKC, line 3541: nit: newline between requests and responses here?

terminal.go
```diff
@@ -304,6 +304,7 @@ func (g *LightningTerminal) startSubservers() error {
 			MacaroonDir:           macDir,
 			TLSPath:               tlsPath,
 			BlockUntilChainSynced: true,
+			BlockUntilUnlocked:    true,
```
carlaKC, line 308: 🎉 

One question here in the context of lit. Since we just log that we are waiting for lnd to be unlocked, I imagine there won't be any feedback indicating that the user needs to unlock? Still an improvement over crashing (which was the previous behaviour I think?), just wondering about it longer term.

config.go
```diff
@@ -171,7 +172,14 @@ type RemoteDaemonConfig struct {
...
 	// required for the remote connection.
-	MacaroonDir string `long:"macaroondir" description:"The directory containing all lnd macaroons to use for the remote connection."`
+	MacaroonDir string `long:"macaroondir" description:"DEPRECATED: Use macaroonpath. The directory containing all lnd macaroons to use for the remote connection."`
+
+	// MacaroonPath is the path to the single macaroon that should be used
+	// instead of needing to specify the macaroon directory that contains
+	// all of lnd's macaroons. The specified macaroon MUST have all
+	// permissions that all the subservers use, otherwise permission errors
+	// will occur.
+	MacaroonPath string `long:"macaroonpath" description:"The full path to the single macaroon to use, either the admin.macaroon or a custom baked one. Cannot be specified at the same time as macaroondir. A custom macaroon must contain ALL permissions required for all subservers to work, otherwise permission errors will occur."`
```
carlaKC, line 182: Are we going to add the bake command in the readme for a custom macaroon? I remember we discussed this a while back but don't remember exactly what we came up with. Could be nice to point people there if that's the plan.

carlaKC (approved):
Lndclient changes LGTM!

terminal.go
guggero: Yeah, you'd only see this in the log at the moment. This will be fully solved by implementing the unlock in the UI itself.

config.go
guggero: Yes, we'll probably do that eventually. But because things with macaroons will probably change again to support the "stateless remote mode", we'll probably look at it in that context again.

PR #179, opened 2021-02-16
Add remote mode for faraday, loopd and poold

guggero:
This PR makes it possible to also run `farday`, `loopd` and `poold` in remote mode.
The UI then only acts as a grpc-web reverse proxy that converts the basic authentication into a macaroon authentication.
Users can also use the command line tools like `frcli`, `loop` or `pool` against the UI port and they'll be proxied to the remote node.

To simplify some parts of the code we also do some cleanup and refactoring. For instance we now remove the deprecated `--lnd.remote.macaroondir` flag. And we move the `--lnd.remote.network` flag to a new top-level `--network` flag who's value will be propagated into all sub structs automatically.


rpc_proxy.go
```diff
@@ -200,7 +263,29 @@ func (p *rpcProxy) director(ctx context.Context,
...
+	// mode. For all other daemons the request shouldn't get here in
+	// integrated mode (after all, the director only picks up calls that the
+	// gRPC server itself would throw a 404 for) so we throw an error
+	// message for them if they're not in remote mode.
+	switch {
+	case isLndURI(requestURI):
+		return outCtx, p.lndConn, nil
+
+	case isFaradayURI(requestURI) && p.cfg.FaradayMode == ModeRemote:
+		return outCtx, p.loopConn, nil
```
jamaljsr, line 277: Should this be `p.faradayConn`?

jamaljsr (commented):
Outstanding update!! 

I quickly skimmed through the diff just to see what changes were made. The config updates seem simple enough to follow. Great work.

I'll try to run it locally later today.

rpc_proxy.go
jamaljsr: Actually meant to select line #280 below.
```go
	case isFaradayURI(requestURI) && p.cfg.FaradayMode == ModeRemote:
		return outCtx, p.loopConn, nil
```

guggero: Most definitely. Great catch, thank you!

rpc_proxy.go
```diff
@@ -199,7 +263,29 @@ func (p *rpcProxy) director(ctx context.Context,
 
 	outCtx := metadata.NewOutgoingContext(ctx, mdCopy)
 
-	return outCtx, p.lndConn, nil
+	// Direct the call to the correct backend. For lnd we _always_ have a
```
halseth, line 266: I might be missing some context here on why this is needed/done this way, but why wouldn't we unify this such that all backends were treated the same (lnd, loop, pool, faraday)?

halseth (commented):
Very cool, I've been wanting this for a while! Will be testing this :)

jamaljsr (approved):
LGTM 🚀

Tested on regtest. It works as advertised. Great work!

rpc_proxy.go
```diff
@@ -135,6 +143,41 @@ func (p *rpcProxy) Start() error {
...
+				p.cfg.Remote.Faraday.TLSCertPath,
+			),
+		)
+		if err != nil {
+			return fmt.Errorf("could not dial remote faraday: %v",
+				err)
+		}
+	}
+
+	if p.cfg.LoopMode == ModeRemote {
```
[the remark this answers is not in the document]
guggero: I was trying to make it as similar as possible to `lnd-mode`. But because that flag defaults to the value `"remote"` and has other implications as well (how/where the config/logs/certs are stored), it's probably okay to use a different pattern for the other daemons. Switched them over to a `bool`.

rpc_proxy.go
guggero: It's because of the grpc-web proxy. We need to create our own `grpc.Server` instance with the custom codec enabled. We can hook the "guest" daemons into that one directly because we added the `StartAsSubserver()` method to them. The gRPC server in `lnd` is a black box to us and that's why we always have to forward the call to a connection, even if that connection is to a port in the same process.

We could probably simplify this by exposing the RPC server in `lnd`. But for this PR it actually made it way easier to add the remote modes of the other daemons because this was already in place.

rpc_proxy.go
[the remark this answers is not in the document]
halseth: I mean, why not use bool for `lnd-mode` also? (rename it to `remote-lnd` or similar.

rpc_proxy.go
halseth: Ah, of course. Would it be an idea to add the `StartAsSubserver` option to lnd going forward? Can easily be done together with the `grpc` refactoring I'm doing for lnd (I think).

rpc_proxy.go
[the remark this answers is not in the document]
guggero: The initial idea was to make it more verbose for the user that just looks at the CLI help output. With an explicit mode you can have better descriptions in the section headers.
For example:
```go
Remote *RemoteConfig `group:"Remote mode options (use when lnd-mode=remote)" namespace:"remote"`
```

Booleans are also tricky because they need to be used differently depending on whether you're in a config file or not (`bool-flag=false` vs. not specifying the flag on the CLI). Additionally, the default for `lnd` is remote. So if we added a `--lnd-integrated` boolean flag, the above help text would become a bit less easy to understand IMO:
```go
Remote *RemoteConfig `group:"Remote mode options (use when lnd-integrated=false or not present as a flag)" namespace:"remote"`
```

And finally, I didn't want to break existing users that currently use the `--lnd-mode=integrated` flag.

rpc_proxy.go
guggero: Yes, sure. That would make it a bit nicer. And LiT would actually become the main process and not just a barnacle around `lnd` :joy:

rpc_proxy.go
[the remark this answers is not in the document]
halseth: ah, I see. I would say use the string variant for all then, nice to have them be similar :)

halseth (approved):
Great work! Tested this on my mainnet node, was able to successfully connect to my existing lnd, loop, pool and faraday nodes. Even did a successful loop and bought a channel on pool 😀 

LGTM 🚀

halseth:
@guggero  Only thing was perhaps revert to the `pool-mode=remote` config option. Your call.

guggero:
Okay, I reverted the mode flags back to string on the CLI but kept an internal `bool` variable around to make it easier to read. And now we get the best of both worlds :)
I also added a comment why we do it this way.

guggero:
@halseth re-requesting a review just to make sure I didn't miss anything in the reversal of the flags.

config.go
```diff
@@ -37,7 +37,10 @@ const (
 	ModeIntegrated = "integrated"
 	ModeRemote     = "remote"
 
-	defaultLndMode = ModeRemote
+	defaultLndMode     = ModeRemote
```
halseth, line 40: pre-existing, but is it correct? Documentation indicates that default is `integrated`

halseth (approved):
Looks good, only a (pre-esisting) question about default lnd mode

config.go
guggero: It's correct. But we changed the default, so if the doc mentions `integrated` then we need to update it. Where did you find that reference to `integrated`?

guggero: Ah, I found where it was still wrong. Added a commit to fix it.

PR #185, opened 2021-02-24
build: update to lnd v0.12.1-beta

Roasbeef:
Locally I used Go 1.16 to make these changes, but I don't think that
should be an issue. It seems to handle the replace directive, and
updating `go.sum` better than it did in the past.


go.mod
```diff
@@ -30,4 +30,4 @@ require (
 
 go 1.13
 
-replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.12.0-beta
+replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.12.1-beta
```
guggero, line 33: Because the version is now higher, the replace apparently isn't needed anymore. I included this commit in the bugfix PR #186 but removed the replace. So we can close this if you don't mind me taking it over.

Roasbeef:
Closing in favor of #186

PR #189, opened 2021-02-26
Fix main lnd port usage for daemons in lnd-mode=integrated

guggero:
The call flow diagram wasn't accurate what lead to an incorrect
assumption in the last PR.
In integrated mode, lnd spins up its gRPC server as the main entry point
and the other daemons need to hook into it.

To make sure bugs like this don't happen anymore, we also add a test scenario/protocol document that covers all the existing use cases.

### DIFF
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -1,0 +1,89 @@
+# Test scenarios
+
+The possible number of combinations of local/remote configurations has become
+quite large over time.
+
+The following tests should be run whenever something in the local/remote config
+or proxying changes:
+
+## `lnd` integrated
+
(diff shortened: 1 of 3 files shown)


jamaljsr (approved):
Tested locally on regtest. Thanks for adding the test cases. LGTM 👍

doc/testing.md
```diff
@@ -0,0 +1,89 @@
+# Test scenarios
```
Roasbeef, line 1: 👍

Roasbeef (approved):
LGTM 🤺

PR #190, opened 2021-02-27
pool+accounts: display warning if node has no channels

jamaljsr:
When the `lnd` node has no open channels, an account cannot be opened due to an inevitable LSAT payment failure. If the account creation is attempted, the resulting error message is also very cryptic.

This PR just adds a notice message when the node doesn't have any channels, stating "Your node must already have open channels before you can fund your Pool account.". It also disables the "Open an Account" button in this case as well.

image


guggero (approved):
tACK, LGTM :100:

PR #193, opened 2021-03-01
loop+history: display failure reason on History Page

jamaljsr:
When a swap fails, the Loop History page currently displays the label "Failed" without any more details. Recent versions of `loopd` include a `failureReason` field for each swap. This field provides a bit more informations as to why a swap has failed which can be useful for the user to troubleshoot.

This PR updates the Loop History page to display the failure reason in the Status column for failed swaps. It also adds a tooltip to provide a longer explanation of the failure.

## Steps to test
To test this PR, you can either induce a swap failure or run the following in the browser console:
```
// set the state to FAILED
store.swapStore.sortedSwaps[1].state = 4;
// set the failure reason to FAILURE_REASON_SWEEP_TIMEOUT
store.swapStore.sortedSwaps[1].failureReason = 2
```

## Screenshot
image


app/src/i18n/locales/en-US.json
```diff
@@ -33,6 +33,13 @@
   "cmps.history.HistoryRowHeader.type": "Type",
   "cmps.history.HistoryRowHeader.created": "Created",
   "cmps.history.HistoryRowHeader.updated": "Updated",
+  "cmps.history.HistoryRow.failure-0": "The swap has failed to complete successfully. Check the logs for additional information.",
```
guggero, line 36: I think `FAILURE_REASON_NONE` (value 0) means "no error", e.g. success. This message won't be displayed in that case, but just for completeness' sake we should add the correct description here.

app/src/store/models/swap.ts
```diff
@@ -81,6 +82,29 @@ export default class Swap {
...
+        return 'On-chain Timeout';
+      case LOOP.FailureReason.FAILURE_REASON_SWEEP_TIMEOUT:
+        return 'Sweep Timeout';
+      case LOOP.FailureReason.FAILURE_REASON_INSUFFICIENT_VALUE:
+        return 'Insufficient Value';
+      case LOOP.FailureReason.FAILURE_REASON_TEMPORARY:
+        return 'Temporary Failure';
+      case LOOP.FailureReason.FAILURE_REASON_INCORRECT_AMOUNT:
+        return 'Incorrect Amount';
+      case LOOP.FailureReason.FAILURE_REASON_NONE:
```
guggero, line 102: `FAILURE_REASON_NONE` is the success case. I'd return 'Success' here and 'Failed` only in the default case. Just to be consistent.

guggero (approved):
tACK, LGTM, just two small nits :tada:

app/src/store/models/swap.ts
jamaljsr: This is a catch-all case that theoretically should never be hit. In [HistoryRow.tsx#L122](https://github.com/lightninglabs/lightning-terminal/blob/bb40620d8bc4de74bfe23aca212e1aad431083b0/app/src/components/history/HistoryRow.tsx#L122) it checks if the `swap.state` value is Failed. Only in this case will it display the `failureLabel`. The only way this value would be displayed in the UI is if `swap.state === FAILED && swap.failureReason === FAILURE_REASON_NONE`, which should not be possible. If this does occur, then something is terribly wrong, which is why the message says to check the logs. IMO I don't think it is correct to display "Success" when the swap's `state` is Failed.

guggero: Okay, I see. If you see this as a fall back error message for a case that should never happen, it makes sense :+1:

PR #221, opened 2021-05-06
mod+tools: add dummy import for CLI dependencies

guggero:
With go 1.16 something changed in how the content of go.sum is
calculated. The "go mod tidy" command now removes dependencies for
packages that aren't directly referenced through the main package.
Therefore if we compile another binary (such as the CLI binaries we ship
together with LiT) some of their dependencies might have been pruned
from go.sum. To avoid that, we add dummy imports for those packages
under a build tag that we don't actually use. The "tools" tag seems to
be the idiomatic tag for such tooling depencencies.

Thanks for reporting this, @jamaljsr!


jamaljsr (approved):
tACK 👍 

Thanks for the fix

PR #242, opened 2021-06-30
Register Sidecar Channel

jamaljsr:
This PR introduces some basic functionality to register a [sidecar ticket](https://github.com/lightninglabs/pool/blob/master/docs/sidecar_channels.md#example-flow-command-line-only) with the LND node. A Pool account is not required in order to obtain a channel from the Pool market. The UX can be improved in the future if https://github.com/lightninglabs/pool/issues/263 is completed. We'd be able to display the decoded ticket data in the UI before submitting it to the server.

sidecar-ui

PR #243, opened 2021-06-30
pool+order: display fee estimates in the order form

jamaljsr:
Based on #242 

This PR makes use of the new `QuoteOrder` RPC in Pool to display fee estimates in the order form. The values automatically update then the user modifies the order form fields. The new fee rows also have tooltips to provide more helpful information.

image


app/src/store/views/orderFormView.ts
```diff
@@ -229,6 +239,39 @@ export default class OrderFormView {
...
+    );
+
+    runInAction(() => {
+      this.executionFee = totalExecutionFeeSat;
+      this.worstChainFee = worstCaseChainFeeSat;
+    });
+  }
+
+  /** quote order at most once every 2 seconds when using this func  */
+  quoteOrderThrottled = debounce(this.quoteOrder, 2000);
```
guggero, line 273: This tripped me up a bit. I changed some values and expected the quote to be updated. Was almost ready to enter yet another number when I noticed the number finally updated. Maybe I'm just too impatient.

If it's easy to implement (probably not worth doing if it requires a lot of effort), we could invalidate the quote fields (for example by showing a loading animation) until the next debounce run comes along. That way the user knows something's going to happen soon.

guggero (approved):
tACK, LGTM :100: 

Just one small UX improvement idea that is totally non-blocking.

app/src/store/views/orderFormView.ts
jamaljsr: ah, good point. I like the idea of displaying a loader. I can do that quickly.

jamaljsr: I added the loaders and decreased the debounce to 1 second. Thanks for the suggestion

PR #248, opened 2021-07-16
mod: bump lndclient version to fix loop in

guggero:
Loop in was failing with `Swap error: EOF` because we didn't pull in the
latest version of lndclient that properly handles the single invoice
subscription EOF that is being returned with lnd 0.13.x.


carlaKC (approved):
oof these dependencies will be the death of us all

bhandras (approved):
LGTM

PR #249, opened 2021-07-16
proto: use Docker for compiling protos

guggero:
To remove the need for a dev to install a specific version of protoc we
now use a docker image for compiling the protos.


jamaljsr (approved):
LGTM 🥇 

Great idea to simplify the build process. I'm sure almost all devs have Docker installed already these days.

PR #265, opened 2021-09-15
README: update version to v0.5.1-alpha

guggero:
We create a new minor release for the UI bug fixes and dependency
updates that were merged recently, as well as a fux to the default
macaroon path handling.
We cannot yet update any of the bundled libraries because of their
dependency on lnd v0.14.0.


entone:
Any movement on this?

Issue #160, opened 2020-12-14
Add "stateless remote mode"


guggero:
# Goal

It should be possible to run LiT in a way that it requires no sensitive information to be available to it on disk and also does not store sensitive information to disk itself.
To enter that "stateless remote mode", I'd like to start LiT with just one flag: `--lnd-mode=stateless-remote`

This would just start the UI and then wait for user interaction there. In this mode, the UI would not ask for a password first but for a `lndconnect` format connection string. The connection string contains all information that is needed to connect to the remote `lnd` mode (IP address, TLS cert if necessary and macaroon). Once the string is provided, the LiT binary connects to the given `lnd` node and starts the integrated daemons (Loop, Pool, Faraday).

# Challenges

1. Loop, Pool, Faraday (="the daemons") still require each subserver macaroon to be available as a file to connect to `lnd`. This was fixed in https://github.com/lightninglabs/lndclient/pull/19 but that fix hasn't been pulled into the daemons yet. We can even expand on that and make it possible to pass in a macaroon directly as a hex string instead of needing to pass a file name.
   - Needs a small change to `lndclient` that then needs to be pulled into Pool, Loop, Faraday and LiT.
2. The daemons don't have a stateless init mode yet, meaning they'll leave their own RPC macaroons laying around on the disk (unencrypted).
   - When running in `stateless-remote` mode, the daemons wouldn't use their own macaroons (wouldn't even create a macaroon DB) but instead use the remote `lnd` node to validate their macaroons.
   - Because the macaroons for the daemons will contain permissions and/or RPC URIs that `lnd` doesn't know, this requires a new RPC in `lnd` that allows us to validate a macaroon without checking the permissions part of it.
   - When baking a macaroon, a new flag would be necessary to instruct `lnd` to not complain about permission pairs or URIs it doesn't know itself.
   - Users of the stateless remote mode would need to bake a "super admin macaroon" that gives access to `lnd` **and** all daemons. Basically containing all permissions that exist of all the 4 daemons.
3. The daemons don't have an unlock mechanism yet, meaning their macaroon DB is unencrypted (or to be precise: the encryption uses a default password).
   - Would also be solved by turning off macaroons for the daemons in the stateless remote mode.
4. LiT gives the user access to all RPCs with just a password. That Basic Auth mechanism is turned into a macaroon internally. But it means the LiT backend needs access to all daemon macaroons (and `lnd`'s) internally.
[...]

PR #269, opened 2021-09-28
Stateless integrated mode

orbitalturtle:
This PR puts integrated stateless mode into practice, as is described here: https://github.com/lightninglabs/lightning-terminal/issues/160#issuecomment-888689859

- Allows Terminal to call lnd rpc via a bufconn when in integrated mode.
- Doesn't require a TLS certificate for connecting to lnd in integrated mode.
- Using admin macaroon passed to Terminal, we are able to bake a supermacaroon with permission to access lnd, pool, loop, and faraday.
- Uses lnd to validate that the supermacaroon has the correct permissions with each rpc call.

I think the backend is working. But it seems the UI is failing to come up after entering a password in this mode, which I still need to look into. 
        
Requires:
https://github.com/lightningnetwork/lnd/pull/5777/
https://github.com/lightninglabs/pool/pull/272
https://github.com/lightninglabs/faraday/pull/127
https://github.com/lightninglabs/loop/pull/406


guggero:
Replaced by https://github.com/lightninglabs/lightning-terminal/pull/278, I hope you don't mind me taking this over.

Issue #254, opened 2021-07-23
Allow for deployment in subdirectory


dennisreimann:
**Is your feature request related to a problem? Please describe.**

Currently the app assumes to be deployed on a separate (sub)domain. For the [BTCPay Server integration](https://github.com/btcpayserver/btcpayserver/discussions/2706) I'd like to use our existing FE proxy (nginx) and certificates to provide an easier setup for the user. I imaging deploying the LiT app in a subdirectory of the existing BTCPay domain, like `mybtcpay.com/lit/`.

Right now the asset and API requests target the root `/`, whereas we would need to have them prefixed with `/lit/`.

**Describe the solution you'd like**

I started looking into this a bit and found the option of setting the [`homepage` option](https://create-react-app.dev/docs/deployment/#serving-the-same-build-from-different-paths) for create-react-app. By using `"homepage": "."` as a setting the build uses a relative URL for the asset references, which would solve that part. I'm unsure though whether or not this also works for the API calls. Before proceeding on that front I wanted to inquire if this feature would be wanted.

**Describe alternatives you've considered**

I've started using nginx `sub_filter` directives in our setup to rewrite certain URLs, but this is error prone and brittle.

jamaljsr:
> I'm unsure though whether or not this also works for the API calls. 

The urls for the API calls will remain `https://x.x.x.x/lnrpc.Lightning/GetInfo` when you modify the homepage option. I do not believe there will be any issues with this.

> Before proceeding on that front I wanted to inquire if this feature would be wanted.

Exactly what feature are you proposing? modifying the `homepage` field in this repo?

An alternative option to customize the url path is the `PUBLIC_URL` environment variable when running `yarn build`. Unfortunately, this is also something that cannot be specified at runtime, so you wouldn't be able to use the `lightninglabs/lightning-terminal` docker image as is. You'd need to create a new docker image that uses the one of these methods.

dennisreimann:
The general request would be to allow for the app to be run/deployed in a subdirectory. I think the homepage setting would make it work more universally and solve my current issue with the asset paths not working with a subdir deployment.

As for the API calls I think the config would need to be adapted to not only use the [host](https://github.com/lightninglabs/lightning-terminal/blob/master/app/src/config.ts), but also a potential path (as in our case). 

I'd like to avoid shipping a separate docker image and think we can achieve this via the homepage setting and determining the API base URL at runtime.

jamaljsr:
[...]

PR #272, opened 2021-10-06
Use PUBLIC_URL throughout the app

dennisreimann:
See #254.

### DIFF
--- a/app/src/Routes.tsx
+++ b/app/src/Routes.tsx
@@ -3,6 +3,7 @@ import { Route, Router, Switch } from 'react-router';
 import { useStore } from 'store';
 import Loading from 'components/common/Loading';
 import { Layout } from 'components/layout';
+import { PUBLIC_URL } from 'config';
 
 const LazyAuthPage = React.lazy(() => import('components/auth/AuthPage'));
 const LazyLoopPage = React.lazy(() => import('components/loop/LoopPage'));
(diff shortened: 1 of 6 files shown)


app/src/config.ts
```diff
@@ -7,6 +7,8 @@ export const IS_PROD = process.env.NODE_ENV === 'production';
 // flag to check if the app is running in a a test environment
 export const IS_TEST = process.env.NODE_ENV === 'test';
 
+export const PUBLIC_URL = process.env.PUBLIC_URL;
```
jamaljsr, line 10: To avoid confusion, it would be a good idea to sanitize the value of `PUBLIC_URL` provided by the user. It would be very common to omit or add an unnecessary slash (ex: `/lit/`, `/lit`, or `lit/`). Specifying the wrong one will likely result in a blank page with a cryptic console error message. 

Since you are using paths like `${PUBLIC_URL}/settings` for the routes, I suggest we ensure that `PUBLIC_URL` always has a leading slash and no trailing slash after it's sanitized.

app/src/Routes.tsx
```diff
@@ -12,20 +13,19 @@ const LazySettingsPage = React.lazy(() => import('components/settings/SettingsPa
 
 const Routes: React.FC = () => {
   const { router } = useStore();
-
   return (
     <Suspense fallback={<Loading delay={500} />}>
       <Router history={router.history}>
         <Switch>
-          <Route path="/" exact component={LazyAuthPage} />
+          <Route path={PUBLIC_URL || '/'} exact component={LazyAuthPage} />
```
jamaljsr, line 20: If the `PUBLIC_URL` is sanitized as I mentioned below, then you can simply use 
```
path={`${PUBLIC_URL}/`}
```

jamaljsr (changes requested):
Thanks for getting the ball rolling on this. I tested this locally and have some suggestions.

There is a conditional in [store.ts](https://github.com/lightninglabs/lightning-terminal/blob/master/app/src/store/store.ts#L128) that also needs to be updated, otherwise the page will not properly redirect after login.
```
if (document.location.pathname === '/') {
```

app/src/config.ts
dennisreimann: That's already done by the framework. You can test it with different PUBLIC_URL settings, it'll remove the trailing slash if it is present.

dennisreimann:
> There is a conditional in store.ts that also needs to be updated, otherwise the page will not properly redirect after login.

Thanks for the pointer, I must have overlooked this. 

Will update the PR later today!

guggero:
Cool, thank you! Will take a look first thing tomorrow.

guggero (commented):
Looks good and works well, thanks for the fixes!
Just one instance where the auth wouldn't work if an empty public URL is set.
Also it would be nice if you could squash your commits.

I'm going to add the necessary changes in golang and the docker image/build process to create a secondary image that uses the `PUBLIC_URL=/lit` during the build.

app/src/store/views/appView.ts
```diff
@@ -40,13 +41,13 @@ export default class AppView {
 
   /** Change to the Auth page */
   gotoAuth() {
-    this.goTo('/');
+    this.goTo(PUBLIC_URL);
```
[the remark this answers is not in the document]
dennisreimann: Thanks, fixed in d2dfd3b!

guggero:
> > Also it would be nice if you could squash your commits.
> 
> I think you can do this via the GitHub merge settings, choosing the strategy via the merge button.

Yes, it can be done. We just normally don't do it to preserve the author's original commits. But if you prefer that way I can use squash+merge.

jamaljsr (approved):
LGTM 👍 

Thanks for the updates.

PR #275, opened 2021-11-01
docker: create and push secondary image with path prefix

guggero:
Creates a secondary docker image with the tag `:v0.x.y-alpha-path-prefix` that is auto built and pushed to Docker Hub and assumes that LiT is being run under the custom URL path `/lit`.

An example nginx configuration for setting up a reverse proxy that responds to both LiT `grpc-web` requests and native `lnd` gRPC requests under the same domain (and therefore TLS cert) is included. This mainly addresses setups like BTCPayServer (see https://github.com/btcpayserver/btcpayserver-docker/pull/512).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Set env
         run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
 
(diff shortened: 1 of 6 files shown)


jamaljsr (approved):
Very cool. Tested locally and works as advertised.

LGTM 👍

Roasbeef (approved):
LGTM 🏄‍♂️

PR #276, opened 2021-11-03
multi: update lnd dependency to v0.14.0-beta.rc1

guggero:
Depends on https://github.com/lightninglabs/faraday/pull/133.
Depends on https://github.com/lightninglabs/loop/pull/426.
Depends on https://github.com/lightninglabs/pool/pull/310.

Updates all daemons to be compatible with `lnd v0.14.0-beta.rc1`.


proto/loop.proto
```diff
@@ -15,25 +17,15 @@ service SwapClient {
     point onwards, progress can be tracked via the SwapStatus stream that is
     returned from Monitor().
     */
-    rpc LoopOut (LoopOutRequest) returns (SwapResponse) {
-        option (google.api.http) = {
-            post: "/v1/loop/out"
-            body: "*"
-        };
-    }
+    rpc LoopOut (LoopOutRequest) returns (SwapResponse);
```
jamaljsr, line 20: I am curious. Where are rest paths defined if they are no longer in the proto file?

jamaljsr (approved):
LGTM 👍 

I tested locally, remote and integrated. Didn't run into any issues with Loop UI or Pool UI.

proto/loop.proto
guggero: They were extracted into their own yaml file. The advantage is that we don't need to import the google annotations protos anymore.

PR #278, opened 2021-11-04
Stateless integrated mode

guggero:
Replaces #269.

I started reviewing #269 and noticed a few things that could be simplified.
Taking over the PR to speed things up and hopefully get this in with the next major LiT release that also includes `lnd v0.14.0`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -177,6 +177,11 @@ type Config struct {
 	faradayRemote bool
 	loopRemote    bool
 	poolRemote    bool
+
+	// lndAdminMacaroon is the admin macaroon that is given to us by lnd
+	// over an in-memory connection on startup. This is only set in
+	// integrated lnd mode.
+	lndAdminMacaroon []byte
 }
 
 // RemoteConfig holds the configuration parameters that are needed when running
@@ -219,7 +224,7 @@ type RemoteDaemonConfig struct {
 // lndConnectParams returns the connection parameters to connect to the local
 // lnd instance.
 func (c *Config) lndConnectParams() (string, lndclient.Network, string,
-	string) {
+	string, []byte) {
 
 	// In remote lnd mode, we just pass along what was configured in the
 	// remote section of the lnd config.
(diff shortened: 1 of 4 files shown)


guggero:
I found and fixed the bug. Stateless integrated mode is now working, with the super macaroon :tada: So this is ready for full review and test.

Again, big kudos to @orbitalturtle for tackling this huge multi-part project!

jamaljsr (commented):
LGTM 👍 
The frontend is fully functional from my testing. The only confusion I have is regarding the missing macaroon files for `loop`, `pool`, and `faraday`. Are these intentionally removed? How do we talk to the nodes via CLI? I tried using the `admin.macaroon` in my `.lnd` dir but that didn't work.

guggero:
Thank you for the review, @jamaljsr!

You are totally correct, those macaroons shouldn't be deleted. Only when the wallet is initialized as `stateless` (which is the whole purpose of this PR). So I added a commit that checks what state the wallet is in.

jamaljsr:
Just re-tested using commit https://github.com/lightninglabs/lightning-terminal/pull/278/commits/6700a2cf99a95beed9a727acbfc56038068d9c71 but now im getting the error below.

```
bash-5.1# ls -al ~/.pool/regtest/
total 64
drwxr-xr-x    2 root     root          4096 Nov 18 14:05 .
drwxr-xr-x    4 root     root          4096 Nov 18 14:04 ..
-rw-------    1 root     root         32768 Nov 18 14:05 macaroons.db
-rw-------    1 root     root         32768 Nov 18 14:05 pool.db
-rw-r--r--    1 root     root           144 Nov 18 14:05 pool.macaroon

bash-5.1# pool --rpcserver=localhost:10009 --tlscertpath=/root/.lnd/tls.cert  --macaroonpath=/root/.pool/regtest/pool.macaroon accounts list
[pool] rpc error: code = Unknown desc = macaroon is not valid, returned false

bash-5.1# ls -al ~/.loop/regtest/
total 60
drwxr-xr-x    2 root     root          4096 Nov 18 14:05 .
drwxr-xr-x    4 root     root          4096 Nov 18 14:04 ..
-rw-------    1 root     root         32768 Nov 18 14:05 loop.db
-rw-r--r--    1 root     root           164 Nov 18 14:05 loop.macaroon
-rw-------    1 root     root         32768 Nov 18 14:05 macaroons.db

bash-5.1# loop --rpcserver=localhost:10009 --tlscertpath /root/.lnd/tls.cert  --macaroonpath /root/.loop/regtest/loop.macaroon terms
Loop Out
--------
rpc error: code = Unknown desc = macaroon is not valid, returned false

Loop In
------
rpc error: code = Unknown desc = macaroon is not valid, returned false
bash-5.1#
```

guggero:
Thank you for testing! Oops. I know what the problem is, sorry I should've verified it myself first. Will fix tomorrow.

PR #282, opened 2021-11-18
multi: bump lnd to v0.14.0-beta final, release v0.6.0-alpha

guggero:
This PR bumps all daemon dependencies to their latest tagged versions.

@Roasbeef We bump the minimum compatible version of `lnd` from `v0.12.0-beta` to `v0.13.3-beta`, which IMO is okay because of the CVE.
This will make it possible to run Pool with a version of `lnd` that doesn't understand script enforced channels yet. I think we probably have to live with that, otherwise we'd force all LiT users to upgrade to `lnd v0.14.0-beta`?


jamaljsr (approved):
LGTM 👍 

I noticed that the UI is still using order v2 and not the latest v5 that the CLI uses? Will this cause any problems with the new script enforcements? Should we bump the order version in the UI?

guggero:
Hmm, good question. I think we should bump the order version. Ideally we would use the v5 version if we know the UI is connected to `lnd 0.14.x` or later, otherwise fall back to v4.
I added the bump to v4 to this PR. To actually use v5 we would also need to add the new channel type field along with the version. That probably requires us to add a drop down to the hidden/advanced settings of the order.

PR #289, opened 2021-12-03
Fix error handling with stateless init

guggero:
This is a follow-up commit that fixes an issue introduced with stateless
init where an error would be interpreted incorrectly.

This will bump the version in the README to `v0.6.1-alpha`.


Roasbeef (approved):
LGTM 🪡

jamaljsr (approved):
tACK LGTM 🥇

PR #292, opened 2021-12-09
multi: readonly and admin macaroon session 

ellemouton:
- part of https://github.com/lightninglabs/lightning-node-connect/issues/17
- depends on https://github.com/lightninglabs/lightning-node-connect/pull/27

With this PR, we can create a admin or readonly macaroon session type. This bakes a macaroon for a session and adds it to the Noise `authData` payload so the client can use the mac to authorise requests instead of just the UI password.

After this is in, the UI can start defaulting to creating a readonly session and can add an option for the user to select that they want to create an admin session. Once that is done (ie, there is no way to create a UIPassword session anymore), we can add a DB migration diff that downgrades all existing UIPassword sessions to Readonly sessions


session/tlv.go
```diff
@@ -150,5 +147,46 @@ func DeserializeSession(r io.Reader) (*Session, error) {
...
+	for _, permStr := range perms {
+		perm := strings.Split(permStr, ":")
+		if len(perm) != 2 {
+			return nil, errors.New("invalid macaroon " +
+				"recipe encoding")
+		}
+
+		recipe = append(recipe, bakery.Op{
+			Entity: perm[0],
+			Action: perm[1],
```
ellemouton, line 187: gonna replace all this. just found tlv.MakeDynamicRecord. gonna use that instead. also need to add caveates

ellemouton:
@guggero , i have updated this to include the readonly session type. I think the DB migration diff should be in a separate PR after the UI changes that allow you to pick a session type have been done.

session/tlv.go
```diff
@@ -152,3 +154,178 @@ func DeserializeSession(r io.Reader) (*Session, error) {
...
+		// of the byte slice, followed by the byte slice.
+		for _, perm := range v.Permissions {
+			permBytes := []byte(permissionToString(perm))
+
+			err := tlv.WriteVarInt(w, uint64(len(permBytes)), buf)
+			if err != nil {
+				return err
+			}
+
+			w.Write(permBytes)
```
guggero, line 187: nit: unchecked return value here and a few places below.

session/tlv.go
```diff
@@ -152,3 +154,178 @@ func DeserializeSession(r io.Reader) (*Session, error) {
...
+		if err := tlv.WriteVarInt(w, numPerms, buf); err != nil {
+			return err
+		}
+
+		// Now, we iterate over the permissions and for each, we convert
+		// the permission to a string of form `<entity>:<action>` which
+		// we then convert to a byte slice. We first write the length
+		// of the byte slice, followed by the byte slice.
+		for _, perm := range v.Permissions {
+			permBytes := []byte(permissionToString(perm))
```
guggero, line 180: Nice simplification! I guess we could even have concatenated all permission pairs together, separated by a comma. But this is also good and adds one less restriction on the name/content of the entity/action pair.

session/tlv.go
```diff
@@ -152,3 +154,178 @@ func DeserializeSession(r io.Reader) (*Session, error) {
...
+			err := tlv.WriteVarInt(w, uint64(len(caveat.Id)), buf)
+			if err != nil {
+				return err
+			}
+			w.Write(caveat.Id)
+
+			// Write the length of the caveat verification ID
+			// followed by the verification ID itself.
+			err = tlv.WriteVarInt(
+				w, uint64(len(caveat.VerificationId)), buf,
```
guggero, line 210: Ah I should have mentioned that we're not really going to add third-party caveats any time soon (probably). So the `VerificationId` and `Location` aren't really needed. But let's leave them in now that you've added them, so we're future proof.

session_rpcserver.go
```diff
@@ -82,18 +89,36 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
+
+	case session.TypeMacaroonAdmin:
+		ctx := context.Background()
+		mac, err := s.superMacBaker(
+			ctx, sess.MacaroonRootKey, sess.MacaroonRecipe,
+		)
+		if err != nil {
+			log.Debugf("Not resuming session %x. Could not bake"+
+				"the necessary macaroon: %w",
+				sess.LocalPublicKey.SerializeCompressed(), err)
```
guggero, line 116: return `nil` after logging.

session_rpcserver.go
```diff
@@ -20,33 +21,39 @@ type sessionRpcServer struct {
...
 	}
 
-	expiry = time.Unix(int64(req.ExpiryTimestampSeconds), 0)
+	expiry := time.Unix(int64(req.ExpiryTimestampSeconds), 0)
 	if time.Now().After(expiry) {
 		return nil, fmt.Errorf("expiry must be in the future")
 	}
 
+	var macPerms []bakery.Op
+	if typ == session.TypeMacaroonAdmin {
```
guggero: Unify this `if` over the `typ` with the one above and transform into a nice `switch` statement that combines both tasks?

session_rpcserver.go
```diff
@@ -20,33 +21,39 @@ type sessionRpcServer struct {
...
+		macPerms = getAllPermissions()
```
guggero: I'm still trying to decide if we should define these permissions at _session create time_ or at _bake time_. E.g. if we add a new permission to one of the daemons, does the user expect the session to upgrade automatically (since it's an admin or full read-only session) after restart? Or would the user need to create a new session if something in the permissions changed? The first would be better UX while the second might be a bit more security conscious.

The difference in the code would be: for `TypeMacaroonAdmin` and `TypeMacaroonReadonly` we wouldn't store the recipe in the DB (that would be reserved for the future `TypeMacaroonCustom`) and assemble the list of permissions right before baking the macaroon.

What do you think?

guggero (commented):
Nice, one step closer to getting rid of the UI password :tada:
Did a quick pass and the code looks good! Will do some manual testing on the second pass. In the meantime, we should probably discuss the question of when to assemble the permission list (see inline comment).

session_rpcserver.go
ellemouton: hmmmm yeah I agree with your pro's and con's list of the two options. I guess option 1 is ok given that if the user wants to be more security focused and not let the perms list auto-update, they should use the custom option

ellemouton: option 1 also makes the db migration easier since all we then need to do is change the type & dont need to go populate the recipe field

ellemouton:
Thanks for the review! I have updated accordingly and went for "at bake time" approach.

guggero (commented):
Code looks very good, great feature update!
I tested it out locally and found that the newly baked super macaroons don't work for faraday/loop/pool calls since they're not recognized as super macaroons.
I'll send you a patch offline that you can add to the PR to fix that.

guggero:
Ah, we also discussed adding a DB migration to the session DB to migrate over any existing UI password sessions to admin macaroon ones. But I don't think we can do that just yet, since first we need a way to make the super macaroon work with remote faraday/loop/pool. I have an idea of how to do that, can discuss it offline.

guggero (approved):

session/interface.go
```diff
@@ -39,7 +47,7 @@ type Session struct {
 	ServerAddr      string
 	DevServer       bool
 	MacaroonRootKey uint64
-	Macaroon        *macaroon.Macaroon
```
Roasbeef, line 42: Do we need to do a "migratIon" to properly make this change? Or seems TLV saves the day here...

session/tlv.go
```diff
@@ -58,17 +67,6 @@ func SerializeSession(w io.Writer, session *Session) error {
 		),
 	}
 
-	if session.Macaroon != nil {
```
Roasbeef, line 61: Re backwards compat, would appear we always need to continue reading this. 

With the way TLV works, eventually the new clients just stop writing it and the logic checks to see which of the two have been populated.

session/tlv.go
```diff
@@ -109,12 +115,15 @@ func DeserializeSession(r io.Reader) (*Session, error) {
 		tlv.MakePrimitiveRecord(
 			typeMacaroonRootKey, &session.MacaroonRootKey,
 		),
-		tlv.MakePrimitiveRecord(typeMacaroon, &macaroonBytes),
```
Roasbeef, line 112: See comment re still reading this.

session/tlv.go
```diff
@@ -131,14 +140,7 @@ func DeserializeSession(r io.Reader) (*Session, error) {
 	session.Expiry = time.Unix(int64(expiry), 0)
 	session.ServerAddr = string(serverAddr)
 	session.DevServer = devServer == 1
-
-	if t, ok := parsedTypes[typeMacaroon]; ok && t == nil {
```
Roasbeef, line 135: Here too

session/tlv.go
```diff
@@ -152,3 +154,186 @@ func DeserializeSession(r io.Reader) (*Session, error) {
 
 	return session, nil
 }
+
+// encodeMacaroonRecipe serializes a macaroon recipe so that it can be stored
+// as a tlv value. The serialization will follow the following pattern:
```
Roasbeef, line 159: If we want to make this encoding _more_ flexible. We can us a _nested_ TLV instance here. Sounds scary, but it just means adding a length prefix before the inner TLV blob, and doing TLV as normal within that.

terminal.go
```diff
@@ -141,7 +141,8 @@ type LightningTerminal struct {
 	wg         sync.WaitGroup
 	lndErrChan chan error
 
-	lndClient *lndclient.GrpcLndServices
+	lndClient   *lndclient.GrpcLndServices
+	basicClient lnrpc.LightningClient
```
Roasbeef, line 145: Can likely start to retroactively move some of the new macaroon methods in to a new sub-server, but that's outside the scope of this PR.

subserver_permissions.go
```diff
@@ -55,13 +55,28 @@ func getAllMethodPermissions() map[string][]bakery.Op {
 	return result
 }
 
+// denyList is the list of permission actions that will be denied for readonly
+// macaroons.
+// TODO(elle): the individual daemon libraries should export a readonly
```
Roasbeef, line 60: Which can be done mainly by exporting the perms in the various `*rpc` packages?

Roasbeef: Alternatively, this can function as a white list: iirc they all have "read" in there as the perms right?

cmd/litcli/sessions.go
```diff
@@ -50,6 +50,14 @@ var addSessionCommand = cli.Command{
 			Usage: "set to true to skip verification of the " +
 				"server's tls cert.",
 		},
+		cli.StringFlag{
+			Name: "type",
+			Usage: "session type to be created which will " +
+				"determine the permissions a user has when " +
+				"connecting with the session. Options " +
+				"include readonly|admin",
+			Value: "readonly",
```
Roasbeef, line 59: We'll need to update our docs here as the UI assumes it has all perms (?) atm. 

I think a good slot to settle in would be normal perms w/ certain thresholds.

session/macaroon.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"google.golang.org/protobuf/proto"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+	"gopkg.in/macaroon.v2"
+)
+
+const (
+	// SuperMacaroonRootKeyBitmask is the prefix bitmask we set on a super
+	// macaroon's root key to clearly mark it as such.
+	SuperMacaroonRootKeyBitmask uint64 = 0xFFEEDDCC_00000000
```
Roasbeef, line 17: Where's this come from? 

Couldn't we alternatively go with a prefix? Or maybe I don't understand the implications of the ID: I thought it was just a top-level key we used to read the macaroon secret to validate incoming macaroons?

Roasbeef (changes requested):
Completed an initial pass, main thing I still need to grok is the implications re the identification of the super macaroons. 

I think we should land https://github.com/lightninglabs/lightning-terminal/pull/294 before this, so we ensure we're able to add adequate end to end test coverage.

session/macaroon.go
guggero: We need a way to identify a super macaroon by just reading the macaroon itself. Since a super macaroon is always baked by lnd, we can specify the _ID_ of the root key to use. And that ID (not the root key itself, obviously) is encoded within the macaroon (in a field called `storage_id`). Since the ID is `uint64`, we have 8 bytes to work with and this mask is effectively the prefix we use for any super macaroon.
I guess we could've simplified the code somewhat by generating 4 random bytes, prepending the prefix and only afterwards convert it to `uint64` instead of doing bit mask magic...

ellemouton:
!lightninglabs-deploy mute

ellemouton:
muting this until the dependent PRs are complete. (should probably add an option to lbot so that it can depend on prs outside of the current repo)

session/tlv.go
ellemouton: ok nice! i decided to go all out on the nested-tlv thing after seeing this comment. Let me know if it is too much now 😅

subserver_permissions.go
ellemouton: > Which can be done mainly by exporting the perms in the various *rpc packages?

yes

> Alternatively, this can function as a white list: iirc they all have "read" in there as the perms right?

good idea 👍

ellemouton (commented):
Thanks for the review @Roasbeef 🚀 
See my comments on not reading the macaroon tlv anymore. tl;dr: we never wrote anything there before, so i think we dont need to keep reading from it to remail backwards compatible

ellemouton:
I have pushed a new version based on the latest [dynamic auth pr](https://github.com/lightninglabs/lightning-node-connect/pull/27) & addressed some of the comments in the previous review. Note that for testing: LNC on the client must be updated to the dynamic auth version

guggero (commented):
A couple of comments about the TLV encoding. Other than those, the PR looks good to me!

And I also simplified my commit (cc34d6e8) a bit according to @Roasbeef's comments. See the attached diff. Thanks for applying it to this PR!
[rootkey.diff.txt](https://github.com/lightninglabs/lightning-terminal/files/7995139/rootkey.diff.txt)

session/tlv.go
```diff
@@ -152,3 +167,336 @@ func DeserializeSession(r io.Reader) (*Session, error) {
...
+			}
+
+			err = tlvStream.Encode(&permsTLVBytes)
+			if err != nil {
+				return err
+			}
+
+			// We encode the record with a varint length followed by
+			// the _raw_ TLV bytes.
+			tlvLen := uint64(len(permsTLVBytes.Bytes()))
```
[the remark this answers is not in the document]
ellemouton: but this is encoding an _array_ of tlv streams. So yes, the tlv layer above will know the length of the whole array but not the lengths of the individual tlv streams. Or am i missing something here?

ellemouton (commented):
Thanks @guggero 💪 I applied your patch. and also left a comment on your comment about the mac recipe encoding

ellemouton:
hmmm what 4 byte id do we want to give the in memory super mac? a fixed one or can we just randomly generate?

session/tlv.go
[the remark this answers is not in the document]
guggero: Ah I see. And it looks like this is exactly how we do it with the invoices in `lnd`. Sorry, never mind then :sweat_smile:

guggero (approved):
Nice, LGTM :100: 

> hmmm what 4 byte id do we want to give the in memory super mac? a fixed one or can we just randomly generate? (made it empty for now)

Yeah, a fixed one that equals `prefix || 0` is what we should use IMO, so basically exactly like you currently implemented it :+1:

ellemouton:
This is ready for re-review now that https://github.com/lightninglabs/lightning-node-connect/pull/27 is in 👍

subserver_permissions.go
guggero: I think this might cause an issue with Loop since some calls require a `loop:in` or `loop:out` permission in addition to the `terms:read` one (for example `LoopOutTerms`). But we can adjust that later if we find that we need loop terms in read-only mode.

PR #294, opened 2021-12-14
itest: add integration tests for integrated and remote modes

guggero:
This PR adds a full integration test suite for testing the different proxy and authentication modes of the built-in proxy.


itest/litd_node.go
```diff
@@ -0,0 +1,1704 @@
...
+	}
+	if cfg.RESTPort == 0 {
+		cfg.RESTPort = lntest.NextAvailablePort()
+	}
+	if cfg.ProfilePort == 0 {
+		cfg.ProfilePort = lntest.NextAvailablePort()
+	}
+}
+
+type NodeConfig struct {
```
Roasbeef, line 92: Is there anything we need custom here, or can we just import the current `lnd` instance for something like this?

itest/litd_node.go
```diff
@@ -0,0 +1,1704 @@
...
+		lndArgs = append(lndArgs, "--bitcoin.simnet")
+	case &chaincfg.RegressionNetParams:
+		lndArgs = append(lndArgs, "--bitcoin.regtest")
+	}
+
+	backendArgs := cfg.BackendCfg.GenArgs()
+	lndArgs = append(lndArgs, backendArgs...)
+	lndArgs = append(lndArgs, "--bitcoin.active")
+	lndArgs = append(lndArgs, "--nobootstrap")
+	lndArgs = append(lndArgs, "--debuglevel=debug")
```
Roasbeef, line 215: IIRc, these all need to be prefixed with `lnd`?

itest/litd_node.go
```diff
@@ -0,0 +1,1704 @@
...
+//       ]
+//  },
+//  "chanPoint2": ...
+// }
+type policyUpdateMap map[string]map[string][]*lnrpc.RoutingPolicy
+
+// HarnessNode represents an instance of lnd running within our test network
+// harness. Each HarnessNode instance also fully embeds an RPC client in
+// order to pragmatically drive the node.
+type HarnessNode struct {
```
Roasbeef, line 304: Similar q here as above

itest/litd_mode_integrated_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/stretchr/testify/require"
+)
+
+// testModeIntegrated makes sure that in integrated mode all daemons work
+// correctly.
+func testModeIntegrated(net *NetworkHarness, t *harnessTest) {
+	ctx := context.Background()
+
+	time.Sleep(2 * time.Second)
```
Roasbeef: Is it possible to move this synchronization up to where the tests themselves are actually run?

itest/litd_node.go
```diff
@@ -95,6 +107,15 @@ type NodeConfig struct {
 	AdminMacPath   string
 	ReadMacPath    string
 	InvoiceMacPath string
+	FaradayMacPath string
```
Roasbeef, line 114: Re the prior comment, we could then embed _just_ these new fields within a new struct.

itest/litd_node.go
```diff
@@ -205,7 +248,17 @@ func (cfg NodeConfig) genArgs() []string {
 		lndArgs = append(lndArgs, "--feeurl="+cfg.FeeURL)
 	}
 
-	return lndArgs
+	// All arguments so far were for lnd. Let's namespace them now so we can
+	// add args for the other daemons and LiT itself afterwards.
+	litArgs = append(litArgs, cfg.LitArgs...)
+	for idx := range lndArgs {
```
Roasbeef, line 276: Ah ok here's where the prefix is added

itest/server_harness.go
```diff
@@ -0,0 +1,106 @@
...
+
+const (
+	// DefaultAutogenValidity is the default validity of a self-signed
+	// certificate. The value corresponds to 14 months
+	// (14 months * 30 days * 24 hours).
+	DefaultAutogenValidity = 14 * 30 * 24 * time.Hour
+)
+
+type loopPoolServer struct {
+	auctioneerrpc.UnimplementedChannelAuctioneerServer
```
Roasbeef, line 26: Ohh, very cool hack here to not have to manually declare everything needed to pose as a loop/pool server!

itest/litd_mode_integrated_test.go
```diff
@@ -2,24 +2,279 @@ package itest
...
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+	"gopkg.in/macaroon.v2"
+)
+
+type requestFn func(ctx context.Context,
+	c grpc.ClientConnInterface) (proto.Message, error)
+
+type macaroonFn func(cfg *NodeConfig) string
```
Roasbeef, line 43: Would be nice to have a `godoc` comment here and above.

itest/litd_mode_integrated_test.go
```diff
@@ -2,24 +2,279 @@ package itest
...
+// what we expect them to be.
+func runCertificateCheck(t *testing.T, node *HarnessNode) {
+	// In integrated mode we expect the LiT HTTPS port (8443 by default) and
+	// lnd's RPC port to present the same certificate, namely lnd's TLS
+	// cert.
+	litCerts, err := getServerCertificates(node.Cfg.LitAddr())
+	require.NoError(t, err)
+	require.Len(t, litCerts, 1)
+	require.Equal(
+		t, "lnd autogenerated cert", litCerts[0].Issuer.Organization[0],
```
Roasbeef, line 291: Code was copied over for cert generation so it retains the "lnd" prefix here?

subserver_permissions.go
```diff
@@ -16,6 +16,20 @@ var (
 		"/litrpc.Sessions/ListSessions":  {{}},
 		"/litrpc.Sessions/RevokeSession": {{}},
 	}
+
+	// whiteListedMethods is a map of all lnd RPC methods that don't require
+	// any macaroon authentication.
+	whiteListedMethods = map[string][]bakery.Op{
```
Roasbeef, line 22: 👍

itest/litd_node.go
```diff
@@ -399,9 +425,41 @@ func newNode(cfg NodeConfig) (*HarnessNode, error) {
...
+		)
+		remoteNodeHarness, err = lntest.NewNetworkHarness(
+			harness.Miner, harness.BackendCfg, lndBinary,
+			lntest.BackendBbolt,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		remoteNode, _, _, err = remoteNodeHarness.NewNodeWithSeed(
```
Roasbeef, line 445: Super cool how we're able to re-use everything here

Roasbeef (changes requested):
Great set of initial test coverage! 

This'll enable us to more easily test the final bundle to make sure we haven't introduced any breaking changes across the set of daemons we package, in both the remote and integrated mode. 

Did an initial pass, main question is w.r.t the first few commits: is there anything fundamental preventing us from just using the `lnd/lntest` package as is and not copying over all the structs? I ask as there're a few updates to the `lntest` package itself queued up as `lnd` PRs, and ideally we can just update our commit hash here vs needing to manually cherry-pick the set of changes.

ellemouton (commented):
tAck 🚀  Awesome auth coverage!

One thing i noticed was that `make itest-only` only works if `make build-itest` was run previously (ie the btcd-itest, litd-itest & lnd-itest binaries have been built already). So maybe building these binaries should be extracted into their own make file func so that both those methods can call it?

itest/litd_mode_integrated_test.go
guggero: No, in integrated mode lnd's certificate is used for both ports. In remote mode litd's certificate is used.

guggero:
Thanks for the review and feedback. I'm going to look into back-porting some of the custom stuff we need to make a harness for `litd`. Not sure if making the lnd harness more generic is worth the effort vs. having a copy of some of the code.

Roasbeef:
Is this ready for another round of review? Want to make sure we're able to have proper itest coverage for the set of pending PRs that modify authentication.

guggero:
Yes, it's ready for another round. But merging is now blocked by the release of `lnd v0.14.2-beta` because we need to include https://github.com/lightningnetwork/lnd/pull/6139 and don't want to bump `lnd` to a non-tagged version.

guggero:
I refactored this PR to re-use some of the structs from `lntest`. The rest is too lnd specific to be re-used without a huuuge refactor in the lnd repo.

guggero:
> tACK rocket although , my comment from before remain. Cant run the tests with `make itest-only` without first having run `make itest` which does the whole app build first. This is because `make itest` doesnt build the required btcd & lnd binaries by itself. Maybe it should?

The `itest-only` goal is only meant as a shortcut if you changed something in the tests only and don't want to re-compile the whole UI and dependencies. You need to run `make itest` initially, the same way we do it in `lnd`.

Roasbeef (approved):
LGTM 🐡

PR #298, opened 2021-12-21
pool: display lease durations as human readable times

### LINKED ISSUE #296: Pool UI: Markets are sorted incorrectly and inactive market is selected by default
pool-incorrect-market-sorting

Since adding the new market durations the `12096` block market is selected by default, since it's being sorted alphabetically (instead of numerically). This also causes the "last batches" panel to be empty since there currently never was a match in the `12096` block market.

To see the last batches, a user must manually click the `2016` market button.

jamaljsr:
Closes #296 

This PR updates the labels for the Lease Durations to be more user friendly. Instead of showing `12096`, it now displays `3 months`. The durations also are now sorted numerically ascending.

### DIFF
--- a/app/src/store/views/orderFormView.ts
+++ b/app/src/store/views/orderFormView.ts
@@ -1,4 +1,4 @@
-import { entries, makeAutoObservable, runInAction } from 'mobx';
+import { makeAutoObservable, runInAction } from 'mobx';
 import {
   DurationBucketState,
   NodeTier,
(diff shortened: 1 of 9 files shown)


app/src/store/views/orderFormView.ts
```diff
@@ -93,9 +94,9 @@ export default class OrderFormView {
   /** the markets currently open or accepting orders */
   get marketsAcceptingOrders() {
     const { MARKET_OPEN, ACCEPTING_ORDERS } = DurationBucketState;
-    return entries(this._store.batchStore.leaseDurations)
-      .map(([duration, state]) => ({ duration, state }))
-      .filter(({ state }) => state === MARKET_OPEN || state === ACCEPTING_ORDERS);
+    return this._store.batchStore.sortedDurations.filter(
```
guggero, line 97: nit: The import `entries` is now unused (linter is complaining).

guggero (approved):
LGTM, much easier to read and parse that way :tada:

app/src/store/views/orderFormView.ts
jamaljsr: Ah, good catch! Fixed in https://github.com/lightninglabs/lightning-terminal/pull/298/commits/e52a2cb67bd878f7f51cf52456044a5795a85fda

PR #300, opened 2021-12-23
multi: revoke expired sessions

ellemouton:
Ensure that expired sessions are revoked.

- when `resumeSession` is called but session is expired, it is revoked. 
- if the session is not yet expired, a timeout goroutine is used to ensure that it is stoped and revoked after expiry.
- the first commit adds the ability to subscribe to a session being stopped manually so that the above goroutine can exit when it is no longer needed

### DIFF
--- a/session/server.go
+++ b/session/server.go
@@ -19,7 +19,14 @@ type GRPCServerCreator func(opts ...grpc.ServerOption) *grpc.Server
 type mailboxSession struct {
 	server *grpc.Server
 
-	wg sync.WaitGroup
+	wg   sync.WaitGroup
+	quit chan struct{}
+}
+
+func newMailboxSession() *mailboxSession {
+	return &mailboxSession{
+		quit: make(chan struct{}),
+	}
 }
 
 func (m *mailboxSession) start(session *Session,
@@ -62,6 +69,7 @@ func (m *mailboxSession) run(mailboxServer *mailbox.Server) {
 
 func (m *mailboxSession) stop() {
 	m.server.Stop()
+	close(m.quit)
 	m.wg.Wait()
 }
 
@@ -82,7 +90,9 @@ func NewServer(serverCreator GRPCServerCreator) *Server {
 	}
 }
 
-func (s *Server) StartSession(session *Session, authData []byte) error {
+func (s *Server) StartSession(session *Session, authData []byte) (chan struct{},
+	error) {
+
 	s.activeSessionsMtx.Lock()
 	defer s.activeSessionsMtx.Unlock()
 
@@ -91,11 +101,13 @@ func (s *Server) StartSession(session *Session, authData []byte) error {
 
 	_, ok := s.activeSessions[id]
 	if ok {
-		return fmt.Errorf("session %x is already active", id[:])
+		return nil, fmt.Errorf("session %x is already active", id[:])
 	}
 
-	s.activeSessions[id] = &mailboxSession{}
-	return s.activeSessions[id].start(session, s.serverCreator, authData)
+	sess := newMailboxSession()
+	s.activeSessions[id] = sess
+
+	return sess.quit, sess.start(session, s.serverCreator, authData)
 }
 
 func (s *Server) StopSession(localPublicKey *btcec.PublicKey) error {
--- a/session_rpcserver.go
+++ b/session_rpcserver.go
@@ -105,7 +105,8 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
 	}
 
 	authData := []byte("Authorization: Basic " + s.basicAuth)
-	return s.sessionServer.StartSession(sess, authData)
+	_, err := s.sessionServer.StartSession(sess, authData)
+	return err
 }
 
 // ListSessions returns all sessions known to the session store.
--- a/session_rpcserver.go
+++ b/session_rpcserver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec"
@@ -20,6 +21,17 @@ type sessionRpcServer struct {
 
 	db            *session.DB
 	sessionServer *session.Server
+
+	quit chan struct{}
+
+	stopOnce sync.Once
+}
+
+// stop cleans up any sessionRpcServer resources.
+func (s *sessionRpcServer) stop() {
+	s.stopOnce.Do(func() {
+		close(s.quit)
+	})
 }
 
 // AddSession adds and starts a new Terminal Connect session.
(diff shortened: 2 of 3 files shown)


ellemouton:
will request review in the new year :)

session/server.go
```diff
@@ -71,14 +71,17 @@ type Server struct {
 	activeSessions    map[sessionID]*mailboxSession
 	activeSessionsMtx sync.Mutex
 
+	sessionClosedSubs map[sessionID]chan struct{}
```
guggero, line 74: Should we instead add a `quit` chan to the `mailboxSession` so we don't need another map?

session_rpcserver.go
```diff
@@ -88,12 +95,49 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
-	return s.sessionServer.StartSession(sess, authData)
+	if err = s.sessionServer.StartSession(sess, authData); err != nil {
+		return err
+	}
+
+	go func() {
+		sessionClosedSub := s.sessionServer.SubscribeToSessionClosed(
+			pubKey,
+		)
+		go func() {
```
guggero, line 116: Do we really need a second goroutine here? Can't we just continue in the same one?

session_rpcserver.go
```diff
@@ -88,12 +95,49 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
+			ticker := time.NewTimer(time.Until(sess.Expiry))
+			defer ticker.Stop()
+
+			select {
+			case <-sessionClosedSub:
+			case <-ticker.C:
```
guggero, line 122: Also select on session RPC server quit?

guggero (commented):
Nice, this should also reduce the load on our servers somewhat if we stop expired sessions :tada:

session_rpcserver.go
ellemouton: oops. mistake 🙈

session/server.go
ellemouton: indeed. dont know why i didnt think of that. Thanks!

session_rpcserver.go
ellemouton: cool, will do.

ellemouton:
Thanks @guggero! updated 👍

session/server.go
```diff
@@ -62,6 +63,7 @@ func (m *mailboxSession) run(mailboxServer *mailbox.Server) {
 
 func (m *mailboxSession) stop() {
 	m.server.Stop()
+	close(m.quit)
```
Roasbeef, line 66: Is this closed in the simple case of a mailbox server restart? Haven't seen an instance where we'll re-establish our connections when that flaps...

terminal.go
```diff
@@ -838,6 +839,7 @@ func (g *LightningTerminal) shutdown() error {
 		}
 	}
 
+	g.sessionRpcServer.stop()
```
Roasbeef, line 842: Doesn't' look to be protected against a double chan close.

session_rpcserver.go
```diff
@@ -95,12 +102,49 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
 		return nil
 	}
 
 	authData := []byte("Authorization: Basic " + s.basicAuth)
-	return s.sessionServer.StartSession(sess, authData)
+	if err = s.sessionServer.StartSession(sess, authData); err != nil {
+		return err
+	}
+
+	go func() {
```
Roasbeef, line 119: Should be tracked with a wait group.

Roasbeef (changes requested):

terminal.go
ellemouton: cool, added a `sync.Once` to protect against that

session/server.go
```diff
@@ -125,3 +129,20 @@ func (s *Server) Stop() {
...
+
+	s.activeSessionsMtx.Lock()
+	defer s.activeSessionsMtx.Unlock()
+
+	var id sessionID
+	copy(id[:], localPublicKey.SerializeCompressed())
+
+	session, ok := s.activeSessions[id]
+	if !ok {
+		return nil
```
guggero, line 144: To avoid this nil channel from being returned, should we instead return `chan struct{}` in `StartSession`?

session_rpcserver.go
```diff
@@ -100,12 +108,52 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
+		if err := s.db.RevokeSession(pubKey); err != nil {
+			return fmt.Errorf("error revoking session: %v", err)
+		}
+
 		return nil
 	}
 
 	authData := []byte("Authorization: Basic " + s.basicAuth)
-	return s.sessionServer.StartSession(sess, authData)
+	if err = s.sessionServer.StartSession(sess, authData); err != nil {
```
guggero, line 121: This is the only place where we call `StartSession` so maybe it's a bit cleaner to return the `sessionClosedSub` here? See previous comment.

session_rpcserver.go
```diff
@@ -84,6 +85,13 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
 // resumeSession tries to start an existing session if it is not expired, not
 // revoked and a LiT session.
 func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
+	localPubKey := sess.LocalPublicKey.SerializeCompressed()
+
+	pubKey, err := btcec.ParsePubKey(localPubKey, btcec.S256())
```
guggero, line 90: Why do we parse the public key here again? Shouldn't this instead just be a nil check?
Or even just:
```
pubKey := sess.LocalPublicKey
pubKeyBytes := pubKey.SerializeCompressed()
```

guggero (approved):
LGTM, only nits left :tada:

session_rpcserver.go
ellemouton: ya i have no idea why i did that 🙈

session_rpcserver.go
ellemouton: great idea!

ellemouton (commented):
updated 🚀

guggero (approved):

jamaljsr (approved):
tACK LGTM 👍 

The only issue I noticed, which may need to be fixed in a separate PR, is now that the sessions revoke in realtime, the UI just stops receiving data and doesn't know that the session is no longer valid. When the app calls `wasmClientInvokeRPC()`, the callback function is never executed so the UI shows infinite loading icons. It would be great to return an error if the session is no longer active.

This issue never presented itself before because it was not possible to login to the app with a revoked session.

guggero:
Thank you for testing, @jamaljsr! You are right, currently we cannot distinguish between LiT not being started or being unreachable from a session being invalid.
Though I think this might be a bit more tricky to solve indeed, so I agree this should be done in a separate PR. I'm going to create an issue so we can track it in a next sprint.

PR #306, opened 2022-01-05
docs: Match and pass /litrpc.* locations in nginx

dongcarl:
```
Also escape the "." for more precise regex matching
```

Not matching /litrpc.* locations results in lost grpc calls, which looks like this to the user:
image


guggero (approved):
Thanks for the fix and simplification! LGTM :tada:

jamaljsr (approved):
LGTM 👍

PR #308, opened 2022-01-07
multi: add macaroon service and remove ui password use in litcli

ellemouton:
Add a macaroon service to litd for its session rpc server and use this to validate incoming LitURI calls. Remove ui password requirement from litcli.


config.go
```diff
@@ -119,6 +119,16 @@ var (
 		lndDefaultConfig.DataDir, defaultLndChainSubDir,
 		defaultLndChain, DefaultNetwork, defaultLndMacaroon,
 	)
+
+	// DefaultMacaroonFilename is the default file name for the
+	// autogenerated lit macaroon.
+	DefaultMacaroonFilename = "lit.macaroon"
```
Roasbeef, line 125: IIRC, we'll need to also add these caveats to the super macaroon, correct?

config.go
```diff
@@ -141,6 +151,8 @@ type Config struct {
 	LitDir     string `long:"lit-dir" description:"The main directory where LiT looks for its configuration file. If LiT is running in 'remote' lnd mode, this is also the directory where the TLS certificates and log files are stored by default."`
 	ConfigFile string `long:"configfile" description:"Path to LiT's configuration file."`
 
+	MacaroonPath string `long:"macaroonpath" description:"Path to write the macaroon for pool's RPC and REST services if it doesn't exist."`
```
Roasbeef, line 154: pool?

session_rpcserver.go
```diff
@@ -11,6 +12,21 @@ import (
 	"github.com/lightninglabs/lightning-node-connect/mailbox"
 	"github.com/lightninglabs/lightning-terminal/litrpc"
 	"github.com/lightninglabs/lightning-terminal/session"
+	"github.com/lightninglabs/lndclient"
+	"github.com/lightningnetwork/lnd/keychain"
+)
+
+var (
+	// TODO(elle): Temp values. change these before landing pr.
+	nBytes, _ = hex.DecodeString(
```
Roasbeef, line 21: Should instead be a key derived from the user's seed?

terminal.go
```diff
@@ -738,12 +738,12 @@ func (g *LightningTerminal) ValidateMacaroon(ctx context.Context,
...
-		wrap := fmt.Errorf("invalid basic auth")
-		_, err := g.rpcProxy.convertBasicAuth(ctx, fullMethod, wrap)
-		if err != nil {
+		if err := g.sessionRpcServer.macaroonService.ValidateMacaroon(
+			ctx, requiredPermissions, fullMethod,
+		); err != nil {
 			return &proxyErr{
 				proxyContext: "lit",
-				wrapped: fmt.Errorf("invalid auth: %v",
+				wrapped: fmt.Errorf("invalid macaroon: %v",
```
Roasbeef, line 746: Can use `%w` here

config.go
ellemouton: oops. fixed

config.go
ellemouton: yes. the perms in `litPermissions` are already added to the super mac and it is populated in the next commit. (i think this is what you mean?)

session_rpcserver.go
ellemouton: from the users seed in LND? yes this is the idea. So we have a static pubkey here (need to still gen a proper one) and then we derive a shared secret with lnd using this static key and a seed in LND using the `DeriveSharedSecret` endpoint. the macaroon service does this. 
Is this what you mean or is there a litd user seed that you mean?

session_rpcserver.go
```diff
@@ -3,6 +3,7 @@ package terminal
 import (
 	"context"
 	"fmt"
+	"google.golang.org/grpc"
```
guggero, line 7: nit: move to block with external dependencies.

session_rpcserver.go
guggero: Here's the public key derived with `numsgen` from LNC with the phrase "Shared Secret": 0215b5a3e0ef58b101431e1e513dd017d1018b420bd2e89dcd71f45c031f00469e
Should probably mention the phrase and link to https://github.com/lightninglabs/lightning-node-connect/tree/master/mailbox/numsgen

session_rpcserver.go
```diff
@@ -11,6 +12,21 @@ import (
...
+)
+
+var (
+	// TODO(elle): Temp values. change these before landing pr.
+	nBytes, _ = hex.DecodeString(
+		"0254a58cd0f31c008fd0bc9b2dd5ba586144933829f6da33ac4130b555fb5ea32c",
+	)
+	N, _ = btcec.ParsePubKey(nBytes, btcec.S256())
+
+	keyLocator = &keychain.KeyLocator{
```
guggero, line 26: We currently use key family `99` for Loop and `220` for Pool. It doesn't really matter what value we choose. I vote for family 21 and index 0.

cmd/litcli/main.go
```diff
@@ -102,11 +97,12 @@ func fatal(err error) {
 
 func getClient(ctx *cli.Context) (litrpc.SessionsClient, func(), error) {
 	rpcServer := ctx.GlobalString("rpcserver")
-	tlsCertPath, err := extractPathArgs(ctx)
+	tlsCertPath, macPath, err := extractPathArgs(ctx)
 	if err != nil {
 		return nil, nil, err
 	}
-	conn, err := getClientConn(rpcServer, tlsCertPath)
+	fmt.Println(macPath)
```
guggero, line 104: nit: stray `fmt.Println()`

cmd/litcli/main.go
```diff
@@ -1,32 +1,29 @@
...
 	"github.com/lightningnetwork/lnd/lncfg"
+	"github.com/lightningnetwork/lnd/macaroons"
 	"github.com/urfave/cli"
-	"golang.org/x/term"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
-	"google.golang.org/grpc/metadata"
+	"gopkg.in/macaroon.v2"
+	"io/ioutil"
+	"os"
```
guggero, line 18: nit: imports. Sigh, I wish there was a way to tell GoLand to do this split automatically. This happens to me all the time.

guggero (commented):
Great next step towards getting rid of the UI password!
Code looks very good, did an initial pass.

Will do some manual testing next.

cmd/litcli/main.go
guggero: TIL there is :)

ellemouton:
this is pretty much ready for re-review. Just waiting on tagged releases for pool & faraday so the go.mod here can be bumped to include the macaroon service changes in those repos.

guggero:
Can you rebase this when you have time please so I can do another pass?

ellemouton:
Mmmm ok I rebased but need to still figure out how to update the itests to use the macaroon path instead of ui password....

rpc_proxy.go
```diff
@@ -537,7 +537,7 @@ func (p *rpcProxy) basicAuthToMacaroon(basicAuth, requestURI string,
 		}
 
 	case isLitURI(requestURI):
-		return EmptyMacaroonBytes, nil
+		macPath = p.cfg.MacaroonPath
```
guggero, line 540: nit: remove the now unused variable `EmptyMacaroonBytes`.

terminal.go
```diff
@@ -749,12 +749,12 @@ func (g *LightningTerminal) ValidateMacaroon(ctx context.Context,
 		}
 
 	case isLitURI(fullMethod):
-		wrap := fmt.Errorf("invalid basic auth")
-		_, err := g.rpcProxy.convertBasicAuth(ctx, fullMethod, wrap)
-		if err != nil {
+		if err := g.sessionRpcServer.macaroonService.ValidateMacaroon(
```
guggero, line 752: This should also be guarded by `if !g.sessionRpcServerStarted {` like the other sub servers above.

itest/litd_mode_integrated_test.go
```diff
@@ -198,16 +201,16 @@ var (
...
 		requestFn:  litRequestFn,
 		// In some test cases we actually expect some sessions, so we
 		// don't explicitly check for an empty array but just the
 		// existence of the array in the response.
 		successPattern:              "\"sessions\":[",
-		supportsMacAuthOnLndPort:    false,
-		supportsMacAuthOnLitPort:    false,
-		supportsUIPasswordOnLndPort: true,
-		supportsUIPasswordOnLitPort: true,
+		supportsMacAuthOnLndPort:    true,
```
guggero, line 210: All these boolean variables can now be removed since they should always be `true, true, false, true` respectively with litrpc having caught up with the rest of the sub RPC servers.

itest/litd_mode_integrated_test.go
```diff
@@ -198,16 +201,16 @@ var (
...
+		supportsMacAuthOnLitPort:    true,
+		supportsUIPasswordOnLndPort: false,
+		supportsUIPasswordOnLitPort: false,
```
guggero, line 213: That last variable should still be `true`. See comment above about the variable values now all being the same across the different RPC sub servers.

guggero (approved):
tACK, LGTM. Only nits and itest changes required. Also made sure the UI still works with the UI password.

rpc_proxy.go
ellemouton: cool. Removed in next commit since it is still used in litcli at this stage

ellemouton (commented):
Thanks @guggero 🚀  updated 👍

ellemouton:
erg the itests are failing now. dont think i quite understand what do want/dont want on the lnd vs litd port...

ellemouton:
nvm, i found the issue. will push up in a minute

guggero (dismissed):

terminal.go
```diff
@@ -200,46 +199,28 @@ func (g *LightningTerminal) Run() error {
...
-				),
-				grpc.ChainUnaryInterceptor(
-					g.rpcProxy.UnaryServerInterceptor,
-				),
-				grpc.UnknownServiceHandler(
-					grpcProxy.TransparentHandler(
-						// Don't allow calls to litrpc.
-						g.rpcProxy.makeDirector(false),
-					),
+	g.sessionRpcServer, err = newSessionRPCServer(&sessionRpcServerConfig{
```
guggero, line 202: Error needs to be handled here!

session_rpcserver.go
```diff
@@ -11,16 +11,18 @@ import (
...
 type sessionRpcServer struct {
 	litrpc.UnimplementedSessionsServer
 
-	cfg           *sessionRpcServerConfig
-	db            *session.DB
-	sessionServer *session.Server
+	cfg             *sessionRpcServerConfig
+	db              *session.DB
+	sessionServer   *session.Server
+	macaroonService *lndclient.MacaroonService
```
guggero, line 25: After starting with the accounts feature, I think it would make sense to keep the macaroon service as part of the main terminal struct itself. Because the service will also validate macaroons for the accounts RPC service.

guggero:
Sorry if I'm spamming this PR with comments :sweat_smile: Just leaving some thoughts here so we don't forget.
I think with this PR we can also switch over the LNC connection to use the super macaroon instead of the UI password (also in the itests).

ellemouton:
will update this today 👍

terminal.go
ellemouton: good catch!

session_rpcserver.go
ellemouton: cool, will move it out 👍

ellemouton:
> I think with this PR we can also switch over the LNC connection to use the super macaroon instead of the UI password (also in the itests).

oh are you saying i should add that in this PR?

ellemouton:
updated @guggero 😊

itest/litd_mode_integrated_test.go
```diff
@@ -618,7 +620,7 @@ func runLNCAuthTest(t *testing.T, hostPort, uiPassword, tlsCertPath string,
 	// endpoint, unless it is explicitly disallowed (we currently don't want
 	// to support creating more sessions through LNC until we have all
 	// macaroon permissions properly set up).
-	resp, err := makeRequest(ctxm, rawLNCConn)
+	resp, err := makeRequest(ctxt, rawLNCConn)
```
guggero, line 623: Shouldn't this still be `ctxm`?

session_rpcserver.go
```diff
@@ -181,33 +180,29 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
 
+	readOnly := sess.Type == session.TypeMacaroonReadonly
+	mac, err := s.cfg.superMacBaker(
+		context.Background(), sess.MacaroonRootKey,
+		&session.MacaroonRecipe{
+			Permissions: GetAllPermissions(readOnly),
+		},
+	)
+	if err != nil {
+		log.Debugf("Not resuming session %x. Could not bake"+
```
guggero, line 199: nit: missing space in error message before line break.

guggero (approved):
Two small things, otherwise looks good to me. Nice work :100:

session_rpcserver.go
ellemouton: lazer eyes you have there

itest/litd_mode_integrated_test.go
ellemouton: as far as I am aware, no. 

`ctxm` contains the litd macaroon. Not the session macaroon. So `ctxm` is used for creating the session yes since that command is using the litd API. The session macaroon is already injected into the request via the authdata which is received during the handshake

guggero: Aaah, you're totally right.

guggero (approved):

ellemouton:
@Roasbeef , could you take a quick look at this pr? I think we are just waiting for your ACK here since you previously left some review comments on it

itest/litd_mode_integrated_test.go
```diff
@@ -335,9 +335,8 @@ func testModeIntegrated(net *NetworkHarness, t *harnessTest) {
 			endpoint := endpoint
 			tt.Run(endpoint.name+" lit port", func(ttt *testing.T) {
 				runLNCAuthTest(
-					ttt, cfg.LitAddr(), cfg.UIPassword,
-					cfg.TLSCertPath,
-					endpoint.requestFn,
+					ttt, cfg.LitAddr(), cfg.TLSCertPath,
```
Roasbeef, line 338: Typo in commit message: depricate -> deprecate

session_rpcserver.go
```diff
@@ -181,33 +180,29 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
-			ctx, sess.MacaroonRootKey, &session.MacaroonRecipe{
-				Permissions: GetAllPermissions(readOnly),
-			},
-		)
-		if err != nil {
-			log.Debugf("Not resuming session %x. Could not bake"+
-				"the necessary macaroon: %w", pubKeyBytes, err)
-			return nil
-		}
+	if sess.Type != session.TypeMacaroonAdmin &&
```
Roasbeef, line 183: Below we check `TypeMacaroonReadonly`, so presumably we want to allow it here as well?

Roasbeef (changes requested):
PR is looking pretty good! Only one main comment re allowing a read only session type when we go to validate the session type itself.

terminal.go
```diff
@@ -775,12 +775,17 @@ func (g *LightningTerminal) ValidateMacaroon(ctx context.Context,
 		}
 
 	case isLitURI(fullMethod):
-		wrap := fmt.Errorf("invalid basic auth")
-		_, err := g.rpcProxy.convertBasicAuth(ctx, fullMethod, wrap)
-		if err != nil {
+		if !g.macaroonServiceStarted {
```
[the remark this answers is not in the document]
ellemouton: LitURI calls are not allowed through TW.

session_rpcserver.go
ellemouton: see line 184?

Roasbeef (approved):
LGTM 🥟

guggero (approved):

PR #309, opened 2022-01-07
loop: increase maximum miner fee to reduce swap failures

### LINKED ISSUE #303: loop out stuck in "initiated"
Hello, I started a loop out from my channel to send some sats onchain and its stuck.. 

Thats what i See in LIT
Can anyone help me?

I just went straight with the defaults.. No channels picked.. 
Thanks.

IMAGE 2021-12-27 14:29:40

### LINKED ISSUE #307: Match max miner setting to CLI config in Loop Out
The CLI will set the max miner fee to 250x the current fee level and this or higher is advisable for all clients

The max miner fee is currently intended as only a sanity check and not a fee control system. Setting the max miner fee to the current fee level will have adverse impact on the Loop Out experience in the case that the chain backend fee estimation increases over the duration of the swap.

The GUI for Loop Out should match the API access pattern of the CLI to maintain consistency in behavior.

jamaljsr:
This PR increases the maximum miner fee to mimic the behavior of the `loop` CLI [main.go](https://github.com/lightninglabs/loop/blob/master/cmd/loop/main.go#L263). The `MaxMinerFee` value should be set higher in order to reduce the chances of a swap failure when onchain fees increase after the swap is initiated.

Closes #307, closes #303

### DIFF
--- a/app/src/api/loop.ts
+++ b/app/src/api/loop.ts
@@ -91,7 +91,8 @@ class LoopApi extends BaseApi<LoopEvents> {
     const req = new LOOP.LoopInRequest();
     req.setAmt(amount.toString());
     req.setMaxSwapFee(quote.swapFee.toString());
-    req.setMaxMinerFee(quote.minerFee.toString());
+    // mimic the behavior of loop CLI to avoid swap failures due to rising chain fees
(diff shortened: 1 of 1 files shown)


guggero (approved):
utACK, LGTM :tada:

PR #320, opened 2022-02-08
session: add 2 min mailbox ping

ellemouton:
Ensure that the mailbox grpc connection sends a ping every 2 mins to
ensure that the connection remains open.

### DIFF
--- a/session/server.go
+++ b/session/server.go
@@ -4,12 +4,14 @@ import (
 	"crypto/tls"
 	"fmt"
 	"sync"
+	"time"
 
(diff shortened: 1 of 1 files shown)


guggero (approved):
tACK! Tested against staging and observed pings and ACKs from the server for at least 6 ping cycles. So we know the server isn't going to complain about too many pings in a short interval (the hard coded limit of "ping strikes" is 2, so seeing 6 successful pings and the connection still being alive proves this is working).

PR #334, opened 2022-02-23
cmd/litcli: change default session expiry to 3 months

guggero:
The default expiry of one hour was way too short for anything practical
and we increase it to 90 days which is almost three months.
The user can still override this value manually by specifying the flag.

### DIFF
--- a/cmd/litcli/sessions.go
+++ b/cmd/litcli/sessions.go
@@ -9,6 +9,12 @@ import (
 	"github.com/urfave/cli"
 )
 
+var (
+	// defaultSessionExpiry is the default time a session can be used for.
+	// The current value evaluates to 90 days.
+	defaultSessionExpiry = time.Hour * 24 * 90
+)
+
 var sessionCommands = []cli.Command{
 	{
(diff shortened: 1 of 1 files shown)


jamaljsr (approved):
tACK LGTM 👍

guggero:
Hmm, I think I missed or misunderstood that part. So the expiry was always thought of as "maximum time until first use/pairing" and not as "make sure a session cannot be used indefinitely by adding a maximum lifetime"? Or would we instead implement the maximum lifetime (once the second handshake is implemented) through a macaroon caveat? Or would a session be "valid until manually revoked"?

Roasbeef:
> Hmm, I think I missed or misunderstood that part. 

Ah no it's me that's misunderstanding, you're correct in that after we introduce the _second_ handshake, this all still applies.

Roasbeef:
The only other change that'll come along with the second handshake is that we'd disallow using the pairing phrase again (naive impl would be just to delete/forget it on the server side).

ellemouton (approved):
tACK 🚀

PR #342, opened 2022-03-16
multi: persist remote static key for handshakev2

ellemouton:
In this commit, we update the go mod to point to the version of LNC that
contains the logic for handshake version 2. This requires that we pass
in a call back that LNC can call to persist the remote static key once
it is received. This then needs to be provided each time we start up the
session again.

depends on https://github.com/lightninglabs/lightning-node-connect/pull/35

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,4 +34,6 @@ require (
 	gopkg.in/macaroon.v2 v2.1.0
 )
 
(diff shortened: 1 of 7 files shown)


guggero:
This PR was merged by accident. I reverted commit a2f12450 in 46e9424b (but I kept 53b00a4a since that was useful on its own).
@ellemouton can you please rebase your branch and open a new PR?

PR #345, opened 2022-03-22
Add support for read-only macaroons

itsrachelfish:
This PR updates the `AddSession` component, allowing users to generate either admin or read-only macaroons for connecting to their node.

lightning-node-connect

### DIFF
--- a/app/src/components/connect/AddSession.tsx
+++ b/app/src/components/connect/AddSession.tsx
@@ -2,11 +2,13 @@ import React, { useCallback, useState } from 'react';
 import { observer } from 'mobx-react-lite';
 import styled from '@emotion/styled';
 import { usePrefixedTranslation } from 'hooks';
+import * as LIT from 'types/generated/lit-sessions_pb';
 import { MAX_DATE } from 'util/constants';
 import { useStore } from 'store';
 import { Button, Column, HeaderFour, Row } from 'components/base';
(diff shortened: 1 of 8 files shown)


app/src/components/connect/AddSession.tsx
```diff
@@ -32,18 +45,25 @@ const AddSession: React.FC<Props> = ({ primary }) => {
...
+        ? LIT.SessionType.TYPE_MACAROON_ADMIN
+        : LIT.SessionType.TYPE_MACAROON_READONLY;
+
+    const session = await sessionStore.addSession(label, sessionType, MAX_DATE);
+
     if (session) {
       setLabel('');
       setEditing(false);
     }
   }, [label]);
```
jamaljsr, line 64: I ran into a small issue. First I created an Admin session followed by a Read Only session without reloading the page. The second session was created but with the permissions set to Admin. 

The problem is that this `useCallback` is memoized based on the values passed into the dependencies array. You'll need to add `permissions` into the array to resolve the problem.

jamaljsr (changes requested):
Great work! This PR is virtually flawless. 

I just ran into a minor bug which is easy to fix. It's described in the comment below.

jamaljsr (approved):
Thanks for the fixes. This PR looks good to go.

PR #350, opened 2022-04-03
deps: update react-scripts to v5.0.0

jamaljsr:
This PR updates the `react-scripts` npm dependency along with a few other related packages. I also needed to resolve some breaking changes due to the upgrades of `webpack` and `jest`. 

Replaces the following PRs: 
- #317 
- #324 
- #336

The dependabot PRs above apply updates to the dependencies of our dependencies by patching the `yarn.lock` file. I typically prefer not to use those because you end up playing a game of wack-a-mole. Once you close one PR, another one pops up shortly afterwards. The better approach is to figure out which top level dependencies need to be updated, and update them. That is what this PR is doing. All of those package updates, plus more, are included in the update to `react-scripts`.

To test, you just need to load the site and click around a bit to ensure there are no console errors.


proto/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Start with a NodeJS base image that also contains yarn.
-FROM node:12.17.0-buster as nodejsbuilder
+FROM node:16.14.2-buster as nodejsbuilder
```
guggero, line 2: We also need to bump this version in the `Dockerfile` and `dev.Dockerfile`, otherwise I get `error react-scripts@5.0.0: The engine "node" is incompatible with this module. Expected version ">=14.0.0". Got "12.17.0"`

guggero (commented):
tACK, just needs a small change in the Docker files.

proto/Dockerfile
jamaljsr: Ah, good catch. I updated those two files in https://github.com/lightninglabs/lightning-terminal/pull/350/commits/0233f27f33262cf6aad4e792196dedb16e4ab33c

guggero (approved):
Nice, LGTM :tada:

PR #355, opened 2022-04-21
docker: fix build by removing unused python install

guggero:
We bumped the NodeJS version which also pulled in a new version of Alpine that no longer knows the `python` package (it's now explicitly called `python3`).
But since we don't require the binary any longer anyway, we remove it.


jamaljsr (approved):
tACK LGTM 👍

PR #356, opened 2022-04-21
README+version: bump version to v0.6.7-alpha

guggero:
We forgot to bump the app version, so we bump everything to the next one to avoid confusion. Will create a new release and basically replace 0.6.6 with this one.


jamaljsr (approved):
tACK LGTM 👍

PR #363, opened 2022-05-12
Add support for custodial off-chain accounts

guggero:
Early preview of the LiT custodial off-chain accounts feature.

~Depends on #308.~
Depends on #432.

EDIT: All TODOs have been resolved :tada:


guggero:
This still has a bunch of TODOs and things that can be improved, that's why it's still in draft. But would be nice if you could take a first look at the general architecture, @ellemouton and @Roasbeef.

ellemouton:
yep! currently doing so :)

rpcmiddleware/interface.go
```diff
@@ -0,0 +1,345 @@
...
+	// HandlesResponse returns true if the checker can handle protobuf
+	// response messages of the given type. This is mainly a safety feature
+	// to make sure a round trip checker is implemented correctly.
+	HandlesResponse(protoreflect.MessageType) bool
+
+	// HandleRequest is called for each incoming gRPC request message of the
+	// type declared to be accepted by AcceptsRequest. The handler can
+	// accept (=return nil) or refuse (=return non-nil error with rejection
+	// reason) an incoming request.
+	HandleRequest(proto.Message) error
```
ellemouton, line 91: could there maybe be a situation where we want to edit the request before it goes through? (like we do for the response)

ellemouton (commented):
still need to do a more in depth pass. But so far this looks amazing 😱 🔥 leaving a few comments so long

guggero:
Taking this out of WIP state since this is now a bit better tested and also received its own set of integration tests!

The following TODOs are still in place:
 - add unit tests
 - take in-flight payments into account when calculating remaining account balance
 - add support for creating a custom "account" LNC session by adding custom caveats to the AddSession RPC.

rpcmiddleware/interface.go
```diff
@@ -0,0 +1,345 @@
...
+
+	// HandleResponse is called for each outgoing gRPC response message of
+	// the type declared to be handled by HandlesResponse. The handler can
+	// pass through the response (=return nil, nil), replace the response
+	// with a new message of the same type (=return non-nil message, nil
+	// error) or abort the call by returning a non-nil error.
+	HandleResponse(proto.Message) (proto.Message, error)
```
ellemouton, line 98: qq: should we maybe pass a `context.Context` through to these so that the `Handle` functions can be canceled/timedout?

rpcmiddleware/interface.go
```diff
@@ -0,0 +1,345 @@
...
+	// intercepted message.
+	Intercept(context.Context,
+		*lnrpc.RPCMiddlewareRequest) (*lnrpc.RPCMiddlewareResponse,
+		error)
+}
+
+// requestHandler is a function type for a generic gRPC request message handler
+// that can accept (=return nil) or refuse (=return non-nil error with rejection
+// reason) an incoming request.
+type requestHandler func(req proto.Message) error
```
Roasbeef, line 66: Should we also permit the request handler to modify the request (so return a new one), as we did in `lnd` recently?

litrpc/lit-accounts.proto
```diff
@@ -0,0 +1,101 @@
...
+
+    /*
+    The expiration date of the account as a timestamp. Set to zero to never
+    expire.
+    */
+    int64 expiration_date = 2;
+}
+
+message Account {
+    /*
```
Roasbeef, line 50: Should we also return the raw macaroon here? Otherwise how will it be obtained? 

One architecture I have in mind is:
  * Service maintains an external DB of the macaroon for each user's accoutn 
  * Each time a user wants to do something, the services get the macaroon then issues the response as normal

So the service just needs to handle the mapping from user to account macaroon. 

One other alternative  architecture would be the discard model: the account service actually encodes the latest balance in the _macaroon itself_. This minimizes the state, as its the clients jobs to maintain this macaroon, and they can't modify it w/ breaking the HAMC check.

doc/accounts.md
```diff
@@ -0,0 +1,190 @@
...
+                "last_update": "1652353332",
+                "expiration_date": "0"
+        }
+}
+```
+
+This created a new account (ID ``) with an initial balance of 50k satoshis and
+no expiration.
+
+### Bake a new macaroon for the user
```
Roasbeef, line 120: Related to my comment above: what if we did this for the user?

Roasbeef (commented):
Amazing PR!

Rather large, so I did an initial pass through, focusing on internalizing the porposed architecture. To main things popped into my head:
  1. Why not return the macaroon from the account creation/list calls so the caller can store that along side the account ID? 
  2. What if we instead stored the latest account details in the macaroon itself? This would make the service more stateless as the balances would be encoded in the macaroon itself. The persistence gets a bit simpler, but then we need to sort of burden the user with storing that new updated macaroon. Also this wouldn't really work for the invoices unfortunately, since we'd want to keep the user's macaroon somewhat compact.

accounts/store.go
```diff
@@ -0,0 +1,443 @@
...
+	}
+
+	account.Invoices[hash] = struct{}{}
+	return s.storeAccount(account)
+}
+
+// CreditAccount checks if a paid invoice was associated with any existing
+// account. If an account is found that is still valid, it gets the invoice
+// amount credited.
+func (s *Store) CreditAccount(paidInvoice lntypes.Hash,
```
ellemouton, line 343: would it perhaps be worth keeping an in-memory map of `map[invoice-hash] accountID` so that this is quicker to check?

ellemouton (commented):
tACK 🔥  Leaving some initial comments :)

guggero:
!lightninglabs-deploy mute

accounts/checkers.go
```diff
@@ -0,0 +1,407 @@
...
+				t.PendingOpenRemoteBalance.Msat = 0
+				t.UnsettledLocalBalance.Sat = 0
+				t.UnsettledLocalBalance.Msat = 0
+				t.UnsettledRemoteBalance.Sat = 0
+				t.UnsettledRemoteBalance.Msat = 0
+
+				return t, nil
+			},
+		),
+		"/lnrpc.Lightning/WalletBalance": WalletBalanceEmptyRewriter,
```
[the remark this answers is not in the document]
guggero: Yeah, mainly to reduce the scope. Would the account balance be a unified balance (e.g. you can spend it either on-chain or off-chain)? Or would the user have two balances?
Maybe for an unified balance the diff wouldn't be too large. But maybe still something to do in a later PR?

accounts/interceptor.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	acctID, err := accountFromMacaroon(mac)
+	if err != nil {
+		return mid.RPCErrString(
+			req, "error parsing account from macaroon: %v", err,
+		)
+	}
+
+	// No account lock in the macaroon, not our concern!
```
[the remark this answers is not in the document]
guggero: True, changed this to an error.

accounts/checkers.go
```diff
@@ -0,0 +1,407 @@
...
+				return nil
+			},
+		),
+
+		// Payments:
+		"/lnrpc.Lightning/SendPayment": rpcmiddleware.NewRequestChecker(
+			&lnrpc.SendRequest{},
+			&lnrpc.SendResponse{},
+			func(r *lnrpc.SendRequest) error {
+				return s.checkSend(
```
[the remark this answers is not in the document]
guggero: The problem is that we can only map an incoming payment to an account if we see the invoice being created with the macaroon. We don't know that a keysend payment is for a specific account. Unless it would encode that information somehow, which is something very custom that each wallet would need to implement... I guess we would need to use the HTLC interceptor (and a bunch of other tricks) to generate a custom fake destination node pubkey and then show that to the user so someone else could keysend them.
Definitely out of scope of this PR.

rpcmiddleware/interface.go
```diff
@@ -0,0 +1,345 @@
...
+		),
+		responseHandler: passThroughResponseHandler,
+	}
+}
+
+// NewResponseRewriter returns a round trip checker that allows the incoming
+// request and inspects and potentially modifies the response.
+func NewResponseRewriter(requestSample proto.Message,
+	responseSample proto.Message,
+	typedResponseHandler interface{}) *DefaultChecker {
```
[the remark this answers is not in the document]
guggero: I tried that but failed. Maybe I'll try again after I've completed everything else.

accounts/checkers.go
```diff
@@ -0,0 +1,407 @@
...
+			&lnrpc.AddInvoiceResponse{},
+			func(t *lnrpc.AddInvoiceResponse) (proto.Message,
+				error) {
+
+				hash, err := lntypes.MakeHash(t.RHash)
+				if err != nil {
+					return nil, fmt.Errorf("error parsing "+
+						"invoice hash: %v", err)
+				}
+
```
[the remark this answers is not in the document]
guggero: Exactly, a `nil` return value in a response re-writer means "don't change anything".

litrpc/lit-accounts.proto
```diff
@@ -0,0 +1,101 @@
...
+    actually spend that amount.
+    */
+    rpc CreateAccount (CreateAccountRequest) returns (CreateAccountResponse);
+
+    /*
+    ListAccounts returns all accounts that are currently stored in the account
+    database.
+    */
+    rpc ListAccounts (ListAccountsRequest) returns (ListAccountsResponse);
+
```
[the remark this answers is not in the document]
guggero: Since you currently can't have two accounts in the same macaroon, it makes sense to allow an account to be updated. Added that to the RPC and CLI.

accounts/checkers.go
```diff
@@ -0,0 +1,407 @@
...
+		&lnrpc.GetTransactionsRequest{},
+		&lnrpc.TransactionDetails{},
+		func(r *lnrpc.TransactionDetails) (proto.Message, error) {
+			r.Transactions = make([]*lnrpc.Transaction, 0)
+
+			return r, nil
+		},
+	)
+
+	ListPeersEmptyRewriter = rpcmiddleware.NewResponseRewriter(
```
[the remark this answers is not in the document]
guggero: My aim was to make this work with a remote app like Zeus or Zap. And those apps probably won't work properly if they receive errors in some of the less important calls. The goal is to eventually support most calls that apps use to not break any of them.

accounts/checkers.go
```diff
@@ -0,0 +1,407 @@
...
+			r.Channels = []*lnrpc.ChannelCloseSummary{}
+
+			return r, nil
+		},
+	)
+
+	WalletBalanceEmptyRewriter = rpcmiddleware.NewResponseRewriter(
+		&lnrpc.WalletBalanceRequest{},
+		&lnrpc.WalletBalanceResponse{},
+		func(t *lnrpc.WalletBalanceResponse) (proto.Message, error) {
```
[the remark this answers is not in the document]
guggero: Good point! I updated the code to instead return an empty message where possible.

accounts/store.go
guggero: Yeah, good point. Since we need to listen for all invoices anyway in the service we might as well keep track of them there.

accounts/checkers.go
```diff
@@ -0,0 +1,407 @@
...
+
+				return nil, s.Store.associateInvoice(
+					acct.ID, hash,
+				)
+			},
+		),
+		"/lnrpc.Lightning/ListInvoices": rpcmiddleware.NewResponseRewriter(
+			&lnrpc.ListInvoiceRequest{},
+			&lnrpc.ListInvoiceResponse{},
+			func(t *lnrpc.ListInvoiceResponse) (proto.Message,
```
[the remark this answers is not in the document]
guggero: I think that's okay or even necessary to support pagination. Otherwise apps that paginate with the help of these indexes will do infinite requests? With the indexes intact, it just looks to the app like we deleted a lot of invoices because there are many gaps. But with the correct logic the app should still be able to paginate correctly.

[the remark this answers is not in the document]
ellemouton: ah yes, good point 👍

accounts/checkers.go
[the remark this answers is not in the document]
ellemouton: ah yes, good point 👍

accounts/checkers.go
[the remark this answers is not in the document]
ellemouton: shweeeeet 🚀

accounts/checkers.go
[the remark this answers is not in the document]
ellemouton: makes sense 👍

guggero:
Yayy, I was finally able to resolve all TODOs :rocket: 
This PR is now fully ready for final reviews.

accounts/checkers.go
```diff
@@ -0,0 +1,407 @@
...
+
+	return s.CreditAccount(invoice.Hash, invoice.AmountPaid)
+}
+
+func (s *Service) checkSend(amt, amtMsat int64, invoice string,
+	acct *OffChainBalanceAccount) error {
+
+	// TODO(guggero): Take in-flight payments into account!
+
+	sendAmt := lnwire.NewMSatFromSatoshis(btcutil.Amount(amt))
```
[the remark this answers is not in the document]
guggero: Was heavily refactored in the meantime, should be more clear now.

litrpc/lit-accounts.proto
guggero: Added the baking of an account macaroon to the RPC server and updated the docs.

doc/accounts.md
guggero: Yeah, great idea. Did that.

session_rpcserver.go
```diff
@@ -122,16 +125,36 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid account ID: %v", err)
+		}
+
+		cav := checkers.Condition(macaroons.CondLndCustom, fmt.Sprintf(
+			"%s %x", accounts.CondAccount, id[:],
+		))
+		caveats = append(caveats, macaroon.Caveat{
+			Id: []byte(cav),
+		})
```
ellemouton, line 147: 👍

doc/accounts.md
```diff
@@ -0,0 +1,177 @@
...
+        ]
+}
+```
+
+The important part is the `lnd-custom account ...` part in the `caveats`
+section.
+
+Example of using `lncli` to check the account balance:
+```shell
+$ lncli --macaroonpath=/tmp/accounts.macaroon channelbalance
```
ellemouton, line 162: maybe worth noting that this is only in integrated mode (right?)

litrpc/lit-accounts.proto
```diff
@@ -0,0 +1,122 @@
...
+    int64 expiration_date = 5;
+
+    /*
+    The list of invoices created by the account. An invoice created by an
+    account will credit the account balance if it is settled.
+    */
+    repeated bytes invoices = 6;
+
+    /*
+    The list of payments made by the account. A made by an account will debit
```
ellemouton, line 88: nit: second sentence starts weird

litrpc/lit-accounts.proto
```diff
@@ -0,0 +1,122 @@
...
+    the account balance if it is settled.
+    */
+    repeated bytes payments = 7;
```
ellemouton, line 91: also: for both payments and invoices about, could be nice to show the amounts associated with them and also show which have completed vs are still pending.

ellemouton (approved):
Such a stunning PR!!! 🤩 tACK! This is soooo cool. And awesome that it could be linked to LNC session creation so nicely.

accounts/checkers.go
```diff
@@ -0,0 +1,648 @@
...
+				if _, ok := acct.Invoices[hash]; !ok {
+					return fmt.Errorf("invoice does not " +
+						"belong to this account")
+				}
+
+				return nil
+			},
+		),
+
+		// Payments:
```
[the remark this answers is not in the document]
guggero: Hmm... Good point. I think it's doable and not too complex but IMO out of context for this PR.

doc/accounts.md
guggero: No, this should also work in remote mode. But then of course the server and TLS cert need to be adjusted. Added that as a note.

litrpc/lit-accounts.proto
guggero: Yeah, that might be nice. Though we don't have a state or amount for the invoice currently. Maybe a nice optimization for the future. I changed both response types to their own message types so we can easily extend this in the future.

ellemouton: awesome

accounts/store.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// Close closes the underlying bolt DB.
+func (s *BoltStore) Close() error {
+	return s.db.Close()
+}
+
+// NewAccount creates a new OffChainBalanceAccount with the given balance and a
+// randomly chosen ID.
+func (s *BoltStore) NewAccount(balance lnwire.MilliSatoshi,
```
Roasbeef, line 105: Are we assuming the account manager is always holding a mutex here? Or that a random ID collision will be caught when we got to update the account itself?

accounts/store.go
```diff
@@ -0,0 +1,311 @@
...
+// uniqueRandomAccountID generates a new random ID and makes sure it does not
+// yet exist in the DB.
+func (s *BoltStore) uniqueRandomAccountID() (AccountID, error) {
+	var newID AccountID
+	if _, err := rand.Read(newID[:]); err != nil {
+		return newID, err
+	}
+
+	if _, err := s.Account(newID); err != ErrAccNotFound {
+		return s.uniqueRandomAccountID()
```
Roasbeef, line 164: Ah ok, we'll just retry....👍

Though I wonder if a loop is better to batter manage the stack size? 

On the otherhand, don't really imagine this needing to be executed this many times given the large range.

Roasbeef: Also feels like we could bundle this entire operation in a DB transaction as well (account create, retry, commit).

accounts/store.go
```diff
@@ -0,0 +1,311 @@
...
+	// last known invoice settle index.
+	lastSettleIndexKey = []byte("last-settle-index")
+
+	// byteOrder is the binary byte order we use to encode integers.
+	byteOrder = binary.BigEndian
+)
+
+// BoltStore wraps the bolt DB that stores all accounts and their balances.
+type BoltStore struct {
+	db *bbolt.DB
```
Roasbeef, line 57: What if we modified this to use the `kvdb.DB` interface? We can then have an implementation that in practice will be bolt initially, but would let users hook this into their existing postgres (and later on `sqlite`) DB and inherit replication and some very basic querying (would at least be able to read the account values independently).

accounts/checkers.go
```diff
@@ -0,0 +1,615 @@
...
+			}, mid.PassThroughErrorHandler,
+		),
+		"/lnrpc.Lightning/ListInvoices": mid.NewResponseRewriter(
+			&lnrpc.ListInvoiceRequest{},
+			&lnrpc.ListInvoiceResponse{},
+			func(ctx context.Context,
+				t *lnrpc.ListInvoiceResponse) (proto.Message,
+				error) {
+
+				filteredInvoices, err := filterInvoices(ctx, t)
```
Roasbeef, line 143: Will be nice once we can do this type of filtering within the actual db query ;)

accounts/checkers.go
```diff
@@ -0,0 +1,615 @@
...
+			&lnrpc.SendRequest{},
+			&lnrpc.SendResponse{},
+			func(ctx context.Context, r *lnrpc.SendRequest) error {
+				return checkSend(
+					ctx, chainParams, service, r.Amt,
+					r.AmtMsat, r.PaymentRequest, r.FeeLimit,
+				)
+			}, sendResponseHandler, mid.PassThroughErrorHandler,
+		),
+		// routerrpc.Router/SendPayment is deprecated.
```
Roasbeef, line 198: Is it? 🧐

accounts/checkers.go
```diff
@@ -0,0 +1,615 @@
...
+		if route != nil {
+			totalAmount = route.TotalAmtMsat + route.TotalFeesMsat
+		}
+
+		return checkSendResponse(
+			ctx, service, status, hash, totalAmount,
+		)
+	}
+
+	checkers := CheckerMap{
```
Roasbeef, line 111: Will any calls the aren't part of this map succeed as normal, or error out due to insufficient permissions?

accounts/checkers.go
```diff
@@ -0,0 +1,615 @@
...
+// checkIncomingRequest makes sure the type of incoming call is supported and
+// if it is, that it is allowed with the current account balance.
+func (a *AccountChecker) checkIncomingRequest(ctx context.Context,
+	fullUri string, req proto.Message) error {
+
+	// If we don't have a handler for the URI, it means we don't support
+	// that RPC.
+	checker, ok := a.checkers[fullUri]
+	if !ok {
+		return ErrNotSupportedWithAccounts
```
Roasbeef, line 401: 👍 

Answers the question above.

doc/accounts.md
```diff
@@ -0,0 +1,178 @@
+# Custodial accounts
```
Roasbeef, line 1: 🥇

Roasbeef (commented):
Excellent work! 

Really like how this PR has shaped up over its lifetime. Catching up now with the altest version, but didn't find anything too glaring. 

My main comment here is that we start to use the `kvdb.DB` interface for the current key-value based account storage. This way, ppl can hook into the _existing_ postgres (and soon sqlite) datas stores, which allows them to utilize replication, etc.

accounts/interface.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+const (
+	// AccountIDLen is the length of the ID that is generated as a unique
+	// identifier of an account. It is 8 bytes long so guessing is
+	// improbable, but it's still not mistaken for a SHA256 hash.
+	AccountIDLen = 8
```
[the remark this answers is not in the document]
guggero: No, it's randomly assigned to use the full 8 bytes instead of an incrementing number (so it doesn't give any information about the number of accounts).

accounts/interface.go
```diff
@@ -0,0 +1,204 @@
...
+	if a.ExpirationDate.IsZero() {
+		return false
+	}
+
+	return a.ExpirationDate.Before(time.Now())
+}
+
+// CurrentBalanceSats returns the current account balance in satoshis.
+func (a *OffChainBalanceAccount) CurrentBalanceSats() int64 {
+	return a.CurrentBalance / 1000
```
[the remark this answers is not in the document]
guggero: The `CurrentBalance` is an `int64` while `lnwire.MilliSatoshi` is an `uint64`. But we need the balance to go negative if it is "overdrawn" (which should not happen in practice as we try to account for the routing fees properly). So we can't do `ToSatoshis()` here easily without more conversions/fixes.

accounts/checkers.go
guggero: Yes! Looking forward to native SQL everywhere :sunglasses:

accounts/checkers.go
guggero: Yes, since there's the `routerrpc.Router/SendPaymentV2` which should be used. We still allow the `lnrpc.Lightning/SendPayment` to be used, even though it is also kind of old. Maybe you meant that one?

accounts/store.go
guggero: Yeah, makes sense to limit the number of attempts and do everything in a single transaction. I wrote some of this code back in 2018/2019, good to get another pair of eyes to take a look at it :sweat_smile:

guggero:
Addressed all comments, should be ready for a final look.

ellemouton (commented):
needs a `go mod tidy` one one of the commits it seems. Build is breaking. Otherwise LGTM!

ellemouton (approved):
LGTM! 🔥 epic stuff 🚀

ellemouton (dismissed):

ellemouton (approved):
lgtm 🤓

accounts/service.go
```diff
@@ -0,0 +1,628 @@
...
+		}
+	}
+
+	// The invoice hasn't been settled yet, there is nothing for us to do.
+	// If it eventually settles, we'll be called again.
+	if invoice.State != channeldb.ContractSettled {
+		return nil
+	}
+
+	// The invoice was settled, let's now debit the account. But only if the
```
sangaman, line 388:
```suggestion
	// The invoice was settled, let's now credit the account. But only if the
```

litrpc/lit-accounts.proto
```diff
@@ -0,0 +1,142 @@
...
+    // The payment hash.
+    bytes hash = 1;
+
+    // The state of the payment as reported by lnd.
+    string state = 2;
+
+    /*
+    The full amount in satoshis reserved for this payment. This includes the
+    routing fee estimated by the fee limit of the payment request. The actual
+    credited amount will likely be lower if the fee is below the limit.
```
sangaman, line 109:
```suggestion
    debited amount will likely be lower if the fee is below the limit.
```

accounts/interface.go
```diff
@@ -0,0 +1,230 @@
...
+// account. This basically includes all information required to make sure
+// in-flight payments don't exceed the total available account balance.
+type PaymentEntry struct {
+	// Status is the RPC status of the payment as reported by lnd.
+	Status lnrpc.Payment_PaymentStatus
+
+	// FullAmount is the total amount of the payment which includes the
+	// payment amount and the estimated routing fee. The routing fee is
+	// estimated based on the fee limit set when sending the payment and
+	// might be higher than the actual routing fee.
```
sangaman, line 63:
```suggestion
	// set to the fee limit set when sending the payment and updated to
	// the actual routing fee when the payment settles.
```

sangaman (dismissed):
This is awesome. I reviewed most of the code and left some nits/comments. I also did a basic happy path test of creating an account, and using the macaroon to successfully make `lncli` calls to check channel balance, create an invoice, pay an invoice, list invoices & payments, and everything worked as expected.

One question I have is how robust this system is with an lnd instance that comes online & offline unpredictably. The `Start` logic that tracks the settle index for invoice and tracks all pending payments in the db makes sense, but what happens if we're past `Start` and lnd goes down and comes back (causing us to lose our `TrackPayment` subscriptions). Does the rpc interceptor get re-registered and the `TrackPayment` subscriptions re-established along with it? I figure it does but I couldn't follow where exactly in the code that happens.

guggero:
Thanks a lot for the review and for testing the happy path :rocket: Happy to hear you find this useful.

> Does the rpc interceptor get re-registered and the TrackPayment subscriptions re-established along with it?

Yes, on every startup we re-initialize the [payment tracking for each payment](https://github.com/lightninglabs/lightning-terminal/blob/aeef8bf564d134dd0417d59b03fd005d5127ba65/accounts/service.go#L115) and [invoice update stream](https://github.com/lightninglabs/lightning-terminal/blob/aeef8bf564d134dd0417d59b03fd005d5127ba65/accounts/service.go#L153).

So this should be pretty robust. Though in general "lnd instance that comes online & offline unpredictably" is never recommended in a production environment. Or are you thinking about a mobile setup? Curious about your use case!

Going to address your nits/comments inline.

accounts/interface.go
```diff
@@ -0,0 +1,230 @@
...
+
+// AccountType is an enum-like type which denotes the possible account types
+// that can be referenced in macaroons to keep track of user's balances.
+type AccountType uint8
+
+const (
+	// OneTimeBalance represents an account that has an initial balance
+	// that is used up when it is spent and is not replenished
+	// automatically.
+	OneTimeBalance AccountType = 0
```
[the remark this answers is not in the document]
guggero: Good point. I renamed the initially supported type to `TypeInitialBalance` to make it a bit more precise ("default/normal" isn't very descriptive IMO) and added a TODO to hint at possible future account types.

accounts/checkers.go
```diff
@@ -0,0 +1,615 @@
...
+}
+
+// replaceOutgoingResponse inspects the responses before sending them out to the
+// lightningClient and replaces them if needed.
+func (a *AccountChecker) replaceOutgoingResponse(ctx context.Context,
+	fullUri string, resp proto.Message) (proto.Message, error) {
+
+	// If we don't have a handler for the URI, it means we don't support
+	// that RPC. We shouldn't get here in the first place, since the request
+	// should've already been refused.
```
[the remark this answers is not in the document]
guggero: No, that comment is correct. The `request` here is referring to the client's original gRPC request. If that isn't allowed it is denied in the code above so `lnd` will never generate a response for it. Updated the comment to be a bit more explicit (request and response have many meanings in this context).

accounts/service.go
guggero: I think this is correct: ["A debit entry in an account represents a transfer of value to that account, and a credit entry represents a transfer from the account."](https://en.wikipedia.org/wiki/Debits_and_credits).

litrpc/lit-accounts.proto
guggero: See comment above.

accounts/service.go
sangaman: > I think this is correct: ["A debit entry in an account represents a transfer of value to that account, and a credit entry represents a transfer from the account."](https://en.wikipedia.org/wiki/Debits_and_credits).

Wow, that goes against my understanding of the words but apparently in the accounting world (I am not an accountant) "debit" does indeed mean adding value to an account and "credit" means subtracting value. But that seems to go against the "common" meaning of the words, according to [this article](https://www.sapling.com/4827482/debit-vs-credit-accounting):

> Outside of the accounting world, the term "debit" usually refers to money removed from a consumer bank account, such as money removed from your checking account when you buy groceries. Similarly, "credit" usually refers to money added to a consumer bank account account, or to money that is otherwise freely yours to spend

The dictionary definitions of the words also are in line with debit = payment/subtraction and credit = received/addition.

My 2 cents is that the common definitions of the words would be preferable, since this will be read by software developers and not accountants. It also seems to me that most existing usage of credit/debit is already in line with the common definitions, e.g.: https://github.com/lightninglabs/lightning-terminal/pull/363/files#diff-3bfb289868a7cc73be6cb96f5f75cf79f1b8f14395e62ffa03283ffb2aa07e6cR219

sangaman:
> Yes, on every startup we re-initialize the [payment tracking for each payment](https://github.com/lightninglabs/lightning-terminal/blob/aeef8bf564d134dd0417d59b03fd005d5127ba65/accounts/service.go#L115) and [invoice update stream](https://github.com/lightninglabs/lightning-terminal/blob/aeef8bf564d134dd0417d59b03fd005d5127ba65/accounts/service.go#L153).
> 
> So this should be pretty robust. Though in general "lnd instance that comes online & offline unpredictably" is never recommended in a production environment. Or are you thinking about a mobile setup? Curious about your use case!

Gotcha, I think one thing I was overlooking was that (if I'm not mistaken) `litd` will stop when it sees that `lnd` has stopped. So this `Start` function will be called every time either `litd` or `lnd` restart. My concern was only that `lnd` goes down, all subscriptions are lost, and then when it comes back up we don't re-establish the streams.

I wasn't thinking about a mobile setup or anything like that, just about what might happen if there are unexpected outages in an otherwise stable environment.

accounts/checkers.go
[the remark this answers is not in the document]
guggero: Yes, correct.

accounts/service.go
guggero: Hmm, no wonder I always get those confused then, if even Wikipedia disagrees with most use cases. And yes, after taking a closer look, I did use the terms mostly in the non-Wikipedia way so I fixed the two occurrences you highlighted.

guggero:
> `litd` will stop when it sees that `lnd` has stopped

Ah yes, that is important to know in this context. `litd` is indeed set up to stop (and be re-started by systemd, docker, k8s, whatever) when the connection to `lnd` is lost.

sangaman (approved):
Thanks for taking the time to respond to my feedback, this is great stuff and will be very useful I think.

sangaman:
My last handful of thoughts about this PR & feature:

1. It might be nice to have an RPC call that increments an account's balance. That way there's an easy way to, for example, add 1000 satoshis to an account. Or possibly decrement balance (by passing a negative value), but then there'd need to be the checks for sufficient balance on the account. Right now, you can update an account's balance via RPC, but that's only good if you're trying to set the balance to a specific amount, not add or subtract from the existing balance. Trying to do that by querying the balance and then updating the account balance would be subject to race conditions (maybe the balance updates in between the two calls). I figure this can be done in a follow-up PR, though.

2. It'd be nice if self-payments from one account to another were better supported. Right now, it seems like it does work if I use the `--allow_self_payment` flag, which then sends a payment out one channel and back in on the same channel. I think it'd be cool to be able to make an account-to-account payment without actually needing to make a real lightning payment (which may incur a fee, delay, and/or chance of failure). But I can imagine that would be a complicated change since it would likely break some internal assumptions in lnd about payments, namely that they actually involve the external lightning network. Just something to think about for future work.

guggero:
> My last handful of thoughts about this PR & feature:

Thanks a lot for your feedback! I think both of these features make sense for follow-up PRs (given the added complexity for item 2 and the age of this PR, this has been in the works since [late 2018](https://github.com/lightningnetwork/lnd/pull/2390) in some form or another). Feel free to add feature request issues to the repo so they don't drop off the radar.

litrpc/lit-accounts.proto
```diff
@@ -0,0 +1,142 @@
...
+    The list of invoices created by the account. An invoice created by an
+    account will credit the account balance if it is settled.
+    */
+    repeated AccountInvoice invoices = 6;
+
+    /*
+    The list of payments made by the account. A payment made by an account will
+    debit the account balance if it is settled.
+    */
+    repeated AccountPayment payments = 7;
```
sangaman, line 91: One last thought when I was looking through this PR, and also probably something that could be addressed in a follow-up PR, but it looks like these include all payments and invoices ever made by the accounts. Therefore when calling `ListAccounts`, you're fetching all payment and invoice history for all accounts across all time. That seems like it could potentially be a very heavy query.

It might make sense to have a `ListAccounts` call that just gives you the accounts and balances that exist, and then separate `ListInvoices` and `ListPayments` calls that work for a single account specified in the request? Something like that, just so that there's a way to list the accounts without needing to fetch all invoice & payment history.

I can put this in an issue as well, just wanted to raise it here first.

ellemouton (approved):
🔥

PR #370, opened 2022-05-21
deps: Upgrade `react-router` to v6

rohit-nair:
1/

To upgrade `react-router` from v5 to v6:
* Removed `react-router`
* Install react-router-dom`
* Removed `history` resolution

The major version upgrade had some breaking changes. To address those:
* Updated imports to be from `react-router-dom` as opposed to `react-router`
* `react-router` v6 provides a cleaner way of managing routes and nested
  routes. Refactored `Routes` and `SettingsPage` to reflect this.
* Renamed `Routes` to `AppRoutes` to avoid confict with `react-router`'s
  `Routes` component.
* Using `BrowserRouter` instead of `Router`
* Use `useNavigate` and `useLocation` hooks in the component instead of using
  `routerStore` to avoid `history` dependency and to avoid calling
  `createBrowserHistory`

[Upgrade instructions](https://reactrouter.com/docs/en/v6/upgrading/v5)


app/package.json
```diff
@@ -28,7 +28,6 @@
     "date-fns": "2.14.0",
     "debug": "4.1.1",
     "file-saver": "2.0.2",
-    "history": "4.10.1",
```
rohit-nair, line 31: Removed direct dependency to `history` library.

app/package.json
```diff
@@ -42,7 +41,7 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-i18next": "11.7.0",
-    "react-router": "5.2.0",
```
rohit-nair, line 45: Bumped `react-router` library version and added the `*-dom` library as dependency.

app/package.json
```diff
@@ -89,9 +88,6 @@
     "ts-protoc-gen": "0.12.0",
     "typescript": "4.1.6"
   },
-  "resolutions": {
```
rohit-nair, line 92: Removed resolution for `history` as no longer required.

app/src/AppRoutes.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { Suspense } from 'react';
```
rohit-nair, line 1: Renamed `Routes` component to `AppRoutes` to avoid conflict with `react-router.Routes`.

app/src/AppRoutes.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <Suspense fallback={<Loading delay={500} />}>
+      <BrowserRouter>
+        <Routes>
+          <Route path="/" element={<LazyAuthPage />} />
+          <Route>
+            <Route
+              path="loop"
+              element={
+                <Layout>
```
rohit-nair, line 23: Can't have `Layout` as child of `Routes` so wrapped lazy loaded components with it.
Optionally, I can have `Layout` in the components itself instead of here.

app/src/__stories__/StoryWrapper.tsx
```diff
@@ -70,13 +70,13 @@ const StoryWrapper: React.FC<{
   return (
     <StoreProvider store={store}>
       <ThemeProvider>
-        <Router history={store.router.history}>
+        <BrowserRouter>
```
rohit-nair, line 73: Using `BrowserRouter` instead of generic `Router` as suggested in their [guidance](https://reactrouter.com/docs/en/v6/routers/router).

app/src/components/common/PageHeader.tsx
```diff
@@ -43,7 +44,7 @@ const Styled = {
 
 interface Props {
   title: ReactNode;
-  onBackClick?: () => void;
+  showBackButton?: boolean;
```
rohit-nair, line 47: As we can use `useNavigate` hook here, refactored to have the navigation helper function defined in just this component instead of all components that use this component and want a back button.

app/src/components/layout/NavMenu.tsx
```diff
@@ -49,42 +51,44 @@ const Styled = {
...
     <Styled.NavItem className={className}>
-      <span onClick={onClick}>
-        {l(page)}
-        {badge && (
-          <sup>
-            <Badge muted>{badge}</Badge>
-          </sup>
-        )}
-      </span>
+      <Link to={`/${page}`} onClick={onClick}>
```
rohit-nair, line 66: Using `react-router` `Link` for routing as opposed to having a callback that invokes `useNavigate`.

app/src/components/settings/GeneralSettings.tsx
```diff
@@ -18,75 +20,82 @@ const Styled = {
...
-          icon="copy"
-        />
-        <SettingItem
-          name={l('alias')}
-          value={nodeStore.alias}
-          onClick={handleCopyAlias}
-          icon="copy"
-        />
-        {nodeStore.urlLabel && (
+    <Layout>
```
rohit-nair, line 39: Wrapping this component in `Layout` as only `GeneralSetting` needs the `Hamburger` menu while others show `<- Settings`.

app/src/components/settings/SettingsPage.tsx
```diff
@@ -19,12 +18,12 @@ const SettingsPage: React.FC = () => {
 
   return (
     <Wrapper>
-      <Switch>
-        <Route path={`${PUBLIC_URL}/settings`} exact component={GeneralSettings} />
-        <Route path={`${PUBLIC_URL}/settings/unit`} component={UnitSettings} />
-        <Route path={`${PUBLIC_URL}/settings/balance`} component={BalanceSettings} />
-        <Route path={`${PUBLIC_URL}/settings/explorers`} component={ExplorerSettings} />
-      </Switch>
+      <Routes>
```
rohit-nair, line 21: `v6` gives us much cleaner route definitions.

app/src/store/store.ts
```diff
@@ -49,8 +49,8 @@ export class Store {
   settingsStore = new SettingsStore(this);
   sessionStore = new SessionStore(this);
 
-  /** the store which synchronizes with the browser history */
-  router: RouterStore;
+  // /** the store which synchronizes with the browser history */
+  // router: RouterStore;
```
rohit-nair, line 53: Unsure whether we need `RouteStore` as `useLocation` hook exposes `location` to all components within the `*Router`. Perhaps I can inject `location` when we invoke `createStore` in `App`.

app/src/AppRoutes.tsx
```diff
@@ -0,0 +1,62 @@
...
+const LazyPoolPage = React.lazy(() => import('components/pool/PoolPage'));
+const LazySettingsPage = React.lazy(() => import('components/settings/SettingsPage'));
+const LazyConnectPage = React.lazy(() => import('components/connect/ConnectPage'));
+
+const AppRoutes: React.FC = () => {
+  return (
+    <Suspense fallback={<Loading delay={500} />}>
+      <BrowserRouter>
+        <Routes>
+          <Route path="/" element={<LazyAuthPage />} />
```
rohit-nair, line 18: Think I would have to support `{PUBLIC_URL}`. Will look into whether I can specify a base app path and have other routes be relative to it. Iirc Angular used to support this as you would have SPAs nested deep in your site.

app/src/AppRoutes.tsx
```diff
@@ -0,0 +1,63 @@
...
+              path="pool"
+              element={
+                <Layout>
+                  <LazyPoolPage />
+                </Layout>
+              }
+            />
+            {/* Only the GeneralSettings component needs hamburger menu 
+            hence not wrapping entire settings page in <Layout>. */}
+            <Route path="settings/*" element={<LazySettingsPage />} />
```
jamaljsr, line 47: We currently keep the sidebar (Layout comp) on all of the settings pages. Why do you think we should remove it? The styling breaks (black font) without the Layout parent component on the nested routes.

app/src/components/settings/SettingsPage.tsx
jamaljsr: Yes, these relative routes are much better. 👍

app/src/components/layout/NavMenu.tsx
jamaljsr: I like this approach much better. The `appView.goToXXX` began to smell fishy after it was initially implemented. Now we can open these links in a new tab, which is minor but can be very handy sometimes.

The only drawback is that there is additional logic in `goToLoop` and `goToPool` that is no longer being executed since these functions are not being called. I would suggest to move the logic into `useEffect` hooks on the `LoopPage` and `PoolPage` components. It actually makes more sense there to ensure the code is always executed when the pages are mounted.

app/src/store/views/appView.ts
```diff
@@ -28,15 +29,12 @@ export default class AppView {
     this._store = store;
   }
 
-  get fullWidth() {
-    return this._store.router.location.pathname === `${PUBLIC_URL}/pool`;
-  }
-
   /** navigate to the specified route */
   goTo(route: string) {
```
jamaljsr, line 33: Looks like these `goTo*` functions can all be removed now. 🙌

app/src/store/store.ts
jamaljsr: I'm fine with getting rid of this if we no longer needed.

jamaljsr (changes requested):
Thanks so much for the contribution. This is a pretty big update that was sorely needed. 

Overall the code looks really good. Thanks for being consistent with the existing code base. I made some comments below on issues I've found. Outside of these items, and fixing the unit tests and linter errors, I think this PR is solid.

app/src/AppRoutes.tsx
jamaljsr: I didn't fully test the `PUBLIC_URL` env var but will circle back to it in a follow-up review. This code looks like it should be fine though.

app/src/components/settings/GeneralSettings.tsx
jamaljsr: I think it would be fine to keep the existing behavior of showing the hamburger/sidebar along with the `<- Settings` link.

app/src/AppRoutes.tsx
rohit-nair: Motive was to avoid overlapping icon.

image

Was scratching my head as to what the styling issues you were referring to. Turns out _Dark Reader_ extension was enabled for `localhost:3000` so wasn't seeing the black font color.
I'll wrap settings in `Layout` and add some padding to the `←` icon.

app/src/App.tsx
```diff
@@ -14,9 +17,13 @@ const App = () => {
     <FullHeight>
       <StoreProvider store={store}>
         <ThemeProvider>
-          <Routes />
-          <AlertContainer />
-          <TourHost />
+          <Suspense fallback={<Loading delay={500} />}>
```
rohit-nair, line 20: Refactored `AppRoutes` to just be routes. Moved the `Suspense` and `Router` a level up.
This also helped fix issue with `AppRoutes.spec.tsx` as we can no longer have nested `Router`s (`renderWithProviders` was wrapping `AppRouter` in another `Router`).

app/src/__tests__/AppRoutes.spec.tsx
```diff
@@ -12,7 +13,7 @@ describe('Routes Component', () => {
   });
 
   const render = () => {
-    return renderWithProviders(<Routes />, store);
+    return renderWithProviders(<AppRoutes />, store);
```
rohit-nair, line 16: Renamed `Routes` to `AppRoutes`.

app/src/__tests__/components/settings/SettingsPage.spec.tsx
```diff
@@ -20,7 +21,13 @@ describe('SettingsPage', () => {
   });
 
   const render = () => {
-    return renderWithProviders(<SettingsPage />, store);
+    return renderWithProviders(
+      <Routes>
```
rohit-nair, line 25: Had to wrap `SettingsPage` in a `Routes` component to define the parent `/settings/` path.

app/src/components/common/PageHeader.tsx
```diff
@@ -12,6 +13,7 @@ const Styled = {
   `,
   Left: styled.span`
     flex: 1;
+    padding-left: 16px;
```
rohit-nair, line 16: Fix for overlapping hamburger and arrow icons on smaller viewports.

app/src/util/tests/renderWithProviders.tsx
```diff
@@ -16,7 +16,9 @@ const renderWithProviders = (component: React.ReactElement, withStore?: Store) =
   const result = render(
     <StoreProvider store={store}>
       <ThemeProvider>
-        <Router history={store.router.history}>{component}</Router>
+        <Suspense fallback={<></>}>
```
rohit-nair, line 19: With the refactor of `Routes` (renamed `AppRoutes`) `Suspense` and `Router` are now in the `App` component. In the absence of `Suspense` and `Router`, `AppRoutes.spec.tsx` throws `A React component suspended while rendering, but no fallback UI was specified.` error. Adding `Suspense` here to address that issue and shouldn't affect other components.

app/src/store/views/appView.ts
rohit-nair: I was planning to do it as part of the upgrade. But realized that refactoring navigation to use `useNavigate` and `useLocation` hooks in this PR would significantly increase the delta making it hard to manage and review.
Will refactor in a following standalone PR.

app/src/store/stores/routerStore.ts
```diff
@@ -14,7 +14,7 @@ export default class RouterStore {
     this.history = history;
     this.location = history.location;
 
-    history.listen(location => {
+    history.listen(({ location }) => {
```
rohit-nair, line 17: The argument to the listener is [now](https://github.com/remix-run/history/blob/main/docs/api-reference.md#historylistenlistener-listener) the `Update` interface.

app/src/store/views/appView.ts
jamaljsr: Great idea! This definitely decreased the diff. Thanks for the clarification as well. 👍

jamaljsr (dismissed):
I tested the app and haven't run into any issues. I also tested the `PUBLIC_URL` path. 

The code looks good as well. Thanks for iterating on this PR. 🥇

jamaljsr:
Please rebase this when you have a chance so we can et it merged. 👍

rohit-nair:
> I tested the app and haven't run into any issues. I also tested the PUBLIC_URL path.
Appreciate you testing these scenarios!
Rebased.

jamaljsr (approved):

PR #375, opened 2022-05-31
ui+pool: disable buttons during account changes

jamaljsr:
This PR fixes a recurring issue where users were creating multiple Pool accounts via the UI by clicking on the "Create Account" button while the previous request was in-flight. To resolve this, we disable the buttons immediately after being clicked for account modification requests (init, deposit, renew, close).

### DIFF
--- a/app/src/store/views/closeAccountView.ts
+++ b/app/src/store/views/closeAccountView.ts
@@ -10,6 +10,8 @@ export default class CloseAccountView {
   destination = '';
   satsPerVbyte = 0;
 
+  loading = false;
+
   constructor(store: Store) {
     makeAutoObservable(this, {}, { deep: false, autoBind: true });
 
@@ -55,6 +57,10 @@ export default class CloseAccountView {
     this.satsPerVbyte = satPerVbyte;
   }
 
+  setLoading(loading: boolean) {
+    this.loading = loading;
+  }
+
   //
   // Actions
   //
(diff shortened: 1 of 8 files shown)


app/src/store/views/closeAccountView.ts
```diff
@@ -75,10 +81,11 @@ export default class CloseAccountView {
   async closeAccount() {
     if (!this.isValid) return;
 
+    this.setLoading(true);
     try {
       await this._store.orderStore.cancelAllOrders();
     } catch (error) {
-      this._store.appView.handleError(error, 'Unable to cancel all open orders');
+      this._store.appView.handleError(error as Error, 'Unable to cancel all open orders');
```
guggero, line 88: Do we need to call `this.setLoading(false)` in case of an error? Or is that handled by `handleError()`?

guggero (dismissed):
tACK, LGTM :tada:

app/src/store/views/closeAccountView.ts
jamaljsr: Ah, good catch. Thx for spotting that. I fixed this in https://github.com/lightninglabs/lightning-terminal/pull/375/commits/88ebfc371d84829a60a1da2c3bc152c34e699b00

guggero (approved):

guggero (approved):

PR #378, opened 2022-06-13
session_rpcserver: add session expiry as macaroon caveat

ellemouton:
Bake the session expiry into the macaroon as a caveat. This will allow
the client of a session to get an idea of how long they will be able to
use the session for.


guggero (approved):
Nice, LGTM :tada:

PR #358, opened 2022-04-29
multi: re-enable LNC handshake v2, bump dependencies for lnd 0.15.0-beta release

guggero:
This commit re-enables the second handshake for LNC.

PR #381, opened 2022-06-22
sessions: disable copying pairing phrases that are in use

jamaljsr:
This PR updates the Lightning Node Connect page with the following changes:
- update `State` column to display "In Use" when the `remotePublicKey` field has a value
- disable the copy icon when the session is "In Use"
- add "Expiry" column to show the expiry date or "Never"
- increase the size of the Revoke icon

This PR depends on #358


guggero (approved):
tACK, LGTM :tada: 

I cherry-picked the two commits onto #358 to make sure we don't merge this too early by accident. So I think we can close this PR?

guggero:
Oh, sorry, I didn't see that you added #358 as a base for this PR. Let me merge it then.

PR #382, opened 2022-06-24
ui+sessions: improve when expiry label displays Never

jamaljsr:
Based on #358 

This PR fixes a minor UI issue where the Expiry label displays the far off date values instead of "Never. This occurs when the session was created by a user in a different time zone.

image


guggero (approved):
tACK, LGTM :tada:

PR #396, opened 2022-08-03
chore: update npm dependencies

jamaljsr:
This PR updates the React and Storybook dependencies.

It also adds a `dependabot.yml` config file to instruct it to ignore security issues for packages listed under `devDependencies`. The PRs #385 and #393 that dependabot created are for for child dependencies of Storybook. After updating the `@storybook/*` packages to the latest version, the vulnerable versions of those packages are still included. However, that code is never bundled in the production app build and therefore does not pose an actual security risk. They can be safely ignored.


guggero (approved):
LGTM :tada:

PR #400, opened 2022-08-03
chore: prevent version updates by dependabot

jamaljsr:
It turns out that the `dependabot.yml` config file only applies to *version* updates, not *security* updates. When I added this file in #396, the bot decided to open PRs for some of our production dependencies, even if there isn't a security issue with those packages. I don't like the automated PRs because it turns into a bunch of spam in the repo. I'm deleting this file to turn off the dependabot version updates.


jamaljsr:
ok, I will give that a try. Thanks 👍

jamaljsr:
Yup, you were right. It actually states this right in the [docs](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file), which I originally missed. Closing this in favor of #402.

PR #404, opened 2022-08-04
ui+settings: update theme colors + border colors

dstrukt:
Apologies, should have probably done this in the last PR, but didn't fully know what it would entail, so didn't want to try to do everything in a single PR.

Just addresses `history` and `settings` to button up the basic things for the time being.

ui+settings: update theme colors + update border color on settings and history

Screenshots:
Screen Shot 2022-08-04 at 4 19 29 PM
Screen Shot 2022-08-04 at 4 19 32 PM
Screen Shot 2022-08-04 at 4 20 02 PM

PR #405, opened 2022-08-10
ui+theme: patch theme color

dstrukt:
ui+theme: patch theme color bug introduced in #404

paste (3)


jamaljsr (approved):
Confirmed this fixes the issue. Thanks for the update.

Issue #406, opened 2022-08-18
add time limit on switch to second handshake


ellemouton:
With nodes that have updated to using second handshake for LNC connection we should set a deadline for the initial, pass phrase based, connection. So if a user creates a new session and gets a passphrase for it but then doesnt make an initial connection in a timely manner, then the session should be force-revoked.

more context here: https://github.com/lightninglabs/lightning-terminal/pull/401#issuecomment-1219068788

PR #408, opened 2022-08-18
multi: add deadline for first connection of new LNC conn

ellemouton:
In this PR, we add a deadline for the initial connection of an LNC
connection. So with this, the user is forced to use their pairing phrase
within a certain time frame. After this initial connection, future
connections are made with the second handshake version meaning that the
pairing phrase is rendered useless. By adding a time limit to the time
in which a user can use their pairing phrase, we reduce the risk created
by the users pairing phrase being leaked. The default time limit is set
to 10 minutes but can be customsed with the new `firstlncconndeadline`
flag.

The first commit in this PR adds a new `CreatedAt` field to `Session` so that 
we can have an absolute deadline.

Depends on https://github.com/lightninglabs/lightning-node-connect/pull/53

Fixes https://github.com/lightninglabs/lightning-terminal/issues/406


jamaljsr (approved):
tACK LGTM.. amazing turn around time 🥇

session/tlv.go
```diff
@@ -24,6 +24,7 @@ const (
 	typeLocalPrivateKey tlv.Type = 10
 	typeRemotePublicKey tlv.Type = 11
 	typeMacaroonRecipe  tlv.Type = 12
+	typeCreatedAt       tlv.Type = 13
```
guggero, line 27: Yayy, TLV for the win :star_struck:

session_rpcserver.go
```diff
@@ -214,15 +215,92 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
+			s.cfg.firstConnectionDeadline,
+		)
+
+		if deadlineDuration.Before(time.Now()) {
+			log.Debugf("Deadline for session %s has already "+
+				"passed. Revoking session", pubKeyBytes)
+
+			err = s.db.RevokeSession(pubKey)
+			if err != nil {
+				log.Debugf("error revoking session: "+"%v", err)
```
guggero, line 241: Any reason for not returning the error here? Or was this intended to be a `return fmt.Errorf()`? Also, string concatenation artifacts.

config.go
```diff
@@ -302,21 +304,22 @@ func defaultConfig() *Config {
...
+		MacaroonPath:         DefaultMacaroonPath,
+		ConfigFile:           defaultConfigFile,
+		FaradayMode:          defaultFaradayMode,
+		Faraday:              &faradayDefaultConfig,
+		faradayRpcConfig:     &frdrpcserver.Config{},
+		LoopMode:             defaultLoopMode,
+		Loop:                 &loopDefaultConfig,
+		PoolMode:             defaultPoolMode,
+		Pool:                 &poolDefaultConfig,
+		FirstLNCConnDeadline: 10 * time.Minute,
```
guggero, line 322: nit: extract into constant?

session_rpcserver.go
```diff
@@ -214,15 +215,92 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
...
+			}
+		}
+
+		// Start the deadline timer.
+		firstConnectionDeadline = time.NewTimer(
```
guggero, line 246: I think we could use `time.AfterFunc()` here that closes the `firstConnTimeout`. Then in the `onNewStatus` callback we only stop that timer and we don't need a whole other goroutine (the debug comment could be moved into the `onNewStatus` callback).

guggero (commented):
Nice, looks pretty good and should be a great security improvement!
I left a suggestion for making the logic a bit easier to understand.

session_rpcserver.go
ellemouton: erg yeah sorry 🙈 also, there should be a `return` after this 🙈

session_rpcserver.go
ellemouton: omg yeeeesss!!! thank you :) updated!

ellemouton (commented):
thanks @guggero ! updated as per you suggestion 🚀  much neater the way you have suggested :)

ellemouton:
hmmm the itest keeps failing....i must have broken something on the latest update.  looking into it

ellemouton:
hmmmm having the same flake on master actually.....

ellemouton:
hahaha found the issue 🙈 when I extracted the timeout into a variable, I used seconds instead of minutes 🙈  should be g2g now! I also added another commit that adds LNC client logging to the itests just to make it easier to debug these things in future

guggero (approved):
utACK, LGTM :tada:

jamaljsr (approved):
tACK 🔥

The deadline works as advertised. Thanks for this great update.

PR #413, opened 2022-08-30
subserver_perms: add Lnd's registered subserver perms

ellemouton:
In this commit, we include all LND's rpc subserver permissions in the permission list used to create the macaroons used in Litd (for sessions etc). The way this is done is a temporary solution while we wait for an LND change that exports the permissions required by each subserver. 

Fixes https://github.com/lightninglabs/lnc-web/issues/60

Run alongside https://github.com/lightninglabs/lightning-node-connect/pull/55 to test

### DIFF
--- a/make/release_flags.mk
+++ b/make/release_flags.mk
@@ -11,7 +11,7 @@ windows-386 \
 windows-amd64 \
 windows-arm
 
-LND_RELEASE_TAGS = litd autopilotrpc signrpc walletrpc chainrpc invoicesrpc watchtowerrpc
+LND_RELEASE_TAGS = litd autopilotrpc signrpc walletrpc chainrpc invoicesrpc watchtowerrpc neutrinorpc peersrpc
 
 # By default we will build all systems. But with the 'sys' tag, a specific
 # system can be specified. This is useful to release for a subset of
--- a/rpc_proxy.go
+++ b/rpc_proxy.go
@@ -24,7 +24,6 @@ import (
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	"google.golang.org/grpc/test/bufconn"
-	"gopkg.in/macaroon-bakery.v2/bakery"
 	"gopkg.in/macaroon.v2"
 )
 
@@ -59,8 +58,7 @@ func (e *proxyErr) Unwrap() error {
 // component.
 func newRpcProxy(cfg *Config, validator macaroons.MacaroonValidator,
 	superMacValidator session.SuperMacaroonValidator,
-	permissionMap map[string][]bakery.Op,
-	bufListener *bufconn.Listener) *rpcProxy {
+	permsMgr *PermissionsManager, bufListener *bufconn.Listener) *rpcProxy {
 
 	// The gRPC web calls are protected by HTTP basic auth which is defined
 	// by base64(username:password). Because we only have a password, we
@@ -77,7 +75,7 @@ func newRpcProxy(cfg *Config, validator macaroons.MacaroonValidator,
 	p := &rpcProxy{
 		cfg:               cfg,
 		basicAuth:         basicAuth,
-		permissionMap:     permissionMap,
+		permsMgr:          permsMgr,
 		macValidator:      validator,
 		superMacValidator: superMacValidator,
 		bufListener:       bufListener,
@@ -146,9 +144,9 @@ func newRpcProxy(cfg *Config, validator macaroons.MacaroonValidator,
 //                                    +---------------------+
 //
 type rpcProxy struct {
-	cfg           *Config
-	basicAuth     string
-	permissionMap map[string][]bakery.Op
+	cfg       *Config
+	basicAuth string
+	permsMgr  *PermissionsManager
 
 	macValidator      macaroons.MacaroonValidator
 	superMacValidator session.SuperMacaroonValidator
@@ -345,17 +343,17 @@ func (p *rpcProxy) makeDirector(allowLitRPC bool) func(ctx context.Context,
 		// handled by the integrated daemons that are hooking into lnd's
 		// gRPC server.
 		switch {
-		case isFaradayURI(requestURI) && p.cfg.faradayRemote:
+		case p.permsMgr.IsFaradayURI(requestURI) && p.cfg.faradayRemote:
 			return outCtx, p.faradayConn, nil
 
-		case isLoopURI(requestURI) && p.cfg.loopRemote:
+		case p.permsMgr.IsLoopURI(requestURI) && p.cfg.loopRemote:
 			return outCtx, p.loopConn, nil
 
-		case isPoolURI(requestURI) && p.cfg.poolRemote:
+		case p.permsMgr.IsPoolURI(requestURI) && p.cfg.poolRemote:
 			return outCtx, p.poolConn, nil
 
 		// Calls to LiT session RPC aren't allowed in some cases.
-		case isLitURI(requestURI) && !allowLitRPC:
+		case p.permsMgr.IsLitURI(requestURI) && !allowLitRPC:
 			return outCtx, nil, status.Errorf(
 				codes.Unimplemented, "unknown service %s",
 				requestURI,
@@ -373,7 +371,7 @@ func (p *rpcProxy) UnaryServerInterceptor(ctx context.Context, req interface{},
 	info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{},
 	error) {
 
-	uriPermissions, ok := p.permissionMap[info.FullMethod]
+	uriPermissions, ok := p.permsMgr.URIPermissions(info.FullMethod)
 	if !ok {
 		return nil, fmt.Errorf("%s: unknown permissions "+
 			"required for method", info.FullMethod)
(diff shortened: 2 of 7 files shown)


jamaljsr (dismissed):
tACK LGTM 👌

Thanks for the quick fix.

rpc_proxy.go
```diff
@@ -624,6 +632,20 @@ func (p *rpcProxy) convertSuperMacaroon(ctx context.Context, macHex string,
 	return nil, nil
 }
 
+// getPerms returns a map from URI to permissions required for the URI. The
+// first time it is called, the getPermsMap is used to set the contents of the
+// map. After that, the same map is returned.
+func (p *rpcProxy) getPerms() map[string][]bakery.Op {
```
ellemouton, line 638: i hate that we have to use a lock here... but couldnt figure out a nicer way of doing it. Since the rpcProxy needs to be started _before_ the connection to LND is complete so that it is possible for the user to unlock lnd through the rpcProxy. But we have to connect to lnd before knowing its build tags. so couldnt pass a pre-set list to rpcProxy anymore...

subserver_permissions.go
```diff
@@ -148,3 +210,83 @@ func isLitURI(uri string) bool {
 	_, ok := litPermissions[uri]
 	return ok
 }
+
+func init() {
```
guggero, line 214: Hmm, this is all becoming very hard to read and to follow what happens when...

Might be a slight refactor, but how about we create a new `PermissionManager` struct that runs this code during `NewPermissionManager` and keeps a copy of the full permission set in memory. Then we can call `GetPermissions(readOnly, lndBuildTags)` to get a filtered version.
And we could give it the lnd build tags later, which causes it to become "ready" (call this the "runtime permissions"). 
That way we can create a single instance of the manager before we create the `rpcProxy` and hand an instance to it (instead of a function). Then the rpcProxy would always call the `GetRuntimePerms()` which would just return an empty map while it's not "ready".

So only state and locks in the manager, everything else should become way simpler.
What do you think?

guggero (commented):
The code looks like it would work. But to be honest, this adds a lot of new run time timing dependencies which are very hard to follow. See my other comment for a suggestion.

subserver_permissions.go
ellemouton: great idea :) thanks!

ellemouton:
Thank you for the quick review on this @jamaljsr & @guggero 🙏 

@guggero: You are completely right - it was definitely getting a bit hard to follow the code. I have taken your suggestion and added a `PermissionsManager` (1st commit just adds the manager and 2nd commit does the lnd-subserver things).

Did it _slightly_ different to how you suggested so let me know if you think this way is fine.

subserver_permissions.go
```diff
@@ -41,53 +41,71 @@ var (
...
+	for _, perms := range permissions {
+		for k, v := range perms {
+			allPerms[k] = v
+		}
 	}
-	return result
+
+	return &PermissionsManager{
+		fixedPerms: permissions,
+		perms:      allPerms,
```
guggero: Isn't setting this on start a contradiction to the godoc? Meaning that if we call `GetPermissionsList()` directly after creating the manager, we'd always get all permissions?
Changing this probably means needing to call `OnLNDBuildTags()` in the itest manually (for baking the super macaroon).

terminal.go
```diff
@@ -193,15 +195,20 @@ func (g *LightningTerminal) Run() error {
 	// Show version at startup.
 	log.Infof("LiT version: %s", Version())
 
+	// Construct a new PermissionsManager
```
guggero, line 198: nit: missing punctuation.

subserver_permissions.go
```diff
@@ -41,53 +41,71 @@ var (
...
 }
 
-// GetAllPermissions retrieves all the permissions needed to bake a super
-// macaroon.
-func GetAllPermissions(readOnly bool) []bakery.Op {
-	dedupMap := make(map[string]map[string]bool)
+// GetPermOps returns a list of permission operations for the given URI if the
+// uri is known to the manager. The second return parameter will be false if
+// the URI is unknown to the manager.
+func (pm *PermissionsManager) GetPermOps(uri string) ([]bakery.Op, bool) {
```
guggero, line 190: Naming suggestion: `URIPermissions`?

subserver_permissions.go
```diff
@@ -41,53 +41,71 @@ var (
...
+	ops, ok := pm.perms[uri]
+	return ops, ok
+}
 
-	for _, methodPerms := range getAllMethodPermissions() {
+// GetPermissionsList returns all the available active permissions that the
+// manager is aware of. Optionally, readOnly can be set to true if only the
+// read-only permissions should be returned.
+func (pm *PermissionsManager) GetPermissionsList(readOnly bool) []bakery.Op {
```
guggero, line 201: Naming suggestion: `ActivePermissions`?

subserver_permissions.go
```diff
@@ -190,3 +290,65 @@ func (pm *PermissionsManager) IsLitURI(uri string) bool {
 	_, ok := pm.fixedPerms[litPerms][uri]
 	return ok
 }
+
+// mockConfig implements lnrpc.SubServerConfigDispatcher. It provides th
```
guggero, line 294: nit: s/th/the

terminal.go
```diff
@@ -494,6 +494,16 @@ func (g *LightningTerminal) startSubservers() error {
 		return err
 	}
 
+	// Collect the tags that LND was built with.
+	version, err := g.lndClient.Versioner.GetVersion(ctxc)
```
guggero, line 498: The version RPC is already called on init and cached in `g.lndClient.Version`, we can use that directly.

guggero (commented):
Almost there! Looks way better with the permission manager :tada:

subserver_permissions.go
ellemouton: well, we never _remove_ any perms from this map. We only ever add more perms once we get the build tags from lnd.

ellemouton: Meaning that `GetPermissionsList` doesnt ever provide you with false positives.

ellemouton: so `perms` represents all permissions that we know we definitely  have

terminal.go
ellemouton: oh cool! ok does this mean I can also remove the `verrpc` tag from the readme? sounds like LND is automatically compiled with that tag?

subserver_permissions.go
ellemouton: the godoc for the `perms` member is also updated in the next commit

terminal.go
guggero: Ah, yes. I forgot that some subservers are always active and not behind a build tag. I guess only those that we declare as "experimental" still have a flag.

subserver_permissions.go
```diff
@@ -163,8 +262,9 @@ func (pm *PermissionsManager) GetLitPerms() map[string][]bakery.Op {
 
 // IsLndURI returns true if the given URI belongs to an RPC of lnd.
 func (pm *PermissionsManager) IsLndURI(uri string) bool {
+	_, lndSubServerCall := pm.lndSubServerPerms[uri]
```
guggero, line 265: This needs to go one level deeper:
```
	lndSubServerCall := false
	for _, subserverPermissions := range pm.lndSubServerPerms {
		_, found := subserverPermissions[uri]
		if found {
			lndSubServerCall = true
			break
		}
	}
```

make/release_flags.mk
```diff
@@ -11,7 +11,7 @@ windows-386 \
 windows-amd64 \
 windows-arm
 
-LND_RELEASE_TAGS = litd autopilotrpc signrpc walletrpc chainrpc invoicesrpc watchtowerrpc
+LND_RELEASE_TAGS = litd autopilotrpc signrpc walletrpc chainrpc invoicesrpc watchtowerrpc routerrpc wtclientrpc
```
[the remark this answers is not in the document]
guggero: Ohhh... This is quite sneaky... When we set the `routerrpc` tag here for example, then the integrated `lnd` (and the one used in our itest!) will show the tag in the version tag list. But the real `lnd` won't. So we need to make sure this really reflects the current list of build tags in `lnd`, otherwise our itests might give us false positives.

guggero (commented):
Sorry, found a few other things while adding an itest (sending you my diff out of band)...

subserver_permissions.go
ellemouton: oh wow yeah that is embarrassing 🙈 thanks for catching!

README.md
```diff
@@ -38,7 +38,7 @@ To use LiT with a remote `lnd` instance please [follow these instructions](./doc
 Note that LiT requires `lnd` to be built with **all of its subservers** and requires running at least v0.11.0. Download the latest [official release binary](https://github.com/lightningnetwork/lnd/releases/latest) or build `lnd` from source by following the [installation instructions](https://github.com/lightningnetwork/lnd/blob/master/docs/INSTALL.md). If you choose to build `lnd` from source, use the following command to enable all the relevant subservers:
 
 ```shell
-⛰  make install tags="signrpc walletrpc chainrpc invoicesrpc"
+⛰  make install tags="signrpc walletrpc chainrpc invoicesrpc verrpc"
```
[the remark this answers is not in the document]
ellemouton: ahhh good point

subserver_permissions.go
```diff
@@ -54,20 +87,50 @@ const (
...
+			&mockConfig{},
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		name := subServer.SubServerName
+		lndSubServerPerms[name] = make(map[string][]bakery.Op)
+		for key, value := range perms {
+			lndSubServerPerms[name][key] = value
```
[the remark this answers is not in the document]
ellemouton: yeah great point. I should have checked this 🙈 havent really worked with build tags before so am learning as I go!

Cool, can confirm that the only ones we need to manually include are the ones that you have mentioned.

jamaljsr (approved):
Tested the latest changes. All of the sub-servers are now working in the LNC example app. Thanks so much for working on this so quickly. LGTM 🔥

PR #419, opened 2022-09-16
multi: rpcmiddleware package

ellemouton:
This PR separates the rpcmiddlware package away from the accounts interceptor logic in https://github.com/lightninglabs/lightning-terminal/pull/363

### DIFF
--- a/config.go
+++ b/config.go
@@ -18,6 +18,7 @@ import (
 	"github.com/lightninglabs/faraday"
 	"github.com/lightninglabs/faraday/chain"
 	"github.com/lightninglabs/faraday/frdrpcserver"
+	mid "github.com/lightninglabs/lightning-terminal/rpcmiddleware"
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightninglabs/loop/loopd"
 	"github.com/lightninglabs/pool"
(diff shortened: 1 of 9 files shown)


config.go
```diff
@@ -382,6 +386,14 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 		return nil, err
 	}
 
+	// To enable the rpc middleware interceptor, we need LND's RPC
+	// middleware interceptor to be enabled. In remote mode we can't
+	// influence whether that's enabled on lnd. But in integrated mode we
+	// can overwrite the flag here.
+	if !cfg.lndRemote {
```
guggero, line 393: I think we shouldn't overwrite the value if the middleware is disabled. Maybe the user wants to have their own interceptor without the manager? So basically move the `!cfg.RPCMiddleware.Disabled` into the `if` statement and then just do `cfg.Lnd.RPCMiddleware.Enable = true`.

guggero (approved):
Very nice, LGTM :tada:

guggero:
Hmm, I have a small diff left when rebasing my account PR on top of this. See https://github.com/lightninglabs/lightning-terminal/pull/363/commits/64ffcac407363137529f515b0962658d89f70861.
Is that no longer needed in your internal feature branch? Or should we add that to this PR as well?

config.go
ellemouton: yeah good point. Fixed!

ellemouton:
> Hmm, I have a small diff left when rebasing my account PR on top of this. See [64ffcac](https://github.com/lightninglabs/lightning-terminal/commit/64ffcac407363137529f515b0962658d89f70861).
> Is that no longer needed in your internal feature branch? Or should we add that to this PR as well?

Hmmm I dont think I understand - all the diffs in the commit you linked are in the accounts package so would not affect anything in the rpcmiddleware package or anything using it

guggero:
> > Hmm, I have a small diff left when rebasing my account PR on top of this. See [64ffcac](https://github.com/lightninglabs/lightning-terminal/commit/64ffcac407363137529f515b0962658d89f70861).
> > Is that no longer needed in your internal feature branch? Or should we add that to this PR as well?
> 
> Hmmm I dont think I understand - all the diffs in the commit you linked are in the accounts package so would not affect anything in the rpcmiddleware package or anything using it

LOL, you're absolutely right. Seems like I bran farted :joy:
Probably just occurred to me because the commit was named the same as 40220d5f.

guggero (approved):

PR #430, opened 2022-10-13
multi: use filepath instead of path to work on Windows

guggero:
This commit fixes an issue with file system paths on Windows. The path package is only intended to be used with URIs or non file system paths. For anything OS dependent, filepath should be used instead.


ellemouton (approved):
LGTM 👍

PR #432, opened 2022-10-15
litd: maintenance, code cleanup, middleware fixes

guggero:
This PR contains a set of commits that were extracted from https://github.com/lightninglabs/lightning-terminal/pull/363 that aren't specific to that functionality.


terminal.go
```diff
@@ -935,14 +943,15 @@ func (g *LightningTerminal) shutdown() error {
 
 	g.wg.Wait()
 
-	// The lnd error channel is only used if we are actually running lnd in
-	// the same process.
-	if g.cfg.LndMode == ModeIntegrated {
-		err := <-g.lndErrChan
+	// Do we have any last errors to display?
+	select {
+	case err := <-g.errQueue.ChanOut():
```
ellemouton, line 948: ok cool, so the assumption here is that the first error added to the queue is the one we want to display (right?)

make/testing_flags.mk
```diff
@@ -0,0 +1,6 @@
+ITEST_FLAGS = 
+
```
ellemouton, line 2: yay!

ellemouton (approved):
EPIC PR 🔥 

Absolutely LOVE the error channel pattern

terminal.go
ellemouton: also looks like there are a few other places that could read from this channel first in which case the log here would not necessarily be the initial error.

ellemouton: (mostly just commenting for my self to help with understanding)

guggero: Hmm, the logic here is a bit tricky to understand, yes. I guess some further refactoring of the way we start/stop the subsystems would be a good thing.
But during normal operation we only shut down on 4 triggers:
 - The loop server errors out
 - There is an error in the new error queue
 - Lnd shut down (if with an error, the above will trigger first, if normal shutdown this will trigger)
 - The shutdown interceptor triggers

So in case 1, 2 and 4 there is no error in the queue before we shut down. So it's only possible that something happened during shutdown. But you're right, there might be more than one error, so I added a loop that logs all remaining errors and returns just the last one to the outer component (which will be logged as the causing error).

terminal.go
```diff
@@ -935,15 +943,25 @@ func (g *LightningTerminal) shutdown() error {
...
 	g.wg.Wait()
 
-	// The lnd error channel is only used if we are actually running lnd in
-	// the same process.
-	if g.cfg.LndMode == ModeIntegrated {
-		err := <-g.lndErrChan
-		if err != nil {
-			log.Errorf("Error stopping lnd: %v", err)
-			returnErr = err
+	// Do we have any last errors to display? We use an anonymous function,
```
ellemouton, line 946: 👍

ellemouton (approved):
Yeeeet 🔥

PR #438, opened 2022-10-25
multi: custom URI sessions

ellemouton:
With this PR, you can now create a `custom` session where on creation, you can specify the exact list of URIs that can be called with the session.

Can create a new custom session as follows:

```
litcli  s a --label=blah --type=custom --uri_list=/lnrpc.Lightning/ListChannels,/poolrpc.Trader/GetInfo
```

Can test this with https://github.com/lightninglabs/lightning-node-connect/pull/59


session_rpcserver.go
```diff
@@ -203,10 +231,17 @@ func (s *sessionRpcServer) resumeSession(sess *session.Session) error {
 		Id: []byte(macExpiry.Condition),
 	})
 
+	if sess.Type == session.TypeMacaroonCustom {
```
guggero, line 234: Same here re `switch` statement.

litrpc/lit-sessions.proto
```diff
@@ -76,6 +76,8 @@ message Session {
     bytes remote_public_key = 10;
 
     uint64 created_at = 11 [jstype = JS_STRING];
+
+    repeated MacaroonPermission macaroon_custom_permissions = 12;
```
guggero, line 80: Should we show the full recipe here? So including the caveats?

cmd/litcli/sessions.go
```diff
@@ -115,6 +138,8 @@ func parseSessionType(sessionType string) (litrpc.SessionType, error) {
 		return litrpc.SessionType_TYPE_MACAROON_ADMIN, nil
 	case "readonly":
 		return litrpc.SessionType_TYPE_MACAROON_READONLY, nil
+	case "custom":
```
guggero, line 141: An itest with the custom session would be nice. Just to make sure the permissions actually work correctly.

guggero (commented):
Very nice! Looking forward to this feature :tada:

cmd/litcli/sessions.go
ellemouton: ah yes of course (completely slipped my mind)

ellemouton (commented):
Thanks for the review @guggero!!

Updated as per suggestion. Added 2 more commits. The first makes the run of the existing LNC auth itests more efficient and the second adds new tests for the custom macaroon perms.

Also left one question about custom caveats

litrpc/lit-sessions.proto
ellemouton: hmmm do we also want to add `CustomCaveats` to the `AddSessionRequest` then?

ellemouton: if we were to add the custom-caveats option to the request, then would it be fine just to let the user specify strings that we will use for the caveat `ID`? ie, is it fine to leave the location & verification ID empty ?

cmd/litcli/sessions.go
```diff
@@ -62,9 +63,18 @@ var addSessionCommand = cli.Command{
...
+				"include readonly|admin|custom",
 			Value: "readonly",
 		},
+		cli.StringSliceFlag{
+			Name: "uri",
+			Usage: "A URI that should be included in the " +
+				"macaroon of a custom session. Note that " +
+				"this flat will only be used if the 'type' " +
+				"flag is set to 'custom'. This flag can be " +
+				"specified multiple times if multiple URI's " +
```
guggero, line 75: nit: just URIs, no apostrophe.

itest/litd_mode_remote_test.go
```diff
@@ -151,6 +154,48 @@ func testModeRemote(net *NetworkHarness, t *harnessTest) {
 					ttt, rawLNCConn, endpoint.requestFn,
 					endpoint.successPattern,
 					endpoint.allowedThroughLNC,
+					"unknown service",
+				)
+			})
+		}
+	})
+
+	t.t.Run("lnc auth custom mac perms", func(tt *testing.T) {
```
guggero, line 163: Nice :+1:

guggero (approved):
Very cool, LGTM :tada:

guggero (approved):

jamaljsr (approved):
tACK LGTM 🔥

Works as advertised. I just have a question. I am testing this along with https://github.com/lightninglabs/lightning-node-connect/pull/59. I was able to add a custom session successfully. 

```
$ rt run zanelit sessions add --label mycustom --type=custom --uri=/lnrpc.Lightning/GetInfo --uri=/routerrpc.Router/GetMissionControlConfig
{
	"session": {
		"label": "mycustom",
		"session_state": "STATE_CREATED",
		"session_type": "TYPE_MACAROON_CUSTOM",
		"expiry_timestamp_seconds": "1674590241",
		"mailbox_server_addr": "mailbox.terminal.lightning.today:443",
		"dev_server": false,
		"pairing_secret": "43a472e678984e803ac31e4cba80",
		"pairing_secret_mnemonic": "dry castle ridge van lumber divorce twelve glow erosion staff",
		"local_public_key": "03e728a7ab09810c97b98d47cc792686ab294d139091bfdcca9dcca5bac1628afd",
		"remote_public_key": null,
		"created_at": "1666814241",
		"macaroon_recipe": {
			"permissions": [
				{
					"entity": "uri",
					"action": "/lnrpc.Lightning/GetInfo"
				},
				{
					"entity": "uri",
					"action": "/routerrpc.Router/GetMissionControlConfig"
				}
			],
			"caveats": [
			]
		}
	}
}
```

When I connected with the LNC example-server, the `lnrpc.State.GetState` RPC was still available. is this intentional? If so, why is that?

image

ellemouton:
@jamaljsr , yes that is expected because `lnrpc.State.GetState` does not require any permissions so will always be callable

PR #442, opened 2022-10-28
multi: introduce status server and sub-server manager

### LINKED ISSUE #290: litd exits with code 0 when it stops due to non-0 lnd exit
litd stops with a "successful" exit code 0 when it stops upon lnd exiting due to a failure and a non-0 exit status code.

**Expected behavior**

litd to exit with a non-0 exit code since it's stopping unexpectedly (and lnd presumably stopped unexpectedly as well).

**Actual behavior**

litd exits with a 0 exit code.

**To reproduce**

Have litd and lnd running - I have them running in systemd. Shutting down bitcoind causes lnd to exit with a failure status code while litd exits with a 0 code.

**System information**

I'm using litd binaries downloaded from github with a separate standalone lnd instance. This is on a rpi4 arm64.

Below are systemd status logs from lnd and litd (with some extra log output for litd) that shows lnd having exited with `1/FAILURE` while litd exits with `0/SUCCESS`.

```
$ sudo systemctl status lnd
● lnd.service - LND Lightning Daemon
   Loaded: loaded (/etc/systemd/system/lnd.service; enabled; vendor preset: enabled)
   Active: activating (auto-restart) (Result: exit-code) since Mon 2021-12-06 10:07:32 EST; 55s ago
  Process: 21935 ExecStart=/home/bitcoin/go/bin/lnd (code=exited, status=1/FAILURE)
 Main PID: 21935 (code=exited, status=1/FAILURE)
```

```
$ sudo systemctl status lit
● lit.service - LND Lightning Terminal
   Loaded: loaded (/etc/systemd/system/lit.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Mon 2021-12-06 10:06:51 EST; 3min 24s ago
  Process: 15969 ExecStart=/home/bitcoin/lit/litd (code=exited, status=0/SUCCESS)
 Main PID: 15969 (code=exited, status=0/SUCCESS)

2021-12-06 10:06:48.679 [INF] SGNL: Received terminated
2021-12-06 10:06:48.820 [INF] SGNL: Shutting down...
2021-12-06 10:06:48.856 [INF] SGNL: Gracefully shutting down.
2021-12-06 10:06:48.857 [INF] LITD: Shutdown signal received
2021-12-06 10:06:48.883 [INF] LOOPD: Stopping gRPC server
2021-12-06 10:06:48.883 [INF] LOOPD: Stopping REST server
2021-12-06 10:06:48.883 [INF] LOOPD: Liquidity manager stopped
2021-12-06 10:06:48.885 [INF] LOOP: Swap client terminating
2021-12-06 10:06:48.885 [INF] LOOP: Swap client terminated
2021-12-06 10:06:48.885 [INF] LOOPD: Swap client stopped
2021-12-06 10:06:48.927 [ERR] ACCT: Unable to receive block notification: rpc error: code = Unavailable desc = error reading from server: read tcp 127.0.0.1:39338->127.0.0.1:10009: use of closed network connection
2021-12-06 10:06:48.884 [ERR] RPCS: Unable to receive block notification: rpc error: code = Unavailable desc = error reading from server: read tcp 127.0.0.1:39338->127.0.0.1:10009: use of closed network connection
2021-12-06 10:06:48.979 [ERR] RPCS: Unable to receive block notification: channel acceptor receive failed: rpc error: code = Unavailable desc = error reading from server: read tcp 127.0.0.1:39338->127.0.0.1:10009: use of closed network connection
2021-12-06 10:06:48.979 [INF] LOOPD: Daemon exited
2021-12-06 10:06:48.979 [INF] POOL: Received shutdown signal, stopping server
2021-12-06 10:06:48.979 [INF] AUCT: Shutting down auctioneer client
2021-12-06 10:06:48.979 [INF] RPCS: Trader server stopping
Dec 06 10:06:48 satoshpi litd[15969]: 2021-12-06 10:06:48.979 [INF] AUCT: Shutting down auctioneer client
Dec 06 10:06:48 satoshpi litd[15969]: 2021-12-06 10:06:48.980 [INF] FNDG: Funding manager stopping
Dec 06 10:06:48 satoshpi litd[15969]: 2021-12-06 10:06:48.980 [INF] FNDG: Stopped funding manager
Dec 06 10:06:48 satoshpi litd[15969]: 2021-12-06 10:06:48.981 [INF] RPCS: Stopped trader server
Dec 06 10:06:48 satoshpi litd[15969]: 2021-12-06 10:06:48.981 [ERR] LNDC: Error closing lnd connection: rpc error: code = Canceled desc = grpc: the client connection is closing
Dec 06 10:06:51 satoshpi litd[15969]: 2021-12-06 10:06:51.067 [ERR] LNDC: Error closing lnd connection: rpc error: code = Canceled desc = grpc: the client connection is closing
Dec 06 10:06:51 satoshpi systemd[1]: lit.service: Succeeded.
Dec 06 10:06:51 satoshpi systemd[1]: Stopped LND Lightning Terminal.
Dec 06 10:07:32 satoshpi systemd[1]: Dependency failed for LND Lightning Terminal.
Dec 06 10:07:32 satoshpi systemd[1]: lit.service: Job lit.service/start failed with result 'dependency'.
```

### LINKED ISSUE #420: Show LND/LiT version mismatch error in Web UI
Yes, I installed LiT v0.8.0-alpha on my Umbrel where I had LND v0.5.0-beta installed, and the LiT container fails to start due to this error: `2022-09-16 10:41:36.067 [ERR] LITD: Could not start subservers: error checking connected lnd version. at least version "v0.15.1, build tags 'signrpc,walletrpc,chainrpc,invoicesrpc'" is required`. A user wouldn't know what the issue is without digging into the container logs.

Could we make it so this is error (and potentially others) are shown in the Web UI so that user can more easily debug their setup when something is misconfigured, a version mismatch, etc.

The only alternative atm is for a user to access the container logs to see this error.

### LINKED ISSUE #421: Show dependency issues in the UX
### Problem
As currently architected, at `litd` start up, in order to open the UX, all of the subservers need to startup meaning all of the `litd` dependencies (LND version, etc.) have to be aligned. With that behavior, for users using GUI experiences because of ease or lack of technical skills, when a `litd` update occurs that changes dependencies, they encounter an issue where the app or experience doesn't open due to dependency mismatch. However, given this user type, it can be difficult for them to debug the issue and understand the problem and solution. 

### Potential Solution
We could re-architect litd so that it can actually "startup" without all of the subservers being activated. With this architecture, we could show the user the relevant errors or dependency issues so that they can more easily debug. The example referenced in the attached issue would mean that we show the user in the UX that they need to be running `lnd 0.15.1` for litd to work properly. But, we've also run into issues in the past whereby the entire `litd` experience doesn't start because of a single subserver issue (Pool most of the time). With this architecture, we might be able to have everything else work, but in the Pool UX flag the error to the user to help them and us debug it. This would probably require an always-on RPC for the UI to obtain the current status or startup errors. 

An additional feature that could be a part of this is being able to configure the subservers from the UI. Then, there would be less flags that need to be specified via the CLI.

Referenced Issue: https://github.com/lightninglabs/lightning-terminal/issues/420

ellemouton:
This PR does a few things:

1. Adds a new `subServerMgr` that handles the connections to Loop, Pool & Faraday. This will also make it easy for us to add another sub server in future. This is also nice because it makes it easier in the next steps to treat these sub-servers ,& any errors we may get from them, differently to how we would treat errors from LND or LiTd. 
2. Adds a new `Status` server that is "always-on". This will allow the UI to query info about the status of any of LiTs sub-servers (including LND, Loop, Pool, Faraday and LiT itself).
3. Does a big refactor so that Litd does not ever shutdown unless requested to via a shutdown signal. This ensures that the `Status` server can always be reached.

```
$ litcli status 

{
        "sub_servers": {
                "faraday": {
                        "running": true,
                        "error": ""
                },
                "lit": {
                        "running": true,
                        "error": ""
                },
                "lnd": {
                        "running": true,
                        "error": ""
                },
                "loop": {
                        "running": true,
                        "error": ""
                },
                "pool": {
                        "running": false,
                        "error": "could not start pool: <something something>"
                }
        }
}
```

TODO:
- [ ] documentation explaining the difference (and reasoning) between the lnd 10009 (gRPC only) and the litd 8443
- [ ] More itests to cover the startup process

Fixes #421 
Fixes #420
Fixes #290


ellemouton:
@guggero - as discussed offline, I have requested your review here but am mainly looking for approach comments to begin with. And also, no rush on this as I will be afk for the next 2 weeks

config.go
```diff
@@ -150,6 +150,9 @@ type Config struct {
 	LetsEncryptDir    string `long:"letsencryptdir" description:"The directory where the Let's Encrypt library will store its key and certificate."`
 	LetsEncryptListen string `long:"letsencryptlisten" description:"The IP:port on which LiT will listen for Let's Encrypt challenges. Let's Encrypt will always try to contact on port 80. Often non-root processes are not allowed to bind to ports lower than 1024. This configuration option allows a different port to be used, but must be used in combination with port forwarding from port 80. This configuration can also be used to specify another IP address to listen on, for example an IPv6 address."`
 
+	TLSCertPath string `long:"tlscertpath" description:"Path to write the self signed TLS certificate for LiT's RPC and REST proxy service (if Let's Encrypt is not used)."`
```
guggero, line 153: Do we need to explicitly mention this only applies to the `HTTP(S)Listen` port (default 8443)? Just to distinguish it more from the 10009 lnd RPC port.

config.go
```diff
@@ -280,9 +280,9 @@ func (c *Config) lndConnectParams() (string, lndclient.Network, string,
 func defaultConfig() *Config {
 	return &Config{
 		HTTPSListen: defaultHTTPSListen,
+		TLSCertPath: DefaultTLSCertPath,
```
guggero, line 283: We should also update `doc/config-lnd-integrated.md` and perhaps explain the difference (and reasoning) between the `lnd 10009` (gRPC only) and the `litd 8443` (grpc-web, REST, gRPC proxy, UI) a bit more explicitly?
Also this change should be made very explicit in the release notes (not sure where to put the reminder for us).

rpc_proxy.go
```diff
@@ -253,13 +263,26 @@ func (p *rpcProxy) Stop() error {
...
+
 // isHandling checks if the specified request is something to be handled by lnd
 // or any of the attached sub daemons. If true is returned, the call was handled
 // by the RPC proxy and the caller MUST NOT handle it again. If false is
 // returned, the request was not handled and the caller MUST handle it.
 func (p *rpcProxy) isHandling(resp http.ResponseWriter,
 	req *http.Request) bool {
 
+	if !p.hasStarted() {
+		return false
```
guggero, line 208: I'm not sure if returning `false` here is a good idea, as the request will just be passed to the next handler. Instead I think we should check whether it's a request we _would_ handle (the same conditions as we check below) and return an error page (perhaps 503 service unavailable) if we did and only `false` if we didn't?

terminal.go
```diff
@@ -1437,6 +1450,20 @@ func (g *LightningTerminal) showStartupInfo() error {
 	return nil
 }
 
+// connectLND sets up LiT's LND connection.
+func (g *LightningTerminal) connectLND(bufListener *bufconn.Listener) error {
```
guggero, line 1510: Could turn this into a function by taking the `cfg` as a parameter and returning the conn. Then we could place this function and the `dialBackend()` and `dialBufConnBackend()` into a different file to reduce the size of this one a bit. Maybe `lnd_connection.go` or something like that?

rpc_proxy.go
```diff
@@ -168,21 +166,9 @@ type rpcProxy struct {
 }
 
 // Start creates initial connection to lnd.
-func (p *rpcProxy) Start() error {
+func (p *rpcProxy) Start(lndConn *grpc.ClientConn) error {
```
guggero, line 172: :+1:

terminal.go
```diff
@@ -400,12 +400,39 @@ func (g *LightningTerminal) Run() error {
 		g.cfg.lndAdminMacaroon = <-macChan
 	}
 
-	err = g.startSubservers()
+	// Set up all the LND clients required by LiT.
+	err = g.setUpLNDClients()
 	if err != nil {
-		log.Errorf("Could not start subservers: %v", err)
+		log.Errorf("Could not set up LND clients: %v", err)
```
guggero: nit: we should start wrapping all errors with `%w`.

terminal.go
```diff
@@ -474,7 +499,7 @@ func (g *LightningTerminal) startSubservers() error {
 		return err
 	}, defaultStartupTimeout)
 	if err != nil {
-		return err
+		return fmt.Errorf("could not create basid LND Client: %v", err)
```
guggero, line 676: nit: typo.

terminal.go
```diff
@@ -543,22 +569,17 @@ func (g *LightningTerminal) startSubservers() error {
...
-		macService := g.lndInterceptorChain.MacaroonService()
-		createDefaultMacaroons = !macService.StatelessInit
-	}
+// startIntegratedDaemons starts all embedded daemons as external sub-servers
+// that hook into the same gRPC and REST servers that lnd started.
+func (g *LightningTerminal) startIntegratedDaemons(
+	createDefaultMacaroons bool) error {
 
-	// Both connection types are ready now, let's start our subservers if
+	var err error
```
guggero: nit: no need to declare the variable here, can just use `:=` below.

terminal.go
```diff
@@ -211,15 +211,22 @@ func (g *LightningTerminal) Run() error {
...
 	}
 
+	// Construct the rpcProxy. It must be initialised before the main web
+	// server is started.
+	g.rpcProxy = newRpcProxy(g.cfg, g, g.validateSuperMacaroon, g.permsMgr)
+
+	// Start the main web server that dispatches requests either to the
+	// static UI file server or the RPC proxy. This makes it possible to
+	// unlock lnd through the UI.
+	if err := g.startMainWebServer(); err != nil {
```
guggero, line 221: Since we start the main web server this early, we have to guard the `g.restHandler` from being called if it's nil (. Otherwise a REST request coming in too early would panic. Or we move the call to `createRESTProxy()` to before this call even.

guggero: Hmm, no, creating the REST proxy before the main web server is started will probably not work since the REST proxy will attempt to connect to the main web server on creation.

guggero: That makes me wonder if we could extract the main web server into its own struct that owns the REST proxy and RPC proxy (and probably also the status server)? Then the start/stop would be a bit more abstracted as well. But not sure how large of a diff that is.

state_server.go
```diff
@@ -0,0 +1,185 @@
...
+		m: map[string]*subServerStatus{
+			LitSubServer:     newSubServerStatus(),
+			LNDSubServer:     newSubServerStatus(),
+			FaradaySubServer: newSubServerStatus(),
+			PoolSubServer:    newSubServerStatus(),
+			LoopSubServer:    newSubServerStatus(),
+		},
+	}
+
+	options := []grpcweb.Option{
```
guggero, line 59: Do we really need to spin up our own grpc-web and gRPC server? Can't we instead have this as a sub component of the `rpcProxy` that already does all of this for us?

state_server.go
```diff
@@ -0,0 +1,185 @@
...
+func (s *statusServer) isHandling(resp http.ResponseWriter,
+	req *http.Request) bool {
+
+	if !strings.Contains(req.URL.Path, "/litrpc.Status") {
+		return false
+	}
+
+	// gRPC web requests are easy to identify. Send them to the gRPC
+	// web proxy.
+	if s.grpcWebProxy.IsGrpcWebRequest(req) ||
```
guggero, line 166: Again, it feels like this should just be hooked up in the right order to allow the existing `rpcProxy` to do this.

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/lightningnetwork/lnd/lncfg"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/macaroons"
+	"google.golang.org/grpc"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+)
+
+// subServerCfg defines the config values that must be set in order to properly
+// define a subServer.
+type subServerCfg struct {
```
guggero, line 18: Nice, I like this idea :+1:

terminal.go
```diff
@@ -792,83 +836,18 @@ func (g *LightningTerminal) ValidateMacaroon(ctx context.Context,
...
-		if g.cfg.poolRemote {
-			return nil
-		}
-
-		if !g.poolStarted {
-			return fmt.Errorf("pool is not yet ready for " +
-				"requests, lnd possibly still starting or " +
-				"syncing")
-		}
+	handled, err := g.subServerMgr.ValidateMacaroon(
```
guggero, line 903: We'll need to make sure we don't create any authentication by-passes by accident, for example when starting up the individual daemons. Maybe adding a few extra itests around the whole startup process would be a good idea.

guggero (commented):
Very nice work :100: This is an important step towards making it possible to unlock `lnd` from the UI and also to make everything more user friendly.
Will need to do a second pass with a focus on security but looks pretty good for its WIP status!

ellemouton (commented):
thanks for the initial review @guggero! 

I have addressed most of your comments. 

I just still need to add the extra documentation& also add more itests around the startup process. Will re-request once I have done that.

ellemouton:
!lightninglabs-deploy mute

ellemouton:
muting for the time being. Will pick up again one I have more room on my plate

rpc_proxy.go
```diff
@@ -165,6 +166,9 @@ type rpcProxy struct {
 
 	grpcServer   *grpc.Server
 	grpcWebProxy *grpcweb.WrappedGrpcServer
+
+	started bool
```
Roasbeef, line 166: Could use an atomic bool alternatively.

Roasbeef: Still getting thru the diff, but could also use an event type to allow callers to get notifications for when things are fully up vs polling.

rpc_proxy.go
```diff
@@ -279,17 +298,27 @@ func (p *rpcProxy) isHandling(resp http.ResponseWriter,
 	if p.grpcWebProxy.IsGrpcWebRequest(req) ||
 		p.grpcWebProxy.IsGrpcWebSocketRequest(req) {
 
-		log.Infof("Handling gRPC web request: %s", req.URL.Path)
-		p.grpcWebProxy.ServeHTTP(resp, req)
+		if !p.hasStarted() {
+			resp.WriteHeader(http.StatusServiceUnavailable)
```
Roasbeef, line 222: Why this vs a gRPC status/error like we do on the `lnd` side?

rpc_proxy.go
```diff
@@ -279,17 +298,27 @@ func (p *rpcProxy) isHandling(resp http.ResponseWriter,
...
 	}
 
 	// Normal gRPC requests are also easy to identify. These we can
 	// send directly to the lnd proxy's gRPC server.
 	if isGrpcRequest(req) {
-		log.Infof("Handling gRPC request: %s", req.URL.Path)
-		p.grpcServer.ServeHTTP(resp, req)
+		if !p.hasStarted() {
+			resp.WriteHeader(http.StatusServiceUnavailable)
+			_, _ = resp.Write(make([]byte, 0))
```
Roasbeef, line 237: Comment above may moreso apply here. Another route would be an http middleware chain (gRPC package) that checked this field and returned a response if things weren't fully active yet.

rpc_proxy.go
```diff
@@ -61,7 +56,7 @@ func (e *proxyErr) Unwrap() error {
 // component.
 func newRpcProxy(cfg *Config, validator macaroons.MacaroonValidator,
 	superMacValidator session.SuperMacaroonValidator,
-	permsMgr *perms.Manager, bufListener *bufconn.Listener) *rpcProxy {
+	permsMgr *perms.Manager) *rpcProxy {
```
Roasbeef: Arg passed but not used anywhere (in this commit at least).

litrpc/lit-status.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+package litrpc;
+
+option go_package = "github.com/lightninglabs/lightning-terminal/litrpc";
+
+// The Status server can be used to query the state of various LiT sub-servers.
+service Status {
+    rpc GetSubServerState (GetSubServerStatusReq)
```
Roasbeef, line 9: Could just drop the `Get`?

state_server.go
```diff
@@ -0,0 +1,125 @@
...
+	LoopSubServer    string = "loop"
+	FaradaySubServer string = "faraday"
+)
+
+// statusServer is an implementation of the litrpc.StatusServer which can be
+// queried for the status of various LiT sub-servers.
+type statusServer struct {
+	litrpc.UnimplementedStatusServer
+
+	m  map[string]*subServerStatus
```
Roasbeef, line 24: `subServers`?

state_server.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/lightninglabs/lightning-terminal/litrpc"
+)
+
+const (
+	LitSubServer     string = "lit"
```
Roasbeef, line 12: Not strictly required to declare them all as `string` here.

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+	// name is the string identifier of the sub-server.
+	name string
+
+	// remote must be set if the sub-server is running remotely and so must
+	// not be started in integrated mode.
+	remote bool
+
+	// startIntegrated defines how the sub-server should be started if it
+	// is to be run in integrated mode.
+	startIntegrated func(lnrpc.LightningClient, *lndclient.GrpcLndServices,
```
Roasbeef, line 28: What about lifting most of these closures out into an interface? would make it easier to implement test, and also wrap/implement some of the existing sub-servers with. 

It would also make it clearer imo exactly what a sub-server _is_ from the PoV of the manager.

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+			returnErr = fmt.Errorf("could not close "+
+				"integrated connection: %v", err)
+			return
+		}
+
+		if s.cfg.serverErrChan == nil {
+			return
+		}
+
+		returnErr = <-s.cfg.serverErrChan
```
Roasbeef, line 123: Should also select on quite otherwise can block indefinitely here?

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+	})
+
+	return returnErr
+}
+
+// startIntegrated starts the subServer in integrated mode.
+func (s *subServer) startIntegrated(lndClient lnrpc.LightningClient,
```
Roasbeef, line 130: Meant to be generic but uses the lnd client directly here?

Roasbeef: Or the assumption is that all/most of the sub-servers need to interact w/ lnd anyway?

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+	lndGrpc *lndclient.GrpcLndServices, withMacaroonService bool) {
+
+	err := s.cfg.startIntegrated(lndClient, lndGrpc, withMacaroonService)
+	if err != nil {
+		s.cfg.onStartError(err)
+		return
+	}
+	s.integratedStarted = true
```
Roasbeef, line 138: Doesn't need a mutex to protect access?

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+	s.cfg.onStartSuccess()
+
+	if s.cfg.serverErrChan == nil {
+		return
+	}
+
+	go func() {
```
Roasbeef, line 145: Should be tracked by a wait group.

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+// newSubServer constructs a new subServer using the given config.
+func newSubServer(cfg *subServerCfg) *subServer {
+	return &subServer{
+		cfg:  cfg,
+		quit: make(chan struct{}),
+	}
+}
+
+// subServerMgr manages a set of subServer objects.
+type subServerMgr struct {
```
Roasbeef, line 186: 😎

rpc_proxy.go
```diff
@@ -152,10 +155,10 @@ type rpcProxy struct {
 
 	superMacaroon string
 
-	lndConn     *grpc.ClientConn
-	faradayConn *grpc.ClientConn
-	loopConn    *grpc.ClientConn
-	poolConn    *grpc.ClientConn
+	lndConn *grpc.ClientConn
```
Roasbeef, line 158: Any reason lnd isn't also managed as a sub-server? I guess since it's sort of the "root" sub-server that all the rest depend on?

terminal.go
```diff
@@ -489,6 +472,123 @@ func (g *LightningTerminal) Run() error {
...
+		macValidator: poolServer,
+		ownsURI:      g.permsMgr.IsPoolURI,
+		registerGrpcService: func(server grpc.ServiceRegistrar) {
+			poolrpc.RegisterTraderServer(server, poolServer)
+		},
+		startIntegrated: func(lndClient lnrpc.LightningClient,
+			lndGrpc *lndclient.GrpcLndServices,
+			withMacaroonService bool) error {
+
+			return poolServer.StartAsSubserver(
```
Roasbeef, line 637: Re that comment re converting many of the closures to be an interface, given this is just a pass thru call (same args and return value), a wrapper struct that embeds all the main daemon configs would take care of this nicely.

Roasbeef (changes requested):

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+	registerGrpcService func(grpc.ServiceRegistrar)
+
+	// macValidator will be used to verify any macaroons for requests that
+	// are to be handled by the sub-server.
+	macValidator macaroons.MacaroonValidator
+
+	// onStartError will be called if there is a startup error (in
+	// integrated mode) or if the remote connection could not be made in
+	// remote mode.
+	onStartError func(err error)
```
positiveblue, line 46: Do we plan `onStartError`, `onStartSuccess` to do more things in the future beyond setting the status for this subserver? If not we can directly set the status using `subServerCfg.name`

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+
+// subServerCfg defines the config values that must be set in order to properly
+// define a subServer.
+type subServerCfg struct {
+	// name is the string identifier of the sub-server.
+	name string
+
+	// remote must be set if the sub-server is running remotely and so must
+	// not be started in integrated mode.
+	remote bool
```
positiveblue, line 24: here `remote == (remoteCfg != nil)`?

subserver_mgr.go
```diff
@@ -0,0 +1,399 @@
...
+			continue
+		}
+
+		ss.startIntegrated(lndClient, lndGrpc, withMacaroonService)
+	}
+}
+
+// StartRemoteSubServers creates connections to all the manager's sub-servers
+// that are running remotely.
+func (s *subServerMgr) StartRemoteSubServers() {
```
positiveblue, line 225: The func name can be more accurate if we use something like `ConnectRemoteSubServers` given that we are not starting anything.

positiveblue (commented):
Thank you for these changes @ellemouton , they'll make my life easier for integrating the taro daemon :100:

terminal.go
ellemouton: turned out nicely I think

ellemouton:
I need to switch away from this for a bit now so just making a note of some issues that the current version still has:

- THe itests seem to not exit
- we currently construct the autopilot server and session server before getting to the registration of rpc servers. This means that if the setup of those fail, then we cant access the status server to see this. And the reason we ordered it this way is so that LND can register non-nil servers. so a bit of a catch 22 here... 
- I think to solve the above, we might need to add a new grpc server that just servers the status server. So that it is not dependent on the start up of LND.

state_server.go
ellemouton: Just realised why I did it this way initially - the reason is that if we fail to start LND, then we never get to the registration and serving of the status server. So I think we need to re-add this actually

ellemouton:
replaced by https://github.com/lightninglabs/lightning-terminal/pull/516 & https://github.com/lightninglabs/lightning-terminal/pull/541

PR #444, opened 2022-10-31
Add option to disable the local UI

gkrizek:
This PR adds a new flag called `--disableui`. When this flag is set, it tells the local Terminal instance to not display any of its local UI. In this state, all APIs still function just as they did previously. This is helpful if you only want to generate LNC phrases via the API or CLI to then use them at places like Terminal Web.

### DIFF
--- a/config.go
+++ b/config.go
@@ -144,6 +144,7 @@ type Config struct {
 	UIPassword     string   `long:"uipassword" description:"The password that must be entered when using the loop UI. use a strong password to protect your node from unauthorized access through the web UI."`
 	UIPasswordFile string   `long:"uipassword_file" description:"Same as uipassword but instead of passing in the value directly, read the password from the specified file."`
 	UIPasswordEnv  string   `long:"uipassword_env" description:"Same as uipassword but instead of passing in the value directly, read the password from the specified environment variable."`
+	DisableUI      bool     `long:"disableui" description:"Disable the local Terminal UI so only gRPC and/or REST API access is allowed."`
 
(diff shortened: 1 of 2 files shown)


ellemouton:
Hi @gkrizek ! Thanks for the PR. Im curious: why is an option needed for this? why not just use the CLI & not use the UI?

guggero:
This implementation is also very unsafe because it sets the UI password to `placeholder`. But the password is also used as an authentication method for `grpc-web` and REST calls, so those will continue to work but with this very weak password.

guggero:
Yes, the super macaroon can be used for the CLI, which makes things easier. We just haven't upgraded the UI to use the macaroon too (has some UX related things that would need to be added).
So I'm just saying that we would need to disable the UI authentication a bit more thoroughly to be safe, but think it's a good idea to have an option to disable it in general.

There is no API doc available just yet, only the `litrpc/lit-sessions.proto` file for now.

gkrizek:
Ok thanks @guggero , I’ll take a look at disabling that auth more completely.

ellemouton:
Hi @gkrizek ! Thanks again for the PR & for the idea. Getting rid of the UI-password is defs something we have been wanting to do . Could I possibly ask you to open an issue for this rather? Then I can assign myself & work on the impl as this change will touch quite a few sensitive parts of the codebase. It will also give me an opportunity to add to the Litd integration tests in the same PR :)

gkrizek:
@ellemouton Yes, I can make an issue today describing my request/use case. Thanks for the help!

gkrizek:
Created! https://github.com/lightninglabs/lightning-terminal/issues/448

PR #450, opened 2022-11-04
multi: support using a regex when specifying custom perms

### LINKED ISSUE #439: Wildcard permissions support
As of #438 it is now possible to create custom macaroons with permission to access specific URIs. However, there are hundreds of possible URIs available in `lnrpc`, `looprpc`, `poolrpc`, and `frdrpc`. 

It may be useful in some situations to allow wildcard permissions, for example, to allow access to all of the URIs in `poolrpc` while still enforcing more granular control over `lnrpc`.

```
litcli  s a \
    --label=Example \
    --type=custom \
    --uri_list=/lnrpc.Lightning/ListPeers,/lnrpc.Lightning/ListChannels,/poolrpc.Trader/*
```

ellemouton:
With this PR, you can now use the wild card character when specifying custom URIs that permissions should be added for in an LNC session. This allows you to do the following:

```
litcli s a --label=custom_perm_session  --type=custom --uri=/lnrpc.Lightning/ListChannels --uri="/poolrpc.Trader/*"
```

The above will create a session that has the perms to only access the `ListChannels` call of the `lnrpc.Lightning` but then also has the perms to access any URI that falls under the `poolrpc.Trader` service. 
`
Fixes #439

### DIFF
--- a/cmd/litcli/sessions.go
+++ b/cmd/litcli/sessions.go
@@ -73,7 +73,11 @@ var addSessionCommand = cli.Command{
 				"this flag will only be used if the 'type' " +
 				"flag is set to 'custom'. This flag can be " +
 				"specified multiple times if multiple URIs " +
-				"should be included",
+				"should be included. Note that a regex can " +
+				"also be specified which will then result in " +
+				"all URIs matching the regex to be included. " +
+				"For example, '/lnrpc\\..*' will result in " +
+				"all `lnrpc` permissions being included.",
 		},
 	},
 }
--- a/perms/permissions_test.go
+++ b/perms/permissions_test.go
@@ -1,0 +1,81 @@
+package perms
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+)
+
+// TestMatchRegexURI tests the behaviour of the MatchRegexURI method of the
+// Manager.
+func TestMatchRegexURI(t *testing.T) {
+	// Construct a new Manager with a predefined list of perms.
+	m := &Manager{
+		perms: map[string][]bakery.Op{
+			"/lnrpc.WalletUnlocker/GenSeed":    {},
+			"/lnrpc.WalletUnlocker/InitWallet": {},
+			"/lnrpc.Lightning/SendCoins": {{
+				Entity: "onchain",
+				Action: "write",
+			}},
+			"/litrpc.Sessions/AddSession": {{
+				Entity: "sessions",
+				Action: "write",
+			}},
+			"/litrpc.Sessions/ListSessions": {{
+				Entity: "sessions",
+				Action: "read",
+			}},
+			"/litrpc.Sessions/RevokeSession": {{
+				Entity: "sessions",
+				Action: "write",
+			}},
+		},
+	}
+
+	// Assert that a full URI is not considered a wild card.
+	uris, isRegex := m.MatchRegexURI("/litrpc.Sessions/RevokeSession")
+	require.False(t, isRegex)
+	require.Empty(t, uris)
+
+	// Assert that an invalid URI is also caught as such.
+	uris, isRegex = m.MatchRegexURI("***")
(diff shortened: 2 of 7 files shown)


cmd/litcli/sessions.go
```diff
@@ -73,7 +73,11 @@ var addSessionCommand = cli.Command{
 				"this flag will only be used if the 'type' " +
 				"flag is set to 'custom'. This flag can be " +
 				"specified multiple times if multiple URIs " +
-				"should be included",
+				"should be included. Note that a wild-card " +
+				"URI can also be specified to include all " +
+				"URIs of a specific server, for example: " +
+				"'--uri=\"/poolrpc.Trader/*\"' will include " +
```
guggero, line 79: nit: this pattern technically isn't a valid regex but a "glob" (or just pattern). But maybe we actually want it to be a full regex so we don't need to do the matching/replacing below? Then I could also select all `lnrpc` permissions with `/lnrpc\..*`?

And does it make sense to print the matched URIs to the console just to show the user what permissions they get in the session?

session_rpcserver.go
```diff
@@ -143,18 +143,48 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
...
+				})
+
+				continue
+			}
+
+			// First check if this is a wildcard URI.
+			uris, isWildCard := s.cfg.permMgr.ExpandWildCardURI(
+				op.Action,
+			)
+			if !isWildCard {
```
guggero, line 159: nit: invert this `if` so the wildcard logic comes directly after the `ExpandWildCardURI` and the default case is non-indented?

guggero (commented):
Nice work! Pretty straightforward implementation, just an idea for making it even more generic.

cmd/litcli/sessions.go
ellemouton: > But maybe we actually want it to be a full regex

Great idea!

> And does it make sense to print the matched URIs to the console just to show the user what permissions they get in the session?

It does currently print them out to the console since the URIs are listed in the macaroon recipe which is printed out. Do you mean that perhaps we should add like a "here are the perms you selected, are you sure you want to continue [y/n] type thing?"

ellemouton: > But maybe we actually want it to be a full regex

Done!

guggero (approved):

ellemouton:
@itsrachelfish , this is ready for a final review :)

perms/permissions_test.go
```diff
@@ -0,0 +1,81 @@
...
+	})
+
+	// Now we check that we can include all the `lnrpc` methods.
+	uris, isRegex = m.MatchRegexURI("/lnrpc\\..*")
+	require.True(t, isRegex)
+	require.ElementsMatch(t, uris, []string{
+		"/lnrpc.WalletUnlocker/GenSeed",
+		"/lnrpc.WalletUnlocker/InitWallet",
+		"/lnrpc.Lightning/SendCoins",
+	})
```
itsrachelfish, line 74: This code works and the tests pass, but I don't think `m.MatchRegexURI("/lnrpc.WalletUnlocker/*")` is doing what you think it does. The regex `/lnrpc.WalletUnlocker/*` matches all of the following strings:

```
/lnrpc.WalletUnlocker/GenSeed
/lnrpc.WalletUnlocker/InitWallet
/lnrpc.WalletUnlocker
/lnrpc.WalletUnlocker///////////////////////////
```

itsrachelfish (approved):
tACK, regex support works, but the examples & tests could be updated to be more clear.

perms/permissions_test.go
itsrachelfish: You could use the regex `m.MatchRegexURI("/lnrpc.WalletUnlocker")` and it would have the same effect

ellemouton: sure - can update to be more clear :+1:

ellemouton: updated to match on "/lnrpc.WalletUnlocker/.*" rather

PR #451, opened 2022-11-04
Custom user permissions UI

itsrachelfish:
This PR updates the Lightning Node Connect page to include a new overlay which allows users to create macaroons with customized permissions for specific URIs as well as expiration dates.

## New option added to permissions dropdown
image

## Custom permissions overlay
image

## Custom session with an expiration date
image

Related to: #438
Depends on #450 & #457


app/src/components/common/v2/Text.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+const BaseText = styled.span<TextProps>`
+  // On mobile, default to semi-bold when the bold prop is used
+  font-family: ${props =>
+    props.bold || props.semiBold
+      ? props.theme.fonts.open.semiBold
+      : props.theme.fonts.open.regular};
+  font-weight: ${props => (props.bold || props.semiBold ? 600 : 400)};
+  ${props => props.muted && `color: ${props.theme.colors.gray};`}
```
jbrill, line 35: Should we stay consistent with
`color: ${props => (props.muted && 'props.theme.colors.gray')}`

app/src/components/common/v2/Text.tsx
```diff
@@ -0,0 +1,152 @@
...
+  ${props => props.space && `margin-bottom: ${props.space}px;`}
```
jbrill, line 36: Same as above

jbrill:
One future thing that could be helpful is to create a common component library as a node package for us to consume in both terminal and lit. This way we could avoid duplicated code, and maybe create a storybook for @dstrukt to view what's being used in the codebase

jbrill:
Should `Liquidity Manager` and `Payments Manager` permission roles be in the dropdown as well?

levmi:
> Should `Liquidity Manager` and `Payments Manager` permission roles be in the dropdown as well?

Design/product decision to just have those live in custom as well. Want to see if people use them or not. Will populate the dropdown with the most popular options based on feedback. Still unsure how popular those two will be, but it's our first stab at it.

jbrill:
I can imagine the terminal lift will be significant for some of these custom permissions -- would those changes happen before the merge? Just curious

jbrill:
The code looks very clean and can confirm the feature works very well 👍 

Will do a more thorough pass when it's ready for review!

app/src/components/common/FormDate.tsx
```diff
@@ -0,0 +1,71 @@
...
+  className,
+  onChange,
+}) => {
+  const { Wrapper, Input, Extra } = Styled;
+  return (
+    <Wrapper className={className}>
+      <Input
+        type="date"
+        value={value}
+        onChange={e => onChange && onChange(e.target.value)}
```
jbrill, line 62: Do we want to use the `useCallback` format similar to what's used in Terminal?

app/src/components/connect/CustomSessionPage.tsx
```diff
@@ -0,0 +1,352 @@
...
+  const handleBack = useCallback(() => {
+    addSessionView.cancel();
+    appView.goTo('/connect');
+  }, [appView]);
+
+  const handleSubmit = useCallback(async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    addSessionView.cancel();
+
+    appView.notify('(coming soon)', 'Custom session created', 'success');
```
jbrill, line 98: (reminder to tweak)

dstrukt:
> One future thing that could be helpful is to create a common component library as a node package for us to consume in both terminal and lit. This way we could avoid duplicated code, and maybe create a storybook for @dstrukt to view what's being used in the codebase

This is interesting .. ACK for the shared library, and storybook if we keep it low effort export for time being.

app/src/store/views/addSessionView.ts
```diff
@@ -0,0 +1,264 @@
...
+          // Add all of the URIs for this permission
+          permissions.push(...this.permissionMap[permissionName]);
+        }
+      });
+
+      // Convert all of the permission strings into MacaroonPermission objects
+      return permissions.map(uri => {
+        const mp = new LIT.MacaroonPermission();
+        mp.setEntity(uri);
+        mp.setAction('uri');
```
ellemouton, line 109: these are the wrong way around :)

ellemouton (changes requested):

dstrukt:
Design Review:

- [ ] Add `24px` padding above Proxy Server input
Screen Shot 2022-11-28 at 12 10 16 PM

- [ ] Make the calendar icon our `offWhite` 
Screen Shot 2022-11-28 at 12 14 24 PM

- [ ] Make the calendar text copy click trigger the dropdown, only the icon does this (ideal, but not required)

- [ ] Conversely, the icon click on the select dropdown should trigger the dropdown too .. but this is an issue across LITD/TW, so another ideal, but not required, and might be worth a sep. issue, but defer to you

- [ ] Add a `200ms ease-in-out` transition to the background-color when hovering/selecting a Permission Type (just a nice to have)

dstrukt (changes requested):
See above.

app/src/components/common/v2/FormSwitch.tsx
```diff
@@ -0,0 +1,30 @@
...
+  Switch: styled(Switch)`
+    &.rc-switch-checked {
+      border: 1px solid ${props => props.theme.colors.iris};
+      background-color: ${props => props.theme.colors.iris};
+    }
+  `,
+};
+
+interface Props {
+  checked?: boolean;
```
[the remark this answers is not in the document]
itsrachelfish: Yep :+1:

app/src/components/common/FormDate.tsx
itsrachelfish: This was copied from the FormInput component. If we change it here we should probably update all of the input components to be consistent?

app/src/components/common/v2/Text.tsx
itsrachelfish: The difference is that the font weight always has a value. If the muted property doesn't exist there would be no color in the code you suggested.

levmi:
That should be okay. Agree with Andy on the ideal behavior, but understand the tradeoff we made wrt usage of the browser controlled date picker and that's fine.

app/src/components/connect/CustomSessionPage.tsx
```diff
@@ -0,0 +1,351 @@
...
+    FormDate,
+    Small,
+    Button,
+    ToggleAdvanced,
+    ProxyField,
+  } = Styled;
+  return (
+    <Wrapper>
+      <OverlayFormWrap
+        title="Custom Permissions"
```
jamaljsr, line 132: The text labels shown in the UI should be extracted into the `en-US.json` file.

jamaljsr (approved):
tACK 🔥 Tested many different permission configurations and didn't run into any issues. 

Overall, the code looks great. I commented with a bunch of nits which aren't blocking this from being merged. Just some minor improvements to make it easy to grasp what's happening in the code.

app/src/components/connect/CustomSessionPage.tsx
```diff
@@ -0,0 +1,351 @@
...
+                  value={addSessionView.expiration}
+                  onChange={addSessionView.setExpiration}
+                  options={[
+                    { label: '7 Days', value: '7' },
+                    { label: '30 Days', value: '30' },
+                    { label: '60 Days', value: '60' },
+                    { label: '90 Days', value: '90' },
+                    { label: 'Never', value: 'never' },
+                    { label: 'Custom', value: 'custom' },
+                  ]}
```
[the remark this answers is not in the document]
dstrukt: @levmi Want to make sure there's visibility on the selection, do all of these options look good?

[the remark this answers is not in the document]
levmi: This is for expiry, correct? If so, LGTM. With custom, they can grant 1 day access, which I think people may want, but don't need to make a change to surface that at this stage given the existence of custom.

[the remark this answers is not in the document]
dstrukt: Yup, for expiry, and sounds good .. figured we'd adjust based on feedback, but wanted to make sure we were on the same page about initial MVP options.

dstrukt (approved):
LGTM!

Excellent work!

itsrachelfish:
@dstrukt I added some code to automatically generate labels when none are defined. Users can still specify their own label when first viewing the add session option. However if no label is specified, the custom permissions page will automatically generate a label based on the permission type selected.

"My read-only session", "My liquidity session", "My loop session", etc.

jamaljsr:
Thanks for the updates. The latest commit LGTM 🚀

levmi (approved):
All LGTM, great work!

ellemouton:
needs a rebase :)

itsrachelfish:
@levmi @ellemouton updated with the latest from master :+1:

ellemouton (approved):
tACK 🔥

guggero:
@itsrachelfish do you want to rebase to get a clean commit history (all authored and committed by you) or should we just merge?

PR #457, opened 2022-11-29
session_rpcserver+litrpc: allow custom session with all read-only perms

### LINKED ISSUE #456: Allow all read-only endpoints for custom macaroons
Related to: #439 & #450

We now have the ability to create custom sessions with macaroons that only allow users to access specific URIs. However, when a macaroon is created with limited permissions (like only being able to access loop or pool) the `terminal-web` interface becomes unusable.

Screenshot from 2022-11-28 11-41-40

Every page has many error messages pop up because requests are being made to URIs that the user does not have permission to view.

It would be helpful if there was an option when creating a custom macaroon to allow all "read-only" endpoints in addition to the custom URIs desired.

It is currently possible to work around this issue by making a list of all the "read-only" endpoints and including them with every custom macaroon created by the UI. The drawback for this method is that whenever new read-only endpoints are added, all custom macaroons would need to be recreated.

ellemouton:
A special case is added to the creation of a custom session to 
allow the user to specify custom URIs as well as the permissions 
for all read-only endpoints. To use this option when creating a 
custom session, a user should set the `entity` of the macaroon 
permission to `"uri"` and the `action` to `"***readonly***"`
The reason for the using the three '*'s in the string is to ensure 
that the string never collides with a valid URI or a valid regex. 

`litcli` example:
```
litcli  s a --label=test --type=custom --uri=/lnrpc.Lightning/ListChannels --uri="/lnrpc.State/.*" --uri="***readonly***"
```

Fixes #456

### DIFF
--- a/litrpc/lit-sessions.proto
+++ b/litrpc/lit-sessions.proto
@@ -36,10 +36,18 @@ message AddSessionRequest {
 }
 
 message MacaroonPermission {
-    // The entity a permission grants access to.
+    // The entity a permission grants access to. If a entity is set to the
+    // "uri" keyword then the action entry should be one of the special cases
+    // described in the comment for action.
     string entity = 1;
 
-    // The action that is granted.
+    // The action that is granted. If entity is set to "uri", then action must
+    // be set to either:
+    //  - a particular URI to which access should be granted.
+    //  - a URI regex, in which case access will be granted to each URI that
+    //    matches the regex.
+    //  - the "***readonly***" keyword. This will result in the access being
+    //    granted to all read-only endpoints.
     string action = 2;
 }
 
--- a/perms/permissions_test.go
+++ b/perms/permissions_test.go
@@ -78,4 +78,10 @@ func TestMatchRegexURI(t *testing.T) {
 	uris, isRegex = m.MatchRegexURI("/poolrpc.Trader/.*")
 	require.True(t, isRegex)
 	require.Empty(t, uris)
+
+	// Assert that the read-only permission's keyword is not seen as a valid
+	// regex.
+	uris, isRegex = m.MatchRegexURI("***readonly***")
+	require.False(t, isRegex)
+	require.Empty(t, uris)
 }
--- a/session_rpcserver.go
+++ b/session_rpcserver.go
@@ -134,7 +134,21 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
 		return nil, err
 	}
 
-	var permissions []bakery.Op
+	// Store the entity-action permission pairs in a map in order to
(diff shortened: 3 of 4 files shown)


perms/permissions_test.go
```diff
@@ -78,4 +78,10 @@ func TestMatchRegexURI(t *testing.T) {
 	uris, isRegex = m.MatchRegexURI("/poolrpc.Trader/.*")
 	require.True(t, isRegex)
 	require.Empty(t, uris)
+
+	// Assert that the read-only permission's keyword is not seen as a valid
+	// regex.
+	uris, isRegex = m.MatchRegexURI("***readonly***")
```
ellemouton, line 84: reason for not using the defined constant is that it causes an import cycle

litrpc/lit-sessions.proto
```diff
@@ -36,10 +36,18 @@ message AddSessionRequest {
...
     string entity = 1;
 
-    // The action that is granted.
+    // The action that is granted. If entity is set to "uri", then action must
+    // be set to either:
+    //  - a particular URI to which access should be granted.
+    //  - a URI regex, in which case access will be granted to each URI that
+    //    matches the regex.
+    //  - the "***readonly***" keyword. This will result in the access being
+    //    granted all read-only endpoints.
```
guggero, line 50: nit: granted **to** all read-only endpoints?

session_rpcserver.go
```diff
@@ -152,6 +160,20 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
...
+			// If the action specified was equal to the
+			// readOnlyAction keyword, then this is taken to mean
+			// that the permissions for all read-only URIs should be
+			// granted.
+			if op.Action == readOnlyAction {
+				readPerms := s.cfg.permMgr.ActivePermissions(
+					true,
+				)
+
+				permissions = append(permissions, readPerms...)
```
guggero, line 172: Do we do a de-duplication somewhere? Because maybe the `***readonly***` overlaps with URIs matched by a regex?

guggero (approved):
Nice, LGTM :tada:

session_rpcserver.go
ellemouton: (added in the last commit so that the first 2 commits dont need to be re-reviewed. Hope that that is ok)

ellemouton:
Thanks for the review @guggero ! Added one more commit to address your comment about de-duplicating the perms.

guggero (approved):
LGTM :tada:

itsrachelfish (approved):
tACK. I rebased my pr #451 based on this branch and was able to create a custom session using the new `***readonly***` flag

PR #464, opened 2022-12-06
multi: change account session type and reserve type 4

ellemouton:
In this commit, the Account session type is changed to be type 5 in both the session proto and in the persisted Session. This is done so that type 4 can be reserved for future use.


litrpc/lit-sessions.proto
```diff
@@ -15,11 +15,12 @@ service Sessions {
 }
 
 enum SessionType {
+    reserved 4;
```
guggero, line 18: nit: keep this in the original order? So between 3 and 5? Or does the compiler complain then?

litrpc/lit-sessions.proto
```diff
@@ -15,11 +15,12 @@ service Sessions {
...
     TYPE_MACAROON_READONLY = 0;
     TYPE_MACAROON_ADMIN = 1;
     TYPE_MACAROON_CUSTOM = 2;
     TYPE_UI_PASSWORD = 3;
-    TYPE_MACAROON_ACCOUNT = 4;
+    TYPE_MACAROON_ACCOUNT = 5;
```
guggero, line 23: I think this might also require a re-build of the JS stuff?

litrpc/lit-sessions.proto
ellemouton: done 👍

litrpc/lit-sessions.proto
ellemouton: aaahhh good catch!! updated!

guggero (approved):
LGTM :tada:

bitromortac (approved):

PR #473, opened 2023-01-05
tools: fix lint VSC permission error

guggero:
Fixes the linter failure that appeared out of nowhere in all our projects: https://github.com/lightninglabs/lightning-terminal/actions/runs/3843975095/jobs/6547862365

### DIFF
--- a/.github/actions/setup-go/action.yml
+++ b/.github/actions/setup-go/action.yml
@@ -1,0 +1,34 @@
+name: "Setup Golang environment"
+description: "A reusable workflow that's used to set up the Go environment and cache."
+inputs:
+  go-version:
+    description: "The version of Golang to set up"
+    required: true
+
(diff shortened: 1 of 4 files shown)


positiveblue (approved):
Thank you, I did [this](https://github.com/lightninglabs/pool/pull/427) for pool yesterday. In LND @yyforyongyu was able to [fix it without marking the folder "safe" for git](https://github.com/lightningnetwork/lnd/pull/7254), but I wasn't able to replicate it in other projects (even after using the same docker images)

jamaljsr (approved):
LGTM 🥇 

Thanks so much for resolving this. I was on a completely different path investigating usage of the `-buildvcs` flag. It's still not clear to me what the exact problem was, but happy it's fixed nevertheless.

PR #475, opened 2023-01-10
multi: add RevokedAt field to Session

ellemouton:
In this PR, a new `RevokedAt` field is added to the `Session` struct and it is set to the current time when `RevokeSession` is called. Note that since it is a _new_ field, old sessions will have a zero value `RevokedAt` time and so readers of a session should still use the sesssion's `Status` field to determine if a session is revoked or not.


session/tlv.go
```diff
@@ -105,6 +111,10 @@ func SerializeSession(w io.Writer, session *Session) error {
 		tlvRecords, tlv.MakePrimitiveRecord(typeCreatedAt, &createdAt),
 	)
 
+	tlvRecords = append(
+		tlvRecords, tlv.MakePrimitiveRecord(typeRevokedAt, &revokedAt),
```
guggero, line 115: nit: could combine this with the `append` right above it.

session_rpcserver.go
```diff
@@ -519,6 +519,11 @@ func marshalRPCSession(sess *session.Session) (*litrpc.Session, error) {
 
 	macRecipe := marshalRPCMacaroonRecipe(sess.MacaroonRecipe)
 
+	var revokeAt uint64
```
guggero, line 522: nit: s/revokeAt/revokedAt/

guggero (approved):
Nice, LGTM :tada:

session/tlv.go
ellemouton: indeed! will update

ellemouton: done :)

bitromortac (approved):
LGTM :rocket:

PR #476, opened 2023-01-10
multi: add some litd client helpers

ellemouton:
This PR does 2 things: 

1. Adds a `litclient` package and adds and exports a list of JSON callbacks that a client connecting to lit should register. 
2. Adds a `withAllSubServers` param to perms.NewManager. This is useful for a client who wants to get a list of all possible permissions available to it without knowing which sub-servers LND was compiled with. This will lead to some false positives if the client tries to call the method of a sub-server that was not compiled within LND.

required for https://github.com/lightninglabs/lightning-node-connect/pull/66


guggero (approved):
LGTM :tada:

kaloudis (approved):

PR #477, opened 2023-01-11
session+litrpc: change RevokedAt tlv and proto number

ellemouton:
In this PR, we change the tlv type number and proto field number of the new RevokedAt field to 16 and we reserve numbers 14 and 15 for future use.


guggero (approved):
LGTM :tada:

bitromortac (approved):

PR #487, opened 2023-02-13
multi: autopilot firewall and privacy map interceptors

ellemouton:
This PR makes extensive use of LNDs RPC middleware interceptors in order to implement a macaroon firewall. 

Calls made with a macaroon containing the `lnd-custom privacy` caveat will be intercepted and any private info (such as pub keys, outpoints etc) will be mapped out to random values for requests and mapped back to real values for responses.

Calls made with a macaroon containing the `lnd-custom lit-mac-fw`caveat will be intercepted and the call will be validated agains various rules that the caveat specifies. The current set of defined rules are: Channel restrictions, peer restrictions, channel policy bounds, history limit, rate limit. 

An Autopilot client is also added which handles registering an LNC session with the Autopilot server. The access given to the autopilot server is restricted using the above mentioned macaroon caveats.


guggero (approved):
We reviewed and tested this code internally for many weeks, so I can give my thumbs up rather quickly.
Amazing work on this, kudos to everyone involved :100:

jamaljsr (approved):
tACK. Did a test run just to sanity check. Everything still works as expected. Awesome job on this feature @ellemouton 🔥

bitromortac (approved):
tACK :rocket: Awesome work!

Issue #448, opened 2022-11-02
Add option to disable the local UI


gkrizek:
**Describe the solution you'd like**

I'd like the ability to run litd without having any local UI available, however still have all the APIs available. This would allow me to not have to worry about exposing only a password protected webpage to the internet, but still be able to generate LNC sessions to connect to my Terminal instance from something like Terminal Web.

**Describe alternatives you've considered**
Some sort of reverse proxy in front of litd that blocks requests to the UI. It'd be a lot of work and a big hack.

**Additional context**
I started to work on this, but realized that currently you set the uipassword in your litd config file. This password is used for authentication to the grpc-web and REST proxies. So we would need to come up with a solution for that or just enable macaroon based authentication for those APIs when the UI is disabled. 

Created based on https://github.com/lightninglabs/lightning-terminal/pull/444

Roasbeef:
AFAICT, it's off by default since it's only on local host? https://github.com/lightninglabs/lightning-terminal/blob/7497edf443762d1542356f2cbea579d972ada2bf/config.go#L282

gkrizek:
@Roasbeef but wouldn't that disable the Terminal API as well? We'd want to still use the Terminal API on 8443 but not serve any of the UI

levmi:
I just want to clarify the ask a little bit on this one so we can prioritize it on our end and figure out how to deliver what y'all need. It sounds like the main priority is actually just not having the password functionality? The removal of UI sounds like a path to get to not having a password. But, correct me if I'm wrong there. 

If that's the case, one thing we could explore is replacing the password with something else. Ideally, this could be the super macaroon that Voltage is already baking for their user. Would a path like that work for y'all? It would get rid of the need for a password, but not meet everything outlined here with regard to removing/disabling the UX.

PR #506, opened 2023-02-25
config: add flag to disable main UI http web server

kaloudis:
This PR adds startup flag `--disable-web-server` for `litd` that disables the main UI web server.

This is particularly useful for users that want to use LNC but don't want to expose a web server on their machine, and don't want to put a firewall in place as a workaround.

### DIFF
--- a/config.go
+++ b/config.go
@@ -147,6 +147,8 @@ type Config struct {
 	UIPasswordFile string   `long:"uipassword_file" description:"Same as uipassword but instead of passing in the value directly, read the password from the specified file."`
 	UIPasswordEnv  string   `long:"uipassword_env" description:"Same as uipassword but instead of passing in the value directly, read the password from the specified environment variable."`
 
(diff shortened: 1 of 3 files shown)


guggero:
Unfortunately, it's not that simple. In lnd remote mode, everything runs over that main web server connection (including the RPC for creating LNC sessions). I'm not sure that is the desired outcome?

kaloudis:
I see now. I've just been using it to toggle it on and off - manage it through the UI when necessary, but there is definitely a more graceful way to handle this.

I saw Graham took a stab at this earlier by just blocking the static files which is a bit closer what we really want https://github.com/lightninglabs/lightning-terminal/pull/444/files#diff-76dcb3ff0bf98acc001558927365f8a3b7f0d2253580a49dc5973dfa398372c7R1048

kaloudis:
Full tracking of the issue is here: https://github.com/lightninglabs/lightning-terminal/issues/448

dstadulis:
#519 superseded this PR and addressed the issue #448

PR #522, opened 2023-03-18
litrpc: add REST annotations for Litd RPC services

ellemouton:
Add REST annotations for all LitD RPC services and register them with the REST handler. Update all itests accordingly. 

Also filled in lots of missing comments & added `litcli` annotations :)


jamaljsr:
Not sure if you want to address it in this PR or separately, but many of the RPC methods are missing comments in the proto files. It would improve the API docs site if each one had a comment which included the CLI command ([example](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/lightning.proto#L27)) so that the Code Samples `Shell` tab will display the help output.

ellemouton:
Great point @jamaljsr! Added them to this PR :)

litrpc/lit-accounts.proto
```diff
@@ -5,7 +5,7 @@ package litrpc;
 option go_package = "github.com/lightninglabs/lightning-terminal/litrpc";
 
 service Accounts {
-    /*
+    /* litcli: `accounts create`
```
guggero, line 8: :pray:

cmd/litcli/autopilot.go
```diff
@@ -20,14 +20,20 @@ var autopilotCommands = cli.Command{
 		{
 			Name:      "features",
 			ShortName: "f",
-			Usage:     "List available features",
-			Action:    listFeatures,
+			Usage:     "List available Autopilot features.",
+			Description: `
+			List available Autopilot features.
```
guggero, line 25: Hmm, this creates an indentation in the generated text of 3 levels. I know it looks weird in the code here, but I think we only want one indentation as we have in the accounts commands (and lncli in general).

litrpc/firewall.yaml
```diff
@@ -0,0 +1,11 @@
...
+config_version: 3
+
+http:
+  rules:
+
+    # firewall.proto
+    - selector: litrpc.Firewall.ListActions
+      get: "/v1/firewall/actions"
+    - selector: litrpc.Firewall.PrivacyMapConversion
+      get: "/v1/firewall/privacy_map/convert"
```
guggero, line 11: Should these calls be POST to avoid needing to use special base64url format for the `bytes` parameter (session_id)?

litrpc/lit-accounts.yaml
```diff
@@ -0,0 +1,17 @@
+type: google.api.Service
+config_version: 3
+
+http:
+  rules:
+
+    # lit-accounts.proto
```
guggero, line 7: The more RESTy way for this CRUD pattern would be:

```
    - selector: litrpc.Accounts.CreateAccount
      post: "/v1/accounts"
      body: "*"
    - selector: litrpc.Accounts.UpdateAccount
      post: "/v1/accounts/{id}"
      body: "*"
    - selector: litrpc.Accounts.ListAccounts
      get: "/v1/accounts"
    - selector: litrpc.Accounts.RemoveAccount
      delete: "/v1/accounts/{id}"
      body: "*"
```

litrpc/lit-autopilot.yaml
```diff
@@ -0,0 +1,17 @@
+type: google.api.Service
+config_version: 3
+
+http:
+  rules:
+
+    # lit-autopilot.proto
+    - selector: litrpc.Autopilot.ListAutopilotFeatures
```
guggero, line 8: Same here, could be a bit more aligned to common REST patterns (which we don't use consistently everywhere, I know)...

guggero (commented):
Very nice, thanks a lot for the added documentation and REST mapping :100: 
Just a few comments about the REST mapping, other than that LGTM :taco:

cmd/litcli/autopilot.go
ellemouton: ahhhh good catch! took me a while to figure out what you mean. I have moved each command into its own named `var` so that it doesnt need to look weird :)

litrpc/firewall.yaml
ellemouton: ahhhh ok cool - TIL! thanks!

litrpc/lit-accounts.yaml
ellemouton: ok very cool - thanks!

litrpc/lit-autopilot.yaml
ellemouton: what would you recommend for the "revoke" one? since it is not really a delete but rather an update...

ellemouton: should it be like this? or should it just be a delete? and is the "local_public_key" part ok?

```
post: "/v1/autopilot/sessions/{local_public_key}"
```

litrpc/proxy.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+http:
+  rules:
+
+    # proxy.proto
+    - selector: litrpc.Proxy.StopDaemon
+      post: "/v1/proxy/stop"
+      body: "*"
+    - selector: litrpc.Proxy.GetInfo
+      get: "/v1/proxy/info"
```
ellemouton, line 12: thoughts on the RESTy-ness of these @guggero ?

litrpc/lit-autopilot.yaml
ellemouton: note that the local_public_key is bytes.... so would require the base64url format you mentioned right?

guggero: I think `delete` still makes sense here, even if it doesn't fully remove the session. The public key being `bytes` does mean the base64url format is required. We solve this in `lnd` by adding a specific comment to the field itself, like something like `When using REST, this field must be encoded as base64url.` which will then show up in the API docs.

ellemouton: cool - updated to add that comment to the RevokeSession and RevokeAutopilotSession methods 👍  thanks for the insights!

guggero (approved):
LGTM :tada:

litrpc/lit-autopilot.proto
```diff
@@ -101,6 +117,10 @@ message ListAutopilotFeaturesResponse {
 }
 
 message RevokeAutopilotSessionRequest {
+    /*
+    The local static public key of the Autopilot session to be revoked.
+    When using REST, this field must be encoded as base64.
```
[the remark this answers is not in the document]
ellemouton: gotcha! updated :)

jamaljsr (commented):
tACK 👌

Tested a handful of REST calls across each of the services. Didn't run into any issues.

proto/proxy.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+package litrpc;
+
+option go_package = "github.com/lightninglabs/lightning-terminal/litrpc";
+
+service Proxy {
+    /* litcli: `proxy getinfo`
```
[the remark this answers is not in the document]
ellemouton: good catch!!! will update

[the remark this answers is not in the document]
ellemouton: updated :)

jamaljsr (approved):
LGTM 🚀 

Great update!

PR #530, opened 2023-04-17
Fix "no matching files found" compilation error when litd is used as a library

guggero:
Fixes the `terminal.go:99:13: pattern app/build/*: no matching files found` error when litd is used as a library or module in another project.

Even though the content of the app/build directory should be ignored by
git, we need it to exist in order for the Golang embedded file system
compilation to succeed on an empty project. This is required to fix
compilation issues when lightning-terminal is used as a module
dependency in another project, where the frontend build process can't be
triggered.

### DIFF
--- a/app/build/.gitkeep
+++ b/app/build/.gitkeep
@@ -1,0 +1,1 @@
+# Keep directory in git.
--- a/app/package.json
+++ b/app/package.json
@@ -8,6 +8,7 @@
     "start": "BROWSER=none react-scripts start",
     "develop": "REACT_APP_USE_SAMPLE_DATA=true yarn start",
     "build": "react-scripts build",
+    "postbuild": "echo -n '# Keep directory in git.' > build/.gitkeep",
     "test": "react-scripts test --env=jest-environment-jsdom",
     "test:ci": "cross-env CI=true yarn test --coverage",
     "eject": "react-scripts eject",


app/build/.gitkeep
```diff
@@ -0,0 +1 @@
+# Keep directory in git.
```
ellemouton, line 1: TIL about .gitkeep 🎉

app/package.json
```diff
@@ -8,6 +8,7 @@
     "start": "BROWSER=none react-scripts start",
     "develop": "REACT_APP_USE_SAMPLE_DATA=true yarn start",
     "build": "react-scripts build",
+    "postbuild": "echo -n '# Keep directory in git.' > build/.gitkeep",
```
ellemouton, line 11: so the front end build deletes the directory?

ellemouton (approved):
LGTM :) nice trick!

app/build/.gitkeep
guggero: Not really a feature but more of a convention:

> A common, standardised practice to solve this exact issue, however, is to push a file called .gitkeep into your empty folders.

> This isn't a feature of Git! So you could name it anything. There's nothing special about the name .gitkeep – some developers add .gitignore instead, for example.

https://www.freecodecamp.org/news/what-is-gitkeep/

app/package.json
guggero: Yes, the react build scripts clear out the whole build folder. So this is a bit hacky but IMO gets the job done.

jamaljsr (approved):
tACK LGTM 🚀

Noticed this also fixes those warnings i the CI logs. Nice update! 👌

ViktorTigerstrom:
Great, thanks for the fix @guggero  🚀!

PR #535, opened 2023-04-26
doc: add note for remote lnd credentials

GeorgeTsagk:
## Description

Adds a small note in the doc for connecting `litd` to remote `lnd`. It is not clear that the LND `admin.macaroon` and `tls.cert` should not be placed inside the `.lit/` directory, as they may be overwritten.


guggero (approved):
LGTM, thanks :tada:

PR #537, opened 2023-05-02
multi: allow disabling of sub-servers

ellemouton:
With this PR, a new "mode" is added for subservers to complement the existing "integrated" and "remote" modes. 
The new mode is "disable" which will allow users to pick and choose which subservers they want Litd to start up with. 

For example:
```
--loop-mode=disable
--pool-mode=disable
--faraday-mode=disable
```

This mode addition is added in the last commit of the PR. The prior commits prepare the code for the change by making all subserver logic a bit more contained within the subserver manager

PR #539, opened 2023-05-03
multi: modularise subserver handling 

ellemouton:
This PR is the first couple of commits of https://github.com/lightninglabs/lightning-terminal/pull/537 without the disabling logic.

### DIFF
--- a/perms/manager.go
+++ b/perms/manager.go
@@ -292,27 +292,3 @@ func (pm *Manager) isLndURI(uri string) bool {
 	_, lndCall := pm.fixedPerms[lndPerms][uri]
 	return lndCall || lndSubServerCall
 }
-
-// IsLoopURI returns true if the given URI belongs to an RPC of loopd.
-func (pm *Manager) IsLoopURI(uri string) bool {
-	_, ok := pm.fixedPerms[loopPerms][uri]
-	return ok
-}
-
-// IsFaradayURI returns true if the given URI belongs to an RPC of faraday.
-func (pm *Manager) IsFaradayURI(uri string) bool {
-	_, ok := pm.fixedPerms[faradayPerms][uri]
-	return ok
-}
-
-// IsPoolURI returns true if the given URI belongs to an RPC of poold.
-func (pm *Manager) IsPoolURI(uri string) bool {
-	_, ok := pm.fixedPerms[poolPerms][uri]
-	return ok
-}
-
-// IsLitURI returns true if the given URI belongs to an RPC of LiT.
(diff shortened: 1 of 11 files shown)


subservers/manager.go
```diff
@@ -259,3 +288,22 @@ func dialBackend(name, dialAddr, tlsCertPath string) (*grpc.ClientConn, error) {
 	}
 	return cc, nil
 }
+
+// readMacaroon tries to read the macaroon file at the specified path and create
+// gRPC dial options from it.
```
positiveblue, line 323: Does `...and create gRPC dial options from it.` apply here?

terminal.go
```diff
@@ -494,7 +491,10 @@ func (g *LightningTerminal) start() error {
 
 	// Initialise any connections to sub-servers that we are running in
 	// remote mode.
-	g.subServerMgr.ConnectRemoteSubServers()
+	if err := g.subServerMgr.ConnectRemoteSubServers(); err != nil {
```
positiveblue, line 494: :pray:

positiveblue (approved):
These changes make it so much cleaner. Thank you @ellemouton 

Note: i did not test the integrate/remote combination that panic in #537

subservers/manager.go
ellemouton: good catch!

ellemouton: fixed

ellemouton:
Thanks @positiveblue! 

> Note: i did not test the integrate/remote combination that panic in https://github.com/lightninglabs/lightning-terminal/pull/537

No worries - that panic is only (it seems) related to the situation where one sub-server is disabled. Which is not the case here.

ellemouton:
ah - ok. I _thiiiiink_ I may have found something that would cause the panic. Will push up a change later today that will hopefully fix it 🤞

ellemouton:
note: i have added one more commit on the end which I think will fix the panic that was noted by Jamal 🤞

ellemouton:
one more commit with the same nature

perms/manager.go
```diff
@@ -1,25 +1,18 @@
...
 	"gopkg.in/macaroon-bakery.v2/bakery"
 )
 
 const (
-	poolPerms    string = "pool"
-	loopPerms    string = "loop"
-	faradayPerms string = "faraday"
-	litPerms     string = "lit"
-	lndPerms     string = "lnd"
+	litPerms string = "lit"
```
guggero, line 14: Nice! A little bit less duplication :tada:

guggero (approved):
Very nice, LGTM :tada:

PR #541, opened 2023-05-03
multi: add a Status server

### LINKED ISSUE #290: litd exits with code 0 when it stops due to non-0 lnd exit

### LINKED ISSUE #420: Show LND/LiT version mismatch error in Web UI

### LINKED ISSUE #421: Show dependency issues in the UX

ellemouton:
This PR introduces a `Status` server to Lit. This server can be queried at any time to determine the 
status of any of the following servers: Lit, LND, Faraday, Loop, Pool. With this PR, Lit will also not shutdown 
unless stopped purposefully. The reason being so that if any parts of Litd fail to start, the status server will 
still be available for a front end to query to see what is up. 

Let's say for example, that Lit is started with Loop disabled (`loop-mode=disabled`) & then some error causes Pool
not to start up properly, then you can query the status server to see the following:
```
$ litcli status 

{
        "sub_servers": {
                "faraday": {
                        "disabled": false,
                        "running": true,
                        "error": ""
                },
                "lit": {
                        "disabled": false,
                        "running": true,
                        "error": ""
                },
                "lnd": {
                        "disabled": false,
                        "running": true,
                        "error": ""
                },
                "loop": {
                        "disabled": true,
                        "running": false,
                        "error": ""
                },
                "pool": {
                        "disabled": false,
                        "running": false,
                        "error": "pool not working"
                },
                "taproot-assets": {
                        "disabled": false,
                        "running": true,
                        "error": ""
                }
        }
}


```

Depends on: #616 

Replaces https://github.com/lightninglabs/lightning-terminal/pull/442

Fixes #421 
Fixes #420
Fixes #290


itsrachelfish (changes requested):
I was able to get this integrated into terminal but while testing I noticed that when a subserver is disabled it completely disappears from the status list. We might want to change this because it means we won't be able to use the status server for feature detection.

For example when taproot assets are added to litd, it could be useful to query the status server to see if tapd is available. Currently there is no way to differentiate between "litd hasn't been upgraded, so tapd does not exist" and "litd has been upgraded, and tapd exists, but it is disabled"

ellemouton:
thanks for the suggestion @itsrachelfish. Updated accordingly

litrpc/lit-status.proto
```diff
@@ -0,0 +1,32 @@
+
```
guggero, line 1: nit: empty line at file start.

status/manager.go
```diff
@@ -0,0 +1,174 @@
...
+	litrpc.UnimplementedStatusServer
+
+	subServers map[string]*SubServerStatus
+	mu         sync.RWMutex
+}
+
+// NewStatusManager constructs a new Manager.
+func NewStatusManager() *Manager {
+	return &Manager{
+		subServers: map[string]*SubServerStatus{},
```
guggero, line 46: nit: should use `make()` instead if we're not pre-assigning values.

status/manager.go
```diff
@@ -0,0 +1,174 @@
...
+func newSubServerStatus() *SubServerStatus {
+	return &SubServerStatus{
+		Disabled: true,
+	}
+}
+
+// Manager managers the status of any sub-server registered to it. It is also an
+// implementation of the litrpc.StatusServer which can be queried for the status
+// of various LiT sub-servers.
+type Manager struct {
```
guggero, line 36: I'm probably missing something, but why can't the `subservers.Manager` keep track of disabled/running/error and we just have a simple RPC server struct that gets an instance of said manager?

subservers/manager.go
```diff
@@ -49,6 +54,9 @@ func (s *Manager) AddServer(ss SubServer, enable bool) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	// Register all sub-servers with the status server.
+	s.statusServer.RegisterSubServer(ss.Name())
```
guggero, line 58: I guess this is another hint that we might want to merge the two managers instead.

itest/litd_mode_integrated_test.go
```diff
@@ -416,7 +418,7 @@ func integratedTestSuite(ctx context.Context, net *NetworkHarness, t *testing.T,
 					endpoint.macaroonFn(cfg),
 					endpoint.requestFn,
 					endpoint.successPattern,
-					endpointDisabled,
+					endpointDisabled || endpoint.litOnly,
```
guggero, line 444: Shouldn't this have exactly the opposite meaning as in the proxy server should never be disabled? So we should never get the `disabledErr` returned?

itest/litd_mode_integrated_test.go
```diff
@@ -787,11 +838,19 @@ func runUIPasswordCheck(t *testing.T, hostPort, tlsCertPath, uiPassword string,
...
+	// this is a call that is allowed to be un-authenticated in which case
+	// we expect it to succeed.
 	_, err = makeRequest(ctxt, rawConn)
-	if disabled {
+	switch {
+	case disabled:
 		require.ErrorContains(t, err, disabledErr)
-	} else {
+	case noAuth:
+		require.NoError(t, err)
```
guggero, line 849: Should we assert the success content here before returning?

guggero (commented):
Awesome PR, great work!
I did a first pass to load all the context. I think the main thing we might want to change is merging of the two subserver managers (see inline comments).
I'll do a round of manual testing in the next round.

terminal.go
```diff
@@ -1112,6 +1129,18 @@ func (g *LightningTerminal) shutdownSubServers() error {
...
+		if err := g.rpcProxy.Stop(); err != nil {
+			log.Errorf("Error stopping rpc proxy: %v", err)
+		}
+	}
+
+	if g.httpServer != nil {
+		if err := g.httpServer.Close(); err != nil {
+			log.Errorf("Error stopping UI server: %v", err)
+		}
+	}
```
ViktorTigerstrom, line 1142: I think we should also set the `returnErr` here if `g.rpcProxy.Stop()` or `g.httpServer.Close()` fails?

terminal.go
```diff
@@ -238,8 +246,12 @@ func (g *LightningTerminal) Run() error {
 	// server is started.
 	g.rpcProxy = newRpcProxy(
 		g.cfg, g, g.validateSuperMacaroon, g.permsMgr, g.subServerMgr,
+		g.statusMgr,
 	)
 
+	litrpc.RegisterProxyServer(g.rpcProxy.grpcServer, g.rpcProxy)
+	litrpc.RegisterStatusServer(g.rpcProxy.grpcServer, g.statusMgr)
```
ViktorTigerstrom, line 253: nit: could use a comment for this code block, given that the all the other code blocks in this function are preceded by comments.

status/manager.go
```diff
@@ -0,0 +1,174 @@
...
+	if !ok {
+		return nil, errors.New("a sub-server with name %s has not " +
+			"yet been registered")
+	}
+
+	return status, nil
+}
+
+// SetEnabled marks the sub-server with the given name as enabled.
+func (s *Manager) SetEnabled(name string) {
```
ViktorTigerstrom, line 112: nit:
```suggestion
// SetEnabled marks the sub-server with the given name as enabled.
// This will silently fail if the referenced sub-server has not yet been registered.
func (s *Manager) SetEnabled(name string) {
```

status/manager.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// newSubServerStatus constructs a new SubServerStatus.
+func newSubServerStatus() *SubServerStatus {
+	return &SubServerStatus{
+		Disabled: true,
+	}
+}
+
+// Manager managers the status of any sub-server registered to it. It is also an
```
ViktorTigerstrom, line 33: nit:
```suggestion
// Manager manages the status of any sub-server registered to it. It is also an
```

status/log.go
```diff
@@ -0,0 +1,25 @@
+package status
+
+import (
+	"github.com/btcsuite/btclog"
+	"github.com/lightningnetwork/lnd/build"
+)
+
+const Subsystem = "STAT"
+
+// log is a logger that is initialized with no output filters.  This
```
ViktorTigerstrom, line 10: nit:
```suggestion
// log is a logger that is initialized with no output filters. This
```

ViktorTigerstrom (commented):
Very nice PR 🚀 🔥!!! My only comments are for minor things, so really great work! I've tested the functionality manually as well, and will also conduct thorough manual testing next review round 🚀!

status/manager.go
ellemouton: the main reason is cause `subservers.Manager` is just meant to handle the external daemons and does not handle Lit and LND (both of which are kind of in their own category) .So seemed a bit contrived to try squash those in to the `subservers.Manager` since they dont fit the `SubServer` interface.

subservers/manager.go
ellemouton: i think every subserver should have a status but not each thing that has a status is a subserver

status/manager.go
guggero: Ah, okay, I didn't consider that. Okay, makes sense to keep them separate then.

ellemouton: lemme know what you think :)

itest/litd_mode_integrated_test.go
ellemouton: jip - good catch 👍  also updated the `runGRPCWebAuthTest` test to properly check success content

ellemouton (commented):
Thanks @guggero  & @ViktorTigerstrom  🚀  updated!

terminal.go
```diff
@@ -464,10 +468,18 @@ func (g *LightningTerminal) start() error {
 	case <-readyChan:
 
 	case err := <-g.errQueue.ChanOut():
-		return err
+		g.statusMgr.SetErrored(
```
guggero, line 464: We also need to set `lnd` as errored on line 444. I tested this in integrated mode where `lnd` couldn't start because of an already occupied listener port. The status shows `running: false` but `error: ""`.

guggero: And then, when trying to shut down (Ctrl+C), it just logs `LTND: Gracefully shutting down.` but doesn't quit... This could be hard to pin down though, so low prio/non-blocking.

terminal.go
```diff
@@ -464,10 +468,18 @@ func (g *LightningTerminal) start() error {
 	case <-readyChan:
```
guggero, line 461: I think we need to react to the `unlockChan` here somehow. I tested integrated mode where `lnd` was locked, and in that state it shows `running: false` for `lnd`. And I can't run `lncli --rpcserver localhost:8443 state` because I get `rpc error: code = Unknown desc = lnd is not running:`. Can we maybe count `lnd` as being locked as a successful startup and let `RPC` calls through? Otherwise we also won't be able to create a wallet or unlock using the 8443 port.

cmd/litcli/status.go
```diff
@@ -0,0 +1,36 @@
...
+func getStatus(ctx *cli.Context) error {
+	clientConn, cleanup, err := connectClient(ctx, true)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+	client := litrpc.NewStatusClient(clientConn)
+
+	ctxb := context.Background()
+	resp, err := client.SubServerStatus(ctxb, &litrpc.SubServerStatusReq{})
```
guggero, line 28: I wonder if we also should try calling `lnd`'s `State.GetState` RPC here and print that as well? Just as a nice addition for CLI users.

itest/litd_mode_integrated_test.go
guggero: Ah okay, makes sense, thanks.

terminal.go
guggero: Also, again same issue with not being able to shut down while lnd is not unlocked yet. Maybe something small after all?

guggero (commented):
Code looks great! Nothing to add there.

Did a round of manual tests and found some issues around lnd startup and unlock behavior.

terminal.go
ellemouton: for both of these, the issue is pre-existing. 
Currently (master branch) if you already have LND running & then try starting LiT in integrated mode, things hang on the `lnd.Main(...)` line. If you then shutdown (Ctrl+C) it will also just log `LTND: Gracefully shutting down.` and will log `LTND: Shutdown complete` about 10 seconds later but still the `lnd.Main(...)` method wont return and so Litd doesnt exit.  

I think the issue is that things block here: https://github.com/lightninglabs/lightning-terminal/blob/03f4bf99a9fd0313a8f219492dcdf25a88817169/terminal.go#L547

I'll open a separate PR to fix this. 

The one open question is: if the port is occupied, `lnd.Main` does not return. It hangs and retries. So there really isnt any way for LiT to set the `error:` in the status. So I think for now, just showing "running: false" is good enough since that should indicate to the user that something is wrong.

terminal.go
ellemouton: good catch! I have moved  the `g.statusMgr.SetRunning(subservers.LND)` line a bit earlier 👍

ellemouton: this also pointed to me to another pre-existing issue which is that LND's whitelisted calls are currently still checked for macaroons if they go through 8443. So i'll open another PR to fix that too :)

ellemouton: ok - added this PR: https://github.com/lightninglabs/lightning-terminal/pull/617

state requests should now work even while LND is still locked 👍

ellemouton:
@AndySchroder  - I'd say this PR is pretty close. But just note that we wont merge this PR until the front end ready as it needs to be able to handle the case where some sub-servers are disabled. So if you are using the front-end then maybe wait. But if not, then I think you can try this PR. There are no DB changes or anything so if anything goes wrong (pls let us know!) then you can just revert to the last released version. 

As with any un-merged & un-released PR: please proceed at your own risk

itsrachelfish (approved):
tACK, the latest changes including the `disabled` flag are working with the front-end #620

ViktorTigerstrom (commented):
Looks really good to me!
After some further manual testing, I've noticed some strange behaviour for subservers which are run in remote mode though.
1.
If running LiT with a subserver configured as in `remote` mode (specifically pool, loop & faraday), and if I then manually shutdown the remote subserver, the `lncli status` command will still mark the specific subserver as "running".
2.
Even more strangely, for some reason which I can't figure out why when looking at the code, if I have shutdown a subserver and then start LiT with a subserver configured as in `remote` mode (specifically tried with pool & loop), the `lncli status` command will still mark the specific subserver as "running".
This is despite the LiT logs clearing showing that we can't connect to the subserver. See the log snippet below:

```
2023-09-12 02:11:35 ----------------------------------------------------------
2023-09-12 02:11:35  Lightning Terminal (LiT) by Lightning Labs               
2023-09-12 02:11:35                                                           
2023-09-12 02:11:35  LND Operating mode      remote                               
2023-09-12 02:11:35  LND Node status         online                               
2023-09-12 02:11:35  LND Alias               dave                               
2023-09-12 02:11:35  LND Version             0.16.4-beta commit=v0.16.4-beta                               
2023-09-12 02:11:35  LiT Version             0.10.5-alpha commit=v0.10.5-alpha-68-g5c8986823ae8483e21c5b01b3c7ef7bc9c233282                               
2023-09-12 02:11:35  Web interface           0.0.0.0:8443 (open https://localhost:8443 in your browser)                                    
2023-09-12 02:11:35 ----------------------------------------------------------
2023-09-12 02:11:35 2023-09-12 00:11:35.436 [INF] LITD: Connecting basic lnd client
2023-09-12 02:11:35 2023-09-12 00:11:35.638 [INF] LITD: Connecting full lnd client
2023-09-12 02:11:35 2023-09-12 00:11:35.638 [INF] LNDC: Creating lnd connection to dave:10009
2023-09-12 02:11:35 2023-09-12 00:11:35.638 [INF] LNDC: Connected to lnd
2023-09-12 02:11:35 2023-09-12 00:11:35.640 [INF] LNDC: Waiting for lnd to unlock
2023-09-12 02:11:35 2023-09-12 00:11:35.641 [INF] LNDC: Wallet state of lnd is now: Lnd main server is ready for requests
2023-09-12 02:11:35 2023-09-12 00:11:35.649 [INF] LNDC: lnd version: v0.16.4-beta, build tags 'signrpc,walletrpc,chainrpc,invoicesrpc,peersrpc'
2023-09-12 02:11:35 2023-09-12 00:11:35.649 [INF] LNDC: Using network regtest
2023-09-12 02:11:35 2023-09-12 00:11:35.649 [INF] LNDC: Waiting for lnd to be fully synced to its chain backend, this might take a while
2023-09-12 02:11:35 2023-09-12 00:11:35.659 [INF] LNDC: lnd is now fully synced to its chain backend
2023-09-12 02:11:35 2023-09-12 00:11:35.660 [INF] SSVR: Opening sqlite3 database at: /root/.tapd/data/regtest/tapd.db
2023-09-12 02:11:35 2023-09-12 00:11:35.663 [INF] SRVR: Version: 0.2.3-alpha commit=v0.2.3, build=production, logging=default, debuglevel=info
2023-09-12 02:11:35 2023-09-12 00:11:35.663 [INF] SRVR: Active network: regtest
2023-09-12 02:11:35 2023-09-12 00:11:35.663 [INF] RPCS: Validating RPC requests based on macaroon at: /root/.tapd/data/regtest/admin.macaroon
2023-09-12 02:11:35 2023-09-12 00:11:35.664 [INF] GRDN: Starting ChainPlanter
2023-09-12 02:11:35 2023-09-12 00:11:35.664 [INF] GRDN: Retrieved 0 non-finalized batches from DB
2023-09-12 02:11:35 2023-09-12 00:11:35.665 [INF] GRDN: Starting asset custodian
2023-09-12 02:11:35 2023-09-12 00:11:35.665 [INF] GRDN: Gardener for ChainPlanter now active!
2023-09-12 02:11:35 2023-09-12 00:11:35.665 [INF] GRDN: Loading pending inbound asset events
2023-09-12 02:11:35 2023-09-12 00:11:35.666 [INF] FRTR: Starting ChainPorter
2023-09-12 02:11:35 2023-09-12 00:11:35.666 [INF] UNIV: Starting FederationEnvoy
2023-09-12 02:11:35 2023-09-12 00:11:35.666 [INF] UNIV: Adding new Universe server to Federation, addrs=([]universe.ServerAddr) {
2023-09-12 02:11:35 }
2023-09-12 02:11:35 
2023-09-12 02:11:35 2023-09-12 00:11:35.667 [INF] GRDN: Resuming 0 pending inbound asset events
2023-09-12 02:11:35 2023-09-12 00:11:35.667 [INF] GRDN: Loading wallet transactions starting at block height 0
2023-09-12 02:11:35 2023-09-12 00:11:35.667 [WRN] UNIV: unable to add universe servers: <nil>
2023-09-12 02:11:35 2023-09-12 00:11:35.667 [INF] RPCS: Starting RPC Server
2023-09-12 02:11:35 2023-09-12 00:11:35.667 [INF] LITD: Starting LiT macaroon service
2023-09-12 02:11:35 2023-09-12 00:11:35.668 [INF] GRDN: Checking 4 wallet transactions for inbound assets, this might take a while
2023-09-12 02:11:35 2023-09-12 00:11:35.669 [INF] GRDN: Starting main custodian event loop
2023-09-12 02:11:35 2023-09-12 00:11:35.720 [INF] AUTO: Starting Autopilot Client
2023-09-12 02:11:35 2023-09-12 00:11:35.726 [INF] LITD: Starting LiT session server
2023-09-12 02:11:35 2023-09-12 00:11:35.726 [INF] LITD: Starting LiT account service
2023-09-12 02:11:35 2023-09-12 00:11:35.727 [INF] LITD: Starting LiT middleware manager
2023-09-12 02:11:35 2023-09-12 00:11:35.728 [INF] LITD: Internal sub server startup complete
2023-09-12 02:11:36 2023-09-12 00:11:36.458 [WRN] GRPC: [core] grpc: addrConn.createTransport failed to connect to {davepool:12010 davepool:12010 <nil> 0 <nil>}. Err: connection error: desc = "transport: Error while dialing dial tcp: lookup davepool on 127.0.0.11:53: no such host". Reconnecting...
2023-09-12 02:11:38 2023-09-12 00:11:38.208 [WRN] GRPC: [core] grpc: addrConn.createTransport failed to connect to {davepool:12010 davepool:12010 <nil> 0 <nil>}. Err: connection error: desc = "transport: Error while dialing dial tcp: lookup davepool on 127.0.0.11:53: no such host". Reconnecting...
2023-09-12 02:11:38 2023-09-12 00:11:38.292 [INF] LITD: Handling gRPC request: /litrpc.Status/SubServerStatus
2023-09-12 02:11:38 2023-09-12 00:11:38.293 [INF] LITD: Handling gRPC request: /lnrpc.State/GetState
2023-09-12 02:11:41 2023-09-12 00:11:41.214 [WRN] GRPC: [core] grpc: addrConn.createTransport failed to connect to {davepool:12010 davepool:12010 <nil> 0 <nil>}. Err: connection error: desc = "transport: Error while dialing dial tcp: lookup davepool on 127.0.0.11:53: no such host". Reconnecting...
```
The `lncli status` command executed in that log snippet marked `pool` as `"running":true` for some reason.

Other than this, this PR looks great :rocket::fire:!!

ellemouton:
> If running LiT with a subserver configured as in remote mode (specifically pool, loop & faraday), and if I then manually shutdown the remote subserver, the lncli status command will still mark the specific subserver as "running".

Yep - this is expected. This initial version is only useful at startup time and it does not ping the subservers or update the status during the lifetime of the daemon. This is something that can be added in future

ellemouton:
> Even more strangely, for some reason which I can't figure out wh

cool - that sounds wrong :+1: I'll look into that 👍

subservers/manager.go
```diff
@@ -32,37 +33,55 @@ var (
...
+		return
+	}
+
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	// Add the enabled server to the set of servers tracked by the Manager.
 	s.servers = append(s.servers, &subServerWrapper{
 		SubServer: ss,
 		quit:      make(chan struct{}),
```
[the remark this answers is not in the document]
ellemouton: At the moment, this is is only being called when LiT itself is also shutting down. meaning the status server anyways wont be up or meaningful. 

I think in a future PR we can address restarts of sub-servers etc

subservers/manager.go
```diff
@@ -74,19 +93,24 @@ func (s *Manager) StartIntegratedServers(lndClient lnrpc.LightningClient,
...
 
-	return nil
+		s.statusServer.SetRunning(ss.Name())
+	}
 }
 
 // ConnectRemoteSubServers creates connections to all the manager's sub-servers
 // that are running remotely.
-func (s *Manager) ConnectRemoteSubServers() error {
+func (s *Manager) ConnectRemoteSubServers() {
```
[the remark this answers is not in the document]
ellemouton: yep - this PR currently only handles start up errors.

ellemouton:
> (2). Even more strangely, for some reason which I can't figure out why when looking a

Ok so this is cause currently for remote sub-servers, we dont block on the connection being successful. Instead we give it:

```
grpc.WithConnectParams(grpc.ConnectParams{
	Backoff:           backoff.DefaultConfig,
	MinConnectTimeout: defaultConnectTimeout,
})
```

and return the unconnected connection so that LiT continues to maintain the connection if the remote server gets restarted or something. I dont think we want to change this behaviour right now. And I think in most cases, the pool/faraday/loop  etc servers are run in integrated mode in anycase. 

I think in a follow up PR, we can add a heartbeat mechanism to the status server to give more live updates regarding sub-server statuses.

ViktorTigerstrom (approved):
tACK LGTM! Great work :fire: :rocket: :fire:

subservers/manager.go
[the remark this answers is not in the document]
ViktorTigerstrom: Ok makes sense :)!

ViktorTigerstrom:
> > (2). Even more strangely, for some reason which I can't figure out why when looking a
> 
> Ok so this is cause currently for remote sub-servers, we dont block on the connection being successful. Instead we give it:
> 
> ```
> grpc.WithConnectParams(grpc.ConnectParams{
> 	Backoff:           backoff.DefaultConfig,
> 	MinConnectTimeout: defaultConnectTimeout,
> })
> ```
> 
> and return the unconnected connection so that LiT continues to maintain the connection if the remote server gets restarted or something. I dont think we want to change this behaviour right now. And I think in most cases, the pool/faraday/loop etc servers are run in integrated mode in anycase.
> 
> I think in a follow up PR, we can add a heartbeat mechanism to the status server to give more live updates regarding sub-server statuses.

Ok makes sense, thanks for checking! Sound like a good plan

config.go
```diff
@@ -187,16 +187,16 @@ type Config struct {
...
 	Loop     *loopd.Config `group:"Integrated loop options (use when loop-mode=integrated)" namespace:"loop"`
 
-	PoolMode string       `long:"pool-mode" description:"The mode to run pool in, either 'integrated' (default) or 'remote'. 'integrated' means poold is started alongside the UI and everything is stored in pool's main data directory, configure everything by using the --pool.* flags. 'remote' means the UI connects to an existing poold node and acts as a proxy for gRPC calls to it." choice:"integrated" choice:"remote"`
+	PoolMode string       `long:"pool-mode" description:"The mode to run pool in, either 'integrated' (default), 'remote' or 'disabled'. 'integrated' means poold is started alongside the UI and everything is stored in pool's main data directory, configure everything by using the --pool.* flags. 'remote' means the UI connects to an existing poold node and acts as a proxy for gRPC calls to it. 'disable' means that LiT is started without pool." choice:"integrated" choice:"remote" choice:"disable"`
```
ViktorTigerstrom, line 196: nit:
```suggestion
	PoolMode string       `long:"pool-mode" description:"The mode to run pool in, either 'integrated' (default), 'remote' or 'disable'. 'integrated' means poold is started alongside the UI and everything is stored in pool's main data directory, configure everything by using the --pool.* flags. 'remote' means the UI connects to an existing poold node and acts as a proxy for gRPC calls to it. 'disable' means that LiT is started without pool." choice:"integrated" choice:"remote" choice:"disable"`
```

guggero (approved):
Amazing work. tACK, LGTM :tada:

rpc_proxy.go
```diff
@@ -375,8 +377,8 @@ func (p *rpcProxy) UnaryServerInterceptor(ctx context.Context, req interface{},
 	info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{},
 	error) {
 
-	if !p.hasStarted() && !isStatusReq(info.FullMethod) {
-		return nil, ErrWaitingToStart
+	if err := p.checkSubSystemStarted(info.FullMethod); err != nil {
```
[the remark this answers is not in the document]
ellemouton: done :)

status/manager.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	return status, nil
+}
+
+// SetEnabled marks the sub-server with the given name as enabled.
+//
+// NOTE: This will silently fail if the referenced sub-server has not yet been
+// registered.
+func (s *Manager) SetEnabled(name string) {
```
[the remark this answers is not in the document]
ellemouton: done :)

guggero:
I assume we'll want to wait for https://github.com/lightninglabs/lightning-terminal/pull/620 before merging this?

ellemouton:
@guggero  - yes :)

PR #546, opened 2023-05-10
itest: export `ServerHarness` and `HarnessNode`

ViktorTigerstrom:
In this PR we export the itest structs `ServerHarness` and `HarnessNode`. This is useful to make it possible to use the structures independently in other test frameworks that need to run and interact with Lit instance(s), and reuse the structures from the `lightning-terminal` repo.

Let me know if there are other components of the itest framework that you think would be useful to export, and I'll update this PR with those as well :)!


ellemouton (commented):
LGTM!

Let's wait until the dependent PR is ready to go before merging this in just so that we can make any last minute exports here if needed.

ellemouton (approved):
Dependent PR looks good soooo......LGTM 🔥

bitromortac (approved):
LGTM :tada:

PR #553, opened 2023-05-17
ui+auth: updates login page view

dstrukt:
Not sure why the test is failing, I've removed that snippet of code in the test..

Current:
Screen Shot 2023-05-16 at 6 10 39 PM

Proposed:
Screen Shot 2023-05-17 at 9 49 44 AM
Screen Shot 2023-05-17 at 9 49 52 AM
Screen Shot 2023-05-17 at 9 50 02 AM

### DIFF
--- a/app/src/components/auth/AuthPage.tsx
+++ b/app/src/components/auth/AuthPage.tsx
@@ -4,7 +4,8 @@ import styled from '@emotion/styled';
 import { ReactComponent as LogoImage } from 'assets/images/logo.svg';
 import { usePrefixedTranslation } from 'hooks';
 import { useStore } from 'store';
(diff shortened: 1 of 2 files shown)


itsrachelfish (changes requested):
@dstrukt Is the sidebar supposed to look like this? :sweat_smile: It doesn't seem to match the color scheme of the background.

Screenshot from 2023-08-16 00-22-25

Additionally, I noticed that the background gradient only fills 1 viewport height. When the page is longer than the screen and you scroll down, the background color changes

Screenshot from 2023-08-16 00-23-52

guggero:
!lightninglabs-deploy mute

PR #558, opened 2023-05-19
go.mod: bump lndclient to 0.16.0-12

ellemouton:
This PR bumps the `lndclient` version to one that ensures that a more meaningful error
is returned if `RegisterRPCMiddleware` fails. Currently, the wanted error will be logged but
the official error will be `context cancelled` which is not very helpful. 

This will be especially important once https://github.com/lightninglabs/lightning-terminal/pull/541 is merged since we want 
meaningful errors to be displayed.


guggero (approved):

ViktorTigerstrom (approved):
LGTM

PR #560, opened 2023-05-29
make: install litcli with `make go-install-cli`

ViktorTigerstrom:
This PR adds installation of the `litcli` executable when running the `make go-install-cli` command. The command should install all cli executables.

Happy to address any feedback! :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,6 +138,7 @@ go-install-cli:
 	$(GOINSTALL) -trimpath github.com/lightninglabs/faraday/cmd/frcli
 	$(GOINSTALL) -trimpath -ldflags "$(LDFLAGS)" github.com/lightninglabs/pool/cmd/pool
 	$(GOINSTALL) -trimpath -ldflags "$(LDFLAGS)" github.com/lightninglabs/taproot-assets/cmd/tapcli
+	$(GOINSTALL) -trimpath -tags="$(LND_RELEASE_TAGS)" -ldflags "$(LDFLAGS)" $(PKG)/cmd/litcli
 
 app-build: yarn-install
 	@$(call print, "Building production app.")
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -138,6 +138,7 @@ go-install-cli:
 	$(GOINSTALL) -trimpath github.com/lightninglabs/faraday/cmd/frcli
 	$(GOINSTALL) -trimpath -ldflags "$(LDFLAGS)" github.com/lightninglabs/pool/cmd/pool
 	$(GOINSTALL) -trimpath -ldflags "$(LDFLAGS)" github.com/lightninglabs/taproot-assets/cmd/tapcli
+	$(GOINSTALL) -tags="$(LND_RELEASE_TAGS)" -ldflags "$(LDFLAGS)" $(PKG)/cmd/litcli
```
ViktorTigerstrom, line 141: I was a bit unsure if I should include the `-trimpath` argument here, as we're only adding the cli executable. Though I figured that users will likely run both `make go-install `and `make go-install-cli`. I therefore opted to not do so, as I didn't want that the order the commands are run in to effect the resulting `litcli` executable after running both commands.

guggero: The `-trimpath` removes local directory paths to golang modules and replaces them with static ones, helping with the binary reproducibility of the release binaries. So IMO we should add the `-trimpath` flag to every binary that is not a debug/dev build. Meaning that we should probably also have them in `go-install`.

ViktorTigerstrom: Ok, thanks for the feedback! Updated by adding the `trimpath` flag to both this, and `go-install`!

guggero (approved):

jamaljsr (approved):
tACK LGTM 👌

jamaljsr:
Just needs a rebase to merge.

ViktorTigerstrom:
Rebased! No changes made.

ellemouton (approved):

PR #568, opened 2023-06-02
cmd/litcli: add super macaroon helper commands

ellemouton:
This PR adds a few super macaroon related commands:

-  `litcli helper supermacrootkey` command is added which lets a user generate a random root key ID for a super macaroon
-  `litcli helper issupermacaroon` command allows a users to easily confirm if a macaroon is considered a super macaroon or not. 
- `litcli helper bakesupermacaroon` can be used to bake a super macaroon. 

The `supermacrootkey` and `issupermacaroon` commands  dont require a connection to LiTd.


ellemouton:
happy to add any other helpers in this area that would be useful :) ideas welcome!

ellemouton:
i could maybe add a `bakesupermac` endpoint?

CRex15:
> i could maybe add a `bakesupermac` endpoint?

That would be quite helpful. It was quite complicated and error prone trying to bake the super macaroon with the normal BakeMacaroon API as described here: [https://github.com/lightninglabs/lightning-api-ng/issues/19](url). I think it would also help clarify that there is a difference between the two, which I was unaware of before running into that issue.

cmd/litcli/helpers.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			return err
+		}
+
+		binary.BigEndian.PutUint32(suffix[:], uint32(suffixInt))
+	}
+
+	id := session.NewSuperMacaroonRootKeyID(suffix)
+
+	fmt.Println("Root Key ID: ", id)
```
guggero, line 64: It looks like for every other command (except the shutdown command) we return JSON. Should we do that here as well, just to be consistent? Can be an anonymous struct with JSON tags passed into the `printJSON()` function.

cmd/litcli/helpers.go
```diff
@@ -0,0 +1,101 @@
...
+	Name: "supermacrootkey",
+	Usage: "Generate a valid super macaroon root key ID from scratch " +
+		"or from a given root key ID suffix",
+	Description: `
+	This command can be used to generate a valid super macaroon root key ID
+	from scratch or from a given root key ID suffix.
+	`,
+	Action: superMacRootKey,
+	Flags: []cli.Flag{
+		cli.StringFlag{
```
guggero, line 35: It's not really clear to me just from the text how to specify this. Should we allow this to be 4 bytes (=8 hex characters) so we can simplify the checks a bit?

cmd/litcli/helpers.go
```diff
@@ -0,0 +1,101 @@
...
+	isNotSuperMac := ""
+	if !isSuperMac {
+		isNotSuperMac = "not "
+	}
+
+	resp := fmt.Sprintf(
+		"The given macaroon is %sa super macaroon", isNotSuperMac,
+	)
+
+	fmt.Println(resp)
```
guggero, line 98: nit: print JSON instead?
```
	printJSON(struct{
		IsSuperMacaroon bool `json:"is_super_macaroon"`
	}{
		IsSuperMacaroon: isSuperMac,
	})

...


func printJSON(resp interface{}) {
	b, err := json.Marshal(resp)
	if err != nil {
		fatal(err)
	}

	var out bytes.Buffer
	json.Indent(&out, b, "", "\t")
	out.WriteString("\n")
	out.WriteTo(os.Stdout)
}

```

guggero (commented):
Very nice addition! Just a few usability/UX comments, otherwise looks great.

cmd/litcli/helpers.go
ellemouton: good idea!

cmd/litcli/helpers.go
ellemouton: good point :) updated!

ellemouton (commented):
Thanks for the review @guggero ! I've updated and have added a `BakeSuperMacaroon` endpoint along with a `bakesupermacaroon` litcli command.

cmd/litcli/proxy.go
```diff
@@ -22,6 +27,29 @@ var litCommands = []cli.Command{
...
 	},
+	{
+		Name: "bakesupermacaroon",
+		Usage: "Bake a new super macaroon with all of LiT's active " +
+			"permissions.",
+		Category: "LiT",
+		Action:   bakeSuperMacaroon,
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name: "suffix",
```
guggero, line 38: nit: I think in this context the flag should be called "root_key_suffix", otherwise it might be confusing.

guggero (approved):
Very nice, LGTM :tada:

guggero:
Ah, just one request: Can we please add the new RPC to one of the itests? Just to make sure authentication and everything works as expected.

ellemouton:
> Can we please add the new RPC to one of the itests? Just to make sure authentication and everything works as expected.

is this not covered by the existing test for `litrpc.Proxy.GetInfo`? since they are both for the same rpc service. 

or do you mean that we should use the new endpoint to bake a macaroon in the itests & then test that that macaroon works for other endpoints?

guggero:
> is this not covered by the existing test for litrpc.Proxy.GetInfo? since they are both for the same rpc service.

Ah yes, you're right. So all the security restrictions should apply to the new call as well. SGTM.

cmd/litcli/proxy.go
```diff
@@ -22,6 +27,29 @@ var litCommands = []cli.Command{
...
+			"permissions.",
+		Category: "LiT",
+		Action:   bakeSuperMacaroon,
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name: "root_key_suffix",
+				Usage: "A 4-byte suffix to use in the " +
+					"construction of the root key ID. " +
+					"If not provided, then a random one " +
+					"will be generated. This must be " +
```
ViktorTigerstrom, line 42: Nit:
```suggestion
					"will be generated. This must be a " +
```

cmd/litcli/helpers.go
```diff
@@ -0,0 +1,125 @@
...
+	This command can be used to generate a valid super macaroon root key ID
+	from scratch or from a given root key ID suffix.
+	`,
+	Action: superMacRootKey,
+	Flags: []cli.Flag{
+		cli.StringFlag{
+			Name: "root_key_suffix",
+			Usage: "A 4-byte suffix to use in the construction " +
+				"of the root key ID. If not provided, then a " +
+				"random one will be generated. This must be " +
```
ViktorTigerstrom, line 42: Nit:
```suggestion
				"random one will be generated. This must be a " +
```

litrpc/proxy.pb.go
```diff
@@ -220,20 +331,24 @@ func file_proxy_proto_rawDescGZIP() []byte {
 	return file_proxy_proto_rawDescData
 }
 
-var file_proxy_proto_msgTypes = make([]protoimpl.MessageInfo, 4)
+var file_proxy_proto_msgTypes = make([]protoimpl.MessageInfo, 6)
 var file_proxy_proto_goTypes = []interface{}{
-	(*StopDaemonRequest)(nil),  // 0: litrpc.StopDaemonRequest
```
ViktorTigerstrom, line 225: This is probably a stupid comment, but just wanted to make sure this isn't a potential issue:

There is no problem here that this overwrites the old `msgTypes` values here with new types right (ie. value `0` was the type `litrpc.StopDaemonRequest` but now becomes `litrpc.BakeSuperMacaroonRequest`)? Just thinking if we'd have external services or something that uses the protos and that way become incompatible, or if this could somehow become a backwards compatibility issue if one downgrades LiT or so.

This is likely not something that could become an issue, but thought I'd just leave this comment to make sure that's the case.

ViktorTigerstrom (approved):
Awesome, LGTM once rebased 🚀! Should be really useful commands!!

Only thing I noticed are a minor typo, and a sanity check to ensure my comment below isn't a potential issue.

litrpc/proxy.pb.go
ellemouton: good question!

So I think that either:
- these indices are only used within this file. So if you scroll up to where `file_proxy_proto_msgTypes` is used, you can see that it is only used by the message types once the types are known and are not used on the wire. 
- _or_, they are used on the wire but the mapping is conveyed via the `file_proxy_proto_rawDesc` descriptor above.

either way -  gRPC protos are backwards compatible & you can test this by running Lit with this PR but compiling litcli with master & then see that the `litcli getinfo` or `litcli stop` calls work just the same no matter if litcli is compiled with this branch or master branch :)

PR #571, opened 2023-06-08
litclient: add taprpc to registrations

ViktorTigerstrom:
This PR adds the taprpc to the litclient's `Registrations` array, to ensure that the TaprootAssets callbacks are registered and available in the litclient.


jamaljsr (approved):
Looks like this needs a rebase, but I tested this with https://github.com/lightninglabs/lightning-node-connect/pull/77 and it worked. 👌

levmi:
cc @dstadulis who can help us prio any necessary change on the `tapd` side

ViktorTigerstrom:
Opened a PR for the required fix in the taproot-assets repo, and updated this PR to use a forked version containing the fix.

 Also updated this PR to register the `assetwalletrpc` `universerpc` JSON callbacks in the litclients `Registrations` array as well!

ViktorTigerstrom:
Updated this PR to use the latest version of the main branch for the taproot-asset dependency to include the required fix. Also update the litclient `Registrations` to include `mintrpc` 🚀

ellemouton:
we might not want to include all the other changes in the `taproot-assets` code since the `v0.2.0` release since it could have behaviour changes that the user is not expecting. Users would then be running an un-released version of taproot-assets if they were to run it via LiT (if we were to do another LiT release that did not include a Taproot-assets bump)

This is something that we might want to do quite often in LiT & I think we dont always want to have to wait for a release on the sub-server side just cause we want to include some changes on the code-structure level that wont actually have behaviour changes for the user. So I think we should decide on a pattern for including such changes. 

My suggestion would be: we should create a `v0.2.0-lit-0` tag (using taproot-assets as an example) in that repo which points to a commit that includes the released version (v0.2.0) _plus_ any specific commits we require for Lit. In this case it would be just the two commits added [here](https://github.com/lightninglabs/taproot-assets/pull/349). The next time we add a commit there, we add a `v0.2.0-lit-1` tag etc. Then on the next `taproot-assets` release, we just use the `v0.2.1` tag again etc etc. 

Keen to hear what @guggero things of this idea 🙏

guggero:
Good point, @ellemouton. There weren't any braking changes merged in `taproot-assets` recently, only bug fixes. And `v0.2.1` should be fairly close. But IMO it still makes sense to create a branch, so I did that and pushed the `v0.2.0-lit-0` tag.

ellemouton:
shweet! thanks @guggero  🙏 🚀 

@ViktorTigerstrom  - can you update this PR to point to the new tag? after that, this is g2g!

ViktorTigerstrom:
Thanks for the great feedback @ellemouton! Makes a lot sense. And thanks for creating the tag so quickly @guggero!!

Updated the dependency to use the new` v0.2.0-lit-0` tag with the latest push!

go.mod
```diff
@@ -224,6 +224,11 @@ require (
 
 replace github.com/lightninglabs/lightning-terminal/autopilotserverrpc => ./autopilotserverrpc
 
+// We want to format raw bytes as hex instead of base64. The forked version
+// allows us to specify that as an option. This is required for the
+// taproot-assets dependency to function properly.
+replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-display v1.30.0-hex-display
```
guggero, line 230: :+1:

guggero (approved):
utACK, LGTM :tada:

ellemouton (approved):
tACK 🚀

PR #589, opened 2023-07-05
build: bump lnd to v0.16.4 and tapd to v0.2.2

Roasbeef:


README.md
```diff
@@ -139,6 +139,7 @@ The following table shows the supported combinations:
 
 | LiT               | LND          | Loop         | Faraday       | Pool         | Taproot Assets |
 |-------------------|--------------|--------------|---------------|--------------|----------------|
+| **v0.10.2-alpha** | v0.16.4-beta | v0.24.1-beta | v0.2.11-alpha | v0.6.4-beta  | v0.2.2-alpha   |
```
ViktorTigerstrom, line 142: I think we should include this in the commit that's bumping the LiT version instead, and then have that commit after the different dependencies bumps :).

README.md
```diff
@@ -139,6 +139,7 @@ The following table shows the supported combinations:
 
 | LiT               | LND          | Loop         | Faraday       | Pool         | Taproot Assets |
```
ViktorTigerstrom, line 140: We'd also need to add an entry for `0.10.2-alpha` above for minimum compatibility version of LND in remote mode.

ViktorTigerstrom (commented):
Generally looks good to me :)! Just a few small changes suggested.

Also:
Would we also like to bump `loop` to the latest version in this PR?

levmi:
Seems reasonable to bump Loop as well, it's a pretty major change wrt moving to a SQLite backend option. I'm not sure if that has any impact on our process, but if you're willing @ViktorTigerstrom, it's probably worth you picking this up and pushing it over the finish line? (tomorrow your time obviously not tonight :)

ViktorTigerstrom:
Updated the PR to also bump loop, and addressed my feedback above :)!

ellemouton (approved):
tACK 🎸

guggero (approved):
LGTM :tada:

itsrachelfish (approved):
tACK, created a few sessions and confirmed they worked in terminal-web with loop

Issue #592, opened 2023-07-08
query a specific account for details


AndySchroder:
**Is your feature request related to a problem? Please describe.**

Efficient administration of lnd accounts.

**Describe the solution you'd like**
Would like RPC calls for getting details about a specific account.

**Describe alternatives you've considered**
We have https://lightning.engineering/api-docs/api/lit/accounts/list-accounts , but that returns a list of all accounts. It also returns all payments for all accounts, so after a lot of accounts and activity on the server, this call is going to likely be very inefficient to use regularly.

We could instead generate a macaroon for the account, then use the macaroon to make calls to LND, but that seems like it is a bit more complex considering https://github.com/lightninglabs/lightning-terminal/issues/582 .

Issue #581, opened 2023-07-02
add a name or label option to `litcli accounts`


AndySchroder:
**Is your feature request related to a problem? Please describe.**

Way too much work to keep track of custodial accounts.

**Describe the solution you'd like**
add a name or label option to `litcli accounts`

**Describe alternatives you've considered**

Manually keep track of the id returned when you run `litcli accounts create 50000`.

levmi:
Appreciate the feedback again! This one should be relatively straightforward and easy. We'll start to get this into our queue.

AndySchroder:
I'm assuming it was implied here, but `litcli accounts create`, `litcli accounts update`, and `litcli accounts remove` should all accept the label that was assigned when creating instead of requiring the id to be used (both id or label should be usable).

AndySchroder:
Another thing that would be very useful: If there is a way to make a GRPC call by the user of the macaroon to find out the name of the account that they are using. For example, say a macaroon is used by some machine to machine interfaces like I described in https://github.com/lightninglabs/lightning-terminal/issues/583, the machine might want to be able to identify the name of the account that it is using so that it can remind the users (of the service that it is enabling) what account they need to refill if its allocated balance starts to get low (and maybe it can tell them on a monitoring screen built into that machine). This takes the burden off the users to keep track of what account is used where and it is one less piece of information to need to configure on the device. For example, say the machine is used in an organization (family, business, club, etc.) and it provides some critical service to the group, there are going to be various people that may be responsible for making sure everything is always running. They might have the ability to send more funds to each machine's account, but they don't have any control of the machines themselves. This feature would also allow you to rename the account and then that new name propagate out to all the devices or wallets that use the account.

Implementing this may be tricky because the RPC Middleware Interceptor doesn't really provide new grpc calls to LND, it just limits existing ones, so you may need to add some new message type to LND for metadata that can be intercepted to make it work.

PR #605, opened 2023-07-26
accounts: add label and AccountInfo RPC

guggero:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/592.
Fixes https://github.com/lightninglabs/lightning-terminal/issues/581.

Adds a label to custodial accounts and allows individual accounts to be queried through the RPC or CLI (instead of needing to list all accounts).


cmd/litcli/accounts.go
```diff
@@ -293,29 +348,59 @@ func removeAccount(ctx *cli.Context) error {
...
-	}
-	if _, err := hex.DecodeString(accountID); err != nil {
-		return err
-	}
+		// Since we have a positional argument, we cannot be sure it's
+		// an ID. So we check if it's an ID by trying to hex decode it
+		// and by checking the length. This will break if the user
+		// chooses labels that are also valid hex encoded IDs. But since
+		// the label is supposed to be human-readable, this should be
+		// unlikely.
```
ViktorTigerstrom, line 392: nit: It shouldn't be possible for the user to set such labels, no?

ViktorTigerstrom (approved):
LGTM! Great additional features to add 🚀!

Just leaving one tiny nit that's non blocking.

cmd/litcli/accounts.go
guggero: Good point, I added a check here: https://github.com/lightninglabs/lightning-terminal/blob/568d7ad508933df0bcb4603b48594d3d0558f57b/accounts/store.go#L119

ViktorTigerstrom: Ah yeah, awesome! I realize that my comment above was a bit unclear, but I ment to say that the part of this comment I've marked isn't necessary, and can be removed as it shouldn't be possible for users to set such labels :). Though feel free to ignore this, as this isn't blocking at all and very nit picky!

bitromortac (approved):
LGTM :zap:, great work!

PR #616, opened 2023-08-07
terminal: dont block indefinitely on macaroon channel

ellemouton:
In this commit, we ensure that when we wait on the macaroon channel for the macaroon from the integrated LND, that we also listen to other exit signals such as the interceptor, lndQuit chanel and the error channel. This is to prevent shutdown from blocking forever in the case where there was an issue preventing LND from sending the initial macaroon.

The issue was caught in this comment: https://github.com/lightninglabs/lightning-terminal/pull/541#discussion_r1285486027

### DIFF
--- a/terminal.go
+++ b/terminal.go
@@ -536,31 +536,56 @@ func (g *LightningTerminal) start() error {
 		return fmt.Errorf("error displaying startup info: %v", err)
 	}
 
+	// waitForSignal is a helper closure that can be used to wait on the
+	// given channel for a signal while also being responsive to an error
+	// from the error Queue, LND quiting or the interceptor receiving a
+	// shutdown signal.
+	waitForSignal := func(c chan struct{}) error {
+		select {
+		case <-c:
+			return nil
+
+		case err := <-g.errQueue.ChanOut():
+			return err
+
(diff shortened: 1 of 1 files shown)


terminal.go
```diff
@@ -526,31 +526,57 @@ func (g *LightningTerminal) start() error {
...
-	// Wait for lnd to be unlocked, then start all clients.
-	select {
-	case <-readyChan:
+	// waitForSignal is a helper closure that can be used to wait on the
+	// given channel for a signal while also being responsive to an error
+	// from the error Queue, LND quiting or the interceptor receiving a
+	// shutdown signal.
+	waitForSignal := func(c chan struct{}) error {
+		select {
+		case <-c:
```
guggero, line 535: nit: return `nil` here (and remove the `return` at the end of the closure) to make things even more explicit?

guggero (approved):
tACK, LGTM :tada: 
Very nice fix, thanks!

terminal.go
ellemouton: good idea :) done!

ViktorTigerstrom (approved):
tACK LGTM 🚀 🔥!

PR #617, opened 2023-08-07
multi: allow LND and subserver whitelisted calls

ellemouton:
This PR does a few things:

1. Ensures that Lit register's LND's State server to it's REST server 
2. Allow LND's & Lit's and all Lit's subserver's whitelisted endpoints to pass through without requiring a macaroon

### DIFF
--- a/perms/manager.go
+++ b/perms/manager.go
@@ -49,8 +49,13 @@ func NewManager(withAllSubServers bool) (*Manager, error) {
 	permissions := make(map[string]map[string][]bakery.Op)
 	permissions[litPerms] = RequiredPermissions
 	permissions[lndPerms] = lnd.MainRPCServerPermissions()
-	for k, v := range whiteListedLNDMethods {
-		permissions[lndPerms][k] = v
+
+	for url := range whiteListedLitMethods {
+		permissions[litPerms][url] = []bakery.Op{}
+	}
+
+	for url := range whiteListedLNDMethods {
+		permissions[lndPerms][url] = []bakery.Op{}
 	}
 
 	// Collect all LND sub-server permissions along with the name of the
@@ -94,6 +99,18 @@ func NewManager(withAllSubServers bool) (*Manager, error) {
 		fixedPerms:        permissions,
 		perms:             allPerms,
 	}, nil
+}
+
+// IsWhiteListedURL returns true if the given URL has been whitelisted meaning
+// that it does not require a macaroon for validation. A URL is considered
+// white-listed if it has no operations associated with a URL.
+func (pm *Manager) IsWhiteListedURL(url string) bool {
+	pm.permsMu.Lock()
+	defer pm.permsMu.Unlock()
+
+	ops, ok := pm.perms[url]
+
+	return ok && len(ops) == 0
 }
 
 // RegisterSubServer adds the permissions of a given sub-server to the set
--- a/perms/manager.go
+++ b/perms/manager.go
@@ -116,7 +116,7 @@ func (pm *Manager) IsWhiteListedURL(url string) bool {
 // RegisterSubServer adds the permissions of a given sub-server to the set
 // managed by the Manager.
 func (pm *Manager) RegisterSubServer(name string,
-	permissions map[string][]bakery.Op) {
+	permissions map[string][]bakery.Op, whiteListURLs map[string]struct{}) {
 
 	pm.permsMu.Lock()
 	defer pm.permsMu.Unlock()
@@ -125,6 +125,15 @@ func (pm *Manager) RegisterSubServer(name string,
 
 	for uri, ops := range permissions {
 		pm.perms[uri] = ops
+	}
+
+	for url := range whiteListURLs {
+		pm.perms[url] = nil
+
+		if pm.fixedPerms[name] == nil {
+			pm.fixedPerms[name] = make(map[string][]bakery.Op)
+		}
+		pm.fixedPerms[name][url] = []bakery.Op{}
 	}
 }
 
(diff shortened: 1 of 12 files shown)


guggero (approved):
LGTM :tada:

jamaljsr:
When integrating with `tapd`'s universe stats HTTP endpoints, I ran into an issue with `litd` requiring a macaroon for [whitelisted](https://github.com/lightninglabs/taproot-assets/blob/v0.2.3/perms/perms.go#L208-L209) RPCs. 

Can this be addressed in this PR or should I open a separate issue?

ellemouton:
ah! great catch @jamaljsr ! Yeah I think it defs makes sense to add that in this PR. Will do asap 👍

perms/manager.go
```diff
@@ -96,6 +101,18 @@ func NewManager(withAllSubServers bool) (*Manager, error) {
...
+// IsWhiteListedURL returns true if the given URL has been whitelisted meaning
+// that it does not require a macaroon for validation. A URL is considered
+// white-listed if it has no operations associated with a URL.
+func (pm *Manager) IsWhiteListedURL(url string) bool {
+	pm.permsMu.Lock()
+	defer pm.permsMu.Unlock()
+
+	ops, ok := pm.perms[url]
+
+	return ok && len(ops) == 0
```
ellemouton, line 113: is this an ok axiom?

ellemouton: compared to a previous version of this pr which kept an explicit whitelist. Just realised it might not be needed if this assumption is ok

ellemouton:
@guggero - apologies, im re-requesting your review cause quite a bit has changed now so that whitelisted calls from other sub-servers are also taken into account

subservers/taproot-assets.go
```diff
@@ -174,3 +174,11 @@ func (t *taprootAssetsSubServer) MacPath() string {
 func (t *taprootAssetsSubServer) Permissions() map[string][]bakery.Op {
 	return perms.RequiredPermissions
 }
+
+// WhiteListedURLs returns a map of all the sub-server's URLs that can be
+// accessed without a macaroon.
+//
+// NOTE: this is part of the SubServer interface.
+func (t *taprootAssetsSubServer) WhiteListedURLs() map[string]struct{} {
+	return perms.MacaroonWhitelist(true)
```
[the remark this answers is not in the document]
guggero: We should probably use the value of `t.cfg.RpcConf.AllowPublicStats` instead.

perms/manager.go
```diff
@@ -49,8 +49,13 @@ func NewManager(withAllSubServers bool) (*Manager, error) {
 	permissions := make(map[string]map[string][]bakery.Op)
 	permissions[litPerms] = RequiredPermissions
 	permissions[lndPerms] = lnd.MainRPCServerPermissions()
-	for k, v := range whiteListedLNDMethods {
-		permissions[lndPerms][k] = v
+
+	for k := range whiteListedLitMethods {
```
guggero, line 53: If we're using the `IsWhiteListedURL` is being used, do we still need this? I assume yes, just to still recognize the URI as a valid call?

guggero, line 53: nit: `k` vs `url` below.

guggero (commented):
Very nice, just one small change then we're good to go.

perms/manager.go
ellemouton: resolving as discussed offline :)

subservers/taproot-assets.go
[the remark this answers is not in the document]
ellemouton: cool, added the `t.cfg.RpcConf.AllowPublicStats | t.remote` as discussed

ellemouton:
Thanks @guggero 🎉  updated!

subservers/taproot-assets.go
```diff
@@ -174,3 +174,19 @@ func (t *taprootAssetsSubServer) MacPath() string {
...
 	return perms.RequiredPermissions
 }
+
+// WhiteListedURLs returns a map of all the sub-server's URLs that can be
+// accessed without a macaroon.
+//
+// NOTE: this is part of the SubServer interface.
+func (t *taprootAssetsSubServer) WhiteListedURLs() map[string]struct{} {
+	// If the taproot-asset daemon is running in integrated mode, then we
+	// the cfg.RpcConf.AllowPublicStats to determine if the public stats
```
guggero, line 184: nit: "then we _use_ the ..."?

guggero (approved):

ViktorTigerstrom (approved):
LGTM, very clean PR :fire::rocket:!

Just for clarification, I just want to verify as well that it's intentional that the `LightningTerminal.Permissions` function won't include any whitelisted urls, as will happen with this PR. That's intended behaviour, correct? I'm asking because I'm not 100% what "permissions for which the external validator of the terminal is responsible." really entails.

ellemouton:
> Just for clarification, I just want to verify as well that it's intentional that the LightningTerminal.Permissions function won't include any whitelisted urls, as will happen with this PR. That's intended behaviour, correct? I'm asking because I'm not 100% what "permissions for which the external validator of the terminal is responsible." really entails.

As far as I can see, `LightningTerminal.Permissions()` calls `perms.Manager.GetLitPerms()` which returns everything in  `fixedPerms`. If a Lit itself has whitelist perms, these will be added to `fixedPerms` in the `NewManager` function. Then, if any subserver registers whitelist URLs via `RegisterSubServer`, then these will also be added to `fixedPerms`. So `Lightning.Permissions` right now _will_ include the whiltelist calls. Or am I misunderstanding?

ViktorTigerstrom:
> As far as I can see, `LightningTerminal.Permissions()` calls `perms.Manager.GetLitPerms()` which returns everything in `fixedPerms`.

Ah wait, I realised that I was confused sorry. I interpreted yesterday  that `GetLitPerms` looped over the specific `[]bakery.Op` list for an url to add it to `result`, which would have resulted in whitelisted urls not getting added as a url is whitelisted by having an empty `[]bakery.Op` list.
But I see now that `GetLitPerms` loops over the outer `map[string][]bakery.Op`, which should mean that the specific url does indeed get added.

Sorry for the confusion :).

PR #620, opened 2023-08-08
Front-end integration for disabled subservers

itsrachelfish:
This PR creates a new SubServerStatus store which checks which services are being run by litd. When loop and pool are not running or are in an error state, the page displays an error message:

image

Depends on #541


app/src/store/store.ts
```diff
@@ -157,11 +159,18 @@ export class Store {
...
     await this.channelStore.fetchChannels();
-    await this.swapStore.fetchSwaps();
     await this.nodeStore.fetchBalances();
     await this.sessionStore.fetchSessions();
+
+    if (
+      this.subServerStore.subServers.loop?.running &&
+      !this.subServerStore.subServers.loop?.error
+    ) {
+      await this.swapStore.fetchSwaps();
```
jbrill, line 172: If our subserver loop state changes, will the store be refreshed ?

jbrill (changes requested):
Looks great! Just a few minor things

jbrill:
Also for the svg, we might want to tweak the color to be `fill="currentColor"` instead of `fill="none"`

dstrukt:
ACK to @jbrill's SVG comment!

app/src/store/store.ts
itsrachelfish: @jbrill After talking with Elle she explained that the subserver status server does not update when runtime errors are encountered. Currently the subserver status server only catches errors on startup.

Since we aren't polling the subserver status server for changes, there's no way for the state to change in the background. I should make sure that we fetch updated subserver information when navigating between pages though.

ellemouton: yeah - currently we can only check if it started successfully and there isnt currently a way to ping the server to see if it has crashed after startup (at least not at the moment)

jbrill: Gotcha -- thanks for the explanation 👍

app/src/store/stores/subServerStore.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.serverStatus.subServersMap.map(([serverName, serverStatus]) => {
+      subServers[serverName] = serverStatus;
+    });
+
+    return subServers;
+  }
+
+  /** fetch subserver statuses from the lit API */
+  async fetchStatus() {
+    const serverStatus = await this._store.api.lit.listSubServerStatus();
```
jbrill, line 34: Might make sense to have `try`/`catch` error handling here. Could also add a loading property to the class

jamaljsr (approved):
tACK. Great work on this @itsrachelfish!  I tested and confirmed both the `disabled` and `error` statuses. The functionality is good to go, so approved from my end.

I left some comments on code improvements that could be made if time permits, or saved for a follow-up PR.

app/src/store/views/appView.ts
```diff
@@ -36,6 +36,9 @@ export default class AppView {
   goTo(route: string) {
     if (this._store.router.location.pathname !== route) {
       this._store.router.push(route);
+
+      // Check for updates to subservers when navigating between pages
+      this._store.subServerStore.fetchStatus();
```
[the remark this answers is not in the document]
itsrachelfish: I included this in case the node restarts while the UI is still open.

This often happens for me during development. For example, while testing the subserver changes I often leave my lightning-terminal browser window open while I make changes to my `regtest.yml` file and restart the containers. Without this change my subserver status won't update unless I manually refresh the page. Maybe this isn't likely to happen in a real-world scenario :thinking:

jbrill (approved):

app/src/store/views/appView.ts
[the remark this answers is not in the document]
jamaljsr: Yea, I don't think it's a good trade-off to have all users fetch so aggressively just to streamline this rare case for average users. The streaming RPCs also disconnect when the node is restarted, so users need to reload the page to regain full functionality of the frontend app.

app/src/components/loop/LoopPage.tsx
```diff
@@ -26,8 +27,20 @@ const LoopPage: React.FC = () => {
...
+  if (subServerStore.subServers.loop?.disabled) {
+    return <SubServerStatus isDisabled={true} />;
+  } else if (subServerStore.subServers.loop?.error) {
+    return (
+      <SubServerStatus
+        isDisabled={false}
+        errorMessage={subServerStore.subServers.loop.error}
+      />
+    );
+  }
```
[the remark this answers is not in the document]
itsrachelfish: @jamaljsr As per your suggestion all of the subserver error message logic has been moved into a new `SubServerRequired` component 

https://github.com/lightninglabs/lightning-terminal/blob/54d04cb1288e962df8d8456d0d5d6631ccfa1c10/app/src/components/common/SubServerRequired.tsx#L67-L75

app/src/store/stores/subServerStore.ts
```diff
@@ -0,0 +1,51 @@
...
+  [key: string]: STATUS.SubServerStatus.AsObject;
+}
+
+export default class SubServerStore {
+  private _store: Store;
+
+  loading = false;
+
+  /** an array map of available subservers and their current status */
+  serverStatus: STATUS.SubServerStatusResp.AsObject = { subServersMap: [] };
```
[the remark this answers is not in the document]
itsrachelfish: I defined `SubServerStatus` like you suggested and put it in `types/state.ts`. I refactored all of the subserver logic to use this type instead of storing the RPC response object.

app/src/types/state.ts
```diff
@@ -69,3 +69,9 @@ export enum ChannelStatus {
  * duration is not just a random number.
  */
 export type LeaseDuration = number;
+
+export interface SubServerStatus {
+  disabled: boolean;
+  running: boolean;
+  error: string;
+}
```
jamaljsr, line 77: Perect location for this 👌

app/src/store/stores/subServerStore.ts
```diff
@@ -0,0 +1,47 @@
...
+  loop: SubServerStatus;
+  pool: SubServerStatus;
+}
+
+export default class SubServerStore {
+  private _store: Store;
+
+  loading = false;
+
+  subServers: SubServers = { loop: {} as SubServerStatus, pool: {} as SubServerStatus };
```
jamaljsr, line 16: I'd highly recommend initializing the fields in these obects with default values instead of empty object. You'll tend to run into hard to catch bugs when the types don't match the actual data stored in these variables.

```ts
  subServers: SubServers = {
    loop: {
      disabled: false,
      running: true,
      error: '',
    },
    pool: {
      disabled: false,
      running: true,
      error: '',
    },
  };
```

jamaljsr (commented):
Fantastic update! The code looks so much cleaner now.

Just have one more small suggestion.

app/src/store/stores/subServerStore.ts
itsrachelfish: Safer code is now being used :+1: 

https://github.com/lightninglabs/lightning-terminal/blob/106897df2a17595c8481c6e743d231124f469704/app/src/store/stores/subServerStore.ts#L16-L27

jbrill (approved):
Looks great! Thanks for the updates! 🚀

PR #621, opened 2023-08-08
multi: Autopilot session linking

ellemouton:
With this PR, we add the ability for a client to link a new autopilot session with
a previously revoked one. If this is done, then the same privacy mapper is used
for the linked sessions & the actions and rules db is also shared across sessions. 

## PR Flow: 

1. Update `autopilotserver.proto` to have the new fields required to link sessions. 
   These include the previous key along with a sig to prove ownership of the linked
   session.
2. update the autopilot server client to send the new fields if they are provided. 
    initially we do not provide them. 
3. Add `PrevLocalPub` and `GroupID` to the `Session` store. The `PrevLocalPub` 
    is the local public key of the session being linked to. The group ID is the local pub
    key of the very first session in the group of linked sessions. 
4. Update `litrpc`s `Session` and `AddAutopilotSessionRequest` messages to include 
    a `previous linked session key`. 
5. We need a mapping from session ID to group ID (one to one) and vice versa (one to 
    many) and so we add a new session-id-index to the firewall db. 
6. We then make sure to always write an entry to the new index when creating a session. 
7. Then we backfill all previous sessions so that there is an entry in the session-id-index for
    all sessions created prior to this PR. This is easy since currently the mapping is 
    one-to-one _both ways_. Only after this PR will the group-to-session index be one-to-many.
8. add a new `ListGroupActions` method to the firewall DB that allows us to query the DB
    for all actions of a specific _group_. 
9. Let the firewall query group actions instead of session actions. 
10. Allow ListActions rpc call to specify groupID. 
11. Let privacy map conversion rpc call specify groupID. 
12. let firewall get the correct privacy mapper by group ID on interception of calls. 
13. Finally, allow the client to link sessions via the cli. 

TODO:

- [ ] add pagination to the new ListGroupActions method (might not need to do this in this PR...).


bitromortac (commented):
First round done, it's a very impressive PR :1st_place_medal:, nice commit structure! Will also do manual tests in the next round.

session_rpcserver.go
```diff
@@ -317,12 +319,21 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
...
-		uniquePermissions, caveats, nil, false,
+		uniquePermissions, caveats, nil, false, nil,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("error creating new session: %v", err)
 	}
 
+	// Ensure that the session ID to group ID index is updated with the
+	// new pair before storing the session.
+	err = s.cfg.sessionIDIndex.AddSessionAndGroupIDPair(
```
ViktorTigerstrom, line 330: Hmm what if this succeeds when we have multiple linked sessions, and we successfully store the `session_id` for the `group_id` fin the index db, and then `s.db.StoreSession(sess)` below fails.
We'll then have a stored `session_id` in the index for the group that no longer exists, and also the latest sequence number in the index will be pointing to a non existing session.
Currently, this isn't an issue as I see that `ListGroupActions` won't fail due to this and the `GetSessionIDs` function isn't used outside of tests. But I see a risk that we will expand the use of the `GetSessionIDs` function and that people will expect that the results always points to existing sessions, or that we add a `GetLatestSessionIdForGroup` function to the index db that gets the `session_id` for the highest `sequence`, which then might be a non existing session.

So if we want to keep it as is, I think we should at least add comments to the indexes and the `GetSessionIDs` that there's no guarantee that the stored `session_ids` in the index maps to existing stored sessions.

session_rpcserver.go
```diff
@@ -979,16 +1011,62 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+	var linkSig []byte
+	if prevSess != nil {
+		msg := sess.LocalPublicKey.SerializeCompressed()
+		linkSig = ecdsa.Sign(prevSess.LocalPrivateKey, msg).Serialize()
+	}
+
+	// Ensure that the session ID to group ID index is updated with the
+	// new pair before storing the session or writing any values to the
+	// privacy mapper.
+	err = s.cfg.sessionIDIndex.AddSessionAndGroupIDPair(
```
ViktorTigerstrom, line 1061: Same comment as above here, that this might successfully store this in the index, but then fail below :).

session_rpcserver.go
```diff
@@ -979,16 +1011,62 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+
+		prevSess, err = s.db.GetSession(prevPub)
+		if errors.Is(err, session.ErrSessionNotFound) {
+			return nil, fmt.Errorf("linked session(%x) not found",
+				req.PrevLocalPub)
+		} else if err != nil {
+			return nil, err
+		}
+
+		if prevSess.State != session.StateRevoked &&
```
ViktorTigerstrom, line 1032: Shouldn't we check here that there's no active session for whole group, and not just that the session for the `prevPub` isn't active? Else this will allow people to set the `PrevLocalPub` to an old session in the group that has already been replaced by a new session, and this will instead fail in on the `AutoPilot` server side. I think we should have checks on both the client + server side?

session_rpcserver.go
```diff
@@ -1219,13 +1227,17 @@ func (s *sessionRpcServer) marshalRPCSession(sess *session.Session) (
 			for feature, rules := range info.FeatureRules {
 				ruleMap := make(map[string]*litrpc.RuleValue)
 				for name, rule := range rules {
-					val, err := s.cfg.ruleMgrs.InitRuleValues(name, []byte(rule))
+					val, err := initRuleValues(
```
ViktorTigerstrom, line 1308: Just a question so I understand for my own future PRs, when is it ok to add these types of just code reformatting changes that isn't strictly related to the commit they're included to into a commit that's touching related code, and when should they be added to a separate commit? Is it mainly if it's in the same file and it's just 1-2 lines of code fixes, or where do we normally draw the line :)?

ViktorTigerstrom (commented):
First round of code review looks good :fire: :rocket:!! Really nice PR. 

Will also do more deep manual testing for next round :)!

bitromortac (commented):
Nice, this looks very close :rocket:. Tested out the migration manually, also tried to go back to an older code version, which was prevented as expected.

session_rpcserver.go
ellemouton: ok i've added option 2 in the commit before this one 🤓

ellemouton (commented):
thanks for the review so far @ViktorTigerstrom & @bitromortac !!

I've updated accordingly. 

I've added one new commit called `session: add session ID to Key index`

session_rpcserver.go
ViktorTigerstrom: Got it! Yeah agree that sqlite would make this so much easier! We should really migrate :)

session_rpcserver.go
ViktorTigerstrom: Awesome, thanks! Agree that this is the best option for now :)!
I guess the only case this can happen now is if a user tries to link a session to a group where all sessions are expired, but then doesn't link the latest session in the group, but a previous one. That will succeed here, but fail on the `AutoPilot` server side. Though I think that's a fair tradeoff to not add checks for that in LiT and instead just error on the `AutoPilot` server side.

firewalldb/session_id_index.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/lightninglabs/lightning-terminal/session"
+	"go.etcd.io/bbolt"
+)
+
+/*
+	The session ID index DB has the following structure:
+
+	session-id-index -> session-to-group -> <session id> -> <group id>
+			 -> group-to-session -> <group id>   -> sequence -> session id
```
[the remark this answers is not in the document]
ViktorTigerstrom: Thanks :)

session_rpcserver.go
ViktorTigerstrom: Ok thanks! Will keep this as the guideline going forward :)!

ViktorTigerstrom (commented):
New changes looks great! Still manually not done with second round of reviewing, but reviewing the changes first :)

Leaving one comment regarding an issue introduced with the latest push.

session_rpcserver.go
```diff
@@ -979,16 +1016,83 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Now, check that they are all revoked or expired.
+		for _, id := range prevSessionIDs {
+			prevSess, err := s.db.GetSessionByID(id)
+			if err != nil {
+				return nil, err
+			}
```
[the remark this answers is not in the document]
ellemouton: ah! yeah true 😂  good catch! will update tomorrow.

ellemouton:
gonna add something else shortly 🤓

session_rpcserver.go
ellemouton: Ok cool - this should no longer be an issue with the latest update :)

ellemouton:
Ok ya'll! thanks for your patience! The PR has been updated quite a bit. Here are the changes:

- a session ID to session local pub key index is added to the session store. 
- the session ID to group ID and group ID index that was originally in the firewall DB is now in the session store. This makes things a lot simpler and more idempotent. 
- an index from previous session ID to linking session ID is added to the session store. This makes it quick to check that no other session has linked to the one you are trying to link to.
- finally, to link a session from the CLI, you now specify the prev session ID (instead of the prev local pub)

session_rpcserver.go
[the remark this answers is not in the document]
ellemouton: think this is now fine with latest updates :)

ellemouton:
ok I tacked on a temporary commit here now labeled `temp:...` i'll squash it in in a nicer way tomorrow

session/store.go
```diff
@@ -13,22 +14,59 @@ var (
...
 // getSessionKey returns the key for a session.
 func getSessionKey(session *Session) []byte {
 	return session.LocalPublicKey.SerializeCompressed()
 }
 
-// StoreSession stores a session in the store. If a session with the
-// same local public key already exists, the existing record is updated/
-// overwritten instead.
-func (db *DB) StoreSession(session *Session) error {
+func (db *DB) CreateSession(session *Session) error {
```
ViktorTigerstrom, line 69: Really nice work finding this issue 🚀 🔥!!
Not sure if this is just cause this is a temporary commit and you intend to fix this soon, but I think the `Store` interface also needs to be updated with the `CreateSession` & `UpdateSessionRemotePubKey` functions, and the removal of the `StoreSession` function.
These two new functions also need to have some docs added :)!

ViktorTigerstrom (commented):
Really nice updates!! Thinks became so much cleaner once the indexes are stored in the same file as the sessions 🚀 🔥 🚀 !!

Unfortunately though, I realised an issue that exists that'll likely require some redesigning... Sorry to find this one out so late in the review cycle. I've explained the issue in my first comment of this review.

session/store.go
ellemouton: It's a temp commit meaning that it is just for functionality rn and that it will be cleaned up later as mentioned offline 👍

session_rpcserver.go
```diff
@@ -1015,7 +1075,7 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
 	// We only persist this session if we successfully retrieved the
 	// autopilot's static key.
 	sess.RemotePublicKey = remoteKey
-	if err := s.db.StoreSession(sess); err != nil {
+	if err := s.cfg.db.CreateSession(sess); err != nil {
```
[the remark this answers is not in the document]
ellemouton: updated to link via group ID

session_rpcserver.go
```diff
@@ -979,16 +998,57 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+		} else if err != nil {
+			return nil, err
+		}
+
+		// Now we need to check that the linked session is no longer
+		// active.
+		if prevSess.State != session.StateRevoked &&
+			prevSess.State != session.StateExpired {
+
+			return nil, fmt.Errorf("linked session(%x) "+
```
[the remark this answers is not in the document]
ellemouton: >  Perhaps we can allow revoking of sessions through the session_id in a later PR as well :

Yes was gonna add this to this PR but think it is better to just do it in a follow up to not expand the scope too much. Think we should provide a way to revoke by both session_ID and group ID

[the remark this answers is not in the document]
ellemouton: > (Image if the user has 100s of sessions linked)

Yeah but it will be the latest session in the group. So not too hard to grep for at the moment. Can add helpers for this later

ellemouton:
re-requesting since the change to only link by group ID. 

Requesting from @guggero  for architectural ACK

session/interface.go
```diff
@@ -141,4 +141,7 @@ type Store interface {
 	// to the session with the given local pub key.
 	UpdateSessionRemotePubKey(localPubKey,
 		remotePubKey *btcec.PublicKey) error
+
+	// GetSessionByID fetches the session with the given ID.
+	GetSessionByID(id ID) (*Session, error)
```
guggero, line 174: Since a session is only 4 bytes, there could be collisions... Do we need to return multiple sessions and then do the final comparison on the actual unique identifier which is the local key?

session/migration1/id_to_key_index.go
```diff
@@ -0,0 +1,71 @@
...
+		if err != nil {
+			return err
+		}
+
+		idBkt, err := idIndexBkt.CreateBucket(session.ID[:])
+		if err != nil {
+			return err
+		}
+
+		return idBkt.Put(sessionKeyKey[:], key)
```
guggero, line 69: So this won't work if there are multiple sessions with the same ID (because of collisions). Or do we make sure we don't generate colliding session IDs somewhere else?

session_rpcserver.go
```diff
@@ -997,16 +998,78 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+	// session's public key in order to prove to the Autopilot server that
+	// the two session's belong to the same owner.
+	var (
+		linkSig        []byte
+		prevSessionPub *btcec.PublicKey
+	)
+	if linkedGroupID != nil {
+		privKey := linkedGroupSession.LocalPrivateKey
+		msg := sess.LocalPublicKey.SerializeCompressed()
+		linkSig = ecdsa.Sign(privKey, msg).Serialize()
```
guggero, line 1067: `Sign()` signs over a message hash, not the message itself. So we should hash it first.

session/migration1/id_to_key_index.go
ellemouton: think this is addressed now with the latest push

session/interface.go
ellemouton: addressed with latest push

session/migration1/interface.go
```diff
@@ -0,0 +1,121 @@
...
+	privateKey, err := btcec.NewPrivateKey()
+	if err != nil {
+		return nil, fmt.Errorf("error deriving private key: %v", err)
+	}
+	pubKey := privateKey.PubKey()
+
+	// NOTE: after this migration, no new sessions use the first 4 bytes of
+	// the serialised pub key for the macaroon root key. After this
+	// migration, bytes 1-4 are used (instead of 0-3) this is so that we
+	// get the full 4 byte entropy instead of waisting one byte on the 0x02
```
ViktorTigerstrom, line 86: nit:
```suggestion
	// get the full 4 byte entropy instead of wasting one byte on the 0x02
```

session/migration1/id_to_key_index.go
```diff
@@ -0,0 +1,150 @@
...
+
+		// Sort the sessions from oldest to newest.
+		sort.Slice(sessions, func(i, j int) bool {
+			return sessions[i].createdAt.Before(
+				sessions[j].createdAt,
+			)
+		})
+
+		// For each session other than the newest one, we ensure that
+		// the session is revoked.
```
ViktorTigerstrom, line 117: Should we add a comment explaining why we need to revoke the sessions which aren't the newest one, ie. due to the potential of collisions previously.

session_rpcserver.go
```diff
@@ -315,15 +309,31 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
...
 	)
 	if err != nil {
+		cancelIDRes()
+
 		return nil, fmt.Errorf("error creating new session: %v", err)
 	}
 
-	if err := s.db.StoreSession(sess); err != nil {
+	if err := s.cfg.db.CreateSession(sess); err != nil {
 		return nil, fmt.Errorf("error storing session: %v", err)
```
ViktorTigerstrom, line 337: Should we also add `cancelIDRes()` here? As the `db.Update` that deletes the reservation in `CreateSession` won't be committed if `CreateSession` errors.

session_rpcserver.go
```diff
@@ -979,43 +1015,130 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+
 		return nil, fmt.Errorf("error registering session with "+
 			"autopilot server: %v", err)
 	}
 
 	// We only persist this session if we successfully retrieved the
 	// autopilot's static key.
 	sess.RemotePublicKey = remoteKey
-	if err := s.db.StoreSession(sess); err != nil {
+	if err := s.cfg.db.CreateSession(sess); err != nil {
```
ViktorTigerstrom, line 1141: Same here, should we add `cancelIDRes()`?

ViktorTigerstrom (commented):
Amazing work :rocket: :fire: 🚀 !!!!! LGTM, I'd just like to do some final manual testing before lunch tomorrow, before I hit the approve button 🚀!

session_rpcserver.go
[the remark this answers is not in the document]
ViktorTigerstrom: > Think we should provide a way to revoke by both session_ID and group ID

Agree, let's leave this for a follow-up :)!

session/migration1/id_to_key_index.go
ellemouton: good idea - adding!

session_rpcserver.go
ellemouton: indeed! good catch. I think im just doing to change it to a `defer` instead so that we never forget to call it if we exit before creating the session. If the session does get created successfully then the `defer cancelIDRes` will be a no-op :+1:

session_rpcserver.go
ellemouton: updated!

session_rpcserver.go
ellemouton: fixed! thanks for this catch

session/macaroon.go
```diff
@@ -127,3 +128,26 @@ func RootKeyIDFromMacaroon(mac *macaroon.Macaroon) (uint64, error) {
...
+	}
+
+	pubKey := privateKey.PubKey()
+
+	// NOTE: we use bytes [1:4] of the serialised public key to create the
+	// macaroon root key base along with the Session ID. This will provide
+	// 4 bytes of entropy. Previously, bytes [0:3] where used but this
+	// resulted in lower entropy due to the first byte always being either
+	// 0x02 or 0x03.
+	copy(id[:], pubKey.SerializeCompressed()[1:4])
```
guggero, line 150: I'm pretty sure this needs to be `pubKey.SerializeCompressed()[1:5])`, otherwise we only get 3 bytes. The second index is _exclusive_. So the comment should be updated to read `Previously, bytes [0:4] were used` as well.

Maybe we should add a debug log here to show the new generated ID, just as a visual check when testing.

session/migration1/id_to_key_index.go
```diff
@@ -0,0 +1,155 @@
...
+
+		session, err := DeserializeSession(
+			bytes.NewReader(sessionBytes),
+		)
+		if err != nil {
+			return err
+		}
+
+		var id ID
+		copy(id[:], key[0:4])
```
guggero, line 74: This is correct, re my comment above.

session_rpcserver.go
```diff
@@ -315,9 +315,26 @@ func (s *sessionRpcServer) AddSession(_ context.Context,
 		}
 	}
 
+	// Reserve Session Key & ID.
+	id, localPrivKey, err := s.db.ReserveNewSessionID()
```
guggero: We could just hold a session mutex in the session RPC server to avoid needing another reserve construct. That way only one session could be created at a time.

guggero (commented):
Looks very good! Thanks for the quick update.

There's just a small thing with the length of the session ID, other than that LGTM :tada:

ViktorTigerstrom (approved):
tACK LGTM after addressing Oli's comment regarding the last byte of the session id not getting assigned 🚀:fire::rocket:!! Great work!! Can confirm that the last byte of the session id always gets set to `00` currently which i noticed during my testing :).

guggero (approved):
aACK, things around the session ID collision look good, very nice work :100:

bitromortac (approved):
tACK :rocket::rocket: strong work :muscle:!

PR #626, opened 2023-08-16
autopilot: feature configuration

bitromortac:
In this PR we add support for configuration of features within autopilot. Feature configuration can signal preferences to  autopilot which are not enforceable by the `litd` firewall.

* Default configurations for each feature are available via the `ListFeatures` call
* Support for feature configuration is added to `litcli`
* Pubkeys in configuration strings are obfuscated


firewall/privacy_mapper.go
```diff
@@ -833,3 +835,41 @@ func CryptoRandIntn(n int) (int, error) {
 
 	return int(nBig.Int64()), nil
 }
+
+// ObfuscateConfig alters the config string by replacing pubkeys with random
+// values and updates the privacy pair map.
+func ObfuscateConfig(privacyPairs map[string]string, configB []byte) ([]byte,
```
ellemouton, line 841: very cool! should we also match on channel IDs/outpoints?

cmd/litcli/autopilot.go
```diff
@@ -65,6 +66,12 @@ var addAutopilotSessionCmd = cli.Command{
 				"perform actions on. In the " +
 				"form of: peerID1,peerID2,...",
 		},
+		cli.StringSliceFlag{
+			Name: "configuration",
+			Usage: `per feature JSON-serialized configuration, ` +
+				`expected format: {"version":0,"option1":` +
+				`"parameter1","option2":"parameter2",...}`,
```
ellemouton, line 73: the `configuration` flag will apply to all features though right? so should the config values not be indexed by feature name? else we send the same config to all features we subscribe to

ellemouton (commented):
Nice nice niiiiice 🔥

firewall/privacy_mapper.go
bitromortac: yes, will add that :+1:!

session_rpcserver.go
```diff
@@ -984,6 +984,19 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+
+		// The configuration may contain sensitive data like pubkeys,
+		// which we replace here and add to to the privacy map.
+		if privacy {
+			configB, err = firewall.ObfuscateConfig(
+				privacyMapPairs, configB,
+			)
+			if err != nil {
+				return nil, err
+			}
```
ellemouton, line 999: this means that `featureConfig` given to `NewSession` and thus persisted, will be the obfuscated one right? Dont think we want to do that.

bitromortac:
Replaced by https://github.com/lightninglabs/lightning-terminal/pull/629 (PR from upstream branch)

cmd/litcli/autopilot.go
```diff
@@ -216,11 +223,27 @@ func initAutopilotSession(ctx *cli.Context) error {
...
+			"must match")
+	}
+
 	featureMap := make(map[string]*litrpc.FeatureConfig)
-	for _, feature := range ctx.StringSlice("feature") {
+	for i, feature := range ctx.StringSlice("feature") {
+		var config []byte
+
+		// We allow empty configs, to signal the usage of the default
+		// configuration when the session is registered.
```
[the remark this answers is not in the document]
bitromortac: added all the suggestions, hope we have a better description now :). is it a breaking change? we can leave out all config flags and it will use the defaults

session_rpcserver.go
bitromortac: very good point :+1: ! the new version persists the unobfuscated data

cmd/litcli/autopilot.go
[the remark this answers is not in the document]
ellemouton: ah - you are right. Sorry about that 🙈

PR #629, opened 2023-08-21
autopilot: feature configuration

bitromortac:
In this PR we add support for configuration of features within autopilot. Feature configuration can signal preferences to autopilot which are not enforceable by the litd firewall.

* Default configurations for each feature are available via the ListFeatures call
* Support for feature configuration is added to litcli
* Pubkeys, channel ids, and channel points in configuration strings are obfuscated


cmd/litcli/autopilot.go
```diff
@@ -65,6 +66,17 @@ var addAutopilotSessionCmd = cli.Command{
...
 		},
+		cli.StringSliceFlag{
+			Name: "configuration",
+			Usage: `If set, this flag must be repeated for each ` +
+				`feature (corresponding to the feature ` +
+				`order) and contains a JSON-serialized ` +
+				`configuration with the expected format: ` +
+				`{"version":0,"option1":"parameter1",` +
+				`"option2":"parameter2",...}. An empty ` +
+				`configuration is allowed with {} to set the ` +
```
ellemouton, line 77: nit: s/to set/to use the

ellemouton (changes requested):

cmd/litcli/autopilot.go
```diff
@@ -65,6 +66,17 @@ var addAutopilotSessionCmd = cli.Command{
 				"perform actions on. In the " +
 				"form of: peerID1,peerID2,...",
 		},
+		cli.StringSliceFlag{
+			Name: "configuration",
+			Usage: `If set, this flag must be repeated for each ` +
+				`feature (corresponding to the feature ` +
```
ViktorTigerstrom, line 72: Hmm I find this a little strange, and think this will be a bit unintuitive for the user. I think it will be hard for users to understand that they should: 
1. Repeat the configuration flag for every feature in the order they want to specify the config for.
2. If they have set a config for ANY feature, they must set the config to `{}` for every other feature they don't want to configure.

In my opinion I think it would make more sense to the user to specify this as a json object IF they have more than one feature added to the `autopilot a` cmd.

If they have multiple features in the `autopilot a` cmd, they have to specify which feature a config is for as a json header:
Ie. if they feature_1 & feature_2 in the request it would be:

`configuration={"feature_1":{"version":0,"option1":"parameter1","option2":"parameter2",...}, "feature_2":{"version":0,"option1":"parameter1","option2":"parameter2",...}}`

Or if they wouldn't want to specify any configuration for feature_2 they'd just leave that one out from the json_object:
`configuration={"feature_1":{"version":0,"option1":"parameter1","option2":"parameter2",...}}`

Optionally, if they have just 1 feature in the command, we could make it ok for them to set:
configuration={"version":0,"option1":"parameter1",`"option2":"parameter2",...}
Though this might add some extra confusion but be more convenient for users. Leave this one out if you think it's too confusing.

Given that the `channel-restrict-list` & the `peer-restrict-list` is applied to every feature of the `autopilot a` cmd, I think users will expect the `configuration` to be as well.

This isn't blocking though, and if you both think the implemented version is better, feel free to ignore this :).
If you choose to ignore this, I think the first words in this description should be changed from:
"If set" to "If set for any feature"

cmd/litcli/autopilot.go
```diff
@@ -216,11 +228,27 @@ func initAutopilotSession(ctx *cli.Context) error {
...
 	featureMap := make(map[string]*litrpc.FeatureConfig)
-	for _, feature := range ctx.StringSlice("feature") {
+	for i, feature := range ctx.StringSlice("feature") {
+		var config []byte
+
+		// We allow empty configs, to signal the usage of the default
+		// configuration when the session is registered.
+		if len(configs) > 0 && configs[i] != "{}" {
+			config = []byte(configs[i])
+		}
```
ViktorTigerstrom, line 247: nit: Maybe we should also validate here that this is configs[i] is a valid json? After you update this PR to adhere to Elle's firewall storing comment https://github.com/lightninglabs/lightning-terminal/pull/629#discussion_r1300196618, we'll fail there before sending the config to the `AutoPilot server` side, but we'll still execute quite a bit of code and call `ListFeatures` for the `Autopilot server` before getting there. So it might be good to fail early here for something that'll definitely fail.

This isn't blocking either though, so feel free to ignore if you don't think it's worth it.

ViktorTigerstrom (commented):
First review looks good after Elle's comment has been addressed 🚀! Really nice work :)!!

Leaving two suggestions for improvements that are non blocking. I'd especially want to hear you're opinion on how the configuration should be specified in the `autopilot a` cmd.

cmd/litcli/autopilot.go
ellemouton: just want to leave one note here that we really dont expect anyone but power users to use LiT via Lncli. The vast vast vast majority of users are expected to use LiT via the UI

bitromortac: yes, I agree with @ellemouton , that this is more for power users, which is why I kept the repeated config flags. I think it's also more difficult to use with every layer of json nesting that's added, and we would add redundancy with respect to the feature keys (if we would be consequential, we would also need to remove the `--feature` flags). However, I see the need to change the channel and peer restrict lists to repeated flags as they may need to be interpreted per feature, which is a good point :+1:!

cmd/litcli/autopilot.go
bitromortac: done! agree it's better to check things rather early

cmd/litcli/autopilot.go
```diff
@@ -51,20 +56,29 @@ var addAutopilotSessionCmd = cli.Command{
...
+			Name: "feature-channel-restrict-list",
 			Usage: "list of channel IDs that the " +
 				"Autopilot server should not " +
 				"perform actions on. In the " +
 				"form of: chanID1,chanID2,...",
 		},
-		cli.StringFlag{
-			Name: "peer-restrict-list",
+		cli.StringSliceFlag{
+			Name: "feature-peer-restrict-list",
```
bitromortac, line 67: note that there's an issue with this flag and the privacy mapper, if repeated with the same pubkey, the mapping fails. will fix this in the next round

ellemouton:
requesting @guggero too for architecture ack

guggero (approved):
aACK, very nice!

ViktorTigerstrom (approved):
LGTM after fixing the issue mentioned in https://github.com/lightninglabs/lightning-terminal/pull/629/#discussion_r1307249951 🔥:rocket:, great work!!

cmd/litcli/autopilot.go
ViktorTigerstrom: Ok makes sense!

> I see the need to change the channel and peer restrict lists to repeated flags as they may need to be interpreted per feature, which is a good point 

Thanks for fixing this with the latest push :)

ellemouton (approved):
LGTM! 🔥  great work

cmd/litcli/autopilot.go
bitromortac: this is going to be fixed in https://github.com/lightninglabs/lightning-terminal/pull/634

cmd/litcli/autopilot.go
```diff
@@ -51,15 +55,15 @@ var addAutopilotSessionCmd = cli.Command{
 			Name:     "feature",
 			Required: true,
 		},
-		cli.StringFlag{
-			Name: "channel-restrict-list",
+		cli.StringSliceFlag{
+			Name: "feature-channel-restrict-list",
```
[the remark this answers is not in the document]
bitromortac: cool, will keep it in mind :+1:

cmd/litcli/autopilot.go
```diff
@@ -39,8 +40,11 @@ var addAutopilotSessionCmd = cli.Command{
 	ShortName: "a",
 	Usage:     "Initialize an Autopilot session.",
 	Description: `
-	Initialize an Autopilot session.	
-	`,
+	Initialize an Autopilot session.
+
+	If set for any feature, the channel and peer restrict lists need to be
+	repeated for each feature that is registered, corresponding to the order
+	of features.`,
```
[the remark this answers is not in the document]
bitromortac: thank you :), have opened another PR to limit the scope here: https://github.com/lightninglabs/lightning-terminal/pull/634

bitromortac:
Rebased on https://github.com/lightninglabs/lightning-terminal/pull/621

bitromortac:
Rebased on https://github.com/lightninglabs/lightning-terminal/pull/637

firewall/privacy_mapper_test.go
```diff
@@ -718,6 +719,196 @@ func TestHideBool(t *testing.T) {
...
+			config: []byte(`{"version":1,"channels":` +
+				`[12345,` +
+				`"e092708c9e737115ff14a85ab65466561280d77c1b8cd666bc655536ad81ccca:1"]}`),
+			expectErr:        true,
+			expectedNewPairs: 0,
+		},
+		{
+			// A list of numbers is not obfuscated. Those can be
+			// useful to submit histograms for example.
+			name: "channel ids",
```
ellemouton, line 823: we would want to obfuscate channel IDs though. Even though they are numbers. This makes me think we should allow the user to specify specify somehow that a list/item in a config struct should be obfuscated. We need to think about how to do this though... We can do this in future though :) we just need to do it before ever asking for channel IDs in the config

firewall/privacy_mapper.go
```diff
@@ -845,3 +858,140 @@ func CryptoRandIntn(n int) (int, error) {
...
+
+		// We only substitute items in lists of strings.
+		stringList := make([]string, len(list))
+		anyString := false
+		allStrings := true
+		for i, item := range list {
+			item, ok := item.(string)
+			allStrings = allStrings && ok
+			anyString = anyString || ok
+			stringList[i] = item
```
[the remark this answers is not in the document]
ellemouton: ah ok it doesnt panic since `item` at this point is empty. But maybe just for clarity, let's do:
```
if !ok {
    continue
}

stringList[i] = item
```

ellemouton (approved):
LGTM!

firewall/privacy_mapper_test.go
bitromortac: yes, agree that we need to think more about these cases and that such a signal could make sense, for now the signal for obfuscation is to have a list of strings, which includes the channel id case as well, just the numbers converted to strings

firewall/privacy_mapper.go
```diff
@@ -845,3 +858,140 @@ func CryptoRandIntn(n int) (int, error) {
...
+	err := json.Unmarshal(configB, &configMap)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	privMapPairs := make(map[string]string)
+	newConfigMap := make(map[string]any)
+	for k, v := range configMap {
+		// We only substitute items in lists.
+		list, ok := v.([]any)
```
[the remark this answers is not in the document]
bitromortac: I tried that, but it didn't work, see https://go.dev/play/p/iA-IUn5VHR4

firewall/privacy_mapper.go
```diff
@@ -845,3 +858,140 @@ func CryptoRandIntn(n int) (int, error) {
...
+	var configMap map[string]any
+	err := json.Unmarshal(configB, &configMap)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	privMapPairs := make(map[string]string)
+	newConfigMap := make(map[string]any)
+	for k, v := range configMap {
+		// We only substitute items in lists.
```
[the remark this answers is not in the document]
bitromortac: so far I didn't have any better applications than for lists of items, similar to the peer deny list and channel deny lists and didn't want to be too restrictive. a fallback for a single field could be to use a list with a single item (hacky)

ViktorTigerstrom (approved):
tACK LGTM!!! 🔥:rocket:🔥

Awesome 🥳!!!

PR #633, opened 2023-08-24
build: update to lnd v0.17.0 and latest version of Pool

Roasbeef:
We point to the rc for 0.17.0 for now. Also we had to update Pool as well, due to a replace directive inserted into lnd.


ellemouton (changes requested):
The README needs to be updated. One of the CI checks is also failing. I also think we will wait till the final 0.17.0-beta is released before merging this

guggero:
Replaced by https://github.com/lightninglabs/lightning-terminal/pull/645.

PR #634, opened 2023-08-29
autopilot: generic rules per feature

bitromortac:
We would like to be able to set different rules for different features. This is solved by introducing JSON-serialized (per-feature) rule flags where all the rules from `litcli autopilot features` can be set.

### DIFF
--- a/cmd/litcli/autopilot.go
+++ b/cmd/litcli/autopilot.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/lightninglabs/lightning-terminal/litrpc"
 	"github.com/lightninglabs/lightning-terminal/rules"
+	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/urfave/cli"
 )
 
@@ -39,10 +40,48 @@ var addAutopilotSessionCmd = cli.Command{
 	Name:      "add",
 	ShortName: "a",
 	Usage:     "Initialize an Autopilot session.",
-	Description: "Initialize an Autopilot session.\n\n" +
-		"   If set for any feature, configuration flags need to be " +
-		"repeated for each feature that is registered, corresponding " +
-		"to the order of features.",
+	Description: `
+	Initialize an Autopilot session.
+
+	If one of the 'feature-' flags is set for any 'feature', then that flag
+	must be provided for each 'feature'.
+
+	The rules and configuration options available for each feature can be
+	seen in the 'autopilot features' output. For a rule, all fields must be
+	set since the unset ones are interpreteded as zero values. Rule values
+	must adhere to the limits found in 'autopilot features'. If a rule is
+	not set, default values are used.
+	
+	An example call for AutoFees reads:
+	
+	#!/bin/bash
+	./litcli autopilot add --label=customRules \
+	--feature=AutoFees \
+	--feature-rules='{
+		"rules": {
+			"channel-policy-bounds":  {
+				"chan_policy_bounds":  {
+					"min_base_msat":  "0",
+					"max_base_msat":  "10000",
+					"min_rate_ppm":   10,
+					"max_rate_ppm":   5000,
+					"min_cltv_delta": 60,
+					"max_cltv_delta": 120,
+					"min_htlc_msat":  "1",
+					"max_htlc_msat":  "100000000000"
+				}
+			},
+			"peer-restriction":  {
+				"peer_restrict":  {
+					"peer_ids":  [
+						"abcabc", 
+						"defdef"
+					]
+				}
+			}
+		}
+	}' \
+	--feature-config='{}'`,
 	Action: initAutopilotSession,
 	Flags: []cli.Flag{
 		labelFlag,
@@ -53,19 +92,21 @@ var addAutopilotSessionCmd = cli.Command{
 			Name:     "feature",
 			Required: true,
 		},
-		cli.StringFlag{
+		cli.StringSliceFlag{
 			Name: "channel-restrict-list",
-			Usage: "List of channel IDs that the " +
+			Usage: "[deprecated] List of channel IDs that the " +
 				"Autopilot server should not " +
 				"perform actions on. In the " +
 				"form of: chanID1,chanID2,...",
+			Hidden: true,
 		},
-		cli.StringFlag{
+		cli.StringSliceFlag{
 			Name: "peer-restrict-list",
-			Usage: "List of peer IDs that the " +
+			Usage: "[deprecated] List of peer IDs that the " +
 				"Autopilot server should not " +
 				"perform actions on. In the " +
 				"form of: peerID1,peerID2,...",
+			Hidden: true,
 		},
 		cli.StringFlag{
 			Name: "group_id",
@@ -80,6 +121,13 @@ var addAutopilotSessionCmd = cli.Command{
 				"\"option2\":\"parameter2\",...}. An empty " +
 				"configuration is allowed with {} to use the " +
 				"default configuration.",
+		},
+		cli.StringSliceFlag{
+			Name: "feature-rules",
+			Usage: `JSON-serialized rule map (see main ` +
+				`description for a format example).` +
+				`An empty rule map is allowed with {} to ` +
+				`use the default rules.`,
 		},
 	},
 }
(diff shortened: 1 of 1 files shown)


cmd/litcli/autopilot.go
```diff
@@ -51,25 +54,36 @@ var addAutopilotSessionCmd = cli.Command{
...
-			Usage: "list of channel IDs that the " +
-				"Autopilot server should not " +
-				"perform actions on. In the " +
-				"form of: chanID1,chanID2,...",
+		cli.StringSliceFlag{
+			Name: "feature-channel-restrict-list",
+			Usage: `List of channel IDs that the ` +
+				`Autopilot server should not ` +
+				`perform actions on. In the ` +
+				`form of: [chanID1,chanID2,...]. ` +
```
ViktorTigerstrom, line 62: nit: 
Should be as below, correct?

```suggestion
				`form of: ["chanID1","chanID2",...]. ` +
```

ViktorTigerstrom (commented):
First code review looks good :rocket:fire:! Awaiting rebase before testing.

> how do we signal empty restrictions for more than a single feature? currently we have to repeat the restrictions as often as the number of features, which requires to set several flags (a flag --feature-peer-restrict-list="" does not count)

Given that we opted for this version with feature configurations, I think it makes sense to do it like you've implemented it in this PR.

> could we transition to json-style lists like "[item1,item1]" to signal empty restrictions with "[]", similarly to the feature configuration?

No strong opinion here, as long as we keep it consistent. 

> We change the flags to have a `feature` prefix, which is a breaking change.

Just a thought, would it make sense to keep the old flags as well for now, i.e. "channel-restrict-list" & "peer-restrict-list" but mark them as deprecated and remove them later to not make this a breaking change with this PR? And if the old flags are set, we keep the old behaviour, but also make sure that only one version of the flags are set and error if both versions are set. Do you think that would make sense?

cmd/litcli/autopilot.go
bitromortac: ah that was intentional, we would error to parse into []int64 otherwise

bitromortac:
Rebased the PR.

> Just a thought, would it make sense to keep the old flags as well for now, i.e. "channel-restrict-list" & "peer-restrict-list" but mark them as deprecated and remove them later to not make this a breaking change with this PR? And if the old flags are set, we keep the old behaviour, but also make sure that only one version of the flags are set and error if both versions are set. Do you think that would make sense?

Am with you there, will check if it's worth to keep them backwards compatible.

I also found a small issue which is preexisting: `ListChannels error: rpc error: code = Unknown desc = error parsing rules: invalid channel ID`, which is related to the channel restrictions. It happens when an unknown channel id restriction, say channel id of `123456` is passed. Will also try to provide a fix.

ellemouton:
>  Will also try to provide a fix.

@bitromortac  - is this ready for review?

cmd/litcli/autopilot.go
```diff
@@ -43,8 +41,9 @@ var addAutopilotSessionCmd = cli.Command{
 	Description: `
 	Initialize an Autopilot session.
 
-	If set for any feature, configuration flags need to be repeated for each
-	feature that is registered, corresponding to the order of features.`,
+	If set for any feature, configuration flags and channel and peer
+	restrict lists need to be repeated (empty if necessary) for each feature
```
ellemouton, line 45: how about something like:

```
If one of the `feature-` flags is set for any `feature`, then that flag must be provided for each `feature`
```

cmd/litcli/autopilot.go
ellemouton: while we are here, let's clear up this confusion by specifying that this should be a list of numbers. Since even in LND, there is a `ChannelID` which is a uint64 and another which is a string ([32]byte)

cmd/litcli/autopilot.go
```diff
@@ -55,19 +54,21 @@ var addAutopilotSessionCmd = cli.Command{
...
 			Required: true,
 		},
-		cli.StringFlag{
-			Name: "channel-restrict-list",
-			Usage: "list of channel IDs that the " +
-				"Autopilot server should not " +
-				"perform actions on. In the " +
-				"form of: chanID1,chanID2,...",
+		cli.StringSliceFlag{
+			Name: "feature-channel-restrict-list",
```
ellemouton, line 58: technically this rename is a backwards incompatible change... imo it is fine since I dont think anyone is using this at the command line level yet. But perhaps something we should at least include in the release notes? (btw, I think we should have a running issue per release where we keep track of notes we want to make in the release so that we dont forget cc @ViktorTigerstrom )

ellemouton: can we move these to be next to `feature-config` just so that all the per-feature flags are together?

cmd/litcli/autopilot.go
```diff
@@ -196,45 +197,77 @@ func initAutopilotSession(ctx *cli.Context) error {
...
+					},
+				},
+			}
 	}
 
-	peerRestrictList := ctx.String("peer-restrict-list")
-	if peerRestrictList != "" {
-		peerIDs := strings.Split(peerRestrictList, ",")
+	peerRestrictLists := ctx.StringSlice("feature-peer-restrict-list")
+	if len(peerRestrictLists) > 0 && len(features) != len(peerRestrictLists) {
```
ellemouton, line 242: nit: lll

cmd/litcli/autopilot.go
```diff
@@ -196,45 +197,77 @@ func initAutopilotSession(ctx *cli.Context) error {
...
-				},
-			},
+		if len(chanIDs) == 0 {
+			continue
 		}
+
+		perFeatureRules[i].Rules[rules.ChannelRestrictName] =
+			&litrpc.RuleValue{
+				Value: &litrpc.RuleValue_ChannelRestrict{
+					ChannelRestrict: &litrpc.ChannelRestrict{
```
ellemouton, line 234: nit: lll

cmd/litcli/autopilot.go
ellemouton: ah - I saw @ViktorTigerstrom's suggestion now re keeping the old flags for now. 

We could do that and then add a waring and a todo that they will be removed in the release after the next one. That's definitely the more friendly way of doing things :)

ellemouton (commented):
Looks good! will test on next round 👍

bitromortac:
I changed this PR to include generic feature rule setting via JSON flags. Legacy flags are preserved and marked for deprecation.

cmd/litcli/autopilot.go
bitromortac: cool, I changed it to remain backwards compatible

cmd/litcli/autopilot.go
bitromortac: a disadvantage of the generic flags is that it's not as well documented

cmd/litcli/autopilot.go
```diff
@@ -39,10 +40,48 @@ var addAutopilotSessionCmd = cli.Command{
...
+	If one of the 'feature-' flags is set for any 'feature', then that flag
+	must be provided for each 'feature'.
+
+	The rules and configuration options available for each feature can be
+	seen in the 'autopilot features' output.
+	
+	An example call reads:
+	
+	./litcli autopilot add --label=customRules \
+	--feature=SampleFeature \
```
ViktorTigerstrom, line 55: nit:
```suggestion
	--feature=AutoFees \
```

Probably makes more sense to mention Autofees instead.

cmd/litcli/autopilot.go
```diff
@@ -39,10 +40,48 @@ var addAutopilotSessionCmd = cli.Command{
...
+			},
+			"channel-restriction": {
+				"channel_restrict": {
+					"channel_ids": [
+						18283782789, 
+						22891928322, 
+						31878293727
+					]
+				}
+			}
```
ViktorTigerstrom, line 78: nit:
Since this enables users to set any of the rules now, I do think that:

1.  It makes sense to give an example for all types rules the user can set, i.e. also include examples of the `channel-policy-bounds`, `history-limit` & `peer-restriction` rule types here.
2. Clarify in the docs that if a specific rule type is not set, the default values will be used.

ViktorTigerstrom (approved):
tACK LGTM :fire::rocket:!!

Great work!! Thanks for addressing the feedback about adding deprecating support :rocket:!

Just adding some minor feedback in my comments below that's not blocking.

Also, I'd just like to show two examples of adding an autopilot session that intuitively probably shouldn't be accepted by the parsing, but that is. We probably don't really need to address these examples though, but feel free to address them if you want:

1. When using the legacy version of setting the `peer-restrict-list` arg with multiple features, if you add multiple `peer-restrict-list` args where the last arg is set to "", that is accepted:
`litcli autopilot a --label="puppet" --feature=AutoFees --peer-restrict-list="peerID1" --feature=SampleFeature --peer-restrict-list=""`
2. If the same feature is specified more than once when adding a autpilot session, that's accepted and only the last version is used:
`litcli autopilot a --label="puppet" --feature=AutoFees --feature-rules="{\"rules\":{}}" --feature=AutoFees --feature-rules="{\"rules\":{\"channel-restriction\":{\"channel_restrict\":{\"channel_ids\":[chanID1]}}}}"`

cmd/litcli/autopilot.go
bitromortac: right, I updated to have an AutoFees example

cmd/litcli/autopilot.go
bitromortac: Good calls, concerning 1, I have added the most interesting rules for `AutoFees`, but left others away since I think it would add too much help output and it should be clear how to apply the pattern. I updated the help text a bit more to give more hints on how to correctly set rules. I've added that comment for 2.

bitromortac:
> Also, I'd just like to show two examples of adding an autopilot session that intuitively probably shouldn't be accepted by the parsing, but that is. We probably don't really need to address these examples though, but feel free to address them if you want:
> 
>     1. When using the legacy version of setting the `peer-restrict-list` arg with multiple features, if you add multiple `peer-restrict-list` args where the last arg is set to "", that is accepted:
>        `litcli autopilot a --label="test" --feature=AutoFees --peer-restrict-list="peerID1" --feature=SampleFeature --peer-restrict-list=""`
> 
>     2. If the same feature is specified more than once when adding a autpilot session, that's accepted and only the last version is used:
>        `litcli autopilot a --label="test" --feature=AutoFees --feature-rules="{\"rules\":{}}" --feature=AutoFees --feature-rules="{\"rules\":{\"channel-restriction\":{\"channel_restrict\":{\"channel_ids\":[chanID1]}}}}"`

Thanks for the review and testing :pray: @ViktorTigerstrom, those issues should be fixed in the latest version, great calls!

ellemouton (approved):
LGTM! thanks for this 🔥

ViktorTigerstrom:
> > Also, I'd just like to show two examples of adding an autopilot session that intuitively probably shouldn't be accepted by the parsing, but that is. We probably don't really need to address these examples though, but feel free to address them if you want:
> > ```
> > 1. When using the legacy version of setting the `peer-restrict-list` arg with multiple features, if you add multiple `peer-restrict-list` args where the last arg is set to "", that is accepted:
> >    `litcli autopilot a --label="test" --feature=AutoFees --peer-restrict-list="peerID1" --feature=SampleFeature --peer-restrict-list=""`
> > 
> > 2. If the same feature is specified more than once when adding a autpilot session, that's accepted and only the last version is used:
> >    `litcli autopilot a --label="test" --feature=AutoFees --feature-rules="{\"rules\":{}}" --feature=AutoFees --feature-rules="{\"rules\":{\"channel-restriction\":{\"channel_restrict\":{\"channel_ids\":[chanID1]}}}}"`
> > ```
> 
> Thanks for the review and testing 🙏 @ViktorTigerstrom, those issues should be fixed in the latest version, great calls!

Thanks for the fixes, they look great :pray:!

cmd/litcli/autopilot.go
ViktorTigerstrom: > but left others away since I think it would add too much help output and it should be clear how to apply the pattern

Ok thanks for the edits :)! Sure, I see your point, let's avoid adding too much output. In case we end up getting lot's of user questions of how to apply a specific rule in the future, we can always revisit this and add more info :)!

PR #639, opened 2023-09-04
itests: add session linking test

ellemouton:
In this PR, we add a session linking itests which aims to test behaviour across linked sessions. Particularly:
1) that calls through linked sessions make use of the same privacy mapper
2) that obfuscated rule values (like channel IDs) use the same privacy mapper across linked sessions
3) that the rule enforcer uses the same DB across linked sessions

### DIFF
--- a/itest/litd_firewall_test.go
+++ b/itest/litd_firewall_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strconv"
 	"strings"
 	"testing"
@@ -13,7 +14,6 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
-	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/lightninglabs/lightning-node-connect/mailbox"
 	"github.com/lightninglabs/lightning-terminal/autopilotserver/mock"
 	"github.com/lightninglabs/lightning-terminal/firewall"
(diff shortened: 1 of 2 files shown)


itest/litd_firewall_test.go
```diff
@@ -155,24 +161,492 @@ func testFWRateLimitAndPrivacyMapper(net *NetworkHarness, t *harnessTest) {
...
+		},
+	)
+	require.NoError(t.t, err)
+
+	// Also close the autopilot connection to the first session so that it
+	// doesn't continue to try and connect.
+	require.NoError(t.t, pilotConn.Close())
+	pilotConn = nil
+
+	sessResp2, err := litAutopilotClient.AddAutopilotSession(
```
ViktorTigerstrom, line 547: Would we clarify the linkage even further if we before we open this session:
1. Open a new channel.
2. Show that after opening the new session here, `getPseudo` get's "no such key found" for that ChanPoint, using the "original" `sessResp.Session.GroupId` (even though it's the same for all linked sessions).
3. Then at the end of the function after the `FeeReport` has been made with the new `lndConn`, an entry for that new chanPoint exists in the privacy map DB, and is found through `getPseudo` with the "original" `sessResp.Session.GroupId`.

Do you think that would add value?

ViktorTigerstrom (approved):
LGTM, really nice coverage :fire::rocket:!!

itest/litd_firewall_test.go
ellemouton: in not sure if this adds anything tbh - cause we already call `FeeReport` with the second connection and assert that it is receiving the same result as the first one. So this already proves that the same privacy map is being used for the two sessions

ViktorTigerstrom: Fair enough, good point, agree :)!

ellemouton:
ping @bitromortac  - would be nice to get this in for test coverage 🙏

itest/litd_firewall_test.go
```diff
@@ -177,6 +179,472 @@ func testFirewallRules(ctx context.Context, net *NetworkHarness,
...
+	defer pilotConn.Close()
+
+	lndConn = lnrpc.NewLightningClient(pilotConn)
+	caveatCreds1 = metaDataInjector.addCaveat(caveat1)
+
+	// List the channels and ensure that the same mapping was used as for
+	// the previous session.
+	feeReport3, err := lndConn.FeeReport(
+		ctx, &lnrpc.FeeReportRequest{}, caveatCreds1,
+	)
```
bitromortac, line 621: nit: needs error check

bitromortac (approved):
Great itest! LGTM :rocket:

ellemouton (commented):
thanks y'all! 🌵

Issue #638, opened 2023-09-02
LITD: Received critical error from subsystem, shutting down: payment isn't initiated


gStart9:
Hi, we encountered a user running into this issue today on v0.10.5:
```
2023-09-02T20:19:21+02:00  2023-09-02 18:19:21.804 [INF] LITD: LiT is running in mainnet, the taproot assets subserver do not support the `mainnet` network yet, disabling taproot assets subserver
2023-09-02T20:19:21+02:00  2023-09-02 18:19:21.805 [INF] LITD: LiT version: 0.10.5-alpha commit=v0.10.5-alpha
2023-09-02T20:19:21+02:00  2023-09-02 18:19:21.806 [INF] LITD: Listening for http_tls on: 127.0.0.1:8443
2023-09-02T20:19:21+02:00  2023-09-02 18:19:21.807 [INF] LITD: Listening for http on: 172.18.0.10:8443
2023-09-02T20:19:21+02:00  2023-09-02 18:19:21.859 [INF] SESS: Checking for schema update: latest_version=0, db_version=0
2023-09-02T20:19:21+02:00  2023-09-02 18:19:21.860 [INF] LITD: Dialing lnd gRPC server at lnd.embassy:10009
2023-09-02T20:19:21+02:00  ----------------------------------------------------------
2023-09-02T20:19:21+02:00   Lightning Terminal (LiT) by Lightning Labs
2023-09-02T20:19:21+02:00  
2023-09-02T20:19:21+02:00   LND Operating mode remote
2023-09-02T20:19:21+02:00   LND Node status online
2023-09-02T20:19:21+02:00   LND Alias xxxxxxxxxxxxxxxxxxxx
2023-09-02T20:19:21+02:00   LND Version 0.16.4-beta commit=v0.16.4-beta
2023-09-02T20:19:21+02:00   LiT Version 0.10.5-alpha commit=v0.10.5-alpha
2023-09-02T20:19:21+02:00   Web interface 127.0.0.1:8443, lightning-terminal.embassy:8443 (open https://127.0.0.1:8443 or http://lightning-terminal.embassy:8443 in your browser)
2023-09-02T20:19:21+02:00  ----------------------------------------------------------
2023-09-02T20:19:21+02:00  2023-09-02 18:19:21.905 [INF] LITD: Connecting basic lnd client
2023-09-02T20:19:22+02:00  2023-09-02 18:19:22.116 [INF] LITD: Connecting full lnd client
[...]

PR #640, opened 2023-09-05
accounts: don't error out on payment not initiated

guggero:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/638.

It can happen that we see the payment request for a payment coming in, register it, but then it is failed on the lnd level and basically never exists in the DB.
In that case we can just stop tracking the payment.

### DIFF
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -2,12 +2,14 @@ package accounts
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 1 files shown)


ellemouton (approved):
Looks good 👍

ViktorTigerstrom (approved):
LGTM as well :)!

PR #642, opened 2023-09-21
Don't shutdown lit on accounts service critical error, and register to status server

ViktorTigerstrom:
This PR is based on #541

In this PR we ensure that if the accounts service critically errors, we won't shutdown LiT, but instead log the error and reject any new requests to the accounts service.

We also add the accounts service to the status manager, and use the status from the status manager to determine if we should allow gRPC requests for the accounts service through or not.

Finally, we also add a config option to allow users to disable the accounts service, which ensures that the accounts service isn't started if the config option is set to `disable`.

TODO:
- [x] add tests

### DIFF
--- a/accounts/checkers.go
+++ b/accounts/checkers.go
@@ -522,6 +522,7 @@ func checkSend(ctx context.Context, chainParams *chaincfg.Params,
 	}
 
 	// The invoice is optional.
+	var paymentHash lntypes.Hash
 	if len(invoice) > 0 {
 		payReq, err := zpay32.Decode(invoice, chainParams)
 		if err != nil {
@@ -530,6 +531,10 @@ func checkSend(ctx context.Context, chainParams *chaincfg.Params,
 
 		if payReq.MilliSat != nil && *payReq.MilliSat > sendAmt {
 			sendAmt = *payReq.MilliSat
+		}
+
+		if payReq.PaymentHash != nil {
+			paymentHash = *payReq.PaymentHash
 		}
 	}
 
@@ -547,6 +552,14 @@ func checkSend(ctx context.Context, chainParams *chaincfg.Params,
 	err = service.CheckBalance(acct.ID, sendAmt)
 	if err != nil {
 		return fmt.Errorf("error validating account balance: %w", err)
+	}
+
+	emptyHash := lntypes.Hash{}
+	if paymentHash != emptyHash {
+		err = service.AssociatePayment(acct.ID, paymentHash, sendAmt)
+		if err != nil {
+			return fmt.Errorf("error associating payment: %w", err)
+		}
 	}
 
 	return nil
--- a/accounts/interceptor.go
+++ b/accounts/interceptor.go
@@ -52,6 +52,17 @@ func (s *InterceptorService) Intercept(ctx context.Context,
 	s.requestMtx.Lock()
 	defer s.requestMtx.Unlock()
 
+	// If the account service is not running, we reject all requests.
+	// Note that this is by no means a guarantee that the account service
+	// will be running throughout processing the request, but at least we
+	// can stop requests early if the service was already disabled when the
+	// request came in.
+	if !s.IsRunning() {
+		return mid.RPCErrString(
+			req, "the account service has been stopped",
+		)
+	}
+
 	mac := &macaroon.Macaroon{}
 	err := mac.UnmarshalBinary(req.RawMacaroon)
 	if err != nil {
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -160,6 +160,12 @@ var (
 	ErrNotSupportedWithAccounts = errors.New("this RPC call is not " +
 		"supported with restricted account macaroons")
 
+	// ErrAccountServiceDisabled is the error that is returned when the
+	// account service has been disabled due to an error being thrown
+	// in the service that cannot be recovered from.
+	ErrAccountServiceDisabled = errors.New("the account service has been " +
+		"stopped")
+
 	// MacaroonPermissions are the permissions required for an account
 	// macaroon.
 	MacaroonPermissions = []bakery.Op{{
@@ -240,4 +246,7 @@ type Service interface {
 	// longer needs to be tracked. The payment is certain to never succeed,
 	// so we never need to debit the amount from the account.
 	RemovePayment(hash lntypes.Hash) error
+
+	// IsRunning returns true if the service can be used.
+	IsRunning() bool
 }
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -247,6 +247,9 @@ type Service interface {
 	// so we never need to debit the amount from the account.
 	RemovePayment(hash lntypes.Hash) error
 
-	// IsRunning returns true if the service can be used.
-	IsRunning() bool
+	// AssociatePayment associates a payment (hash) with the given account,
+	// ensuring that the payment will be tracked for a user when LiT is
+	// restarted.
+	AssociatePayment(id AccountID, paymentHash lntypes.Hash,
+		fullAmt lnwire.MilliSatoshi) error
 }
--- a/accounts/rpcserver.go
+++ b/accounts/rpcserver.go
@@ -53,10 +53,6 @@ func (s *RPCServer) CreateAccount(ctx context.Context,
 	log.Infof("[createaccount] label=%v, balance=%d, expiration=%d",
 		req.Label, req.AccountBalance, req.ExpirationDate)
 
-	if !s.service.IsRunning() {
-		return nil, ErrAccountServiceDisabled
-	}
-
 	var (
 		balanceMsat    lnwire.MilliSatoshi
 		expirationDate time.Time
@@ -119,10 +115,6 @@ func (s *RPCServer) UpdateAccount(_ context.Context,
 	log.Infof("[updateaccount] id=%s, label=%v, balance=%d, expiration=%d",
 		req.Id, req.Label, req.AccountBalance, req.ExpirationDate)
 
-	if !s.service.IsRunning() {
-		return nil, ErrAccountServiceDisabled
-	}
-
 	accountID, err := s.findAccount(req.Id, req.Label)
 	if err != nil {
 		return nil, err
@@ -145,10 +137,6 @@ func (s *RPCServer) ListAccounts(context.Context,
 	*litrpc.ListAccountsRequest) (*litrpc.ListAccountsResponse, error) {
 
 	log.Info("[listaccounts]")
-
-	if !s.service.IsRunning() {
-		return nil, ErrAccountServiceDisabled
-	}
 
 	// Retrieve all accounts from the macaroon account store.
 	accts, err := s.service.Accounts()
@@ -175,10 +163,6 @@ func (s *RPCServer) AccountInfo(_ context.Context,
 
 	log.Infof("[accountinfo] id=%v, label=%v", req.Id, req.Label)
 
-	if !s.service.IsRunning() {
-		return nil, ErrAccountServiceDisabled
-	}
-
 	accountID, err := s.findAccount(req.Id, req.Label)
 	if err != nil {
 		return nil, err
@@ -198,10 +182,6 @@ func (s *RPCServer) RemoveAccount(_ context.Context,
 	error) {
 
 	log.Infof("[removeaccount] id=%v, label=%v", req.Id, req.Label)
-
-	if !s.service.IsRunning() {
-		return nil, ErrAccountServiceDisabled
-	}
 
 	accountID, err := s.findAccount(req.Id, req.Label)
 	if err != nil {
--- a/accounts/rpcserver.go
+++ b/accounts/rpcserver.go
@@ -53,6 +53,10 @@ func (s *RPCServer) CreateAccount(ctx context.Context,
 	log.Infof("[createaccount] label=%v, balance=%d, expiration=%d",
 		req.Label, req.AccountBalance, req.ExpirationDate)
 
+	if !s.service.IsRunning() {
+		return nil, ErrAccountServiceDisabled
+	}
+
 	var (
 		balanceMsat    lnwire.MilliSatoshi
 		expirationDate time.Time
@@ -115,6 +119,10 @@ func (s *RPCServer) UpdateAccount(_ context.Context,
 	log.Infof("[updateaccount] id=%s, label=%v, balance=%d, expiration=%d",
 		req.Id, req.Label, req.AccountBalance, req.ExpirationDate)
 
+	if !s.service.IsRunning() {
+		return nil, ErrAccountServiceDisabled
+	}
+
 	accountID, err := s.findAccount(req.Id, req.Label)
 	if err != nil {
 		return nil, err
@@ -137,6 +145,10 @@ func (s *RPCServer) ListAccounts(context.Context,
 	*litrpc.ListAccountsRequest) (*litrpc.ListAccountsResponse, error) {
 
 	log.Info("[listaccounts]")
+
+	if !s.service.IsRunning() {
+		return nil, ErrAccountServiceDisabled
+	}
 
 	// Retrieve all accounts from the macaroon account store.
 	accts, err := s.service.Accounts()
@@ -163,6 +175,10 @@ func (s *RPCServer) AccountInfo(_ context.Context,
 
 	log.Infof("[accountinfo] id=%v, label=%v", req.Id, req.Label)
 
+	if !s.service.IsRunning() {
+		return nil, ErrAccountServiceDisabled
+	}
+
 	accountID, err := s.findAccount(req.Id, req.Label)
 	if err != nil {
 		return nil, err
@@ -182,6 +198,10 @@ func (s *RPCServer) RemoveAccount(_ context.Context,
 	error) {
 
 	log.Infof("[removeaccount] id=%v, label=%v", req.Id, req.Label)
+
+	if !s.service.IsRunning() {
+		return nil, ErrAccountServiceDisabled
+	}
 
 	accountID, err := s.findAccount(req.Id, req.Label)
 	if err != nil {
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -16,6 +16,12 @@ import (
 	"github.com/lightningnetwork/lnd/lnwire"
 )
 
+// Config holds the configuration options for the accounts service.
+type Config struct {
+	// Disable will disable the accounts service if set.
+	Disable bool `long:"disable" description:"disable the accounts service"`
+}
+
 // trackedPayment is a struct that holds all information that identifies a
 // payment that we are tracking in the service.
 type trackedPayment struct {
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -416,6 +416,40 @@ func (s *InterceptorService) AssociateInvoice(id AccountID,
 	return s.store.UpdateAccount(account)
 }
 
+// AssociatePayment associates a payment (hash) with the given account,
+// ensuring that the payment will be tracked for a user when LiT is
+// restarted.
+func (s *InterceptorService) AssociatePayment(id AccountID,
+	paymentHash lntypes.Hash, fullAmt lnwire.MilliSatoshi) error {
+
+	s.Lock()
+	defer s.Unlock()
+
+	account, err := s.store.Account(id)
+	if err != nil {
+		return err
+	}
+
+	// If the payment is already associated with the account, we don't need
+	// to associate it again.
+	_, ok := account.Payments[paymentHash]
+	if ok {
+		return nil
+	}
+
+	// Associate the payment with the account and store it.
+	account.Payments[paymentHash] = &PaymentEntry{
+		Status:     lnrpc.Payment_UNKNOWN,
+		FullAmount: fullAmt,
+	}
+
+	if err := s.store.UpdateAccount(account); err != nil {
+		return fmt.Errorf("error updating account: %w", err)
+	}
+
+	return nil
+}
+
 // invoiceUpdate credits the account an invoice was registered with, in case the
 // invoice was settled.
 //
@@ -527,13 +561,33 @@ func (s *InterceptorService) TrackPayment(id AccountID, hash lntypes.Hash,
 		return nil
 	}
 
-	// Okay, we haven't tracked this payment before. So let's now associate
-	// the account with it.
 	account.Payments[hash] = &PaymentEntry{
 		Status:     lnrpc.Payment_UNKNOWN,
 		FullAmount: fullAmt,
 	}
+
 	if err := s.store.UpdateAccount(account); err != nil {
+		if !ok {
+			// In the rare case that the payment isn't associated
+			// with an account yet, and we fail to update the
+			// account we will not be tracking the payment, even if
+			// track the service is restarted. Therefore the node
+			// runner needs to manually check if the payment was
+			// made and debit the account if that's the case.
+			errStr := "critical error: failed to store the " +
+				"payment with hash %v for user with account " +
+				"id %v. Manual intervention required! " +
+				"Verify if the payment was executed, and " +
+				"manually update the user account balance by " +
+				"subtracting the payment amount if it was"
+
+			mainChanErr := s.disableAndErrorfUnsafe(
+				errStr, hash, id,
+			)
+
+			s.mainErrCallback(mainChanErr)
+		}
+
 		return fmt.Errorf("error updating account: %w", err)
 	}
 
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -214,6 +214,18 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 
 // Stop shuts down the account service.
 func (s *InterceptorService) Stop() error {
+	// We need to lock the request mutex to ensure that we don't stop the
+	// service while we're processing a request.
+	// This is especially important to ensure that we don't stop the service
+	// exactly after a user has made an rpc call to send a payment we can't
+	// know the payment hash for prior to the actual payment being sent
+	// (i.e. Keysend or SendToRoute). This is because if we stop the service
+	// after the send request has been sent to lnd, but before TrackPayment
+	// has been called, we won't be able to track the payment and debit the
+	// account.
+	s.requestMtx.Lock()
+	defer s.requestMtx.Unlock()
+
 	s.contextCancel()
 	close(s.quit)
 
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -63,6 +63,8 @@ type InterceptorService struct {
 	mainErrChan chan<- error
 	wg          sync.WaitGroup
 	quit        chan struct{}
+
+	isEnabled bool
 }
 
 // NewService returns a service backed by the macaroon Bolt DB stored in the
@@ -83,6 +85,7 @@ func NewService(dir string, errChan chan<- error) (*InterceptorService, error) {
 		pendingPayments:  make(map[lntypes.Hash]*trackedPayment),
 		mainErrChan:      errChan,
 		quit:             make(chan struct{}),
+		isEnabled:        false,
 	}, nil
 }
 
@@ -93,12 +96,15 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 	s.routerClient = routerClient
 	s.checkers = NewAccountChecker(s, params)
 
+	s.isEnabled = true
+
 	// Let's first fill our cache that maps invoices to accounts, which
 	// allows us to credit an account easily once an invoice is settled. We
 	// also track payments that aren't in a final state yet.
 	existingAccounts, err := s.store.Accounts()
 	if err != nil {
-		return fmt.Errorf("error querying existing accounts: %w", err)
+		return s.disableAndErrorf("error querying existing "+
+			"accounts: %w", err)
 	}
 	for _, acct := range existingAccounts {
 		acct := acct
@@ -116,8 +122,8 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 					acct.ID, hash, entry.FullAmount,
 				)
 				if err != nil {
-					return fmt.Errorf("error tracking "+
-						"payment: %w", err)
+					return s.disableAndErrorf("error "+
+						"tracking payment: %w", err)
 				}
 			}
 		}
@@ -146,8 +152,8 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 		s.currentSettleIndex = 0
 
 	default:
-		return fmt.Errorf("error determining last invoice indexes: %w",
-			err)
+		return s.disableAndErrorf("error determining last invoice "+
+			"indexes: %w", err)
 	}
 
 	invoiceChan, invoiceErrChan, err := lightningClient.SubscribeInvoices(
@@ -157,7 +163,7 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 		},
 	)
 	if err != nil {
-		return fmt.Errorf("error subscribing invoices: %w", err)
+		return s.disableAndErrorf("error subscribing invoices: %w", err)
 	}
 
 	s.wg.Add(1)
@@ -187,8 +193,11 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 				}
 
 			case err := <-invoiceErrChan:
-				log.Errorf("Error in invoice subscription: %v",
-					err)
+				// If the invoice subscription errors out, we
+				// stop the service as we won't be able to
+				// process invoices.
+				err = s.disableAndErrorf("Error in invoice "+
+					"subscription: %w", err)
 
 				select {
 				case s.mainErrChan <- err:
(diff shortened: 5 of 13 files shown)


terminal.go
```diff
@@ -644,9 +649,20 @@ func (g *LightningTerminal) start() error {
 	go func() {
 		for {
 			select {
-			case err := <-g.nonCriticalErrQueue.ChanOut():
-				log.Errorf("Received error from subsystem, "+
-					"keeping litd running: %v", err,
+			case errFunc := <-g.nonCriticalErrQueue.ChanOut():
+				serviceName, err := errFunc()
+				if serviceName == accounts.ACCOUNTS_SERVICE {
```
ViktorTigerstrom, line 658: I wasn't really sure how to do this in a clean way, to be able to tell which subsystem the error is sent from. I'm not a big fan of the channel returning a `func() (string, error)`, but couldn't figure out a better solution. Let me know if you have a better alternative!

I guess one alternative would let it be a `queue.ConcurrentQueue[error]` queue, and send it to the accounts service only and no other service. But that kind of defeats the purpose of having a queue in the first place.

accounts/rpcserver.go
```diff
@@ -53,6 +53,10 @@ func (s *RPCServer) CreateAccount(ctx context.Context,
 	log.Infof("[createaccount] label=%v, balance=%d, expiration=%d",
 		req.Label, req.AccountBalance, req.ExpirationDate)
 
+	if !s.service.IsRunning() {
+		return nil, ErrAccountServiceDisabled
+	}
+
```
ellemouton: cant we do a more generic thing where, similar to the existing sub-server manager, we have have a lit-system manager or something (eventually both the sub-server manager and lit-system manager could be merged or abstracted away into one `subserver manager`) and then we only need to check once in the `checkSubSystemStarted` `rpcProxy` method if the accounts service is up and ready. 

Then the account service itself is just responsible for signalling/reporting back that it is currently not ready for requests and then we dont have to repeat this `s.service.IsRunning` check everywhere here.

accounts/service.go
```diff
@@ -83,16 +85,28 @@ func NewService(dir string, errChan chan<- error) (*InterceptorService, error) {
...
 	s.routerClient = routerClient
 	s.checkers = NewAccountChecker(s, params)
 
+	s.isEnabled = true
+
+	defer func() {
+		if returnErr != nil {
+			s.Lock()
+			s.disable()
+			s.Unlock()
```
ellemouton, line 112: I suggest letting `disable` handle the locking and unlocking of the mutex itself. And then you can have another `disableUnsafe` method which does not grab the mutex so that you can use that one in places where the mutex is already acquired. 

```
func (s *InterceptorService) disable() {
	s.Lock()
	defer s.Unlock()
	
	s.disableUnsafe()
}

func (s *InterceptorService) disableUnsafe() {
	s.isEnabled = false
}
```

accounts/service.go
```diff
@@ -219,6 +240,33 @@ func (s *InterceptorService) Stop() error {
 	return s.store.Close()
 }
 
+// IsRunning checks if the account service is running, and returns a boolean
+// indicating whether it is running or not.
+func (s *InterceptorService) IsRunning() bool {
+	s.RLock()
+	defer s.RUnlock()
+	return s.isEnabled
```
ellemouton, line 254: nit: new line about return

accounts/service.go
```diff
@@ -219,6 +240,33 @@ func (s *InterceptorService) Stop() error {
...
+// RequireRunning checks that the account service is running, and returns an
+// error if it is not.
+//
+// NOTE: The store lock MUST be held as either a read or write lock when calling
+// this method.
+func (s *InterceptorService) requireRunning() error {
+	if !s.isEnabled {
+		return ErrAccountServiceDisabled
+	}
+	return nil
```
ellemouton, line 266: nit: new line above return

accounts/service.go
```diff
@@ -563,6 +655,24 @@ func (s *InterceptorService) paymentUpdate(hash lntypes.Hash,
...
+	}
+	defer func() {
+		// We want to ensure that we disable the service if we error
+		// when making the payment update. We need to do this while
+		// still holding the lock to ensure that other waiting threads
+		// to grab a lock will get informed immediately that the service
+		// is disabled.
+		if returnErr != nil {
+			s.disable()
+		}
```
ellemouton, line 732: related to my previous suggestion: perhaps instead there should just be one place to send critical errors and then that can be in-charge of updating the account-service status which will then ensure that no more requests come through

accounts/rpcserver.go
ellemouton: Interested to hear what you guys think though cause perhaps there would be danger in doing it this way since it is harder to keep things atomic since we then depend on an external system to work correctly etc plus we would need to think about the case where accounts wants to be disabled but then Lit lets through just one more request in the mean time before the "disable" is noted. 

Another (perhaps more simple) suggestion would be to allow lit subservers like the accounts server to provide grpc Unary and Stream interceptors that it expects Lit to register with any gRPC server that it spins up that would give access to accounts. These interceptors can then do the "are we ready for a call" check. But perhaps this also has the "call might slip through" issue. 

Lemme know what you guys think

accounts/service.go
```diff
@@ -348,7 +404,6 @@ func (s *InterceptorService) CheckBalance(id AccountID,
 // paid.
 func (s *InterceptorService) AssociateInvoice(id AccountID,
 	hash lntypes.Hash) error {
-
```
ellemouton, line 351: nit: pls keep the new line here :)

accounts/rpcserver.go
```diff
@@ -53,10 +53,6 @@ func (s *RPCServer) CreateAccount(ctx context.Context,
 	log.Infof("[createaccount] label=%v, balance=%d, expiration=%d",
 		req.Label, req.AccountBalance, req.ExpirationDate)
 
-	if !s.service.IsRunning() {
-		return nil, ErrAccountServiceDisabled
-	}
-
```
ellemouton: oh - do we need the first commit then?

accounts/service.go
```diff
@@ -219,6 +240,33 @@ func (s *InterceptorService) Stop() error {
...
+	defer s.RUnlock()
+	return s.isEnabled
+}
+
+// RequireRunning checks that the account service is running, and returns an
+// error if it is not.
+//
+// NOTE: The store lock MUST be held as either a read or write lock when calling
+// this method.
+func (s *InterceptorService) requireRunning() error {
```
ellemouton, line 262: I think we can just use the IsRunning() method instead of this

terminal.go
ellemouton: yeah this is why I think it would be cool to have a lit subsystem manager type thing which will handle receiving errors from the various subsystems (just accounts for now) and setting the status correctly, logging etc. Then you would not need this work around

terminal.go
```diff
@@ -881,21 +885,24 @@ func (g *LightningTerminal) startInternalSubServers(
...
 	}
-	// Even if we error on accountService.Start, we still want to mark the
-	// service as started so that we can properly shut it down in the
-	// shutdownSubServers call.
-	g.accountServiceStarted = true
+	// Even if we error on accountService.Start, or if the accounts service
+	// is disabled, we still want to mark that the service should be stopped
+	// so that we  can properly shut it down (canceling the ctx, and the
+	// accounts store) in the shutdownSubServers call.
+	g.accountServiceShouldBeStopped = true
```
ellemouton, line 905: do we need it to be stopped if the accounts mode is disabled though? 

I also think the previous name for the variable is fine btw

ellemouton (commented):
Leaving some comments for now - just want to gauge what the consensus of the best direction is. 

I think ideally we want things to be as generic as possible so that this pattern can easily be applied to other LiT sub-systems

ViktorTigerstrom:
Thanks for the review @ellemouton!
Posting a general comment here to respond what much of the feedback is related to :).

The main issues I see with just having one outer check that checks if a requests should be allowed through or not based on if the service is running or not, and then not checking in any of the internal functions in the server itself, is that:
1. The internal functions in the accounts system is not only called through gRPC requests, but we are also subscribed to the `lndclient.LightningClient` which will execute `invoiceUpdate` when an accounts invoice is added or paid, and `lndclient.RouterClient` which will execute `paymentUpdate` when a tracked payment is updated. It's important that these functions won't get executed if the service has already been shutdown, especially for `invoiceUpdate` as we'd then risk the `currentAddIndex` & `currentSettleIndex` going out of sync leading to missed added or settled invoices.
2. The service can shutdown during the time a request that has been allowed through is actually executed. I.e. if an account makes a request to send a payment, the service can be `Running` when the actual request is done, but when the payment has been sent and we handle the `SendPaymentV2` response in `TrackPayment`, the service might have been shutdown through for example an invoice update that came in while the users request was processed.

This is also why it's very important that we often `disable` the service under the same lock in the function execution that actually caused the error. Think of the following scenario:
We get 2 invoice settle updates rapidly that settles two different invoices in a very short timespan. Now if the first invoice update acquires the the lock, and the second update attempts to grab the lock in another thread. It's in that scenario super important that the we `disable` the service under the same lock as the update is attempted for the first invoice, if update fails. If we didn't and dropped the lock, and reacquired it to disable the service, we'd risk that the second invoice update grabs the lock before we've managed to `disable` the service. If the second invoice update then succeeds, we'd end-up with the `currentSettleIndex` out of sync, which would mean that we'd never credit the account for the first invoice update.
So if we remove the `requireRunning` function, we need to add a `IsRunning` function version that doesn't also acquire the lock, i.e. a `UnsafeIsRunning` is running function.

Now it's most clear why we'd need to check if the service is already running before we execute the `invoiceUpdate` function, and we may not need to do that for any other function. But I didn't really know where to draw the line of when we'd require that the service is running if we executed the function, so I settled for functions that update the balance of an account. This is because it feels strange that in a scenario that the service was `disabled` while send payment request by an account was executed, but before the `SendPaymentV2` response from LND was received, that we'd still successfully the handle the payment update. Meaning we'd successfully execute `TrackPayment` which adds the subscription to the `lndclient.RouterClient` and which in turn would then execute `paymentUpdate` once the payment has been updated. It just feels strange that all of those call would successfully execute while the service was `disabled`, which is why I drew the line of requiring that the service is running for any function that updates the user balance.

accounts/rpcserver.go
ViktorTigerstrom: Interested to hear if you think we can make your suggestion work for the issues I explained in https://github.com/lightninglabs/lightning-terminal/pull/642#issuecomment-1729293200 :)!

terminal.go
ViktorTigerstrom: Hmm yeah agree that this is just strange. Perhaps we should just close everything in the `accountService` immediately if the service is disabled :)!

ellemouton:
ok cool! thanks for the explanation @ViktorTigerstrom  - that makes sense. I didnt think about point 1.

accounts/interface.go
```diff
@@ -160,6 +160,9 @@ var (
 	ErrNotSupportedWithAccounts = errors.New("this RPC call is not " +
 		"supported with restricted account macaroons")
 
+	ErrAccountServiceDisabled = errors.New("the account service has been " +
```
guggero, line 163: nit: missing godoc.

accounts/service.go
guggero: Also, it's a bit of a weird pattern with the `defer` and `returnErr` (pretty cool idea in general though!).
I would make it a bit more explicit by just replacing any `return fmt.Errorf()` with `return disableAndErrorf(...)`:
```
disableAndErrorf := func (format string, a ...any) error {
    s.disable()
    return fmt.Errorf(format, a...)
}
```

terminal.go
```diff
@@ -843,9 +870,12 @@ func (g *LightningTerminal) startInternalSubServers(
 		g.lndClient.ChainParams,
 	)
 	if err != nil {
-		return fmt.Errorf("error starting account service: %v",
-			err)
+		log.Errorf("error starting account service: %v, disabling "+
+			"account service", err)
 	}
+	// Even if we error on accountService.Start, we still want to mark the
```
guggero: nit: add newline before comment block.

terminal.go
```diff
@@ -631,6 +639,25 @@ func (g *LightningTerminal) start() error {
 	// We can now set the status of LiT as running.
 	g.statusMgr.SetRunning(subservers.LIT)
 
+	// Spawn a goroutine that will log any non-critical errors from
+	// sub systems, until we receive the main shutdown signal.
+	go func() {
+		for {
+			select {
+			case err := <-g.nonCriticalErrQueue.ChanOut():
+				log.Errorf("Received error from subsystem, "+
```
guggero: If we're just logging the errors, shouldn't we instead just remove the error channel from the account service and turn all places where we would send to that channel into a log call?

accounts/checkers.go
```diff
@@ -521,6 +521,7 @@ func checkSend(ctx context.Context, chainParams *chaincfg.Params,
 		sendAmt = lnwire.MilliSatoshi(amtMsat)
 	}
 
+	var paymentHash lntypes.Hash
```
guggero, line 524: nit: move to after the comment?

accounts/checkers.go
```diff
@@ -549,6 +552,13 @@ func checkSend(ctx context.Context, chainParams *chaincfg.Params,
 		return fmt.Errorf("error validating account balance: %w", err)
 	}
 
+	if len(paymentHash) != 0 {
```
guggero, line 555: `paymentHash` is an array, so its length is always set! Need to compare it to an all-zero hash instead: `if paymentHash != lntypes.Hash{} {`

accounts/checkers.go
```diff
@@ -531,6 +532,8 @@ func checkSend(ctx context.Context, chainParams *chaincfg.Params,
 		if payReq.MilliSat != nil && *payReq.MilliSat > sendAmt {
 			sendAmt = *payReq.MilliSat
 		}
+
+		paymentHash = *payReq.PaymentHash
```
guggero, line 536: Throw a `nil` check in here, just in case? Not sure if an AMP invoice really has a payment hash in every case.

accounts/service.go
```diff
@@ -417,6 +417,38 @@ func (s *InterceptorService) AssociateInvoice(id AccountID,
 	return s.store.UpdateAccount(account)
 }
 
+// AssociatePayment associates a payment for with the given account,
```
guggero, line 426: nit: weird sentence (for with the).

accounts/service.go
```diff
@@ -529,11 +561,32 @@ func (s *InterceptorService) TrackPayment(id AccountID, hash lntypes.Hash,
 		return nil
 	}
 
-	// Okay, we haven't tracked this payment before. So let's now associate
-	// the account with it.
-	account.Payments[hash] = &PaymentEntry{
-		Status:     lnrpc.Payment_UNKNOWN,
-		FullAmount: fullAmt,
+	if !ok {
```
guggero, line 570: I'm wondering if we could ever end up here with an all-zero payment hash (e.g. because it wasn't set in one of the calls). Not sure if we should check for that at least before we associate the payment?

accounts/service.go
```diff
@@ -529,11 +561,32 @@ func (s *InterceptorService) TrackPayment(id AccountID, hash lntypes.Hash,
...
+		// In the rare case that the payment isn't associated with an
+		// account yet, we do that first, to ensure that we will track
+		// the payment in case the service is restarted. This can happen
+		// if the payment is a KeySend or SendToRoute payment.
+		// This is attempted even if the service is disabled, as we'll
+		// then track the payment on next startup.
+		account.Payments[hash] = &PaymentEntry{
```
guggero, line 577: Why are we putting this behind the `!ok` flag now? I think we always want to reset the status to `lnrpc.Payment_UNKNOWN` if we call `TrackPayment`, to avoid missing any updates.

accounts/service.go
```diff
@@ -529,11 +561,32 @@ func (s *InterceptorService) TrackPayment(id AccountID, hash lntypes.Hash,
...
+			Status:     lnrpc.Payment_UNKNOWN,
+			FullAmount: fullAmt,
+		}
+
+		err := s.store.UpdateAccount(account)
```
guggero, line 582: Given that we already do exactly this call below, don't we just want to move the sending of the error to the err chan to the `UpdateAccount` call below?

accounts/service.go
```diff
@@ -60,7 +64,7 @@ type InterceptorService struct {
 	invoiceToAccount map[lntypes.Hash]AccountID
 	pendingPayments  map[lntypes.Hash]*trackedPayment
 
-	mainErrChan chan<- error
+	mainErrChan chan<- func() (string, error)
```
guggero, line 67: What if we instead just turn this into an `onError` callback? Then we don't need to do any filtering in Terminal and can just directly set the subserver status to errored.

guggero: And we still don't need to add the non critical error queue.

terminal.go
```diff
@@ -246,7 +246,11 @@ func (g *LightningTerminal) Run() error {
 	// Register LND, LiT and Accounts with the status manager.
 	g.statusMgr.RegisterAndEnableSubServer(subservers.LND)
 	g.statusMgr.RegisterAndEnableSubServer(subservers.LIT)
-	g.statusMgr.RegisterAndEnableSubServer(subservers.ACCOUNTS)
+	if g.cfg.AccountsMode != ModeDisable {
```
guggero, line 249: nit: always register and only enable on condition? Removes an `else` clause :D

guggero (commented):
Did a first pass. Great work! Looks pretty good, but have a few suggestions to make the diff even smaller.

ViktorTigerstrom:
Thanks a lot for the reviews @ellemouton & @guggero! Working on addressing them.

Though I'd like some more feedback on how we should surface the errors to the status server, so would just like to get consensus on that. Thanks for the suggestion on using a callback instead @guggero, will definitely change to that! Though would like to know if we should introduce the sub system server like @ellemouton suggested, to make it more generalizable for future sub systems we might add, or if we should leave that for now and do that later if needed? I think we will still need the internal `RequireRunning` checks either way though.

ellemouton:
> would like to know if we should introduce the sub system server like @ellemouton suggested, to make it more generalizable for future sub systems we might add, or if we should leave that for now and do that later if needed? 

I'd say that if we are on a bit of a time limit with this rn, then we can worry about making things more generalisable in a follow up 👍 Since it sounds like that would take some design work

guggero:
> > would like to know if we should introduce the sub system server like @ellemouton suggested, to make it more generalizable for future sub systems we might add, or if we should leave that for now and do that later if needed?
> 
> I'd say that if we are on a bit of a time limit with this rn, then we can worry about making things more generalisable in a follow up 👍 Since it sounds like that would take some design work

Was about to suggest the same! Even though turning everything into subservers sounds great, might be overkill for this PR. So let's see how the diff looks like with just the callback for now.

ViktorTigerstrom:
Thanks! Will leave it out for now then, and then we can address it in a follow-up later :)

ViktorTigerstrom:
Addressed the latest feedback, rebased on main now that the status server PR has been merged, and added unit tests!

terminal.go
ViktorTigerstrom: Added that we now stop the account service directly if it is disabled instead!

terminal.go
ViktorTigerstrom: Made this a callback instead :)!

accounts/service.go
ViktorTigerstrom: Hmm I'm not completely sure what that I understood this correctly. I think this should be fixed with the fix for this other feedback:
https://github.com/lightninglabs/lightning-terminal/pull/642#discussion_r1334124267

But please double check if I misunderstood what you ment with this comment, and let me know if I need to fix something else :)

accounts/service.go
ViktorTigerstrom: Ah sorry! Didn't understand that was the intention, I've fixed this now with the latest push!

accounts/service.go
ViktorTigerstrom: Sorry, good catch :)

accounts/service.go
ViktorTigerstrom: Updated with the latest push :)

ViktorTigerstrom:
Hmm sorry, looking into to addressing the `unit-race` errors.

In terms of `itests`, I'm looking into adding an `itest` that triggers the `mainErrCallback` to be called, to shutdown the  accounts service while not `disabled`. Though unfortunately I think it's going to be a bit hard as the `litd` process is started from binary and not through the struct. So in case I can't find any, I'll just add a `disabled` through configuration `itest`.   

`check commits` is expected to fail until the commits are squashed.

ViktorTigerstrom:
Fixed CI errors!

ViktorTigerstrom:
Updated the `itests` to test that we can `disable` the accounts system through configuration.
Also updated the commit that  adds that config option a little, as I noticed some bugs in it when making the `itest`.

accounts/service.go
```diff
@@ -472,7 +468,9 @@ func (s *InterceptorService) invoiceUpdate(
 			s.currentAddIndex, s.currentSettleIndex,
 		)
 		if err != nil {
-			return err
+			return s.disableAndErrorfUnsafe(
```
guggero, line 500: I like this pattern better to be honest. It's more explicit and easier to follow. And one doesn't have to reason about "what comes first, the `defer s.Unlock()` or the `defer func() ... s.disableUnsafe() ...`.

accounts/service.go
```diff
@@ -531,14 +564,32 @@ func (s *InterceptorService) TrackPayment(id AccountID, hash lntypes.Hash,
...
 	account.Payments[hash] = &PaymentEntry{
 		Status:     lnrpc.Payment_UNKNOWN,
 		FullAmount: fullAmt,
 	}
+
 	if err := s.store.UpdateAccount(account); err != nil {
-		return fmt.Errorf("error updating account: %w", err)
+		if ok {
+			return fmt.Errorf("error updating account: %w", err)
+		} else {
```
guggero, line 579: Since the `if` is a `return`, we don't need to have the `else` and the indentation.

terminal.go
```diff
@@ -850,22 +855,33 @@ func (g *LightningTerminal) startInternalSubServers(
...
+				"disabling account service", err)
+			g.statusMgr.SetErrored(subservers.ACCOUNTS, err.Error())
+			stopAccountService()
+		} else {
+			g.statusMgr.SetRunning(subservers.ACCOUNTS)
+			g.accountServiceStarted = true
+		}
 	} else {
-		g.statusMgr.SetRunning(subservers.ACCOUNTS)
+		stopAccountService()
```
guggero, line 883: This `else` seems to be wrong... Why would we stop the service if we haven't started it?

itest/litd_mode_integrated_test.go
```diff
@@ -213,10 +213,21 @@ var (
...
+
+		// disabledPattern represents a substring that is expected to be
+		// part of the error returned when a gRPC request is made to the
+		// disabled endpoint.
+		// TODO: once we have a subsystem manager, we can unify the
+		// returned for disabled endpoints for both subsystems and
+		// subservers by not registering the subsystem URIs to the
+		// permsMgr if it has been disabled. This field will then be
+		// unnecessary and can be removed.
+		disabledPattern string
```
guggero, line 229: :+1:

guggero (commented):
Looks good! Can you please apply the fixup commits and I'll do a final review on the completed diff?

accounts/service.go
ViktorTigerstrom: Ok, squashed the commits now! Thanks for the feedback :)!

accounts/service.go
ViktorTigerstrom: Ah thanks!I noticed when fixing this that I had forgotten to return on the `else` case here, so I changed the order of the `if` check to instead check for `!ok` and then return the `err` below the check for both cases.

terminal.go
ViktorTigerstrom: So agree that this is a bit confusing. The reason why we need to do this is that `s.accountService.Stop()` closes the contexts and db store which are opened when we create the account service with `accounts.NewService` in the beginning of the `LightningTerminal.start` function. This is done regardless of if the accounts service is `disabled` or if we fail to start it, so we need to close the contexts and db store either way, so this is why we run the `stopAccountService` function either way.

I changed the name of the `stopAccountService` function to `closeAccountService` to better represent what it actually does to make it more understandable. Let me know though if you have a suggestion of how I could do this better :)

ViktorTigerstrom:
Thanks for the review at @guggero! Addressed the feedback and squashed the commits :)!

Snap though, I notice that the `check commits` check still fails, so will address that now.

ViktorTigerstrom:
Fixed `check commits` CI failure :)

accounts/service.go
```diff
@@ -83,22 +85,27 @@ func NewService(dir string, errChan chan<- error) (*InterceptorService, error) {
...
 		mainErrChan:      errChan,
 		quit:             make(chan struct{}),
+		isEnabled:        false,
 	}, nil
 }
 
 // Start starts the account service and its interceptor capability.
 func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
-	routerClient lndclient.RouterClient, params *chaincfg.Params) error {
+	routerClient lndclient.RouterClient,
```
guggero, line 100: nit: no longer need to break here.

accounts/service.go
```diff
@@ -187,8 +194,11 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 				}
 
 			case err := <-invoiceErrChan:
-				log.Errorf("Error in invoice subscription: %v",
-					err)
+				// if the invoice subscription errors out, we
+				// stop the service as we won't be able to
+				// process invoices.
+				err = s.disableAndErrorf("Error in invoice "+
+					"subscription: %v", err)
```
guggero, line 203: nit: could wrap the error here now (`%w` instead of `%v`).

accounts/service.go
```diff
@@ -544,8 +642,12 @@ func (s *InterceptorService) TrackPayment(id AccountID, hash lntypes.Hash,
 // associated with, in case it is settled. The boolean value returned indicates
 // whether the status was terminal or not. If it's not terminal then further
 // updates are expected.
+//
+// NOTE! Any code that errors in this function MUST call disableAndErrorfUnsafe
+// while the store lock is held to ensure that the service is disabled under
+// the same lock.
 func (s *InterceptorService) paymentUpdate(hash lntypes.Hash,
-	status lndclient.PaymentStatus) (bool, error) {
+	status lndclient.PaymentStatus) (b bool, returnErr error) {
```
guggero, line 693: nit: can remove the named returns.

accounts/service.go
```diff
@@ -36,6 +36,10 @@ type trackedPayment struct {
 	cancel context.CancelFunc
 }
 
+const (
+	ACCOUNTS_SERVICE = "accounts"
```
guggero, line 40: Looks like this is unused?

terminal.go
```diff
@@ -850,22 +855,33 @@ func (g *LightningTerminal) startInternalSubServers(
...
+				"disabling account service", err)
+			g.statusMgr.SetErrored(subservers.ACCOUNTS, err.Error())
+			closeAccountService()
+		} else {
+			g.statusMgr.SetRunning(subservers.ACCOUNTS)
+			g.accountServiceStarted = true
+		}
 	} else {
-		g.statusMgr.SetRunning(subservers.ACCOUNTS)
+		closeAccountService()
```
guggero, line 883: Ah, I see. That's basically a fail on my part then, a `NewXXX` shouldn't do anything that requires stopping, only `Start()` should do that. Maybe add a short comment here then that explains the situation?

guggero (approved):
Looks great, thanks a lot! Only nits left, LGTM :tada:

terminal.go
ViktorTigerstrom: Ok, agree that makes sense :)! Added a comment for the `closeAccountService` function.

ViktorTigerstrom:
Thanks @guggero 🎉!! Addressed you're latest feedback with the last push :)

ViktorTigerstrom:
Added a small fix that ensures that the node runner can't stop the accounts service while a request by an account user is being processed.
This is especially important to ensure that we don't stop the service exactly after a user has made an rpc call to send a payment we can't know the payment hash for prior to the actual payment being sent (i.e. Keysend or SendToRoute). This is because if we stop the service after the send request has been sent to lnd, but before TrackPayment has been called, we won't be able to track the payment and debit the account.

config.go
```diff
@@ -199,6 +201,8 @@ type Config struct {
 	TaprootAssetsMode string         `long:"taproot-assets-mode" description:"The mode to run taproot assets in, either 'integrated' (default), 'remote' or 'disable'. 'integrated' means tapd is started alongside the UI and everything is stored in tap's main data directory, configure everything by using the --taproot-assets.* flags. 'remote' means the UI connects to an existing tapd node and acts as a proxy for gRPC calls to it. 'disable' means that LiT is started without tapd" choice:"integrated" choice:"disable"`
 	TaprootAssets     *tapcfg.Config `group:"Integrated taproot assets options (use when taproot-assets=integrated)" namespace:"taproot-assets"`
 
+	AccountsMode string `long:"accounts-mode" description:"The mode to run the accounts service in, either 'enable' (default), or 'disable'. 'enable' means that LiT is started with the integrated the accounts service enabled and running. 'disable' means that LiT is started with the accounts service disabled, and all requests to the accounts service will be rejected." choice:"enable" choice:"disable"`
```
ellemouton, line 204: could also just be a `disable` boolean instead?

terminal.go
```diff
@@ -181,8 +181,9 @@ type LightningTerminal struct {
 	middleware        *mid.Manager
 	middlewareStarted bool
 
-	accountService        *accounts.InterceptorService
-	accountServiceStarted bool
+	accountService            *accounts.InterceptorService
+	accountServiceStarted     bool
+	accountServiceErrCallback func(error)
```
ellemouton, line 186: unused variable

ellemouton (approved):
Looks good! Only nits/style comments from my side.

The only one I think we should maybe defs address here is converting the new "accounts-mode" option to a "Accounts.Disable" boolean instead?

config.go
ViktorTigerstrom: Good idea! Thanks, updated with a boolean instead :)!

accounts/service.go
```diff
@@ -219,6 +228,51 @@ func (s *InterceptorService) Stop() error {
...
+}
+
+// disable disables the account service, and marks the service as not running.
+// The function acquires the store write lock before disabling the service.
+// The function returns an error with the given format and arguments.
+func (s *InterceptorService) disableAndErrorf(format string, a ...any) error {
+	s.Lock()
+	defer s.Unlock()
+
+	s.isEnabled = false
```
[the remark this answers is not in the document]
ViktorTigerstrom: Thanks! I need to get better at this, so learning from this :). I've looked through the entire PR, and added a few new lines. So hope the new changes are good :)!

accounts/interceptor.go
```diff
@@ -52,6 +52,17 @@ func (s *InterceptorService) Intercept(ctx context.Context,
 	s.requestMtx.Lock()
 	defer s.requestMtx.Unlock()
 
+	// If the account service is not running, we reject all requests.
```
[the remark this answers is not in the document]
ViktorTigerstrom: Thanks, agree! Keeping this in mind for the future.

accounts/service.go
ViktorTigerstrom: Ah, sorry. I ment the discussion that we need to check if the service is running under the same lock, and not drop the lock to then reacquire it, so we still need two different versions of this function which is why I kept this function before.

But agree that it's confusing with the error, so I changed the PR now to instead include a `isRunningUnsafe` function that doesn't acquire the lock, and use that function instead where it is needed :)

accounts/service.go
```diff
@@ -93,12 +96,15 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
...
 
+	s.isEnabled = true
+
 	// Let's first fill our cache that maps invoices to accounts, which
 	// allows us to credit an account easily once an invoice is settled. We
 	// also track payments that aren't in a final state yet.
 	existingAccounts, err := s.store.Accounts()
 	if err != nil {
-		return fmt.Errorf("error querying existing accounts: %w", err)
+		return s.disableAndErrorf("error querying existing "+
```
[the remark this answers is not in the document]
ViktorTigerstrom: So unless I'm completely wrong, we can't use the status manager to block gRPC requests from account users, which are making are making lnrpc calls with a macaroon with the account caveat attached to it. This is why the `!s.IsRunning()` check is kept in `accounts/interceptor.go` even after the commit that adds the accounts subsystem to status manager, to block the account users calls when the accounts service has been disabled (either through erroring or through config).
So if we didn't `disableAndErrorf` here, the accounts users gRPC calls would still pass through the `interceptor.go`, and fail later in the processing due to the db being closed or similar.

I might be completely wrong in my understanding here though, so please let me know if this can be done in a better way.

accounts/service.go
```diff
@@ -93,12 +96,15 @@ func (s *InterceptorService) Start(lightningClient lndclient.LightningClient,
 	s.routerClient = routerClient
 	s.checkers = NewAccountChecker(s, params)
 
+	s.isEnabled = true
```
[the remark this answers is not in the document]
ViktorTigerstrom: I agree that this would be a much cleaner solution. Unfortunately though, we have calls in the `Start` function that requires that `isRunning` returns true to not error, which is why we need to set it as enabled in the start of the function.

1. The `Start` function calls `TrackPayment` which requires that the service is running to add the payment for tracking to the `lndclient.RouterClient`. Potentially this could be changed, and only require the service to be running when the actual payment update is returned by the `lndclient.RouterClient`. Though that feels a bit wrong, so I'd prefer to keep it as is unless you disagree.

2. The `Start` function spawns a goroutine that calls `invoiceUpdate` that definitely requires that the service is running when being executed. This could potentially be solved though by letting the `Start` function acquire the store lock throughout the execution of the function.

Let me know if you think I should change this, or if the current version of the `Start` function is ok :).

ViktorTigerstrom:
Addressed the latest feedback, and left some comments for some of the comments which weren't addressed :)

accounts/service.go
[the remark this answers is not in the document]
ellemouton: oh ok I see 👍  makes sense

accounts/service.go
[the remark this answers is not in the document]
ellemouton: yeah dont worry too much :) it's just engrained in me from the times before the linter was de-activated :) 

Hopefully we can re-activate it soon :)

accounts/service.go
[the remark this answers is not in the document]
ellemouton: gotcha! thanks for the explanation :)

PR #645, opened 2023-09-28
multi: bump lnd to v0.17.0-beta

guggero:
Replaces https://github.com/lightninglabs/lightning-terminal/pull/633.

Depends on:
 - https://github.com/lightninglabs/faraday/pull/171
 - https://github.com/lightninglabs/loop/pull/641
 - https://github.com/lightninglabs/pool/pull/470
 - The final release of `lnd v0.17.0-beta` (and a bump in `lndclient`).


ellemouton (commented):
LGTM pending LiT version bump and Readme update. 

Think we should also include these readme updates here: https://github.com/lightninglabs/lightning-terminal/pull/543

guggero:
Updated the README and also added a version bump for `litd` itself.

ViktorTigerstrom (approved):
tACK LGTM!

Tested the sessions, loop & pool with the new changes, and everything seems to be working as expected!

Very nitty: I think the last 2 commits can be squashed into 1, but definitely not a blocker :)!

Just wanted to sanity check, since the description of the PR doesn't mention that this is dependent on https://github.com/lightninglabs/taproot-assets/pull/536, but depends on all other bumps, should that not be the case?

README.md
```diff
@@ -146,6 +147,7 @@ The following table shows the supported combinations:
 
 | LiT               | LND          | Loop         | Faraday       | Pool         | Taproot Assets |
 |-------------------|--------------|--------------|---------------|--------------|----------------|
+| **v0.11.0-alpha** | v0.17.0-beta | v0.26.3-beta | v0.2.11-alpha | v0.6.4-beta  | v0.2.3-alpha   |
```
ViktorTigerstrom, line 150: nit: taproot assets included is at v0.2.4.0-XXX in the `go.mod` file

guggero: You're right, Taproot Assets is a bit of an outlier here. Should've added that to the PR description: Since we're very close to releasing `v0.3`, we will be doing another `litd` release soon. But because a lot of things in the `main` branch for the `v0.3` release are breaking changes, we need to stay on a `v0.2.3` base for now. So I just pushed a branch with the `lnd` update and didn't create a PR for it, as we couldn't merge it anywhere anyway. But as soon as `v0.3` is out, everything will be smooth again.

Oh and the `v0.2.4` is just a weird thing go modules does, when it uses a commit based pseudo version, it actually increases the patch version. The closest tag to the commit used is actually `v0.2.3`, there is no `v0.2.4`.

ViktorTigerstrom: Ah ok, makes sense. Thanks for explaining!

jamaljsr (approved):
tACK LGTM 🚀

PR #651, opened 2023-10-11
accounts: fix service check in UpdateAccounts call

GeorgeTsagk:
## Description

This PR fixes a simple check in the `UpdateAccount` RPC handler which checks if the service is running.


guggero (approved):
Thanks for the fix! LGTM :tada:

ViktorTigerstrom (approved):
Oh snap, my bad 😬... Thanks a lot for the fix :tada:! LGTM

PR #663, opened 2023-10-19
config: allow taproot-assets-mode to be remote

jamaljsr:
When starting `litd` with the flag `--taproot-assets-mode=remote`, the following error was being displayed preventing it from starting up.

```
could not load config: error parsing flags: Invalid value `remote' for option `--taproot-assets-mode'. Allowed values are: integrated or disable
```

I realized the `remote` choice was missing in the `go-flags` options. Adding this has resolved the startup issue.


guggero (approved):
Thanks for the fix, LGTM :tada:

jamaljsr:
The user who reported this issue in Slack has tested and confirmed it resolved the problem.

PR #684, opened 2023-11-19
feat: add keysend messenger to custom permissions

secondl1ght:
Added a new option to the Custom Permissions view when creating a new LNC session. It is called 'Keysend Messenger' which allows to send/receive payments, plus sign messages. This preset can be used in apps such as Cipherchat for example. Users will now also be able to add Sign Message to their Custom permissions as well.

image


jamaljsr (approved):
tACK LGTM 🚀

Thanks for submitting the PR. I've tested this and confirmed it works as expected. The code looks great as well.

I'd just like to confirm with @levmi that the copy is ok before merging.

levmi (changes requested):
Thanks for the PR, love to see it! Happy to get this in :) 

Just one nit on copy, ideally, we could adjust to "User can send and receive payments plus sign messages". I'm okay with title remaining "Keysend Messenger".

dstrukt (approved):
Had to check this out - very cool - thanks for the PR @secondl1ght!

And ACK to @levmi's copy suggestion to keep the description consistent.

secondl1ght:
Woot! Thanks for the review guys - I adjusted the copy, I agree it reads better this way.

One other thing I was thinking about is that instead of having the expected values for `permissionType` in a comment, we could change it to use an actual type for these values. (see here for the line I am speaking of: https://github.com/lightninglabs/lightning-terminal/pull/684/files#diff-281dcaa6b4e26633152a0b6ce3832cc66285a0ec1d9f1601c3b5bfa56e7855a4R10)

Happy to push one more commit to this PR to add typing to this variable if you guys think it is a good idea.

jamaljsr:
> One other thing I was thinking about is that instead of having the expected values for `permissionType` in a comment, we could change it to use an actual type for these values (see here for the line I am speaking of: https://github.com/lightninglabs/lightning-terminal/pull/684/files#diff-281dcaa6b4e26633152a0b6ce3832cc66285a0ec1d9f1601c3b5bfa56e7855a4R10).
> 
> Happy to push one more commit to this PR to add typing to this variable if you guys think it is a good idea.

Sounds good. Feel free to make that refactoring here in this PR.

secondl1ght:
Ok cool, I just pushed a commit adding a type. Please let me know what you think! 🙌🏼

levmi:
Awesome! Really love to see the external contribution here. If you ever want to add anything else to `litd` or the `LNC` side, don't hesitate to put up a PR or reach out. Oftentimes, we'll really want to implement something, but not necessarily have the resourcing to prioritize it. This is immensely helpful so thanks!

levmi (approved):

jamaljsr (approved):
Tested the latest commits. Everything looks great!

Thanks for the refactoring as well. LGTM 🚀

secondl1ght:
Awesome thanks everyone! I will definitely continue to contribute to the LL repos when I have some time. 🙂

Issue #683, opened 2023-11-16
Restart LND on lit startup if failure


levmi:
If LND fails on `litd` startup, the `LND` process should automatically restart. If `LND` is running in remote mode, we should attempt to re-connect. We should attempt this restart in perpetuity, but have a timeout with regard to how often restart attempts occur.  When failing to start while restarting, the timeout should be multiplied, so that we attempt to restart less and less often. We can make this configurable.

PR #694, opened 2023-12-12
multi: restart LND Client on setup failure

ViktorTigerstrom:
This PR adds the functionality that in case we fail to setup the LND client(s) when starting `litd`, we will continuously retry to connect the LND clients until we succeed to do so. This partly addresses #683, but currently only restarts the LND clients when failing to setting them up, but not in case we fail to start `lnd`. The `lit` sub-server will only be set to status `Running` after the lnd clients have successfully been set up.
This hopefully addresses most of the issues we've had from users that have experienced issues with the `lnd` sub-server erroring on the start up of `litd`, but likely not all of the errors.

A note to reviewers:

1. Currently when failing to set up and therefore now retrying to create the lnd client(s), the `litcli getinfo` + the `litcli stop` commands are blocked and are not allowed, as we currently require macaroon authentication for those calls. Should I update this PR to allow those calls by adding them to the `LiT` Whitelist?
2. To simplify the local testing of this PR, I've created a branch that very simply mocks 3 failures to set up the `basic lnd client` as well as 3 failures to set up the full `lnd client`. The full `lnd client` set up attempts will also fail after 30 seconds, as that's what users experience when their set up fails due to the `lnrpc.GetInfo` call context times out. You can use that branch to test this PR locally: https://github.com/ViktorTigerstrom/lightning-terminal/tree/2023-12-restart-lndclient-on-failure-mocked-lndclient-fails

Happy to address any feedback!


subservers/manager.go
```diff
@@ -52,7 +52,7 @@ func NewManager(permsMgr *perms.Manager,
 // AddServer adds a new subServer to the manager's set.
 func (s *Manager) AddServer(ss SubServer, enable bool) {
 	// Register all sub-servers with the status server.
-	s.statusServer.RegisterSubServer(ss.Name())
+	s.statusServer.RegisterSubServer(ss.Name(), nil)
```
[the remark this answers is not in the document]
ellemouton: ie, something like this:

```
type SubServerOption func(status *subServer)

// WithIsReadyOverride is a functional option that can be used to set a call
// back function that is used to check if a system is ready _iff_ the system
// running status is not yet true. The call-back will be passed the request URI
// along with any manual status that has been set for the subsystem.
func WithIsReadyOverride(fn func(string, string) (bool, bool)) SubServerOption {
	return func(status *subServer) {
		status.isReadyOverride = fn
	}
}

// RegisterSubServer will create a new sub-server entry for the Manager to
// keep track of.
func (s *Manager) RegisterSubServer(name string, opts ...SubServerOption) {
	s.mu.RLock()
	defer s.mu.RUnlock()

	s.registerSubServerUnsafe(name, true, opts...)
}

// RegisterAndEnableSubServer will create a new sub-server entry for the
// Manager to keep track of and will set it as enabled.
func (s *Manager) RegisterAndEnableSubServer(name string,
	opts ...SubServerOption) {

	s.mu.RLock()
	defer s.mu.RUnlock()

	s.registerSubServerUnsafe(name, false, opts...)
}

func (s *Manager) registerSubServerUnsafe(name string, disabled bool,
	opts ...SubServerOption) {

	ss := newSubServerStatus(disabled)

	for _, o := range opts {
		o(ss)
	}

	s.subServers[name] = ss
}
```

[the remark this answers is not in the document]
ellemouton: then, `IsSystemReady` can be changed to:

```
func (s *Manager) IsSystemReady(name, req string) (bool, bool, error) {
	s.mu.RLock()
	defer s.mu.RUnlock()

	server, ok := s.subServers[name]
	if !ok {
		return false, false, errors.New("a sub-server with " +
			"name %s has not yet been registered")
	}

	if server.disabled {
		return false, true, nil
	}

	if server.running {
		return true, false, nil
	}

	if server.isReadyOverride == nil {
		return false, false, nil
	}

	isReady, handled := server.isReadyOverride(req, server.manualStatus)
	if handled {
		return false, false, nil
	}

	return isReady, false, nil
}
```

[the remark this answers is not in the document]
ellemouton: then the code in `terminal.go` becomes:

```
	lndOverride := func(uri, manualStatus string) (bool, bool) {
		if uri != "/lnrpc.State/GetState" {
			return false, false
		}

		return manualStatus == lndWalletReadyStatus, true
	}

	// Register LND, LiT and Accounts with the status manager.
	g.statusMgr.RegisterAndEnableSubServer(
		subservers.LND, status.WithIsReadyOverride(lndOverride),
	)
	g.statusMgr.RegisterAndEnableSubServer(subservers.LIT)
	g.statusMgr.RegisterSubServer(subservers.ACCOUNTS)
```

status/manager.go
```diff
@@ -78,23 +110,27 @@ func (s *Manager) SubServerStatus(_ context.Context,
...
 // Manager to keep track of and will set it as enabled.
-func (s *Manager) RegisterAndEnableSubServer(name string) {
+func (s *Manager) RegisterAndEnableSubServer(name string,
+	overrides map[string]func() bool) {
+
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
-	ss := newSubServerStatus()
+	ss := newSubServerStatus(overrides)
```
ellemouton, line 133: btw - I think we should remove the `GetStatus` method below & unexport the `SubServerStatus` struct (and probably just rename it to `subServer`) since we dont want to expose the values of the struct to outside this package.

status/manager.go
```diff
@@ -25,12 +25,22 @@ type SubServerStatus struct {
...
 	Err string
+
+	// overrides defines which specific RPC endpoints for the sub-server,
+	// if any, where the default behaviour of only allowing the request
+	// after the sub-server has been set to Running is overridden.
+	// The key is the RPC endpoint and the value is a function that returns
+	// true if the request should be allowed at the current time.
+	// Note that if the sub-server is disabled, all requests will be denied
+	// regardless of if the RPC endpoint has an override entry.
+	overrides map[string]func() bool
```
ellemouton, line 36: with the suggestions below, this then becomes:
```
        // isReadyOverride is a call back that, when set and only if `running`
	// is not yet true, will be used to determine if a system is ready for
	// a call. We will pass the request URI to this method along with the
	// `manualStatus`. The first returned boolean is true if the system
	// should be seen as ready and the second is true if the override does
	// handle the given request. If it does not, then we will fall back to
	// our normal is-ready check.
	isReadyOverride func(string, string) (bool, bool)
```

terminal.go
```diff
@@ -712,19 +712,28 @@ func (g *LightningTerminal) setUpLNDClients() error {
 	// because that doesn't do anything else than just connect. We'll check
 	// if lnd is also ready to be used in the next step.
 	log.Infof("Connecting basic lnd client")
-	err := wait.NoError(func() error {
-		// Create an lnd client now that we have the full configuration.
-		// We'll need a basic client and a full client because not all
-		// subservers have the same requirements.
-		var err error
+
+	for {
```
ellemouton, line 716: should we have some kind of back off just? 
Cause at the moment, if something is actually wrong that needs user intervention, this will spin.

ellemouton: also - if LiT is asked to shutdown during this loop - it should be able to exit from here which it currently wont.

terminal.go
```diff
@@ -752,23 +761,34 @@ func (g *LightningTerminal) setUpLNDClients() error {
...
-			BlockUntilChainSynced: true,
-			BlockUntilUnlocked:    true,
-			CallerCtx:             ctxc,
-			CheckVersion:          minimalCompatibleVersion,
-		},
-	)
-	if err != nil {
-		return fmt.Errorf("could not create LND Services client: %v",
-			err)
+	for {
```
ellemouton, line 764: same here re back-off and being able to exit

ellemouton (commented):
> Currently when failing to set up and therefore now retrying to create the lnd client(s), the litcli getinfo + the litcli stop commands are blocked and are not allowed, as we currently require macaroon authentication for those calls. Should I update this PR to allow those calls by adding them to the LiT Whitelist?

We cannot whitelist calls like `litcli stop` since that means that anyone can come along and shutdown our node without a macaroon. 

I think the question is really: should we consider completely decoupling the Lit macaroon DB from LND's macaroon DB? Since then we can start it up without a dependence on LND. 

I think this is worth considering since in future we might want LiT to connect to multiple LND nodes anyway and in that future we would definitely want LiT to have a decoupled macaroon DB. 

Nice PR! left some suggestions.

status/manager.go
ViktorTigerstrom: Good idea, addressed with the latest push!

status/manager.go
```diff
@@ -19,6 +19,10 @@ type SubServerStatus struct {
 	// been started.
 	Running bool
 
+	// Status will be a non-empty string that details the current status of
+	// the sub-server.
+	Status string
```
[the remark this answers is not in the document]
ViktorTigerstrom: I've updated the comment :)! The reason I currently have it duplicate info for "Registered", "Running", "Stopped" and "Errored", is that I thought it probably will be even more confusing for the user that it will else be blank from time to time in the response for the `litcli status` cmd. If you think that I should remove it I will though :). Else I think a perhaps better solution would have been to remove the booleans for the `litrpc.SubServerStatus` struct and only keep this new status field instead, but I'm not a huge fan of removing them for backwards compatibility reasons..

Let me know what you'd prefer :)!

subservers/manager.go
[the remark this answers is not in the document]
ViktorTigerstrom: Thanks a lot for the feedback @ellemouton :fire::rocket:!! I've updated the PR to implement the above, with smaller modifications.

Definitely agree that the new version is much cleaner!

terminal.go
ViktorTigerstrom: Ah yeah :)! Added checking of if we should shutdown before for every step in the loop.

 > should we have some kind of back off just?
 
 Wasn't exactly sure what this was referring to, but I've now added shutdown check + the user has the ability to end the execution with `CTRL+C`. The user also can see what's going wrong with the `litcli status` command, so I think that should be ok?

ViktorTigerstrom:
Thanks a lot for the feedback and for reviewing this so quickly @ellemouton! I've addressed your feedback with the latest push.

status/manager.go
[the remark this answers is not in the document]
ellemouton: So I'd say that since it is called "manual status" it should be clear to the user that this is something special - so im thinking that we could probably just overwrite it to an empty string if any other status gets set.

status/manager.go
```diff
@@ -53,6 +74,39 @@ func NewStatusManager() *Manager {
...
+			req, server.manualStatus,
+		)
+	}
+
+	if !handled || server.Running {
+		return server.Running, false, nil
+	}
+
+	return isReady, false, nil
+}
```
ellemouton: I find this a bit hard to read. 

How about:

```
        if server.disabled {
		return false, true, nil
	}

	// If this system has an override check set, then we first check that
	// to see if it overrides the "ready" status of the system.
	if server.isReadyOverride != nil {
		isReady, handled := server.isReadyOverride(
			req, server.manualStatus,
		)
		if handled {
			return isReady, false, nil
		}
	}

	return server.running, false, nil
```

terminal.go
```diff
@@ -234,8 +234,18 @@ func (g *LightningTerminal) Run() error {
 		return fmt.Errorf("could not create permissions manager")
 	}
 
+	lndOverride := func(uri, manualStatus string) (bool, bool) {
```
ellemouton, line 236: can we add a short comment to this just to describe what we are enabling here?

terminal.go
ellemouton: > Wasn't exactly sure what this was referring to,

What I mean is that this could spin (ie go around the for loop very very fast  if something non-timeout related is erroring). So between iterations, we should add a bit of a wait.

This can be accomplished by moving your `checkRunning` to the end of the for loop and then replacing the default case with a timeout.

terminal.go
```diff
@@ -710,25 +710,55 @@ func (g *LightningTerminal) setUpLNDClients() error {
 		clientOptions = append(clientOptions, lndclient.Insecure())
 	}
 
+	checkRunning := func() error {
+		select {
+		case err := <-g.errQueue.ChanOut():
+			return fmt.Errorf("Error from subsystem: %v", err)
```
ellemouton, line 716: nit: error strings should not be capitalised

status/manager.go
```diff
@@ -80,26 +134,34 @@ func (s *Manager) SubServerStatus(_ context.Context,
...
 
-	ss := newSubServerStatus()
-	ss.Disabled = false
+	s.registerSubServerUnsafe(name, false, opts...)
+}
+
+func (s *Manager) registerSubServerUnsafe(name string, disabled bool,
+	opts ...SubServerOption) {
+
+	ss := newSubServerStatus(disabled)
```
ellemouton: nit: just for future, this refactor along with the renames later on could have been done in 2 commits at the start of the PR. That way, all the commits that follow are mainly additions. 

but yeah - just heads up for future PRs :) no need to change here

ellemouton (commented):
Nice work @ViktorTigerstrom ⭐ 

This is basically there 👍  last major comment is regarding preventing the loops from spinning - once that is addressed - I'll do a final testing round

litrpc/lit-status.proto
```diff
@@ -28,4 +28,7 @@ message SubServerStatus {
     // error describes an error that might have resulted in the sub-server not
     // starting up properly.
     string error = 3;
+
+    // status details the current status of the sub-server.
+    string status = 4;
```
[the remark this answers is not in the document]
ViktorTigerstrom: Hmmm I'm not sure that if I agree that we should rename this to "manual status" as it's surfaced to the end user through the `litcli status` command. IMO that would kinda give users impression that they should somehow be able to manually modify the status of the subserver. This is partly the reason why I added the duplication of the boolean info to this field.
However we should probably name it to something else to that makes it clearer why this is not always set and then remove all the extra duplicate "Registered", "Running" etc.

Perhaps "starting status" could work? Let me know if you have a better suggestion :)?

status/manager.go
ViktorTigerstrom: Thanks, I've updated to your version, with a slight modification to the first if check as we also need to ensure that the system isn't marked as `running` when we let the override function define if the request should be accepted or not. Hope that didn't ruin the readability too much :)!

terminal.go
ViktorTigerstrom: Ah thanks for clarifying :)! I added a timeout for the duration of the `defaultStartupTimeout` to address this! Let me know if you think I should use some other timeout :)!

ViktorTigerstrom:
Thanks again for reviewing this PR so quickly @ellemouton :rocket::fire:!!! I've addressed your latest feedback with the latest push :)

terminal.go
```diff
@@ -718,25 +718,58 @@ func (g *LightningTerminal) setUpLNDClients() error {
...
+		case err := <-g.errQueue.ChanOut():
+			return fmt.Errorf("error from subsystem: %v", err)
+
+		case <-lndQuit:
+			return fmt.Errorf("LND has stopped")
+
+		case <-interceptor.ShutdownChannel():
+			return fmt.Errorf("received the shutdown signal")
+
+		case <-time.After(defaultStartupTimeout):
```
guggero, line 735: If we want this to be the maximum time we wait in total, then we need to create the `time.After()` channel outside of the callback. Otherwise we wait this long for every re-try (since we create a new ticker channel each time).

guggero: Or do we want to wait 5 seconds every time? Wasn't super clear to me from the comment above.

terminal.go
```diff
@@ -764,23 +797,39 @@ func (g *LightningTerminal) setUpLNDClients() error {
...
+	for {
+		g.lndClient, err = lndclient.NewLndServices(
+			&lndclient.LndServicesConfig{
+				LndAddress:            host,
+				Network:               network,
+				TLSPath:               tlsPath,
+				Insecure:              insecure,
+				CustomMacaroonPath:    macPath,
+				CustomMacaroonHex:     hex.EncodeToString(macData),
+				BlockUntilChainSynced: true,
```
guggero, line 809: Since we are passing in `BlockUntilChainSynced: true` here, we don't actually need to re-try ourselves, as that will already initiate a re-try loop, as long as the `ctxc` context we pass in doesn't expire too early.

guggero: Same goes for the `BlockUntilUnlocked: true` I guess. So maybe we could also just try to initiate the full client first, as that already does all the re-trying we want and need and then can initiate the basic client afterwards where we can be pretty certain that it should succeed?

guggero: I guess what we lose that way would be the extra info exposed to the user with the `SetErrored` field. So I'm fine with leaving the order the way it is since it gives the user more information. But I'm fairly certain we don't need the re-try `for` loop here as well.

guggero (commented):
Nice change, will be very helpful to users to know what's going on under the hood!
Nothing major from my end, just a couple of comments around the re-try loop itself.

status/manager.go
```diff
@@ -19,6 +19,12 @@ type SubServerStatus struct {
 	// been started.
 	Running bool
 
+	// manualStatus will be a non-empty string that details the current
+	// status of the sub-server. This status can be set to a unique status
+	// that only exists for the specific sub-server, and will be displayed
+	// to the user with the litrpc.SubServerStatus.
+	manualStatus string
```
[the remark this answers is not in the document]
ViktorTigerstrom: Good idea! Renamed this to customStatus instead :)!

litrpc/lit-status.proto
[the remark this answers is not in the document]
ViktorTigerstrom: Renamed to customStatus as per https://github.com/lightninglabs/lightning-terminal/pull/694#discussion_r1439988650 discussion!

terminal.go
ViktorTigerstrom: > Or do we want to wait 5 seconds every time? Wasn't super clear to me from the comment above.

Yeah this was my intention, so that we don't spin as per @ellemouton's comment https://github.com/lightninglabs/lightning-terminal/pull/694#discussion_r1426517923, but still try to reinitiate the lndclient(s) on failure to set them up.

terminal.go
ViktorTigerstrom: > Since we are passing in BlockUntilChainSynced: true here, we don't actually need to re-try ourselves, as that will already initiate a re-try loop, as long as the ctxc context we pass in doesn't expire too early.

So unfortunately this isn't really the case, and logs we have seen from users, such as @Liongrass, have had the `ctxc` time out during the chain sync of `lnd`, when `lnd` took for example 2mins+ to successfully sync. The `GetInfo` call that's made during the chain sync in the `lndclient` then times out, and therefore the set up of the full client fails. If we would have retried to set up the full `lndclient` during those cases, the `lnd` sub-server in litd would have eventually successfully been started.
One potential alternative solution could have been to increase the `ctxc` timeout, but as there's no specific context just for the chain sync, that would mean that any future RPC call made after the chain sync would also use the updated timeout, which wouldn't be ideal..

This is why we decided to take the approach to attempt to set up the `lndclient` indefinitely until it is successful as a temporary fix, until we implement a more general solution that restarts `litd` sub-servers when required to restart.

ellemouton (approved):
Looks good to me! Contingent on what Oli thinks re the retry loops :) 

Btw - think you need to update your lndclient test branch as the current redirect commit doesnt compile. 

Notes from testing:

- saw this getting logged: ` Setting the lit sub-server as errored: Error %d when creating LND Services client: %v` - gotta make sure that the values are populated in the log. But maybe this is just a factor of the test? Also saw this:
`STAT: Setting the lit sub-server as errored: Error when setting up basic LND Client: %v`
- this is probably not related to this PR but I get this error for taproot-assets: ` "error": "unable to initialize RPC server: unable to start asset custodian: error listing existing addresses: unknown sqlite error: SQL logic error: no such column: asset_version (1)"` have you seen this before? perhaps something we should report to the assets team?

status/manager.go
```diff
@@ -58,6 +77,39 @@ func NewStatusManager() *Manager {
...
+	if server.isReadyOverride != nil && !server.Running {
+		isReady, handled := server.isReadyOverride(
+			req, server.customStatus,
+		)
+		if handled {
+			return isReady, false, nil
+		}
+	}
+
+	return server.Running, false, nil
```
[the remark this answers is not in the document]
ViktorTigerstrom: Sure! Change to your suggested version instead :)

ViktorTigerstrom:
Thanks a lot for the review @ellemouton :tada:!!

I've addressed your feedback with the latest push!

> Btw - think you need to update your lndclient test branch as the current redirect commit doesn't compile.

I've now updated the branch so that it builds and can be used for testing! One thing to note though, Is that I couldn't get it to work with lndclient `v0.17.0-4` for some reason. There seems to be some conflict with taproot-assets that I just couldn't figure out how to solve within a reasonable time. So In the test branch I've downgraded to `v0.17.0-3`, which subsequently downgraded taproot-assets as well. But it can be used to test the functionality of this PR! 

> saw this getting logged:  Setting the lit sub-server as errored: Error %d when creating LND Services client: %v - gotta make sure that the values are populated in the log.

Thanks! This was a pre-existing error that I've now addressed with 3b83d6e1022ce1ee5e99a929d43c88430dec1de6 :)

> this is probably not related to this PR but I get this error for taproot-assets: 

I hadn't seen this, but as I started downgrading the lndclient and therefore the taproot-assets version to make the test branch compile, and then upgraded again and such, I actually managed to trigger this as well! Did you do something similar before the error occurred for you as well?
However, after I removed `testnet` folder (my configured netwok) in the taproot-assets Application data, and restarted litd again, the error disappeared. I suspect it's some migration thing in taproot-assets going on. But as my way of making the error get triggered is probably nothing a normal user will do, I'm not sure if this is something we should actually report. Though it would be really good to get @guggero's opinion on this as well!

terminal.go
guggero: Ah right, good point. Makes sense.

guggero (approved):
Code looks good to me, very nice work!

> One thing to note though, Is that I couldn't get it to work with lndclient v0.17.0-4 for some reason.

Not exactly sure what this is referring to? Because the current version of the PR doesn't modify the `lndclient` version in the `go.mod` file, so it's currently at `v0.17.0-4`. What test branch of `lndclient` were you talking about?

> I suspect it's some migration thing in taproot-assets going on.

Yes, that's very likely it. If the version of taproot assets was changed between versions of PRs, a downgrade would explain the error. So all good I think.

ViktorTigerstrom:
Thanks a lot for the review @guggero :tada::fire:!

> Not exactly sure what this is referring to? Because the current version of the PR doesn't modify the `lndclient` version in the `go.mod` file, so it's currently at `v0.17.0-4`. What test branch of `lndclient` were you talking about?

Ah sorry, I should have been more clear what I was referring to. This is referring to the test branch to test the restart functionality locally, that I mentioned in the original description message of this PR above:  
https://github.com/ViktorTigerstrom/lightning-terminal/tree/2023-12-restart-lndclient-on-failure-mocked-lndclient-fails

> Yes, that's very likely it. If the version of taproot assets was changed between versions of PRs, a downgrade would explain the error. So all good I think.

Ok great, sounds good :rocket:!

PR #695, opened 2023-12-17
fix: add new v2 methods to custom permissions

secondl1ght:
I spoke with @jamaljsr and we noticed that some of the V2 methods are not currently included when setting custom permissions and the deprecated versions are. This would cause any applications using LNC with the V2 methods to get a permissions error even though the user granted permission for these values. Adding these will now support clients that are using the latest LND API versions.

### DIFF
--- a/app/src/util/constants.ts
+++ b/app/src/util/constants.ts
@@ -105,8 +105,10 @@ export const PermissionUriMap: { [key: string]: string[] } = {
     '/lnrpc.Lightning/SendCoins',
     '/lnrpc.Lightning/SendMany',
     '/lnrpc.Lightning/SendPayment',
+    '/lnrpc.Lightning/SendPaymentV2',
     '/lnrpc.Lightning/SendPaymentSync',
     '/lnrpc.Lightning/SendToRoute',
+    '/lnrpc.Lightning/SendToRouteV2',
     '/lnrpc.Lightning/SendToRouteSync',
   ],
   receive: [
--- a/app/src/util/constants.ts
+++ b/app/src/util/constants.ts
@@ -105,10 +105,10 @@ export const PermissionUriMap: { [key: string]: string[] } = {
     '/lnrpc.Lightning/SendCoins',
     '/lnrpc.Lightning/SendMany',
     '/lnrpc.Lightning/SendPayment',
-    '/lnrpc.Lightning/SendPaymentV2',
+    '/routerrpc.Router/SendPaymentV2',
     '/lnrpc.Lightning/SendPaymentSync',
     '/lnrpc.Lightning/SendToRoute',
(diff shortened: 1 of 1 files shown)


guggero (approved):

app/src/util/constants.ts
```diff
@@ -105,8 +105,10 @@ export const PermissionUriMap: { [key: string]: string[] } = {
     '/lnrpc.Lightning/SendCoins',
     '/lnrpc.Lightning/SendMany',
     '/lnrpc.Lightning/SendPayment',
+    '/lnrpc.Lightning/SendPaymentV2',
     '/lnrpc.Lightning/SendPaymentSync',
     '/lnrpc.Lightning/SendToRoute',
+    '/lnrpc.Lightning/SendToRouteV2',
```
jamaljsr, line 111: These methods are in the `routerrpc` package, so the uris need to be `/routerrpc.Router/SendPaymentV2` and `/routerrpc.Router/SendToRouteV2`.

jamaljsr (changes requested):
Tested this branch and it's not working correctly. After looking into it, i've found the issue.

app/src/util/constants.ts
guggero: Ouch... How did I miss that? Shouldn't do drive-by reviews early in the morning :see_no_evil:

jamaljsr: 😂

I would have missed it too.

secondl1ght: Thanks for testing! I wonder why `SendPayment` is working then because it's also in the `routerrpc` package but is set as `lnrpc` here... 🤔 Or is that also not working?

guggero: No, there are actually 3 `SendPayment` RPC methods... The oldest in `lnrpc` which still can be used but lacks features, the intended replacement in `routerrpc` that was deprecated even while the `routerrpc` package was still labelled as experimental, and then the one that everybody should use: `SendPaymentV2` in `routerrpc`.
Unfortunately that's what you get if you promise your users to never break RPC methods, you need to write new ones if you want to change things...

secondl1ght: Ah ok got it, do you want me to also add the ` /routerrpc.Router/SendPayment` one then?

guggero: No, since that should definitely not be used by anyone.

secondl1ght: It is updated now - sorry I should have noticed that this was under the wrong path I think but I just assumed it was the same as the original one.

jamaljsr (approved):
tACK LGTM 🚀

Thanks for the update. Confirmed this works as expected.

PR #699, opened 2023-12-19
linter: re-enable the linter

guggero:
The linter was disabled before due to a bug. Since the bug was likely
caused by an issue with a go module in a dependency project which was
updated in the meantime, we attempt to re-enable the linter in the CI
now.


ViktorTigerstrom (approved):
Nice 🎉! This will be really helpful!

ellemouton (approved):
Yaaaayyyy 🔥 just needs a rebase

PR #718, opened 2024-02-01
accounts + main: Allow new accounts with 0 balance

ViktorTigerstrom:
This PR updates the `accounts` subsystem to allow creation of accounts with an initial balance set to 0.
As accounts can now receive balance through settled invoices during the lifecycle of the account, it makes sense that we'd allow creation of accounts with an initial balance of 0.


ellemouton (approved):
lgtm!

PR #746, opened 2024-04-16
firewall: session-dependent privacy flags

bitromortac:
This PR adds functionality for the autopilot privacy mapper that enables to selectively reveal data that otherwise is obfuscated by default. This is necessary in order to support more advanced autopilot features, such as the soon to be added automatic channel open service. In this case it is necessary to communicate litd's node pubkey to the service (via `GetInfo`) in order for it to be able to make high-quality node suggestions, taking the nodes's position within the graph into account. Uncovering pubkeys will also be necessary in order to know the peers with which force closes have happened, to not open new channels, or to check whether channels with liquidity already exist. It is nevertheless important to still obfuscate precise channel balances and other information. This is why **privacy flags** are introduced (in the auto open example it is the `ClearPubkeys` flag), which define the precise deobfuscation within a session for the following interaction points:

* Feature configuration:
  Features can be configured (not used yet by any feature) to define algorithmic behavior, for example one can imagine to specify peers to preferentially open channels with. Currently, those node ids are obfuscated, but for automatic channel opens it will be important to know the identity of the nodes in order to respect the user's preference.
  
* Feature rules:
  Feature rules define the boundaries within which autopilot is allowed to operate. This includes a peer restriction rule. In the example of automatic channel opens, the service needs to know the pubkeys for nodes not to open channels with.
  
* Firewall intercepted calls from autopilot to litd:
  As explained above, certain APIs need to reveal the pubkeys of peers in order for the autopilot to function correctly.

Privacy obfuscation remains on by default and `AutoFee`'s privacy properties remain unaltered. Features can be registered together in a single session, whereby the session's privacy flags will be the union of privacy flags of all the features (which is why feature registration is recommended to be done in individual sessions). When linking sessions, privacy flags must be preserved accross the linked sessions, which is checked by autopilot.


ViktorTigerstrom (approved):
tACK LGTM 🚀🔥🚀!!! Really awesome work :tada::tada::tada:!!!

session_rpcserver.go
```diff
@@ -925,6 +925,52 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
+			// server.
+			autopilotFeature, ok := autopilotFeatureMap[f]
+			if !ok {
+				return nil, fmt.Errorf("%s is not a features "+
+					"provided by the Autopilot server", f)
+			}
+
+			// Deserialize and check that we know the privacy flags.
+			featurePrivacyFlags, err := session.Deserialize(
+				autopilotFeature.PrivacyFlags,
```
ellemouton, line 961: ok so I think if we are gonna do the parsing here (and not in the autopilot ListFeatures itself), then we should at least also parse these in `ListFeatures` so that we can set the `RequiresUpgrade` flag accordingly to also be true if the feature has a privacy bit we dont understand. That way the front end can also more easily indicate to the user that they will not be able to register for this feature instead of letting them only find out when they actually attempt to register for the feature.

cmd/litcli/autopilot.go
```diff
@@ -129,6 +129,27 @@ var addAutopilotSessionCmd = cli.Command{
 				`An empty rule map is allowed with {} to ` +
 				`use the default rules.`,
 		},
+		cli.Int64Flag{
```
ellemouton, line 132: wondering how useful this actually is cause the user would need to manually go workout the resulting int. 

How about instead having a flag per `Clear**` `PrivacyFlag` type that we have at the moment and then construct the int on behalf of the user?

ellemouton: so that I can do like `--clear-pubkeys --clear-htlcs` instead of needing to figure out what each int & then doing a manual bitwise OR

session/interface.go
```diff
@@ -130,6 +130,11 @@ func NewSession(id ID, localPrivKey *btcec.PrivateKey, label string, typ Type,
 	return sess, nil
 }
 
+// PrivMapDetails details how privacy mapping should be done.
```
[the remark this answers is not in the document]
bitromortac: yes, agree looks better :tada:! I moved the interface into `firewalldb` (I think I will move it to `firewall` after the next round) as well as the mock for it (but could also live in `session`?). I think that `GetGroupID(sessionID ID) (ID, error)` in `IDToGroupIndex` is now redundant, should I remove it?

session_rpcserver.go
```diff
@@ -1403,7 +1404,10 @@ func (s *sessionRpcServer) marshalRPCSession(sess *session.Session) (
 						db := s.cfg.privMap(
 							sess.GroupID,
 						)
-						val, err = val.PseudoToReal(db)
+						val, err = val.PseudoToReal(
+							db,
+							session.PrivacyFlags{},
```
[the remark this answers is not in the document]
bitromortac: was just an intermediate change, inserted the zero value in this commit for later replacement :+1:

firewall/privacy_mapper.go
```diff
@@ -935,6 +935,14 @@ func ObfuscateConfig(db firewalldb.PrivacyMapReader, configB []byte) ([]byte,
 			// Try to replace with a chan point.
 			_, _, err := firewalldb.DecodeChannelPoint(value)
 			if err == nil {
+				// We don't obfuscate channel points if the flag
```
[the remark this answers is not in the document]
bitromortac: great call :pray:! have changed it and added a test

session_rpcserver.go
bitromortac: great point, added!

cmd/litcli/autopilot.go
bitromortac: right, I had the same thought, but decided against it because it adds a lot of noise to the command documentation and because of reasoning the same way we did previously, thinking it would be mostly used for testing. Another way could be to parse it from the string representation like `--privacy-flags="ClearPubkey|ClearAmounts"`. I would add one of the solutions if you find it important

firewall/privacy_mapper.go
```diff
@@ -950,6 +967,23 @@ func ObfuscateConfig(db firewalldb.PrivacyMapReader, configB []byte) ([]byte,
...
+
+				// We replace the pubkey with a random value,
+				// should we already have it.
+				if obfVal, ok := alreadyHave(); ok {
+					obfuscatedValues[i] = obfVal
+
+					continue
+				}
+
+				// Otherwise we create a new mapping.
```
ellemouton, line 1054: nice 🙏

firewall/privacy_mapper_test.go
```diff
@@ -880,6 +921,12 @@ func TestObfuscateConfig(t *testing.T) {
 			if !tt.notExpectSameLen {
 				require.Equal(t, len(tt.config), len(config))
 			}
+
+			// We expect the config to be unobfuscated if we have
+			// the corresponding privacy flag.
```
ellemouton, line 980: 🚀

firewall/privacy_mapper.go
```diff
@@ -240,64 +244,69 @@ func (p *PrivacyMapper) replaceOutgoingResponse(ctx context.Context, uri string,
...
+					if err != nil {
+						return err
+					}
+
+					return nil
+				},
+			)
+			if err != nil {
+				return nil, err
+			}
```
ellemouton, line 309: nit: can save some vertical space:

```
			err := db.Update(
				func(tx firewalldb.PrivacyMapTx) error {

					var err error
					pseudoPubKey, err = firewalldb.HideString(
						tx, r.IdentityPubkey,
					)

					return err
				})
			if err != nil {
				return nil, err
			}
```

session_rpcserver.go
```diff
@@ -810,13 +810,21 @@ func (s *sessionRpcServer) ListAutopilotFeatures(ctx context.Context,
 			return nil, err
 		}
 
+		// We check whether we should upgrade because of unknown privacy
+		// flags.
+		_, err = session.Deserialize(f.PrivacyFlags)
+		if errors.Is(err, session.ErrUnknownPrivacyFlag) {
+			upgrade = true
+		}
```
ellemouton, line 818: should we return the error otherwise? if we cant parse it here, it will error later on anyways. 
So either should:
1) return the error here or
2) log it and make upgrade = true  so that user doesnt go and try to register to this feature since it will fail anyways right?

cmd/litcli/autopilot.go
ellemouton: cool yeah the "flags="ClearPubkey|ClearAmounts" seems nice. Even if this is just for testing, it is nice if it is easy to quickly get something going (bitwise ORing manually is a bit much)

ellemouton: what's nice about your idea to do the "flags="ClearPubkey|ClearAmounts"" way is also that we dont have to keep updating the CLI code each time we add a new one

itest/litd_firewall_test.go
```diff
@@ -600,19 +557,11 @@ func testSessionLinking(net *NetworkHarness, t *harnessTest) {
...
-	require.NoError(t.t, err)
-
-	pilotConn, metaDataInjector, err = connectMailboxWithRemoteKey(
-		ctx, pilotPriv, litdPub,
+	lndConn, metaDataInjector, cleanup2 := newAutopilotLndConn(
+		ctx, t.t, net, sessResp2.Session,
 	)
-	require.NoError(t.t, err)
-	defer pilotConn.Close()
+	defer func() { require.NoError(t.t, cleanup2()) }()
```
ellemouton, line 855: can we not instead use `t.t.CleanUp()`?

rules/manager_set.go
```diff
@@ -31,7 +31,8 @@ func (m ManagerSet) InitEnforcer(cfg Config, name string,
 
 	mgr, ok := m[name]
 	if !ok {
-		return nil, ErrUnknownRule
+		return nil, fmt.Errorf("%w %s, please upgrade", ErrUnknownRule,
```
ellemouton, line 34: cool - nice change 🙏

ellemouton (approved):
🔥 🚀  LGTM!

itest/litd_firewall_test.go
bitromortac: cool, TIL :)

cmd/litcli/autopilot.go
bitromortac: I implemented and tested string-based flags :+1:

session_rpcserver.go
bitromortac: I returned the error :+1:

PR #750, opened 2024-04-25
Bump litd to version `v0.12.5-alpha`

ViktorTigerstrom:
This PR bumps the litd version to `v0.12.5-alpha`.

This PR also bumps lnd to version `v0.17.5-beta` and loop to `v0.28.1-beta`.

We've not yet decided if we'll wait for #746 to be merged before the release, so I'm therefore pushing this as draft until we've made a decision regarding that.


ViktorTigerstrom:
We've decided to cut the release before merging #746.

I've tested this locally with the terminal web UI, and the Loop & Pool flow worked as expected for me. I've also tested the Autopilot flow locally.

ellemouton (approved):
LGTM! 🎸

ViktorTigerstrom:
Updated the PR with a commit that updates the wording in the Compatibility section in the README, to make it clearer what the first table details.

I'm just re-requesting a review from you @ellemouton, to verify that you agree with the new wording :)!

README.md
```diff
@@ -66,16 +66,18 @@ If you’d prefer to compile from source code please follow [these instructions]
...
-version of `lnd`. View the table below to ensure that you run the correct
+Full Lightning Terminal functionality can be dependent on running a compatible
+version of `lnd`. View the table below to ensure that you run a compatible
 version of `lnd` with the relevant `litd` release.
 
-This version compatibility dependency is only relevant for running remote mode
-`litd`. The bundled version will always come with the correct, [compatible
-versioning](#daemon-versions-packaged-with-lit).
+This version compatibility dependency table is only relevant when running `lnd`
+in remote mode, and shows the **_minimum_** `lnd` version that can be used with
```
ellemouton, line 74: "when running `lnd` in remote mode" -> technically LND cant be run in remote mode. It is a mode of "litd". So perhaps just "when running `litd` in remote mode" or if we want to be more descriptive: "when running `litd` in remote mode (meaning that `lnd` is running as a separate process to `litd`)?

ellemouton, line 74: "mode, and shows..." -> "mode. It shows the ..."

ellemouton (commented):
Just a couple of minor suggestions

README.md
ViktorTigerstrom: Thanks for the good clarification suggestion :)! I updated this, but changed the "(meaning that ... )" section a little to:
"This version compatibility dependency table is only relevant when running litd in remote mode (meaning that `lnd-mode=remote` is set)."

I think that should be pretty clear to most users. The only con with my new version is that we don't mention the operating modes of Litd until after the table, but I think it should be understandable either way.

If we would like to address that con though, I think it would also make sense to move the entire section of "Full Lightning Terminal functionality can be dependent on running a compatible version of `lnd` ..." + the version compatibility dependency table. I'd suggest we'd then move it so it's written after the "supported combinations table".

Let me know if you think the current version is ok, or if you'd like me to move the section!

ellemouton (approved):
lgtm!

guggero (approved):
LGTM :tada:

PR #752, opened 2024-04-26
rules+firewall: rules and privacy mappings for channel creation

bitromortac:
Based on #746

This PR adds functionality to litd in order to make automatic channel opens possible.

We add two new rules:
* `ChannelConstraint`: to limit channel sizes
* `OnChainBudget`: to limit onchain interactions with a per session budget
* update the peer restriction rule to be active for channel opening endpoints

Automatic channel creation involves the following additional endpoints:
* `WalletBalance`: check whether enough funds are available
* `PendingChannels`: check if a channel is already pending with a node
* `ClosedChannels`: for peer sanity checks
* `ConnectPeer`: connect to a peer before opening
* `BatchOpenChannel`: the request for opening


firewall/privacy_mapper.go
```diff
@@ -812,6 +832,405 @@ func handleUpdatePolicyResponse(db firewalldb.PrivacyMapDB,
...
+	flags session.PrivacyFlags,
+	_ func(int) (int, error)) func(ctx context.Context,
+	r *lnrpc.BatchOpenChannelRequest) (proto.Message, error) {
+
+	return func(_ context.Context, r *lnrpc.BatchOpenChannelRequest) (
+		proto.Message, error) {
+
+		err := db.Update(func(tx firewalldb.PrivacyMapTx) error {
+			// Modify channels in place.
+			for _, c := range r.Channels {
```
bitromortac, line 1112: A general question here is whether we should take the override vs creation of a new struct approach. Could it be an advantage for forward compatibility if we don't have to manually add new fields that may be added to the api? Of course there's also a privacy tradeoff here as well.

firewall/privacy_mapper.go
```diff
@@ -1087,6 +1094,80 @@ func handleClosedChannelsResponse(db firewalldb.PrivacyMapDB,
...
+						tx, c.NodePubkey,
+					)
+					if err != nil {
+						return err
+					}
+				}
+			}
+
+			return nil
+		})
```
ellemouton, line 1129: a thought popped up while looking at this commit: I wonder if we'd perhaps want to add a "Private chans only" option to the ChannelConstraint? for privacy reasons

ellemouton (commented):
First pass - looking soooo good 🔥  very clean.

Gonna need a second pass to just really get into the budget logic. 

Main suggestion on first pass is to perhaps add s "private only" restriction to open-chan constraints?

firewall/privacy_mapper.go
```diff
@@ -1168,6 +1173,46 @@ func handleBatchOpenChannelResponse(db firewalldb.PrivacyMapDB,
...
+func handleConnectPeerRequest(db firewalldb.PrivacyMapDB,
+	flags session.PrivacyFlags) func(ctx context.Context,
+	r *lnrpc.ConnectPeerRequest) (proto.Message, error) {
+
+	return func(_ context.Context, r *lnrpc.ConnectPeerRequest) (
+		proto.Message, error) {
+
+		err := db.View(func(tx firewalldb.PrivacyMapTx) error {
+			var err error
+			// Note, this only works if the pubkey alias was
```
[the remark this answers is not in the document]
bitromortac: right, clarified it :)

firewall/privacy_mapper_test.go
```diff
@@ -526,6 +527,23 @@ func TestPrivacyMapper(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:    "ConnectPeer Request",
```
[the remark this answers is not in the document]
bitromortac: added :+1:

firewall/privacy_mapper.go
bitromortac: good idea! I added two fields to the rule, `PrivateAllowed` and `PublicAllowed` to be able to specify exactly which channel types are allowed, because one may want one or the other

rules/channel_constraints.go
```diff
@@ -0,0 +1,360 @@
...
+	// channel opening.
+	MaxPushSat uint64 `json:"max_push_sat"`
+
+	// PrivateAllowed allowed indicates that the constraint should be
+	// applied to private channels.
+	PrivateAllowed bool `json:"private_allowed"`
+
+	// PublicAllowed allowed indicates that the constraint should be
+	// applied to public channels.
+	PublicAllowed bool `json:"public_allowed"`
```
bitromortac, line 102: docstrings are not accurate, will fix in the next round

firewall/privacy_mapper.go
```diff
@@ -812,6 +808,89 @@ func handleUpdatePolicyResponse(db firewalldb.PrivacyMapDB,
...
+			)
+		}
+
+		return &lnrpc.WalletBalanceResponse{
+			TotalBalance:       int64(totalBalance),
+			ConfirmedBalance:   int64(confirmedBalance),
+			UnconfirmedBalance: int64(unconfirmedBalance),
+			LockedBalance:      int64(lockedBalance),
+			ReservedBalanceAnchorChan: int64(
+				reservedBalanceAnchorChan,
```
ellemouton, line 893: i think all these int64 conversions are redundant (the types are already int64s)

firewall/privacy_mapper.go
```diff
@@ -812,6 +808,89 @@ func handleUpdatePolicyResponse(db firewalldb.PrivacyMapDB,
...
+
+		reservedBalanceAnchorChan, err := maybeHideAmount(
+			flags, randIntn, r.ReservedBalanceAnchorChan,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		accountBalance := r.AccountBalance
+		if !flags.Contains(session.ClearAccountBalances) {
```
ellemouton, line 879: in what scenario would the "clear account balances" be useful?

ellemouton: or rather: why is that separate from "ClearAmounts"?

ellemouton: just checking for understanding: do we ever expect to have the WalletBalances API available for a feature where the clear amounts flag is not set?

rules/peer_restrictions_test.go
```diff
@@ -150,6 +150,49 @@ func TestPeerRestrictCheckRequest(t *testing.T) {
 		},
 	)
 	require.NoError(t, err)
+
+	// Test that we can open a channel to a non-restricted peer.
```
ellemouton, line 154: nice nice

rules/channel_constraints.go
```diff
@@ -0,0 +1,354 @@
...
+		return nil, fmt.Errorf("incorrect RuleValue type %T",
+			value.Value)
+	}
+
+	channelBounds := rv.ChannelConstraint
+
+	return &ChannelConstraint{
+		MinCapacitySat: channelBounds.MinCapacitySat,
+		MaxCapacitySat: channelBounds.MaxCapacitySat,
+		MaxPushSat:     channelBounds.MaxPushSat,
```
ellemouton, line 71: missing the new public/private allowed?

ellemouton: i think both are false here meaning no channels allowed?

should we make the default (false) => allowed & then rather have "NoPublic"/"NoPrivate"?

ellemouton: does a call to OpenChan work with this as is given that both are false here?

ellemouton (commented):
Really great work! Pretty much g2g I think. One main question though:

How does session linking worth with the on-chain budget? Does the same budget carry over & do we prevent them from providing a new one? or do we update the budget when a session is linked?

bitromortac:
> Really great work! Pretty much g2g I think. One main question though:
> 
> How does session linking worth with the on-chain budget? Does the same budget carry over & do we prevent them from providing a new one? or do we update the budget when a session is linked?

Thank you for the reviews :pray:! It's possible to change the budget, it's used as a mechanism to top up funds for a session, so we don't prevent it. We could enforce it that it's only allowed to be increased though.

firewall/privacy_mapper.go
bitromortac: I removed `ClearAccountBalances` as it is too specific and changed it to obfuscate the amounts instead.

> just checking for understanding: do we ever expect to have the WalletBalances API available for a feature where the clear amounts flag is not set?

I'm not sure about that, it could maybe be useful for liquidity management, but hard to say

rules/channel_constraints.go
bitromortac: > does a call to OpenChan work with this as is given that both are false here?

I think `UnmarshalRuleValues` is only used for sanity checking upon session registration, but the actual used final rule values are the ones from the server, so even if those are false, channel openings work.

rules/onchain_budget.go
```diff
@@ -0,0 +1,613 @@
...
+		"/lnrpc.Lightning/OpenChannelSync": mid.NewFullChecker(
+			&lnrpc.OpenChannelRequest{},
+			&lnrpc.ChannelPoint{},
+			handleOpenChannelRequest,
+			func(ctx context.Context,
+				r *lnrpc.ChannelPoint) (proto.Message, error) {
+
+				return nil, o.handlePaymentConfirmed(ctx)
+			},
+			handleOpenChannelError,
```
ViktorTigerstrom, line 243: I'm not sure that this is something we actually can/want to handle, but I thought I'd just point to a potential issue here with using the temp store in between requests and their responses, that could potentially allow on-chain spending above the rule bounds:

Since the temp store is cleared on every restart of litd, we MUST process the response for every successful request in order for it to actually be saved permanently and therefore also be used in the calculation for the current spent on-chain budget after litd is restarted. 

Now, if either:
1. litd is shutdown after a successful request has been sent, but before we've had time to process the response
2. The processing of the response errors (causing a shutdown)

The spent on-chain amount will be lost and will not be used for the future on-chain budget calculations once litd is restarted.

Like I said though, I'm not sure if these edge cases is something we should address, but I just to make you aware of it.

rules/onchain_budget_test.go
```diff
@@ -0,0 +1,281 @@
...
+		name      string
+		values    *OnChainBudget
+		expectErr bool
+	}{
+		{
+			name: "within bounds",
+			values: &OnChainBudget{
+				AbsoluteAmtSats: 2000,
+				MaxSatPerVByte:  40,
+			},
```
ViktorTigerstrom, line 31: A general comment for the bounds tests: IMO when we're testing bounds, it makes sense to also have a test that sets these values at the exact bounds i.e. `1000` & `30`. Then the tests, like "amount too low" below should lower the amount by the least possible value (i.e. `999`) and keep the `MaxSatPerVByte` at the exact limit.

That way the unit tests will be much more precise and catch potential bugs where the code uses a `<=` comparison instead of a `<` etc.

Hope that makes sense?

I do think that it does make sense to keep the current "within bounds" test though, just to prove that the values don't need to be at the exact bounds :)

rules/onchain_budget_test.go
```diff
@@ -0,0 +1,281 @@
...
+	// new request but with a different request id. First ensure that an
+	// amount that would exceed the current budget would fail.
+	cfg.ReqID = 2
+	_, err = enf.HandleRequest(
+		ctx, "/lnrpc.Lightning/OpenChannel", &lnrpc.OpenChannelRequest{
+			LocalFundingAmount: 15,
+		},
+	)
+	require.Error(t, err)
+	assertSpentAmt(30, 60)
```
ViktorTigerstrom, line 226: nit:
I think it maybe would make sense to prove here that if the previous requests that keeps `60` of the `pending_balance` fails, this or a similar request succeeds when attempted once more.
That's just to prove that it **_is_** intended behaviour that when processing concurrent requests the order of the requests does actually matter for the outcome.

Feel free to ignore this if you think this is unnecessary coverage though :)!

rules/channel_constraints_test.go
```diff
@@ -0,0 +1,205 @@
...
+		expectErr error
+	}{
+		{
+			name: "between bounds",
+			rule: &ChannelConstraint{
+				MinCapacitySat: 6,
+				MaxCapacitySat: 6,
+				MaxPushSat:     6,
+				PublicAllowed:  true,
+			},
```
ViktorTigerstrom, line 46: Same here, I think it makes sense to include testing of the exact bounds :)!

rules/peer_restrictions_test.go
```diff
@@ -150,6 +150,49 @@ func TestPeerRestrictCheckRequest(t *testing.T) {
...
+			},
+			{
+				NodePubkey: peerKey3[:],
+			},
+		},
+	}
+	_, err = enf.HandleRequest(
+		ctx, "/lnrpc.Lightning/BatchOpenChannel", batchOpenReq,
+	)
+	require.ErrorContainsf(t, err, "illegal action on peer in peer ", "")
```
ViktorTigerstrom, line 195: I think it makes sense to include test coverage of a `BatchOpenChannelRequest` that actually does succeed.

rules/chan_policy_bounds_test.go
```diff
@@ -368,6 +368,64 @@ func TestChannelPolicyBoundsCheckers(t *testing.T) {
...
+					{
+						UseBaseFee:  true,
+						BaseFee:     6,
+						UseFeeRate:  true,
+						FeeRate:     6000000,
+						MinHtlcMsat: 2,
+					},
+				},
+			},
+			expectErr: true,
```
ViktorTigerstrom, line 427: Same here, it would be nice with coverage of a `BatchOpenChannelRequest` that does succeed!

firewall/privacy_mapper_test.go
```diff
@@ -302,13 +353,224 @@ func TestPrivacyMapper(t *testing.T) {
...
+				UnconfirmedBalance:        950_100,
+				LockedBalance:             950_100,
+				ReservedBalanceAnchorChan: 950_100,
+				AccountBalance: map[string]*lnrpc.WalletAccountBalance{
+					"first": {
+						ConfirmedBalance:   950_100,
+						UnconfirmedBalance: 950_100,
+					},
+				},
+			},
```
ViktorTigerstrom, line 393: Nit:
Would be great with also providing test coverage with for a the `/lnrpc.Lightning/WalletBalance` but with the  `ClearAmounts` flag set!

firewall/privacy_mapper_test.go
```diff
@@ -302,13 +353,224 @@ func TestPrivacyMapper(t *testing.T) {
...
+		},
+		{
+			name:    "PendingChannels Response",
+			uri:     "/lnrpc.Lightning/PendingChannels",
+			msgType: rpcperms.TypeResponse,
+			msg: &lnrpc.PendingChannelsResponse{
```
ViktorTigerstrom, line 399: Same here, would be awesome with test coverage where the privacy flags are set!

firewall/privacy_mapper_test.go
```diff
@@ -302,13 +353,224 @@ func TestPrivacyMapper(t *testing.T) {
...
+							Memo:      "something",
+						},
+					},
+				},
+			},
+		},
+		{
+			name:    "ClosedChannels Response",
+			uri:     "/lnrpc.Lightning/ClosedChannels",
+			msgType: rpcperms.TypeResponse,
```
ViktorTigerstrom, line 440: I think we definitely should add coverage here for a test with the new `ClearClosingTxIds` flag set!

firewall/privacy_mapper.go
```diff
@@ -812,6 +832,471 @@ func handleUpdatePolicyResponse(db firewalldb.PrivacyMapDB,
...
+func handleBatchOpenChannelRequest(db firewalldb.PrivacyMapDB,
+	flags session.PrivacyFlags) func(ctx context.Context,
+	r *lnrpc.BatchOpenChannelRequest) (proto.Message, error) {
+
+	return func(_ context.Context, r *lnrpc.BatchOpenChannelRequest) (
+		proto.Message, error) {
+
+		var reqs = make([]*lnrpc.BatchOpenChannel, len(r.Channels))
+
+		err := db.Update(func(tx firewalldb.PrivacyMapTx) error {
```
ViktorTigerstrom, line 1130: nit: I think these and similar requests should work with `db.View` to make it clear that nothin is muted in the db?

ViktorTigerstrom (commented):
Really nice PR :fire::rocket:!! Just like Elle says, I think this is pretty much ready to go, so great work :tada:!

I mainly added some test coverage request comments, and also commented one potential issue for the `onchain_budget` handling. Let me know what you think about that potential issue. But otherwise this looks great :fire:!

rules/onchain_budget_test.go
bitromortac: Makes sense :+1:!

firewall/privacy_mapper_test.go
bitromortac: Added!

rules/onchain_budget.go
bitromortac: Yes, that is (a minor) issue. I'm not sure how that would be solved, because one would need a message buffer on lnd that would persist them in order for litd to be able to ack them? So unless we don't find an easier solution, I'd suggest to keep it this way.

rules/onchain_budget_test.go
bitromortac: Thanks for thinking about these edge cases :pray:! I have added a test in the end (also noted that there could be issues with accounting if the URI was changed for errors), hope I understood you correctly concerning the edge case.

ViktorTigerstrom (approved):
LGTM, great work :fire::rocket::fire:!

Question, do we want to merge this before deciding to marge automatic channel opens, in case we'd find out that more functionality is needed?

rules/onchain_budget.go
ViktorTigerstrom: Yeah, I agree! Just thought it be worth pointing out.

rules/onchain_budget.go
```diff
@@ -0,0 +1,613 @@
+package rules
```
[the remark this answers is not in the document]
bitromortac: added :+1:

bitromortac:
Thanks a lot for the suggestions and the itest @ellemouton :pray: :100:.

I removed the `OpenChannel` streaming RPC functionality, which was not fully built out and added privacy mapping for `OpenChannelSync`.

During itests I found that there was an issue when combining other rules with the budget rule, which is stateful. It may happen that a request adds a temporary budget reservation while another rule is violated, resulting in the reserved budget not being removed. I tried to fix this by adding a `RollbackRequest` method to the `Enforcer` interface, which is called after any rule is violated for a request. In principle we would need to add something similar to commit a balance reservation after knowing that all *response* rules were followed. We don't have restrictions on responses right now which is why I didn't add that.

rules/onchain_budget.go
```diff
@@ -0,0 +1,603 @@
...
+		}
+
+		// Check that spending amt would not exceed the budget given the
+		// amount already spent as well as any pending amount.
+		if spent.Amount+pending.Amount+request.Amount >
+			o.AbsoluteAmtSats {
+
+			return fmt.Errorf("action with %v sat exceeds budget "+
+				"limit %v sat (%v sat used, %v spent, %v pending)", request.Amount,
+				o.AbsoluteAmtSats, spent.Amount+pending.Amount, spent.Amount, pending.Amount)
```
bitromortac, line 397: forgot to break those lines

rules/onchain_budget.go
ellemouton: this is a good point and I do think we should consider handling it. In the accounts system, we are "restart" proof so probably worth trying to be restart proof here too if possible. 

for the accounts system, we can do this quite easily cause we know the payment hash to track before sending the request through, so then on startup, we call LND to see if the payment with that hash did in fact go through. 

So is there any way here that we can force there to be some identifier? For example, I see that `OpenChannelRequest` has an optional `Memo string` field and same for `BatchOpenChannel. Can we not force the caller to set these (to something unique that we have not seen before) and then on startup, can we then not list channels & pending channels and then subscribe to channel open events to continue tracking these? For example I see that the Channel open event does contain the memo field. 

If we do this then we also would _not_ use the temp store. Instead we would have some kind of permanent "memo -> amount" map that we can check on startup. 

thoughts?

ellemouton:
re-requesting from @ViktorTigerstrom  since quite a bit has changed

rules/onchain_budget.go
bitromortac: I explored the idea of using the memo to keep track of pending channel opens (see commits after the itest). A missing piece is to add the memo field to the `ClosedChannels` api. With that we could then delete any pending actions for which we don't find an outcome (so that must have either been some internal failure to open the channel or a channel open was abandoned). In my opinion, the complexity is not really worth it. I think we should not use the temporary store, but persist all pending actions and remove them once we get a response/error. This way we are conservative and don't allow overspending. In the worst case this leads to a decreased budget, which may lead to UX issues, but I would consider a session to be something ephemeral and a possible problem can be fixed by just creating a new (non-linked) one.

rules/onchain_budget.go
```diff
@@ -0,0 +1,687 @@
...
+
+	reqID := o.GetReqID()
+	reqPref := fmt.Sprintf("%s:%d", onChainPending, reqID)
+
+	b, err = json.Marshal(request)
+	if err != nil {
+		return err
+	}
+
+	return store.Set(ctx, reqPref, b)
```
ViktorTigerstrom, line 673: nit:
Even though lnd should never create another request with the same request ID, should we check if an entry of this key already exists in the Store? Given that there will be pretty bad consequences if we'd overwrite an already existing entry.

rules/onchain_budget.go
```diff
@@ -0,0 +1,687 @@
...
+
+		return r, nil
+	}
+
+	// handleOpenChannelError is a helper function which handles the error
+	// returned for all the channel opening endpoints.
+	handleOpenChannelError := func(ctx context.Context, _ error) (error,
+		error) {
+
+		return nil, o.cancelPendingPayment(ctx)
```
ViktorTigerstrom, line 253: This shouldn't return `nil` for the first return value, right? It should be set to the `error` passed as the second arg to the function, given that `o.cancelPendingPayment(ctx)` doesn't error.

rules/channel_constraints.go
```diff
@@ -0,0 +1,356 @@
...
+//
+// NOTE: this is part of the Enforcer interface.
+func (e *ChannelConstraint) HandleErrorResponse(_ context.Context, _ string,
+	err error) (error, error) {
+
+	return err, nil
+}
+
+// RollbackRequest reverts the enforcer's state after errors.
+func (e *ChannelConstraint) RollbackRequest(_ context.Context, _ string) error {
```
ViktorTigerstrom, line 152: nit: Should be removed?

itest/litd_firewall_test.go
```diff
@@ -891,6 +942,401 @@ func testSessionLinking(net *NetworkHarness, t *harnessTest) {
 	assertStatusErr(t.t, err, codes.ResourceExhausted)
 }
 
+// testChannelOpening tests that a client connected via a firewall can open
+// channels to a peer respecting the channel rules set by the firewall.
+func testChannelOpening(net *NetworkHarness, ht *harnessTest, t *testing.T) {
```
ViktorTigerstrom, line 947: Niiice :fire::rocket::fire:!

ViktorTigerstrom (commented):
I think this looks great! Left a few minor review comments. I agree with the choice that it's better that this could now error on the underspend side on Litd restarts, instead of overspending, given that the use of the Memo field won't work properly until `closedchannels` contains the field. The only thing I dislike a bit with the currently implemented approach is that it'll lead to on-chain balance spends being marked as "pending" forever if the restart edge-case occurs, which feels pretty hacky, but I think that's an ok trade-off. If litd is restarted, can we ever receive a response from `lnd` after the restart that "confirms" the on-chain spend? Else it might make sense to mark all pending spends as confirmed spends on restarts of litd. I'm interested to hear @ellemouton's perspective on this choice as well! Given that this is the approach we'd like to take, I think it makes sense to remove the last commit, until that approach could actually be used.

Will go ahead with final testing, once we've all agreed on this approach!

> As another safety measure, Viktor suggested that we could block shutdown with a timeout if we haven't gotten any response from lnd yet, but I'm not sure how to do that at the moment.

Yes, If we could combine this with the currently implemented approach, I think that would mitigate the edge case of restarting lnd before the response has been processed, so sufficiently that it should hopefully become a non issue for the vast majority of the users! :rocket:

rules/onchain_budget.go
bitromortac: Ah, ok, understand now better how the request id is created (by lnd), added a check.

rules/onchain_budget.go
bitromortac: right, great catch! :pray:

rules/onchain_budget.go
ellemouton: > In my opinion, the complexity is not really worth it. I think we should not use the temporary store, but persist all pending actions and remove them once we get a response/error

agreed. but we should have some identifier persisted with them that is restart safe (ie, not the request ID as used today). We want to make sure that whatever we do now we can potentially recover from in future.  

> but I would consider a session to be something ephemeral and a possible problem can be fixed by just creating a new (non-linked) one.

ah yes that is a good point!

ellemouton (commented):
Cool cool cool - looking good!

so are we moving forward with the pending-chan-ID as discussed offline? 

One thing is that we potentially will want to keep a list of "already-tried-pending-IDs" so that we never re-use an old one

firewall/rule_enforcer.go
```diff
@@ -236,21 +237,35 @@ func (r *RuleEnforcer) handleRequest(ctx context.Context,
...
+	var errs []error
 	for _, rule := range rules {
 		newRequest, err := rule.HandleRequest(ctx, ri.URI, msg)
 		if err != nil {
-			st := status.Errorf(
-				codes.ResourceExhausted, "rule violation: %v",
-				err,
-			)
-			return nil, st
+			errs = append(errs, err)
```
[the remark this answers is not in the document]
bitromortac: yes, I can see that argument of resource usage, the current approach enables testing for different rule violations in a single test, see https://github.com/lightninglabs/lightning-terminal/blob/a6b2b260bd036ee9ddfc81054737bde51d3919f4/itest/litd_firewall_test.go#L1104-L1105 and https://github.com/lightninglabs/lightning-terminal/blob/a6b2b260bd036ee9ddfc81054737bde51d3919f4/itest/litd_firewall_test.go#L1320-L1322
It also gives a complete list of errors, otherwise we'd only know the first successful ones, but there could be others after the first failed one. I can change it to your suggested approach, if you think it makes more sense

rules/onchain_budget.go
```diff
@@ -0,0 +1,647 @@
...
+
+// addPendingAmt updates the temporary KV store with the new amount. The total
+// pending amount is updated, and a request specific entry is added so that
+// we know how much to adjust our budget by if this request where to succeed or
+// fail.
+func (o *OnChainBudgetEnforcer) addPendingAmt(ctx context.Context,
+	tx firewalldb.KVStoreTx, pending *onChainState,
+	request *onChainAction) error {
+
+	store := tx.Local()
```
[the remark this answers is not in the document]
ellemouton: ie, i think what can happen is: 

1) request 1 comes through for open channel. We add pending payment for this.
2) restart lit
3) request 1 comes through again for completely new request but now we dont allow it cause we never cleared the previous entry

ellemouton (changes requested):
a couple points:

1) I dont think the association between action and memo is currently actually being persisted. Can we add a test to cover this? have ya'll tested things manually?
2) I dont think we can/should persist things prefixed with request ID to the main store. This can result in issues on start up as far as I know cause on start up, request IDs may be re-used. So perhaps in stead, on startup (for now) we take the "pending amount" and set that to the new "spent amount" with some persisted action indicating that we've done this. (or perhaps that balance just always remains in the "pending" bucket . 
3)
> Note that an alternative could be to use the request id, but this is
predictable and could manually be prefixed by a user.

dont think we can use request ID cause it is reset on startup I think

rules/onchain_budget.go
```diff
@@ -278,20 +304,33 @@ func (o *OnChainBudgetEnforcer) checkers() map[string]mid.RoundTripChecker {
...
+					)
 				}
 
-				return o.handlePendingPayment(
+				err = o.handlePendingPayment(
 					ctx, &onChainAction{
 						Amount:     uint64(totalAmt),
 						ActionType: actionTypeOpenChannel,
 					},
 				)
```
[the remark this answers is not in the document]
bitromortac: good catch :pray:! I haven't spent the time to fully flesh that commit out because it was still uncertain if we will use the identifier, but I should have done that

rules/onchain_budget.go
```diff
@@ -0,0 +1,703 @@
...
+type onChainAction struct {
+	// Amount is the onchain amount of the action.
+	Amount uint64 `json:"amount"`
+
+	// ActionType is the type of action that the amount is associated with.
+	ActionType actionType `json:"type"`
+
+	// ID is a unique identifier for the request that can be used to
+	// associate the action within LND.
+	ID string `json:"id"`
```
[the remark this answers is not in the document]
bitromortac: ah yes, leftover from a previous iteration where I stored the identifier there

rules/onchain_budget.go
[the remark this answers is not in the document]
bitromortac: > That was the og motivation for using the temp store in the first place
Ah good to know, I was not aware of that, yes, that needs to be fixed. I explored amending the request id with a unique identifier.

rules/onchain_budget.go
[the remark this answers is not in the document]
bitromortac: I added a check in the unit tests and the itest also has a check for the memo.

rules/onchain_budget.go
```diff
@@ -0,0 +1,694 @@
...
+	// A batch open channel request can contain multiple channels, so we
+	// need to sum up the total amount. This operation is atomic since we
+	// publish a single transaction.
+	var totalAmt int64
+	for _, c := range r.Channels {
+		totalAmt += c.LocalFundingAmount + c.PushSat
+
+		// Prepend an identifier to later be able to check the action's
+		// outcome.
+		c.Memo = fmt.Sprintf("%s:%s", reqID, c.Memo)
```
ViktorTigerstrom, line 412: Just leaving a note here, that for batch-channel openings, every channel will have the same Memo. This isn't an issue now, but I think we should consider if this could become an issue for the future if:
1. The memo field could potentially get a unique constraint for channels in the future, effectively breaking this.
2. If lnd is ever updated to enable tracking of channel openings by specifying the memo field, that could become problematic if we don't make it unique.
3. For our intended way of checking on restart if a channel opening was successful or not, I guess this won't become an issue right? I assume that we'd iterate the list of open & closed channels and stop the execution if we find a single channel with this specific memo value, and if so consider all of the pending payment amount to have been confirmed.

rules/onchain_budget.go
```diff
@@ -0,0 +1,694 @@
...
+	// any given time. This prevents db entry race conditions.
+	sync.Mutex
+}
+
+// Stop cleans up the resources held by the manager.
+//
+// NOTE: This is part of the Manager interface.
+func (o *OnChainBudgetMgr) Stop() error {
+	return nil
+}
```
ViktorTigerstrom, line 53: Just noting that I think we should be able to use this function to delay litd's shutdown until we've either received a response from `lnd` for a channel open request, or until we time out.

I also just wanted to check that my understanding is correct that we decided to delay that extra functionality and instead add it in a follow-up PR. Is that correct?

rules/onchain_budget.go
```diff
@@ -0,0 +1,694 @@
...
+// derived from the main rules Config struct.
+type onChainBudgetConfig interface {
+	GetStores() firewalldb.KVStores
+	GetReqID() int64
+	GetLndConnID() string
+}
+
+// OnChainBudgetEnforcer enforces requests and responses against a
+// OnChainBudget rule.
+type OnChainBudgetEnforcer struct {
```
ViktorTigerstrom, line 112: I want to check what our conclusion was regarding clearing all pending amounts and entries in the `onChainPending` local store, on restart of litd (i.e. consider them as confirmed even if they weren't)?

If I understood correctly, the conclusion was that it's nice to not clear them to have historical data showing where balance was deducted. Given that we also now add the memo field, that also gives some historical proof.

I agree with the decision to not clear them for now though, as when the memo field actually becomes useable, we can eventually clear the old pending amounts which were never succeeded in a channel opening, in the future.

rules/onchain_budget.go
```diff
@@ -0,0 +1,694 @@
...
+	}
+
+	err := o.checkFeeRate(r.SatPerVbyte)
+	if err != nil {
+		return nil, err
+	}
+
+	// Derive a unique identifier for the request to be able to associate it
+	// with the repsonse.
+	reqID := formatReqId(o.GetLndConnID(), o.GetReqID())
```
ViktorTigerstrom, line 362: nit: I think it's a little confusing that we name this variable to `reqID` here and at other places, as that can be confused with the request ID returned by the `GetReqID` function. 
Perhaps it should be renamed to something like `prefReqID`?

rules/onchain_budget.go
```diff
@@ -0,0 +1,694 @@
...
+	// Construct the unique key for this request.
+	reqPref := fmt.Sprintf("%s:%s", onChainPending, reqID)
+
+	// Check that we don't already have a pending payment for this request.
+	reqBytes, err := store.Get(ctx, reqPref)
+	if err != nil {
+		return err
+	}
+
+	if len(reqBytes) != 0 {
```
ViktorTigerstrom, line 688: I think it makes sense to add a test case for this edge case in the `onchain_budget_test.go`, and test that this get´s triggered if we send another request with the same `ReqID` before processing the last one.

I think it also makes sense to add coverage that you can make 2 `OpenChannelSync` requests at the same time before either of them has received a response, as long as the `ReqID` is different.

itest/litd_firewall_test.go
```diff
@@ -891,6 +943,416 @@ func testSessionLinking(net *NetworkHarness, t *harnessTest) {
...
```
ViktorTigerstrom, line 948: Perhaps this test isn't the correct place for it, but I think we should add test coverage to prove that the `LndConnId` gets set to a different value if `litd` is restarted, i.e. with `net.RestartNode`.

itest/litd_firewall_test.go
```diff
@@ -891,6 +943,416 @@ func testSessionLinking(net *NetworkHarness, t *harnessTest) {
...
+
+		for _, channel := range channels.Channels {
+			if channel.ChannelPoint == outpoint.String() {
+				parts := strings.Split(channel.Memo, ":")
+				require.GreaterOrEqual(t, len(parts), 2)
+
+				// We expect the format of lndConnID-reqID.
+				parts = strings.Split(parts[0], "-")
+				require.Len(t, parts, 2)
+				require.Len(t, parts[0], terminal.LndConnIdLen)
```
ViktorTigerstrom, line 1192: nit: I think we should add coverage to prove that the `reqID` (`parts[1]`) in the memo field varies for every made OpenChannelRequest.

terminal.go
```diff
@@ -542,6 +548,11 @@ func (g *LightningTerminal) start() error {
 		return fmt.Errorf("could not connect to LND")
 	}
 
+	// In order to be able to create unique middleware request identifiers,
+	// we set a new unique connection ID. This should be refreshed every
+	// time we (re)connect to LND.
+	g.lndConnID = randId(LndConnIdLen)
```
ViktorTigerstrom, line 554: nit: maybe it makes more sense to add this in the `connectLND` function instead? Then you can perhaps add it as a return value of the function.

ViktorTigerstrom (commented):
This looks really great :tada::rocket::tada:!

I left a few comments to my review. The first comment is the only potential thing that could be a somewhat bigger thing to fix if we should address it.
The other comments are either minor fixes, or reaffirming of what we discussed offline.

I will finalize my testing tomorrow!

I think this should be pretty much ready to go :rocket:!!

rules/onchain_budget.go
bitromortac: 1. I think it's unlikely that it becomes unique, as that wouldn't be great UX-wise if a user has to account for this.
2. Good point! I guess it would make sense to track based on batched channel opens (as transactions bundle them atomically). The memo field could still be used for that, returning any batch for which a single memo field matched. The memo field isn't that nice for that, I'd rather use the pending channel id, but better to have the recordings here, in case we need to do something with them (we can still ignore them in the future).
3. Right, that would be the approach, we'd fetch all pending items and cross check against the channel API's memos.

terminal.go
bitromortac: right, I considered that, but I thought better to keep `connectLND` free from side effects and only assign to `g` in this scope

rules/onchain_budget.go
bitromortac: I think that would be ideal to have, but would require a bit more work to check that we don't create side effects and I'm not sure how to test this in an automated way (we'd need to deliberately create a delayed call). I'll experiment around with that, but would suggest to keep this optional for this PR, but to be addressed soon.

firewall/privacy_mapper.go
```diff
@@ -812,6 +841,654 @@ func handleUpdatePolicyResponse(db firewalldb.PrivacyMapDB,
...
+				if !flags.Contains(session.ClearPubkeys) {
+					remotePub, err = firewalldb.HideString(
+						tx, remotePub,
+					)
+					if err != nil {
+						return err
+					}
+				}
+
+				capacity, err := maybeHideAmount(
```
ViktorTigerstrom, line 1056: How come we obfuscate the capacity here and in the Pending Channels response, but we don't obfuscate the capacity in the `listchannels` response? I think the latter is related to the `AutoFees` algo right? Given that though, should we mimic that behaviour here and for the Pending Channels response as well?

firewall/privacy_mapper.go
```diff
@@ -812,6 +841,654 @@ func handleUpdatePolicyResponse(db firewalldb.PrivacyMapDB,
...
+					flags, randIntn, c.Capacity,
+				)
+				if err != nil {
+					return err
+				}
+
+				settledBalance, err := maybeHideAmount(
```
ViktorTigerstrom, line 1063: We should probably cap this at the capacity of the channel.

firewall/privacy_mapper.go
```diff
@@ -812,6 +841,654 @@ func handleUpdatePolicyResponse(db firewalldb.PrivacyMapDB,
...
+				if err != nil {
+					return err
+				}
+
+				remoteBalance, err := maybeHideAmount(
+					flags, randIntn, c.Channel.RemoteBalance,
+				)
+				if err != nil {
+					return err
+				}
```
ViktorTigerstrom, line 969: Should we cap the obfuscated value of this to the sum of of the capacity? We could also make the sum of these become equal to the capacity similar to the listchannels response.

ViktorTigerstrom (commented):
Noticed no real issues furing testing. Great work 🚀:fire::rocket!!!!
I did notice somewhat minor strange behaviour regarding the `ClearAmounts` flag for the closed & pending channels responses. I'm leaving a few comments below describing those. These are non-blocking issues though :)!

firewall/privacy_mapper.go
bitromortac: > I think the latter is related to the AutoFees algo right?
Yes, that's right, we need more work to study the influence on the algo, but the goal would be to obfuscate there as well.

rules/onchain_budget.go
```diff
@@ -0,0 +1,785 @@
...
+
+// formatReqId creates a unique identifier for the request that can be used to
+// associate the action accross request-response handling.
+func formatReqId(lndConnID string, reqID int64) string {
+	return fmt.Sprintf("%s-%s-%d", memoPrefix, lndConnID, reqID)
+}
+
+// removeReqId removes the request ID from the memo if present.
+func removeReqId(memo string) string {
+	parts := strings.Split(memo, ":")
```
ViktorTigerstrom, line 399: It might be worth already implementing removal of the memo in by looping the over the parts and to look for the onBudget prefix to avoid that older versions of litd in the future leak extra privacy, and to have a reference how it should be done if we implement useage of the memo field for fetching the reqId in the future.

rules/onchain_budget_test.go
```diff
@@ -0,0 +1,485 @@
...
+
+	for _, channel := range pending.WaitingCloseChannels {
+		assertMemoInvariant(channel.Channel.Memo)
+	}
+}
+
+// TestRemoveMemo tests that request identifiers are correcly removed from the
+// memo string.
+func TestRemoveMemo(t *testing.T) {
+	tests := []struct {
```
ViktorTigerstrom, line 448: If you implement my suggestion above, I'd add another test that shows that you can have another part before the `onBudget-connID-reqID` part.
If you decide not to do it, I'd probably add another test to show that it's intended behaviour that we can't have another part before it currently.

ViktorTigerstrom (approved):
Everything looks great, awesome job :rocket:fire:!!!!

Just leaving one small suggestion in my comments below. Other than that, there's one last thing I noticed which wasn't really added by this PR:
If obfuscating an amounts field, the exact returned varies for every API request. This behaviour is different than how our normal obfuscation works, as we for example will return the same obfuscated pubkey with every unique API request in the firewall. I'm not sure if this is something we should address for now, but just thought I'd brought it up!

rules/onchain_budget.go
ellemouton: surely pretty easy to test? just extend `TestOnChainBudgetCheckRequest`: call a request, call Stop - see that it waits, then call response & see that Stop exits

ellemouton: this would be a bit complicated to actually implement though afaict since what happens if LND stops before lit? we would need to hook in to that to know which requests we can just not wait for

terminal.go
```diff
@@ -542,6 +544,11 @@ func (g *LightningTerminal) start() error {
 		return fmt.Errorf("could not connect to LND")
 	}
 
+	// In order to be able to create unique middleware request identifiers,
+	// we set a new unique connection ID. This should be refreshed every
+	// time we (re)connect to LND.
+	g.lndConnID = randId(rules.LndConnIdLen)
+
```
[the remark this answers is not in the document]
ellemouton: mabye worth adding a TODO somewhere since we do eventually want to make everything restart safe

rules/onchain_budget.go
ellemouton: yeah. Worth noting that we cant really rely on the memo field not changing though. So this is a nice to have, temp measure. Worst case scenario - user just links a session and updates the budget

rules/onchain_budget.go
```diff
@@ -0,0 +1,785 @@
...
+		return nil, fmt.Errorf("target conf not supported")
+	}
+
+	err := o.checkFeeRate(r.SatPerVbyte)
+	if err != nil {
+		return nil, err
+	}
+
+	// Derive a unique identifier for the request to be able to associate it
+	// with the repsonse.
```
ellemouton, line 452: s/repsonse/response

rules/onchain_budget.go
```diff
@@ -0,0 +1,785 @@
+package rules
```
ellemouton, line 1: perhaps this is a good time to start a doc where we explain small details like the memo field so that it is easy to recap in future?

so perhaps let's add a `rules/docs/onchain_budget.md` where we explain some things here?

ellemouton (approved):
Great work 🔥

firewall/rule_enforcer.go
```diff
@@ -236,21 +237,35 @@ func (r *RuleEnforcer) handleRequest(ctx context.Context,
...
+	// Should we have encountered any errors for rules in the request, we
+	// need to roll back any pending state changes.
+	if len(errs) > 0 {
+		for _, rule := range rules {
+			// We call HandleErrorResponse to undo any persisted
+			// state changes.
+			_, err := rule.HandleErrorResponse(ctx, ri.URI, nil)
+			if err != nil {
+				return nil, err
+			}
```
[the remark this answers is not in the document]
bitromortac: good point :pray:

terminal.go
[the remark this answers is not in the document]
bitromortac: I added the TODO :+1:

rules/onchain_budget.go
bitromortac: cool, I added a small doc, let me know if I missed anything there

rules/onchain_budget.go
bitromortac: I have implemented that idea and tested the current prefix together with a different one

PR #761, opened 2024-05-23
custom channels: use `tapd` as the aux component implementation for `lnd` in integrated mode

guggero:


config.go
```diff
@@ -395,6 +397,17 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 		cfg.Lnd.RPCMiddleware.Enable = true
 	}
 
+	// For the integration of tapd with lnd, we need to allow tapd to send
+	// custom error messages to peers through the SendCustomMessage RPC in
+	// lnd. Since the error messages aren't in the custom range, we
+	// explicitly need to allow them.
+	if !cfg.lndRemote {
```
ellemouton, line 406: ok so for remote mode i assume (like rpc middle ware stuff) that we will just start up but then quickly tap will notice it fails to deliver custom msgs to lnd?

terminal.go
```diff
@@ -1233,6 +1248,46 @@ func (g *LightningTerminal) BuildWalletConfig(ctx context.Context,
...
 }
 
+// buildAuxComponent builds the auxiliary components required by lnd when
+// running in integrated mode with tapd being the service that provides the
+// aux component implementations.
+func (g *LightningTerminal) buildAuxComponents() (*lnd.AuxComponents, error) {
+	tapdWrapper, available := g.subServerMgr.GetServer(subservers.TAP)
+	if !available {
+		return nil, fmt.Errorf("tapd is not available, must be " +
+			"started in integrated mode for Taproot Assets " +
```
ellemouton, line 1261: `started in integrated mode` -> do you mean that tapd must be integrated? cause 'integrated mode' to me means that lnd is integrated

terminal.go
```diff
@@ -818,16 +818,19 @@ func (g *LightningTerminal) setUpLNDClients(lndQuit chan struct{}) error {
...
+				Insecure:           insecure,
+				CustomMacaroonPath: macPath,
+				CustomMacaroonHex: hex.EncodeToString(
+					macData,
+				),
 				BlockUntilChainSynced: true,
 				BlockUntilUnlocked:    true,
 				CallerCtx:             ctxc,
 				CheckVersion:          minimalCompatibleVersion,
+				ChainSyncPollInterval: g.cfg.LndConnectInterval,
```
ellemouton, line 833: nice

cmd/litcli/ln.go
```diff
@@ -0,0 +1,751 @@
...
+	// minAssetAmount is the minimum amount of an asset that can be put into
+	// a channel. We choose an arbitrary value that allows for at least a
+	// couple of HTLCs to be created without leading to fractions of assets
+	// (which doesn't exist).
+	minAssetAmount = 100
+)
+
+var lnCommands = []cli.Command{
+	{
+		Name:     "ln",
```
ellemouton, line 38: would `tap` be better? `ln` to me is general `lightning network`

cmd/litcli/ln.go
```diff
@@ -0,0 +1,751 @@
...
+	"github.com/lightninglabs/taproot-assets/taprpc/rfqrpc"
+	tchrpc "github.com/lightninglabs/taproot-assets/taprpc/tapchannelrpc"
+	"github.com/lightningnetwork/lnd/cmd/commands"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/record"
+	"github.com/urfave/cli"
+)
```
ellemouton, line 26: just skimmed this commit really quickly but it isnt yet immediately obvious to me why this is here and not in the taproot assets `cmd` directory? ie what makes this different from `loopcli` or equivalent?

ellemouton (commented):
Had a quick look to gain familiarity - looks good and mostly plumbing 👍  main question is around why the cli commands belong here

config.go
```diff
@@ -379,6 +380,7 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 			cfg.Remote.LitDebugLevel, cfg.Lnd.LogWriter,
 		)
 	} else {
+		SetupLoggers(cfg.Lnd.LogWriter, interceptor)
```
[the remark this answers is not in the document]
guggero: Yes, you're right. I was debugging something logging related and forgot to remove this again. Confirmed it's not needed.

terminal.go
guggero: Actually, both. I'll update the error message.

terminal.go
guggero: Yeah, should speed up the normal itest as well.

cmd/litcli/ln.go
guggero: Yeah, I agree. Though I think this is temporary and we'll remove them in a future release anyway. But for now I think Polar already uses the commands in some demos, so I don't want to rename them just for renaming sake.

cmd/litcli/ln.go
guggero: We do eventually want the logic that's done in this file be RPCs in `tapd`, then it will make sense to move them to `tapcli`. But for now we need access to both `lnd`'s and `tapd`'s RPCs from within a single command, which is done most conveniently in `tapcli` (see the way we bake a macaroon just to connect to `tapd`).

terminal.go
```diff
@@ -466,13 +474,20 @@ func (g *LightningTerminal) start() error {
...
+		}
+
 		implCfg := &lnd.ImplementationCfg{
 			GrpcRegistrar:       g,
 			RestRegistrar:       g,
 			ExternalValidator:   g,
 			DatabaseBuilder:     g.defaultImplCfg.DatabaseBuilder,
 			WalletConfigBuilder: g,
 			ChainControlBuilder: g.defaultImplCfg.ChainControlBuilder,
+			AuxComponents:       *auxComponents,
```
bitromortac, line 490: Very cool to see how things come together :tada:!

bitromortac (commented):
Really exciting to see how everything fits together :tada::rocket:. Left some comments in this first (quick) pass.

cmd/litcli/main.go
```diff
@@ -244,3 +295,69 @@ func printRespJSON(resp proto.Message) { // nolint
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("error connecting client: %w", err)
+	}
+
+	defer cleanup()
+
+	var suffixBytes [4]byte
+	superMacID := session.NewSuperMacaroonRootKeyID(suffixBytes)
+	lndClient := lnrpc.NewLightningClient(lndConn)
+	ctxb := context.Background()
```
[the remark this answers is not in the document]
ellemouton: personally, I like when these types of resources are defined at the top of a function (so i think this location is good).

cmd/litcli/ln.go
dstadulis: I suppose since it's inducing ln functionality but pairing with tapd it's a bit of both

terminal.go
```diff
@@ -1233,6 +1248,46 @@ func (g *LightningTerminal) BuildWalletConfig(ctx context.Context,
...
+	if !available {
+		return nil, fmt.Errorf("tapd is not available, must be " +
+			"started in integrated mode for Taproot Assets " +
+			"Channels to be available")
+	}
+
+	if tapdWrapper.Remote() {
+		return nil, fmt.Errorf("tapd is not available, must be " +
+			"started in integrated mode for Taproot Assets " +
+			"Channels to be available")
```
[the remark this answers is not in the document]
guggero: Updated the error message according to @ellemouton's review. Hope it makes more sense now.

cmd/litcli/ln.go
```diff
@@ -0,0 +1,751 @@
...
+			r.RejectedQuote.ErrorCode, r.RejectedQuote.ErrorMessage)
+
+	default:
+		return fmt.Errorf("unexpected response type: %T", r)
+	}
+
+	msatPerUnit := acceptedQuote.BidPrice
+	numUnits := uint64(decodeResp.NumMsat) / msatPerUnit
+
+	fmt.Printf("Got quote for %v asset units at %v msat/unit from peer "+
```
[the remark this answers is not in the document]
guggero: Hmm, this message isn't really about consent, more of an information (and I guess initially mainly added for debugging purposes :sweat_smile: ).
The actual validation of the exchange rate is done by the RFQ and price oracle logic of `tapd`. The end user will be able to plug in their exchange API of choice to validate prices against, and then define a maximum percentage (PPM actually) of deviation due to spread.

cmd/litcli/ln.go
```diff
@@ -0,0 +1,751 @@
...
+		FeeProportionalMillionths: uint32(ourPolicy.FeeRateMilliMsat),
+		CltvExpiryDelta:           ourPolicy.TimeLockDelta,
+	}
+
+	invoice := &lnrpc.Invoice{
+		Memo:            ctx.String("memo"),
+		ValueMsat:       int64(numMSats),
+		DescriptionHash: descHash,
+		FallbackAddr:    ctx.String("fallback_addr"),
+		Expiry:          ctx.Int64("expiry"),
```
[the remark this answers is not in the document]
guggero: Ah yes, good point. Fixed!

itest/litd_custom_channels_test.go
```diff
@@ -0,0 +1,1339 @@
...
+	t.Logf("Fabia balance after invoice payment: %v",
+		toJSON(t.t, balanceFabia))
+
+	charlieAssetBalance -= fabiaInvoiceAssetAmount1
+	daveAssetBalance += fabiaInvoiceAssetAmount1
+	erinAssetBalance -= fabiaInvoiceAssetAmount1
+	fabiaAssetBalance += fabiaInvoiceAssetAmount1
+
+	// ------------
+	// Test case 4: Create an asset invoice on Fabian and pay it with just
```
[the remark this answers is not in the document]
guggero: It's a bit hidden, we hard code the max shard size to 80k in the `payInvoiceWithXXX()` helper. So choosing a large enough asset amount will guarantee at least two shards. Updated the comment to make it a bit more obvious.
And yes, we don't currently track the number of attempts. There are a lot of assertions still missing here, mainly due to the fact that the itest library isn't fully compatible with the refactored version in `lnd`, so we can't just re-use a lot of those assertions.

guggero:
Thanks for the reviews! I addressed all comments and fixed the failing CI steps (hopefully).
Taking the PR out of draft, even if we still want to add the force-close code. But that will likely only affect the integration test (and the `go.mod` file).

PR #762, opened 2024-05-28
accounts: address `pendingPayments` issues

ViktorTigerstrom:
This PR addresses two issues with the `pendingPayments` map handling in the accounts system.

1. Prior to this PR, we would deduct any pending payments as in-flight payments from an account's total balance when calculating the available balance, even if the pending payments were not initiated by the specific account itself. This PR adds an ID check to only deduct pending payments that are actually associated with the account.
2. If a payment with status `UNKNOWN` turns out to be uninitiated, we should remove it from the `pendingPayments` map. This ensures that the payment is not considered when calculating the in-flight balance for the account.

### DIFF
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -66,6 +66,8 @@ type InterceptorService struct {
 	invoiceToAccount map[lntypes.Hash]AccountID
 	pendingPayments  map[lntypes.Hash]*trackedPayment
 
+	*requestValuesStore
+
 	mainErrCallback func(error)
 	wg              sync.WaitGroup
 	quit            chan struct{}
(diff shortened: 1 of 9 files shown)


accounts/service.go
```diff
@@ -672,6 +672,16 @@ func (s *InterceptorService) TrackPayment(id AccountID, hash lntypes.Hash,
...
 
+					s.Lock()
+					defer s.Unlock()
+
+					// We also remove the payment from the
+					// pending payments map, so that the
+					// payment won't be seen as in-flight
+					// balance when calculating the
+					// account's available balance.
+					delete(s.pendingPayments, hash)
```
bitromortac, line 683: Why do we keep the payment request in the accounts? Could we alternatively call `s.removePayment`? If a payment wasn't initiated we should track it the next time it is really attempted, not?

accounts/service_test.go
```diff
@@ -444,18 +448,22 @@ func TestAccountService(t *testing.T) {
...
-						FullAmount: 1234,
+						FullAmount: 2000,
 					},
 					testHash2: {
 						Status:     lnrpc.Payment_UNKNOWN,
-						FullAmount: 3456,
+						FullAmount: 1000,
+					},
+					testHash3: {
+						Status:     lnrpc.Payment_UNKNOWN,
```
bitromortac, line 465: Could we also add a case with Status 4 (`lnrpc.Payment_INITIATED`), which is not yet in the proto, but will be for 0.18? Also test the case down below when we receive such an update. I think we may need to invert the inflight check in `paymentUpdate` to check that the status is not success/failure to be able to exit early on the new status code, otherwise status 4 would be treated as a success and may wrongly reduce the balance?

bitromortac (commented):
Nice fixes :fire:! I have a question and a suggestion.

ellemouton (commented):
Looking good once those comments are addressed!

Im also gonna send you the commits i've created that adds some tracing in the accounts package just so that we have more info on these things in future :) 

Once that has been added then i think this is g2g

accounts/service_test.go
ViktorTigerstrom: Ah nice suggestion! I've updated this with the latest push :)!

ellemouton (approved):
LGTM once CI is green! 🔥 

good stuff

ViktorTigerstrom:
Thanks for the reviews :tada:!

And thanks a lot for the commits that adds tracing @ellemouton 🙏! That'll be really useful.

I've addressed the feedback and added your commits with the latest push @ellemouton!

ellemouton:
i merged https://github.com/lightninglabs/lightning-terminal/pull/763 before addressing the review comments there. And viktor is OOO at the moment so im gonna close this PR and open a new one with these commits along with the addressed review comments.

PR #763, opened 2024-05-29
firewalldb: fetch correct bucket for session level temp KV store

ellemouton:
In this commit, we fix a bug which would result in the "perm" store being used for the LocalTemp KV DB _if and only if_ the KV store being requested was a session level one (as opposed to a session feature level one). It is ok for us to just change this here since as of this commit, the session level KV store is not used at all and so nothing will be in that store yet.

### DIFF
--- a/firewalldb/kvstores.go
+++ b/firewalldb/kvstores.go
@@ -9,8 +9,11 @@ import (
 
 /*
 The KVStores are stored in the following structure in the KV db. Note that
-the `perm` and `temp` buckets are identical in structure. The only difference is
-that the `temp` bucket is cleared on restart of the db.
+the `perm` and `temp` buckets are identical in structure. The only difference
+is that the `temp` bucket is cleared on restart of the db. The reason persisting
+the temporary store changes instead of just keeping an in-memory store is that
+we can then guarantee idempotency if changes are made to both the permanent and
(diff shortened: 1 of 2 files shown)


firewalldb/kvstores.go
```diff
@@ -9,8 +9,11 @@ import (
 
 /*
 The KVStores are stored in the following structure in the KV db. Note that
-the `perm` and `temp` buckets are identical in structure. The only difference is
-that the `temp` bucket is cleared on restart of the db.
+the `perm` and `temp` buckets are identical in structure. The only difference
+is that the `temp` bucket is cleared on restart of the db. The reason persisting
+the temporary store changes instead of just keeping an in-memory store is that
+we can then guarantee idempotency if changes are made to both the permanent and
```
bitromortac, line 15: nit: not 100% sure but wouldn't it be called atomicity instead of idempotency in this context?

bitromortac (approved):
nice find! LGTM :tada:

firewalldb/kvstores.go
ellemouton: indeed! you are correct - will update

ViktorTigerstrom (approved):
Nice catch! LGTM :tada:!

PR #764, opened 2024-05-31
Bump litd to version `v0.13.0-alpha`

ViktorTigerstrom:
This PR bumps the litd version to `v0.13.0-alpha`.

This PR also bumps lnd to version `v0.18.0-beta` and all sub-servers except faraday to new versions which support lnd `v0.18.0-beta`.

The Go version is also bumped to version `1.22.3`.

I've tested this locally with the terminal web UI, and the Loop & Pool flow worked as expected for me. I've also tested the Autopilot flow locally.


terminal.go
```diff
@@ -138,7 +138,7 @@ var (
 	// to run LiT in remote mode.
 	minimalCompatibleVersion = &verrpc.Version{
 		AppMajor: 0,
-		AppMinor: 17,
```
ViktorTigerstrom, line 141: I may revert this if it turns out that none of our su-servers require v0.18.0

README.md
```diff
@@ -155,6 +156,7 @@ The following table shows the supported combinations:
 
 | LiT               | LND          | Loop         | Faraday       | Pool         | Taproot Assets |
 |-------------------|--------------|--------------|---------------|--------------|----------------|
+| **v0.13.0-alpha** | v0.18.0-beta | v0.28.3-beta | v0.2.13-alpha | v0.6.5-beta  | v0.3.3-alpha   |
```
ViktorTigerstrom, line 159: Not sure exactly what to set `pool` & `tapd` to here, as those won't be official releases, but rather branches litd will depend on.

autopilotserver/mock/server.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/lightninglabs/lightning-terminal/autopilotserverrpc"
 	"github.com/lightninglabs/lightning-terminal/rules"
 	"github.com/lightninglabs/lightning-terminal/session"
-	"github.com/lightningnetwork/lnd/lntest/node"
+	"github.com/lightningnetwork/lnd/lntest/port"
```
guggero, line 16: Unfortunately this commit needs to be squashed with the previous one to make CI happy.

README.md
guggero: The latest tag versions are okay IMO. So this looks good.

guggero (approved):
Nice, LGTM once CI is fixed.

ViktorTigerstrom:
Updated to the latest loop release, and also dropped the updates to the minimum required lnd version!

guggero (approved):
utACK, LGTM :tada:

ellemouton (approved):
LGTM 🥇

app/build/.gitkeep
jamaljsr, line 1: Just out of curiosity, why is this file no longer needed?

The [commit](https://github.com/lightninglabs/lightning-terminal/commit/4499e45ac533dfad7c34f558786c7811578c0b5f) where this was originally added states:
> Even though the content of the app/build directory should be ignored by git, we need it to exist in order for the Golang embedded file system compilation to succeed on an empty project. This is required to fix compilation issues when lightning-terminal is used as a module dependency in another project, where the frontend build process can't be triggered.

Was the Golang embedded fs updated so this is no longer needed?

jamaljsr (approved):
tACK LGTM 🚀

Just did a quick test of Loops in the `litd` UI and connecting from TW.

app/build/.gitkeep
guggero: Oh, I missed this in my review... Thanks for spotting! We definitely should keep this file.

guggero (changes requested):

app/build/.gitkeep
ViktorTigerstrom: Ah totally my bad! Great catch, thanks!

ViktorTigerstrom:
Addressed @jamaljsr feedback with the latest push!
I also bumped loop to `v0.28.5-beta` that's been released since the last push. Note that this also changes the litd `lnd` dependency to reference the `lnd v0.18.0-beta.1` tag. That tag is referencing the same commit as lnd `v0.18.0-beta` though, so that's not an issue!

ellemouton:
@jamaljsr  - mind giving this another since the update and the rebase?

jamaljsr (approved):
Reviewed the latest changes. 

tACK LGTM 👌

PR #774, opened 2024-06-20
Modify itest make commands

ffranr:
This PR modifies the `build-itest` make command such that it builds `litd-itest`. It also adds a new `clean-itest` command which deletes itest binaries.


jamaljsr (approved):
tACK LGTM 👌

PR #775, opened 2024-06-21
itest: add BTC-only keysend and invoice payments

guggero:
To make sure asset channels also function properly with BTC-only
payments, we add a keysend and invoice payment to the integration tests.

### DIFF
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -377,7 +377,7 @@ func testCustomChannels(_ context.Context, net *NetworkHarness,
 	// Test case 1: Send a direct keysend payment from Charlie to Dave.
 	// ------------
 	const keySendAmount = 100
-	sendKeySendPayment(t.t, charlie, dave, keySendAmount, assetID)
+	sendAssetKeySendPayment(t.t, charlie, dave, keySendAmount, assetID)
 	logBalance(t.t, nodes, assetID, "after keysend")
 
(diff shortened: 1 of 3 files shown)


itest/litd_custom_channels_test.go
```diff
@@ -405,6 +409,11 @@ func testCustomChannels(_ context.Context, net *NetworkHarness,
 	charlieAssetBalance -= paidAssetAmount
 	daveAssetBalance += paidAssetAmount
 
+	// We should also be able to do a multi-hop BTC only payment, paying an
+	// invoice from Erin by Charlie.
+	createAndPayNormalInvoiceWithBtc(t.t, charlie, erin, 2000)
+	logBalance(t.t, nodes, assetID, "after BTC only invoice")
```
GeorgeTsagk, line 415: does this also assert that the balance is as expected?

GeorgeTsagk (approved):
LGTM 🌳

itest/litd_custom_channels_test.go
guggero: No, but we do that later. That's just for visual confirmation. And currently we don't track and/or assert the BTC balance. But since we don't really change any logic in the BTC-only flow, just making sure the payment arrives should be enough.

jamaljsr (approved):
I can confirm that the sats are flowing in both directions over custom channels. I tested direct peer and multi-hop payments. Fantastic! 👏

PR #781, opened 2024-06-25
itest: Add custom channels with grouped asset test

GeorgeTsagk:
## Description

This PR adds a new itest that uses a grouped asset, in order to tackle some of the issues discovered in https://github.com/lightninglabs/taproot-assets/issues/947

Closes https://github.com/lightninglabs/taproot-assets/issues/957


guggero:
Looks like the grouped asset itest fails consistently when sending a payment. Does that happen as well for you locally?

GeorgeTsagk:
>  Looks like the grouped asset itest fails consistently when sending a payment. Does that happen as well for you locally?

Nope, locally it passes consistently, and on Github CI it's flaky

guggero:
The flake will be fixed by https://github.com/lightningnetwork/lnd/pull/8871. Not sure if we should get this in while flakey and then wait for the CI to become fully green in https://github.com/lightninglabs/lightning-terminal/pull/782 which bumps all dependencies?

guggero (approved):
LGTM :tada: 

I'm in favor of merging this, even if the itest doesn't pass yet. I've confirmed that https://github.com/lightningnetwork/lnd/pull/8871 will fix this, so it's just a matter of bumping the `lnd` dependency after merge, which https://github.com/lightninglabs/lightning-terminal/pull/782 will do anyway.

GeorgeTsagk:
> The flake will be fixed by https://github.com/lightningnetwork/lnd/pull/8871. Not sure if we should get this in while flakey and then wait for the CI to become fully green in https://github.com/lightninglabs/lightning-terminal/pull/782 which bumps all dependencies?

I'm fine both ways, since we know it's a flake we could hit the merge prematurely

Issue #805, opened 2024-07-25
Bug on LITD instances // unable to use genseed API


weltitob:
JSON Response: {'code': 2, 'message': 'lnd is not ready for: /lnrpc.WalletUnlocker/GenSeed', 'details': []} // The genSeed API service is uncallable for LITD instances after startup eventough it should be the way hwo to start them up - my code:

```
import base64
import json
import requests

REST_HOST = '54.198.183.180:8443'
#TLS_PATH = 'LND_DIR/tls.cert'

url = f'https://{REST_HOST}/v1/genseed?"aezeed_passphrase"=Test'

data = {
    #'aezeed_passphrase': base64.b64encode(mnemonic_seed.encode()).decode('utf-8'),
    #'seed_entropy': base64.b64encode(b'').decode('utf-8'),
}

headers = {
    'Content-Type': 'application/json'
}

r = requests.get(url, headers=headers, data=json.dumps(data), verify=False)


# Try to print JSON response if available
try:
    json_response = r.json()
    print("JSON Response:", json_response)

except requests.exceptions.JSONDecodeError:
    print("Response is not a valid JSON: ", r)
```

endpoint url = f'https://{REST_HOST}/v1/state' retruns {'state': 'NON_EXISTING'}

However: Using lnclli create works completly fine on the same instance, this issue was also reported after I mentioned it in the forum by another user
![image](https://github.com/user-attachments/assets/58362c6a-77af-4101-8aef-0c5219ca9fe7)

Please can someone help me with this as it is crucial for my application development to continue

weltitob:
I need help on this issue please. Might also be a failure in me calling the api but I coudn't find anything so far

guggero:
What are you trying to achieve? Generate a completely new seed? Because that's what the `genseed` call is for. It's completely stateless and doesn't affect the wallet state at all.
But from your example code it looks to me like you're trying to create a wallet with an existing seed? That would then be `InitWallet` (`/v1/initwallet`) call.

weltitob:
> What are you trying to achieve? Generate a completely new seed? Because that's what the `genseed` call is for. It's completely stateless and doesn't affect the wallet state at all. But from your example code it looks to me like you're trying to create a wallet with an existing seed? That would then be `InitWallet` (`/v1/initwallet`) call.

[...]

PR #806, opened 2024-07-29
terminal: allow wallet unlocker services REST calls

guggero:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/805.

When lnd isn't marked as fully ready yet, we didn't allow any REST methods through. But to allow creating a wallet through REST, we also need to allow these methods.

### DIFF
--- a/terminal.go
+++ b/terminal.go
@@ -145,6 +145,27 @@ var (
 			"signrpc", "walletrpc", "chainrpc", "invoicesrpc",
 		},
 	}
+
+	// walletUnlockerServiceMethods defines methods of the wallet unlocker
+	// service that we don't require macaroons to access. We also allow
+	// these methods to be called even if lnd is not yet fully marked as
+	// started up (because it cannot start if it's still locked or no wallet
+	// exists).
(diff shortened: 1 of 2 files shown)


ellemouton (approved):
Noice 🙏

guggero:
I keep forgetting that we now also have file based release notes in `litd`. Added a new file.

terminal.go
```diff
@@ -246,11 +267,17 @@ func (g *LightningTerminal) Run() error {
 	// URI as soon as LND can accept the call, i.e. when the lnd sub-server
 	// is in the "Wallet Ready" state.
 	lndOverride := func(uri, manualStatus string) (bool, bool) {
```
ViktorTigerstrom, line 269: Nit: The comment above here for `lndOverride` is maybe a bit misplaced now that `/lnrpc.State/GetState` URI has been moved to a map. I'm fine to keep it as is if you think that the comment is still ok here :).

ViktorTigerstrom (approved):
Awesome, thanks for the fix :pray:! LGTM :rocket:

PR #807, opened 2024-07-29
[custom channels]: Use new `tapchannelrpc` RPC methods

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1048.
~Depends on https://github.com/lightningnetwork/lnd/pull/8947.~

Uses the new `tapchannelrpc` RPC methods `SendPayment` and `AddInvoice` that do the RFQ part inline, so we don't have to keep that logic in the `litcli` code.

### DIFF
--- a/accounts/service_test.go
+++ b/accounts/service_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/lightningnetwork/lnd/channeldb"
 	invpkg "github.com/lightningnetwork/lnd/invoices"
 	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/stretchr/testify/require"
 )
@@ -31,14 +32,10 @@ type mockLnd struct {
 	mainErrChan chan error
 
 	invoiceReq chan lndclient.InvoiceSubscriptionRequest
-	paymentReq chan lntypes.Hash
 
 	invoiceSubscriptionErr error
-	trackPaymentErr        error
 	invoiceErrChan         chan error
-	paymentErrChan         chan error
 	invoiceChan            chan *lndclient.Invoice
-	paymentChans           map[lntypes.Hash]chan lndclient.PaymentStatus
 }
 
 func newMockLnd() *mockLnd {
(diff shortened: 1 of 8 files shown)


guggero:
I tested all commands manually, everything seems to work as expected. Ready for full review now.

accounts/service_test.go
```diff
@@ -100,23 +100,69 @@ func (m *mockLnd) assertInvoiceRequest(t *testing.T, addIndex,
...
+	if m.invoiceSubscriptionErr != nil {
+		return nil, nil, m.invoiceSubscriptionErr
+	}
+
+	m.invoiceReq <- req
+
+	return m.invoiceChan, m.invoiceErrChan, nil
+}
+
+type mockRouter struct {
```
ffranr, line 118: I think we should add a bit more to this commit message. I'm not certain what's happening here. I think you're removing functionality from the LND mock and putting it into a new mock called `mockRouter`.

Is that split related to the lndclient version bump?

cmd/litcli/ln.go
```diff
@@ -223,8 +224,61 @@ var (
...
+		Usage: "the amount of the asset to send in the asset keysend " +
+			"payment",
+	}
+
+	rfqPeerPubKeyFlag = cli.StringFlag{
+		Name: "rfq_peer_pubkey",
+		Usage: "(optional) the public key of the peer to ask for a " +
+			"quote when converting from assets to sats; must be " +
+			"set if there are multiple channels with the same " +
+			"asset ID present",
```
ffranr, line 177: Currently, this must be set before it reaches the tap RFQ service. If this field is optional, and unset by the user, when does it get populated?

cmd/litcli/ln.go
```diff
@@ -223,8 +224,61 @@ var (
...
+	}
 )
 
+type resultStreamWrapper struct {
```
ffranr, line 181: missing doc for struct, struct fields, and `Recv` method

cmd/litcli/ln.go
```diff
@@ -339,15 +360,20 @@ func sendPayment(ctx *cli.Context) error {
 			"is instead: %v", len(destNode))
 	}
 
+	rfqPeerKey, err := hex.DecodeString(ctx.String(rfqPeerPubKeyFlag.Name))
+	if err != nil {
+		return fmt.Errorf("unable to decode RFQ peer public key: "+
+			"%w", err)
+	}
```
ffranr, line 305: Does this work if the field is optionally unset?

itest/assets_test.go
```diff
@@ -584,31 +584,42 @@ func sendAssetKeySendPayment(t *testing.T, src, dst *HarnessNode, amt uint64,
...
+		PaymentRequest: sendReq,
+	})
 	require.NoError(t, err)
 
-	htlcCarrierAmt := btcAmt.UnwrapOr(500)
-	sendKeySendPayment(
-		t, src, dst, btcutil.Amount(htlcCarrierAmt),
-		encodeResp.CustomRecords,
-	)
+	time.Sleep(time.Second)
```
ffranr, line 615: Unnecessary sleep? Shouldn't `getAssetPaymentResult` block on the recv?

accounts/service_test.go
GeorgeTsagk: +1, fully understand the diff but missing rationale/context

cmd/litcli/ln.go
GeorgeTsagk: Tapd is able to find a peer pubkey, given a specific asset id (it will look for a peer that has that asset channel)

cmd/litcli/ln.go
GeorgeTsagk: yes, decoding an empty string ("") is valid and doesn't return an error

itest/assets_test.go
guggero: Unfortunately and strangely the tests fail without it... It shouldn't make a difference, but somehow it does. Moved it to a single place and added an explanatory comment.

accounts/service_test.go
guggero: Added some rationale to the commit message.

ffranr (approved):

PR #813, opened 2024-08-08
Add nightly schedule to Docker CD workflow

dstadulis:
Added a daily schedule to the Docker workflow to automate builds and
pushes of the `daily-testing` image. This change introduces a new cron
job that runs every day at midnight (UTC) and triggers the workflow.

Additionally, updated the workflow to set different environment
variables for daily builds, including `RELEASE_VERSION` and `IMAGE_TAG`.
These changes enable the automatic creation of daily testing images with
unique tags.

Inspired by:
https://github.com/lightningnetwork/lnd/blob/57a5e4912bced63ce13be06896f64d7e476616b4/.github/workflows/docker.yml#L7-L44

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,7 +4,11 @@ on:
   push:
     tags:
       - 'v*'
-
+  schedule:
+    # Every day at midnight (UTC).
+    - cron: '0 0 * * *'
+  # Manual trigger through the UI for testing.
+  workflow_dispatch:
 
 defaults:
   run:
@@ -13,6 +17,11 @@ defaults:
 env:
   DOCKER_ORG: lightninglabs
   DOCKER_REPO: lightning-terminal
+  NIGHTLY_DOCKER_REPO: lightning-terminal-nightly
+  # TODO(guggero): Change the branch to 'master' once the experimental branch is
+  # merged into master.
+  NIGHTLY_BRANCH_NAME: 0-19-staging
+  NIGHTLY_TAG_NAME: experimental-daily-testing
 
 jobs:
   main:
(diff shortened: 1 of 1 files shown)


guggero (changes requested):
This would now push a tag called `master` every night, which isn't what we want.

Roasbeef:
Yeah this needs to be constrained further, for now we just want that side branch.

guggero:
Pushed up some fixes. According to the [documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch):

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

Which is why this is going into the `master` branch, even though it will only build from the `0-19-staging` branch.

guggero (approved):

.github/workflows/docker.yml
```diff
@@ -11,8 +11,8 @@ defaults:
     shell: bash
 
 env:
-  DOCKER_REPO: lightninglabs
-  DOCKER_IMAGE: lightning-terminal
+  DOCKER_ORG: lightninglabs
+  DOCKER_REPO: lightning-terminal
```
ellemouton, line 19: much nicer, thanks 🙏

.github/workflows/docker.yml
```diff
@@ -30,16 +30,20 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_API_KEY }}
 
+      # Make it possible to use different values for the version (used for git
+      # checkout) and the image tag (used for the docker image tag).
       - name: Set env
-        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+        run: |
+          echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+          echo "IMAGE_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
```
ellemouton, line 45: looks like both are always set to the same thing though?

ellemouton: ah nvm - i see next commit sets them to diff values for the nightly build 👍

.github/workflows/docker.yml
```diff
@@ -13,6 +17,9 @@ defaults:
 env:
   DOCKER_ORG: lightninglabs
   DOCKER_REPO: lightning-terminal
+  NIGHTLY_DOCKER_REPO: lightning-terminal-nightly
+  NIGHTLY_BRANCH_NAME: 0-19-staging
```
ellemouton, line 21: i assume we will eventually change this to master?

ellemouton (approved):
lgtm!

.github/workflows/docker.yml
guggero: Correct. Currently we get the most value out of nightly builds of the experimental branch, because that's where we iterate the fastest. Once that's in master, we'll change this too.
Going to add a TODO to make it clear.

dstadulis:
Nightly builds will be published to this repo
https://hub.docker.com/r/lightninglabs/lightning-terminal-nightly

PR #825, opened 2024-08-22
itest: spend assets after sweeping from force close

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1103.

Demonstrates the issue in https://github.com/lightninglabs/taproot-assets/issues/1099, and that the fix works.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/lightninglabs/loop/swapserverrpc v1.0.8
 	github.com/lightninglabs/pool v0.6.5-beta.0.20240604070222-e121aadb3289
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.2
-	github.com/lightninglabs/taproot-assets v0.4.2-0.20240815180811-2110839696cb
+	github.com/lightninglabs/taproot-assets v0.4.2-0.20240824000229-881ecafbeae1
 	github.com/lightningnetwork/lnd v0.18.0-beta.rc4.0.20240730143253-1b353b0bfd58
 	github.com/lightningnetwork/lnd/cert v1.2.2
 	github.com/lightningnetwork/lnd/fn v1.1.0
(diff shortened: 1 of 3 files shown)


jharveyb:
Locally I observe a different (probably related?) error:

```bash
litd_custom_channels_test.go:1429: 
        	Error Trace:	/home/jhb/lightning-terminal/itest/litd_custom_channels_test.go:1429
        	            				/home/jhb/lightning-terminal/itest/test_harness.go:103
        	            				/home/jhb/lightning-terminal/itest/litd_test.go:100
        	Error:      	Received unexpected error:
        	            	rpc error: code = Unknown desc = unable to sign and commit virtual packet: unable to generate Taproot Asset witness data: invalid split commitment proof
```

AFAICT, the sweep TX by Dave is performing a full-value self-send, which gets broadcast, verified by Dave correctly, and imported into his node. The script key is detected as local and known.

On coin selection, the input is detected as not having a split commitment, but during signing tapd fails on generating a witness. I suspect the sweep output is being stored incorrectly s.t. the split root is set incorrectly, causing the split leaf proof for the send to be invalid.

jharveyb:
AFAICT this is the point of failure:

https://github.com/lightninglabs/taproot-assets/blob/e893dee87e9d8f0de53b8ee9e2527add80df6491/vm/vm.go#L314

From debug sesssions it looks like the split proof verify is failing on the new split root.

Roasbeef:
> AFAICT, the sweep TX by Dave is performing a full-value self-send, which gets broadcast, verified by Dave correctly, and imported into his node. The script key is detected as local and known.


>  I suspect the sweep output is being stored incorrectly s.t. the split root is set incorrectly, causing the split leaf proof for the send to be invalid.

Isn't this contradictory? If it's a full value send, then there's no split commitment needed, unless the asset created is stored incorrectly, referencing a split in the witness when non is required?

Roasbeef:
Pushed a fix for the added assertion. Also have a fix for the underlying issue as well, PR inc there.

PR #826, opened 2024-08-23
chore: add vscode config to support itest debugging

jharveyb:
This PR adds a debug target for VSCode users to make itest debugging easier.

Breakpoints in the itest work as expected, and with some changes you can also step into the subprojects (tapd, lnd) being run as part of litd:

- Update `go.mod` to use a local copy of the subproject repo; e.x.:

`replace github.com/lightninglabs/taproot-assets => ../tap/tap`

- Set a breakpoint in an itest, right after all nodes are started.
- Open a new window in the workspace of the subproject.
- Use the 'Attach to process' debug target in that window to attach to a litd node started from the itest.

At that point, breakpoints in both windows should work as normal.

I think this would need some cleanup to be mergeable; I'm not sure if the itest case name could be fetched dynamiccaly from the task name. The subtasks could also be cleaned up a bit.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,6 +5,22 @@
   "version": "0.2.0",
   "configurations": [
     {
+      "name": "Debug itest",
+      "type": "go",
+      "request": "launch",
+      "mode": "exec",
+      "preLaunchTask": "reset before itest",
+      "program": "itest/itest.test",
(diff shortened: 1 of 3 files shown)


dstadulis:
@sputn1ck Might be qualified to review

Issue #828, opened 2024-08-24
Add version cli flag to litd


dstadulis:
Prior to issue completion:
Only litcli has version reporting while litd doesn't

```shell
./litcli-debug --version
litcli version 0.13.3-alpha commit=v0.13.3-alpha-11-ga9c576494247f2c94c44667a791f4d47a3126395
```

other lightninglabs software reports version of the actual daemon and not the cli interface

PR #830, opened 2024-08-25
config: add a "--version" config option 

ellemouton:
With this commit, a user can now run `litd --version` instead of needing
to run `litd --lnd.version` or `litcli --version` to obtain the LiTd
version.

Fixes https://github.com/lightninglabs/lightning-terminal/issues/828

### DIFF
--- a/config.go
+++ b/config.go
@@ -147,6 +147,8 @@ var (
 // all config items of its enveloping subservers, each prefixed with their
 // daemon's short name.
 type Config struct {
+	ShowVersion bool `long:"version" description:"Display version information and exit."`
+
 	HTTPSListen    string   `long:"httpslisten" description:"The main litd host:port to listen on for incoming HTTP/2 connections. On this port all gRPC services of all enabled daemons are exposed as well as REST (if --enablerest is specified), grpc-web and the web UI itself."`
 	HTTPListen     string   `long:"insecure-httplisten" description:"The host:port to listen on with TLS disabled. This is dangerous to enable as credentials will be submitted without encryption. Should only be used in combination with Tor hidden services or other external encryption."`
 	EnableREST     bool     `long:"enablerest" description:"Also allow REST requests to be made to the main HTTP(s) port(s) configured above."`
@@ -345,6 +347,10 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 	// Show the version and exit if the version flag was specified.
 	appName := filepath.Base(os.Args[0])
 	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
+	if preCfg.ShowVersion {
+		fmt.Println(appName, "version", Version())
+		os.Exit(0)
+	}
 	if preCfg.Lnd.ShowVersion {
 		fmt.Println(appName, "version", build.Version(),
 			"commit="+build.Commit)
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -147,6 +147,8 @@ var (
 // all config items of its enveloping subservers, each prefixed with their
 // daemon's short name.
 type Config struct {
+	ShowVersion bool `long:"version" description:"Display version information and exit"`
```
dstadulis, line 150: tACK c416ca21a056d35506976315ba7d65c5f26dee2c

config.go
```diff
@@ -345,7 +347,7 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 	// Show the version and exit if the version flag was specified.
 	appName := filepath.Base(os.Args[0])
 	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
-	if preCfg.Lnd.ShowVersion {
+	if preCfg.Lnd.ShowVersion || preCfg.ShowVersion {
```
dstadulis, line 350: seems the version number to be reported should be exclusively related to litd, when passing the `--version` flag, rather than lnd.

E.g.
Instead of lnd's 0.18.2 being reported here:
```shell
$ ./litd-debug --version
litd-debug version 0.18.2-beta commit=lightning-terminal-v0.13.3-alpha-1-gc416ca21a056d35506976315ba7d65c5f26dee2c
```

It should be
```shell
$ ./litd-debug --version
litd-debug version 0.13.4-alpha
```

e.g. `0.18.2-beta` changed to `0.13.4-alpha`

dstadulis (changes requested):
Thanks for the fast fix!
lmk your thoughts about which version number should ultimately be reporting

ViktorTigerstrom (approved):
Awesome, thanks :rocket:! tACK LGTM!

config.go
guggero: We've always had the combined version of lnd and litd.

config.go
guggero, line 150: ultra-nit: looks like the other entries in this section have a full stop at the end of their description...

config.go
```diff
@@ -345,6 +347,11 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 	// Show the version and exit if the version flag was specified.
 	appName := filepath.Base(os.Args[0])
 	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
+	if preCfg.ShowVersion {
+		fmt.Println(appName, "version", Version(),
+			"commit="+Commit)
```
guggero, line 352: The `Version()` function already returns `commit=...`.
This currently prints:
```
litd version 0.13.3-alpha commit=v0.13.3-alpha-12-gbaebcbdc4b86cb9453445102295cea8a030cdd57 commit=v0.13.3-alpha-12-gbaebcbdc4b86cb9453445102295cea8a030cdd57
```

guggero (changes requested):
Thanks! One small request (and a nit), otherwise looks good :tada:

dstadulis:
Will approve PR since other comments have addressed blocking fix requests

dstadulis (approved):

guggero:
@dstadulis is correct, it seems like we need to add the dot to the `semanticAlphabet` variable. We have it that way in `lnd` too.

config.go
ellemouton: ah true - thanks!

guggero (approved):

PR #833, opened 2024-08-28
github: allow api workflow dispatch

ellemouton:
so that the CI can be triggered via other repos


guggero (approved):

ellemouton:
looks like we wont need this anymore given the suggestion from @guggero  re how to do this without interacting with the lit repo :) just verifying locally first and then will close this

PR #843, opened 2024-09-13
Bump `litd` to version `v0.13.4-alpha`

ViktorTigerstrom:
This PR bumps the litd version to `v0.13.4-alpha`.

This PR also bumps lnd to version `v0.18.3-beta` & loop `v0.28.7-beta`.

As lnd `v0.18.3-beta` adds explicitly declared optional protos fields with commit https://github.com/lightningnetwork/lnd/commit/3de6c5415a44438632ed29acea2b47fc2f9d413e, we this PR also needed to update the protoc version used for the protos generation script.  Optional proto fields are only available in Proto3 after protoc version `3.15+`.

When doing so, we mimic the protobuf-compiler version used in `lnd`, which is `3.21.12`. As protobuf-compiler 3.21.x+ is only available for Node.js base images with Debian 12 (bookworm), we also update our NodeJS base image to version 22.8.0-bookworm.


proto/Dockerfile
```diff
@@ -1,10 +1,10 @@
 # Start with a NodeJS base image that also contains yarn.
-FROM node:16.14.2-buster as nodejsbuilder
+FROM node:22.8.0-bookworm as nodejsbuilder
 
 RUN apt-get update && apt-get install -y \
   git \
-  protobuf-compiler='3.6*' \
-  clang-format='1:7.0*'
+  protobuf-compiler='3.21.12*' \
```
ViktorTigerstrom, line 6: Let me know if you think we should update to this version for all other Docker files within the repo as well!

proto/Dockerfile
```diff
@@ -1,10 +1,10 @@
 # Start with a NodeJS base image that also contains yarn.
-FROM node:16.14.2-buster as nodejsbuilder
+FROM node:22.8.0-bookworm as nodejsbuilder
```
ViktorTigerstrom, line 2: @jamaljsr, let me know if you think this NodeJS base image version is ok, or if you'd like me to use `18.20-bookworm` instead :)!

ViktorTigerstrom:
Sorry, will solve the failing CI checks asap!

guggero (approved):
utACK, LGTM :tada:

docs/release-notes/release-notes-0.13.4.md
```diff
@@ -16,15 +16,13 @@ with 'make go-build-noui' and 'make go-install-noui'](https://github.com/lightni
 
 ### LND
```
ellemouton, line 18: LND also updated yeah?

ellemouton (approved):
Looks good - just a release notes comment

proto/Dockerfile
jamaljsr: I tested this and it works for me, so this is fine.

I would suggest bumping the image version in [Dockerfile](https://github.com/lightninglabs/lightning-terminal/blob/master/Dockerfile) and [dev.Dockerfile](https://github.com/lightninglabs/lightning-terminal/blob/master/dev.Dockerfile) as well, just to be consistent.

proto/Dockerfile
jamaljsr: We don't install `protoc` using `apt-get` in the other Docker files. We install it using NodeJS which you did already. So we should be all good with just this.

jamaljsr (approved):
tACK LGTM 🚀

Tested the `litd` UI and Terminal UI over LNC. No issues on my end.

docs/release-notes/release-notes-0.13.4.md
ViktorTigerstrom: Hmm yeah, I believe that's already in the release notes :)! Let me know if I'm misunderstanding your feedback though :)

ellemouton: my bad! was not viewing the side by side diff 🙈

docs/release-notes/release-notes-0.13.4.md
```diff
@@ -16,15 +16,13 @@ with 'make go-build-noui' and 'make go-install-noui'](https://github.com/lightni
...
 
-### Loop
-
-### Pool
+* [Update the integrated LND version to
+  v0.18.3-beta] (https://github.com/lightninglabs/lightning-terminal/pull/843)
```
ellemouton, line 20: still not rendering correctly. Need to remove the space between the [] and the ()

same below

ViktorTigerstrom: Thanks :pray:!

PR #844, opened 2024-09-13
multi: bake super macaroon during stateless init mode

ellemouton:
This PR adds the ability to use LiT to bake a super macaroon during stateless init mode. What's nice about this is that the macaroon will also only contain all the permissions for the subservers activated. So for example, if Loop is disabled, then the super macaroon will not contain permissions for loop.

TODO: add an itest. 

## Usage

Let's assume LiT and LND are in stateless init mode & that the `admin.macaroon` generated during wallet creation is stored in `~/mac_test/admin.macaroon`

1) This will fail due to `litcli` trying to find LiT's macaroon & not being able to.
```
litcli  bakesupermacaroon
```

2) This will also fail cause LiT will try to validate the given admin macaroon against its own macaroon service. 
```
litcli --macaroonpath=~/mac_test/admin.macaroon bakesupermacaroon
```

3) This will work. LiT will take the given macaroon & create a new connection to LND with it and attempt to bake the super macaroon with that connection (which will only be valid if the given macaroon was generated by LND and has the bakemacaroon permissions)
```
litcli --macaroonpath=~/mac_test/admin.macaroon bakesupermacaroon --stateless_init
```

## Technical Detail

In this commit, we force the permissions manager to see the
BakeSuperMacaroon call of LiT as whitelisted. This means that when the
initial call comes in, the `LightningTerminal.ValidateMacaroon` method
will return early and not validate the call (which would fail in
stateless init mode since LiT does not have a macaroon service in that
case). So the call ends up going through to the rpcProxy's
BakeSuperMacaroon method. Here we now have the following flow:
- if the request does not have stateless_init set, then we keep the flow
  as it was:
        1.  Use the lit mac validator to check that the call is allowed
           given the required permissions of `BakeSuperMacaroon`.
        2.  if it is, then use the existing connection to LND that Lit
           has to do the macaroon baking call.
- if stateless_init mode is set, then we do the following:
        1.    we extract the macaroon from the call (this should be an LND
           macaroon)
        2.  we create a new connection to LND using this provided
           macaroon.
        3.  use this connection to LND to bake the macaroon.


ellemouton:
will work on an itest for this 👍

ellemouton:
see https://github.com/lightninglabs/lightning-terminal/pull/858 for an alternative approach 🙏

PR #852, opened 2024-09-26
lightning-terminal: ensure reproducible releases

ViktorTigerstrom:
This update modifies the release.sh script to ensure that release files are reproducible, generating identical hashes when rebuilt. Previously, the script did not set fixed timestamps for the files, resulting in different hashes across rebuilds.

Additionally, the release.sh script has been updated to align more closely with the behavior of the lnd release script, ensuring consistency and coherence across repositories for users. Let me know if you think it's unnecessary for us to mimic those behaviour in `litd` :)

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -17,6 +17,61 @@ POOL_PKG="github.com/lightninglabs/pool"
 TAP_PKG="github.com/lightninglabs/taproot-assets"
 PACKAGE=lightning-terminal
 
+# Needed for setting file timestamps to get reproducible archives.
+BUILD_DATE="2020-01-01 00:00:00"
+BUILD_DATE_STAMP="202001010000.00"
+
+# reproducible_tar_gzip creates a reproducible tar.gz file of a directory. This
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -79,16 +141,21 @@ function build_release() {
...
     fi
-
-    rm -r "${dir}"
   done
 
-  shasum -a 256 * >manifest-$tag.txt
+  # Add the hash of the packages too, then sort by the second column (name).
+  shasum -a 256 lightning-terminal-* vendor* >> "manifest-$tag.txt"
+  LC_ALL=C sort -k2 -o "manifest-$tag.txt" "manifest-$tag.txt"
+  cat "manifest-$tag.txt"
```
ViktorTigerstrom, line 158: Let me know if you think this is unnecessary for `litd`, but I figured we'd probably want to mimic the `lnd` behaviour as much as possible to be as coherent between the repos as possible.

guggero:
Replaced by https://github.com/lightninglabs/lightning-terminal/pull/881.

PR #858, opened 2024-09-30
multi: bake super macaroon during stateless init mode (V2) 

### LINKED ISSUE #736: Add `StatelessInit` itests
We should add itests that runs and integrated lnd node in `StatelessInit` mode. These tests should test that all Subservers works as expected while also running in integrated mode.

The motivation for such itests is catch cases where macaroons don't get generated as expected when lnd in running in `StatelessInit` mode.

ellemouton:
An alternative approach to https://github.com/lightninglabs/lightning-terminal/pull/844

Looking for approach ACK here first (ie, ACK-ing this approach over #844) after which I'll work on an itest. 

Replaces #844 

In this commit, we special case the handling of BakeSuperMacaroon so as
to allow a user to make use of `litcli bakesupermacaroon` while LiT is
running in stateless init mode. The handling is as follows:
- if the call to _LiT's_ BakeSuperMacaroon is made while in stateless
  init mode then we can assume that the macaroon provided is either:
        1) an LND native macaroon which may or may not have the
           necessary permissions for the _LND_
           "/lnrpc.Lightning/BakeMacaroon" call.
        2) a baked macaroon (possibly a super macaroon) which may or may
           not have the permissions to the _LiT_
           "/litrpc.Proxy/BakeSuperMacaroon" call.

For case 1: we check that the provided macaroon has the correct perms.
If it does, then we use LiT's existing connection to LND to bake the
super mac.
For case 2: we have a macaroon that doesnt have LND's bakemac call perms
directly but does have LiT's BakeSuperMac perms. So for this, we treat
the call as normal and verify using LiT's macaroon validator as normal.

Fixes #736


ellemouton:
cc @ViktorTigerstrom for approach ACK here. 

Decided to open a new PR instead of pushing to the old PR since the approaches are completely different

ViktorTigerstrom (commented):
Definitely prefer this approach compared to #844, as it's much cleaner and easier to use for users. Awesome job :rocket::fire:!

There's one issue though that neither this or the other PR addresses unfortunately, that'll require some refactoring (potentially quite big, and I'm not quite sure we'll be able be able to address it easily😞):
Both of these PRs will require that the `macaroonService` has started. Now the reason why that's problematic is because that only happens in the `startInternalSubServers`, which is only executed **after** both `lnd` has been **fully synced**, and our internal sub-servers have been started. Now the reason that's problematic is that the feedback we've gotten from our users is that they wouldn't like to wait for `lnd` to be fully synced before macaroons can be generated.

I think we miiight potentially be able to make a `bakesupermacaroon` call in `stateless_init` mode work before `lnd` has been fully synced, as it only require a Basic Lnd Client (and it's the Full Lnd Client in `setUpLNDClients` that requires the chain sync). The perms for all sub-servers are also added in `initSubServers` which is executed way earlier, so the perms manager is aware of all different sub-servers permissions before the sub-servers are started.

If you do think that's a user requirement this PR/functionality shouldn't address, let me know, as I then prefer this approach over the other PR!

ellemouton:
Thanks for the feedback @ViktorTigerstrom  - although, iiuc, This should not be an issue since we should not be hitting LiT's macaroon service for this call if an LND admin macaroon is passed in. We only hit that if there is an error validating the macaroon via _or_ if the `statelessinit` boolean is still false. So I think your concern is easily solved by just moving the spot that the `statelessinit` boolean gets set. I have tested this locally with an LND node that never syncs and am able to bake the macaroon just fine. Let me know if im misunderstanding something

ellemouton:
@ViktorTigerstrom  - see latest commit

ViktorTigerstrom:
thnaks for the response and update @ellemouton! 

> I have tested this locally with an LND node that never syncs and am able to bake the macaroon just fine. Let me know if im misunderstanding something

Oh interesting, I hit the "the macaroon service has not started yet" when testing it locally during the review. I'll test again with your changes though, as I there may have been some issue on my side :)

ellemouton:
> as I there may have been some issue on my side

No i dont think there was a specific issue on your side - it makes sense that we would hit that error if the `statelessinit` boolean is not yet set.

ViktorTigerstrom (approved):
tACK LGTM 🚀🎉, after fixing the release notes issue!

You were right Elle, I think I mistook that error due to the late assignment that's now fixed 🙂!

Really like that this functionally can handle any of lnd admin macs, previously baked macs, and also normal lit.macaroon:s 🔥

ellemouton:
will try get round to adding an itest this week.

terminal.go
```diff
@@ -689,16 +688,16 @@ func (g *LightningTerminal) start() error {
...
 		macService := g.lndInterceptorChain.MacaroonService()
-		createDefaultMacaroons = !macService.StatelessInit
+		g.cfg.statelessInitMode = macService.StatelessInit
 	}
 
 	// Both connection types are ready now, let's start our sub-servers if
 	// they should be started locally as an integrated service.
 	g.subServerMgr.StartIntegratedServers(
-		g.basicClient, g.lndClient, createDefaultMacaroons,
+		g.basicClient, g.lndClient, !g.cfg.statelessInitMode,
```
guggero, line 688: nit: to give this boolean a bit more context for how it's being used by the subservers, should we just do `createDefaultMacaroons := !g.cfg.statelessInitMode` above this function call?

rpc_proxy.go
```diff
@@ -183,6 +186,10 @@ type rpcProxy struct {
 // bakeSuperMac can be used to bake a new super macaroon.
 type bakeSuperMac func(ctx context.Context, rootKeyID uint32) (string, error)
 
+// lndBasicClient can be used to obtain access to an lnrpc.LightningClient if
+// it is available.
+type lndBasicClient func() (lnrpc.LightningClient, error)
```
guggero, line 194: nit: To make it more obvious this is a function type, should it be called `lndBasicClientFn` or `lndBasicClientGetter` or something like that?

rpc_proxy.go
```diff
@@ -410,6 +413,87 @@ func (p *rpcProxy) UnaryServerInterceptor(ctx context.Context, req interface{},
...
 	}
 
+	// We special case the handling of a call to LiT's BakeSuperMacaroon
+	// method if we are in stateless init mode. If a user has called this
+	// method, they are either calling with an LND macaroon or a baked
+	// super macaroon.
+	if p.cfg.statelessInitMode &&
+		info.FullMethod == litBakeSuperMacMethod {
+
+		lndClient, err := p.getBasicLNDClient()
```
guggero, line 423: nit: move this call to just before we use the client?

guggero (approved):
tACK, LGTM :tada:

guggero (changes requested):
Marking this as "request changes" until we have an itest, which would definitely be nice.

itest/network_harness.go
```diff
@@ -348,6 +350,61 @@ func (n *NetworkHarness) newNode(t *testing.T, name string, extraArgs,
...
+	// With the seed created, construct the init request to the node,
+	// including the newly generated seed.
+	initReq := &lnrpc.InitWalletRequest{
+		WalletPassword:     walletPassword,
+		CipherSeedMnemonic: genSeedResp.CipherSeedMnemonic,
+		AezeedPassphrase:   walletPassword,
+		StatelessInit:      statelessInit,
+	}
+
+	// Init the
```
guggero, line 397: nit: incomplete sentence.

guggero (approved):
Very nice :tada:

itest/litd_stateless_init_test.go
```diff
@@ -0,0 +1,62 @@
...
+// is running in stateless-init mode.
+func testStatelessInitMode(_ context.Context, net *NetworkHarness,
+	t *harnessTest) {
+
+	// Set up a new node (charlie) in stateless init mode (create and
+	// unlock).
+	walletPassword := []byte("stateless")
+	charlie, adminMac := net.NewNodeWithSeed(
+		t.t, "Charlie", nil, walletPassword, false, true,
+	)
```
ViktorTigerstrom, line 21: Should we maybe add test coverage for what the intended behaviour is when attempting to use a remote lnd node + stateless init mode, as we've had users who have attempted that? Else feel free to disable setting both of them to true in the `NewNodeWithSeed` function.

ViktorTigerstrom (approved):
tACK LGTM :rocket:! Great work :fire:!

Thanks a lot for the itest + the option to also create a `read_only` macaroon :)

itest/litd_stateless_init_test.go
ellemouton: what would such a test look like though? cause lit would just fail to connect to lnd cause there would be no macaroon to authenticate itself

ellemouton: > Else feel free to disable setting both of them to true in the NewNodeWithSeed function.

dont think we need to explicitly guard against the set-up though - the test will fail either way :)

ellemouton: gonna merge this in the mean time & we can continue discussion offline with possible follow up 👍

Issue #863, opened 2024-10-09
make+build: Create globally-consistent, build-input identifier


dstadulis:
## Background
- During the build process, there is a motivation to create a consistent identifier for what code and resources were used to create build artifacts/binaries.
	- This identifier should communicate
		- What notional code basis was used for the build
		- The presence of any modifications to that code
- Factors affecting such an identifier:
	- Root cause of https://github.com/lightninglabs/lightning-terminal/issues/854 is
		- Inconsistent client-side git.config settings may result in globally inconsistent commit-hash abbreviation between different hosts
		- a git instance's [core.abbrev](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreabbrev) configuration value dictates how many characters git describe will produce

The manifest file's in-situ build-artifacts identifier entries are produced here:
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/release.sh#L91

Assigned from here:
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/release.sh#L33

Called from here
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/Makefile#L164C30-L164C45

Defined here:
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/make/release_flags.mk#L1C23-L1C42

## Proposed Solution
- Modify [VERSION_TAG](https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/make/release_flags.mk#L1C23-L1C42) to use enhanced identifier
	- [`git describe` Documentation:](https://git-scm.com/docs/git-describe)
	- git describe --abbrev=40 --broken --tags --always

## Design-choice interrogation
- --always
	- Produce uniquely abbreviated commit object as fallback, regardless if tags precede checked-out commit
	- Forces `git describe` to produce the identifier, in the rare scenario, for commits which have no proceeding tags
- --broken
	- When the working tree matches HEAD, the output is the same as "git describe HEAD".
	- If the working tree has local modification "-dirty" is appended to it.
	- If a repository is corrupt
		- and Git cannot determine if there is local modification, Git will error out
			- Enumerate cases in which having the build identifier note a corrupted build
[...]

Issue #854, opened 2024-09-26
`./release.sh` produces tags of different lengths, across certain hosts


dstadulis:
# Background
Digest entries in the manifest file are produced here:
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/release.sh#L91

Assigned from here:
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/release.sh#L33

Called from here
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/Makefile#L164C30-L164C45

Defined here:
https://github.com/lightninglabs/lightning-terminal/blob/43ac9fac94cd158e1b7c560555f2a7b565c307f9/make/release_flags.mk#L1C23-L1C42

# Issue
@ViktorTigerstrom produced a build **with** an **extra** character, `a`, at the end of the tag
```log
lightning-terminal-darwin-amd64-v0.13.995-experimental-1-g8da6c29a
```

On two different hosts[1], after running
```bash
make release tag=v0.13.995-experimental
```

a build with tags **without** the `a` character, at the end of the tag are produced
```log
lightning-terminal-darwin-amd64-v0.13.995-experimental-1-g8da6c29
```

[1] Hosts running:
git version 2.46.2
git version 2.30.2

# Reproduce
- [ ] Reproduce git tag with extra `a` character
  - @ViktorTigerstrom report the output of 
git describe --tags
git --version
- [ ] Determine any differences in `git describe` has occurred between the git versions

ViktorTigerstrom:
@dstadulis thanks! 

`git describe --tags` when having pr #853 checked out results in:
`v0.13.995-experimental-1-g8da6c29a`

`git --version` is:
`git version 2.45.2`

guggero:
Just a general/drive-by comment that `litd` release builds are _NOT_ yet fully reproducible. I think there might be other things than just the git tag that could cause differences in the build (biggest unknown being the frontend/UI/NodeJS build, not sure if that can even be made reproducible in the first place).

dstadulis:
have a fix for the git tag issue, targeting to post soon.

Will create an issue cataloging the factors that affect reproducibility

dstadulis:
Fixed in https://github.com/lightninglabs/lightning-terminal/pull/864

PR #864, opened 2024-10-09
make: update VERSION_TAG to a globally-consistent build-input identifier

dstadulis:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/863 https://github.com/lightninglabs/lightning-terminal/issues/854

Additional flags produce:
* Globally consistent commitment hash length, regardless of gitconfig
* `--broken` to create fail-safe tag indicating a dirty or corrupted
  work tree
* `--always` to generate a tag, even if no tags are reachable from the
  current commit.

### DIFF
--- a/make/release_flags.mk
+++ b/make/release_flags.mk
@@ -1,4 +1,5 @@
-VERSION_TAG = $(shell git describe --tags)
+# Create a globally-consistent, build-input identifier.
+VERSION_TAG = $(shell git describe --abbrev=40 --broken --tags --always)
 VERSION_CHECK = @$(call print, "Building master with date version tag")
 
 BUILD_SYSTEM = darwin-amd64 \


make/release_flags.mk
```diff
@@ -1,4 +1,5 @@
-VERSION_TAG = $(shell git describe --tags)
+# Create globally-consistent, build-input identifier
```
guggero, line 1: nit: comments should always be full sentences, including punctuation.

guggero (approved):

make/release_flags.mk
dstadulis: i.e. "Add a full stop"?
CMIIW but "{Imperative verb} {adjective} {direct objective}" is a complete sentence.

ellemouton (approved):
LGTM once other comment is addressed.

make/release_flags.mk
ellemouton: yup, just update it to: `# Create globally-consistent, build-input identifier.`

ellemouton: i'll do it 👍

PR #871, opened 2024-10-15
itest: add test for consistent balance reporting

gijswijs:
We want proper asset balance reporting between lnd and tapd, without asset balances turing up in both the asset balance from tapd and the custom channel balance from lnd.

Currently this behavior is not aligned with that requirement. tapd has currently now way of not reporting asset balances that are actually part of the funding transaction of a custom channel. This is fixed with [taproot-assets PR 1151](https://github.com/lightninglabs/taproot-assets/pull/1151) 

The current PR under review here is dependent on the taproot-assets commit:
https://github.com/lightninglabs/taproot-assets/commit/31d333cce24bf1c20a43b6b3343fd4914713e8c3

Bumping the tap version was included in this commit 5bbf3d7424bcfaffaea3c01c6b47eb4f6fa791f1 and that commit should probably be edited once 1151 is merged.

### DIFF
--- a/docs/release-notes/release-notes-0.13.4.md
+++ b/docs/release-notes/release-notes-0.13.4.md
@@ -4,9 +4,9 @@
 
 ### Lightning Terminal
 
-- [Fixed a bug where REST calls for the `WalletUnlocker` service weren't allowed
+* [Fixed a bug where REST calls for the `WalletUnlocker` service weren't allowed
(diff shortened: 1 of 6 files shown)


guggero:
Needs a rebase after changing the base branch.

guggero (approved):
One `go.mod` fix, other than that looks good to me :100:

guggero (approved):

PR #848, opened 2024-09-19
Prepare for custom channels (lnd v0.18.4-beta)

guggero:
This is the mother of all PRs for the custom channel saga. Merging this PR will be the final step for bringing custom channels to a non-experimental version of litd. There are many dependencies (see list/graph below), so it will remain in draft for a bit.
Most of the code in these PRs have already been reviewed on the `0-19-staging` branch and have just been squashed down into easy-to-review pieces.

The following list is an **ordered** list of dependencies that need to be resolved before this PR can be merged (cc @dstadulis):

 - [x] Review and merge https://github.com/lightninglabs/lndclient/pull/192
 - [x] Review and merge https://github.com/lightningnetwork/lnd/pull/8960
 - [x] Create release branch PR with stable commits (https://github.com/lightningnetwork/lnd/pull/9183)
 - [x] Push tag `lndclient v0.18.4-0` that points to release branch with stable commits
 - [x] Update https://github.com/lightninglabs/pool/pull/488 to point to `lndclient v0.18.4-0` (and with it also bumping the `lnd` version), take out of draft
 - [x] Review and merge https://github.com/lightninglabs/pool/pull/488
 - [x] Update https://github.com/lightninglabs/loop/pull/828 to point to `lndclient v0.18.4-0` (and with it also bumping the `lnd` version), take out of draft
 - [x] Review and merge https://github.com/lightninglabs/loop/pull/828
 - [x] Update https://github.com/lightninglabs/taproot-assets/pull/1130 to point to `lndclient v0.18.4-0` (and with it also bumping the `lnd` version), take out of draft
 - [x] Review and merge https://github.com/lightninglabs/taproot-assets/pull/1130
 - [x] Code, review and merge HTLC sweeping saga:
   - [x] https://github.com/lightningnetwork/lnd/issues/9176
   - [x] https://github.com/lightninglabs/taproot-assets/pull/1154
   - [x] https://github.com/lightningnetwork/lnd/pull/9194
 - [x] Wait for `lnd v0.18.4-beta` to be tagged (at least an RC tag) https://github.com/lightningnetwork/lnd/issues/9173
 - [x] https://github.com/lightninglabs/lndclient/issues/194 point to the tagged version of `lnd`
 - [x] Push tag `lndclient v0.18.4-6`
 - [x] ~Create and merge Pool PR to reference `lndclient v0.18.4-6`~ (not required, `litd` compiles without)
[...]

PR #874, opened 2024-10-16
itest: add <354 sats payment coverage for custom channels

GeorgeTsagk:
## Description

This PR adds coverage for the scenario where an asset invoice worth less than 354 sats is paid for with sats (from a node that is not the RFQ peer).

This case triggers a specific code check in lnd where the incoming/outgoing sats produce a negative fee
(see [related issue](https://github.com/lightninglabs/taproot-assets/issues/1150))

The current expected status of the payment is set to `lnrpc.PAYMENT_FAILED` as this has not yet been patched.
This itest will be updated in order to expect a `lnrpc.PAYMENT_SUCCEEDED` once this behavior is fixed in lnd
(see [related PR](https://github.com/lightningnetwork/lnd/pull/8804) for tracking)

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -682,7 +682,7 @@ func createAndPayNormalInvoiceWithBtc(t *testing.T, src, dst *HarnessNode,
 	})
 	require.NoError(t, err)
 
-	payInvoiceWithSatoshi(t, src, invoiceResp)
+	payInvoiceWithSatoshi(t, src, invoiceResp, lnrpc.Payment_SUCCEEDED)
 }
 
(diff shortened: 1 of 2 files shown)


dstadulis:
Coverage for the current behavior.
Expected behavior will change.
Need to link to a negative fee check.

guggero:
This PR should now base against the `update-to-lnd-18-4` branch, there shouldn't be anything going to the `0-19-staging` branch anymore, with how close we are to merging the "real" branch.

guggero:
Cherry picked commit 9c104455 onto https://github.com/lightninglabs/lightning-terminal/pull/848.

PR #877, opened 2024-10-21
subserver+status: check map entries before adding new servers

ellemouton:
1) For the subserver manager: use a map instead of an array for subservers. Check that there is no name collision before adding a new server. This is also nice if we want to get a specific subserver by its name (as is done in https://github.com/lightninglabs/lightning-terminal/pull/848). 

2) for the status manager, first check that we are not overriding an existing subserver before we register a new one

### DIFF
--- a/docs/release-notes/release-notes-0.13.5.md
+++ b/docs/release-notes/release-notes-0.13.5.md
@@ -32,6 +32,9 @@
 ### Technical and Architectural Updates
 
 * [Convert litrpc package into a module](https://github.com/lightninglabs/lightning-terminal/pull/823).
+
+* Check internal maps before registering new sub-servers for both the 
+  [status and sub-server manager.](https://github.com/lightninglabs/lightning-terminal/pull/877)
  
 ### Autopilot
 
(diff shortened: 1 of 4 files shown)


guggero (approved):
LGTM :tada:

docs/release-notes/release-notes-0.13.5.md
```diff
@@ -1,21 +1,44 @@
...
+  - [Bug Fixes](#bug-fixes)
+  - [Functional Changes/Additions](#functional-changesadditions)
+  - [Technical and Architectural Updates](#technical-and-architectural-updates)
+- [Integrated Binary Updates](#integrated-binary-updates)
+  - [LND](#lnd)
+  - [Loop](#loop)
+  - [Pool](#pool)
+  - [Faraday](#faraday)
+  - [Taproot Assets](#taproot-assets)
+- [Contributors](#contributors-alphabetical-order)
```
ViktorTigerstrom, line 13: Niiiice 😎!

ViktorTigerstrom (approved):
Thanks! LGTM!

PR #883, opened 2024-10-23
itest: add RPC oracle with spread based asset rates

guggero:
Depends on https://github.com/lightninglabs/lightning-terminal/pull/848.

This commit adds a fully configurable RPC oracle that can serve asset rates that differ for buy and sell.
We then add a test that shows and asserts proper asset conversion to satoshis and back.

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -61,8 +61,8 @@ func createTestAssetNetwork(t *harnessTest, net *NetworkHarness, charlieTap,
 	daveTap, erinTap, fabiaTap, yaraTap, universeTap *tapClient,
 	mintedAsset *taprpc.Asset, assetSendAmount, charlieFundingAmount,
 	daveFundingAmount,
-	erinFundingAmount uint64, pushSat int64) (*tchrpc.FundChannelResponse,
-	*tchrpc.FundChannelResponse, *tchrpc.FundChannelResponse) {
+	erinFundingAmount uint64, pushSat int64) (*lnrpc.ChannelPoint,
+	*lnrpc.ChannelPoint, *lnrpc.ChannelPoint) {
 
 	ctxb := context.Background()
 	assetID := mintedAsset.AssetGenesis.AssetId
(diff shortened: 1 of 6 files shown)


itest/assets_test.go
```diff
@@ -586,6 +606,65 @@ func getAssetChannelBalance(t *testing.T, node *HarnessNode, assetID []byte,
...
+
+	require.Len(t, assetBalance.Assets, 1)
+
+	require.InDelta(t, local, assetBalance.Assets[0].LocalBalance, 1)
+	require.InDelta(t, remote, assetBalance.Assets[0].RemoteBalance, 1)
+}
+
+func routingFee(amt lnwire.MilliSatoshi) lnwire.MilliSatoshi {
+	return amt + (amt / 1000_000) + 1000
+}
```
ffranr, line 666: consider adding a comment to explain this math: why is that the routing fee?

itest/assets_test.go
```diff
@@ -825,19 +908,103 @@ func createAssetInvoice(t *testing.T, dstRfqPeer, dst *HarnessNode,
...
+		"SCID %d", decodedInvoice.NumMsat, mSatPerUnit,
+		dstRfqPeer.PubKey[:], resp.AcceptedBuyQuote.Scid)
 
 	return resp.InvoiceResult
 }
 
+// assertInvoiceHtlcAssets makes sure the invoice with the given hash shows the
+// individual HTLCs that arrived for it and that they show the correct asset
+// amounts when decoded.
+func assertInvoiceHtlcAssets(t *testing.T, node *HarnessNode,
```
ffranr, line 929: I don't think this function is used yet in the codebase. I think we should add this to the PR where its used.

ffranr: I think this came from your other PR. It's also in your other PR.

itest/assets_test.go
guggero: Updated.

jharveyb (approved):
LGTM! Took me a sec to be convinced that the rates are being used correctly but the comments are very thorough.

PR #892, opened 2024-10-30
Add itest for MinRelayFee check in tapd

gijswijs:
This commit adds an integration test for the MinRelayFee check. The test ensures that transactions with fees below the minimum relay fee are rejected.

This PR depends on https://github.com/lightninglabs/taproot-assets/pull/1163 which in turn depends on  https://github.com/lightninglabs/lndclient/pull/200

It uses pseudoversions in `go.mod` for ~both~ `taproot-assets` ~and `lndclient`~ that should be updated before merging.

### DIFF
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/lightningnetwork/lnd/lntest"
 	"github.com/lightningnetwork/lnd/lntest/port"
 	"github.com/lightningnetwork/lnd/lntest/wait"
+	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/lightningnetwork/lnd/lnwire"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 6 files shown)


gijswijs:
I've removed the default value of 1 for `sat_per_vbyte` that's being passed to `litd` rpc endpoint `fundchannel`. `litd` handles this now perfectly on its own, and adding logic here only adds to the confusion.

itest/litd_custom_channels_test.go
```diff
@@ -2637,3 +2638,87 @@ func testCustomChannelsOraclePricing(_ context.Context,
...
+	assetID := cents.AssetGenesis.AssetId
+
+	t.Logf("Minted %d lightning cents, syncing universes...", cents.Amount)
+	syncUniverses(t.t, charlieTap, dave)
+	t.Logf("Universes synced between all nodes, distributing assets...")
+
+	// Fund a channel with a fee rate of zero.
+	zeroLowFeeRate := uint32(0)
+
+	_, err = charlieTap.FundChannel(
```
jharveyb, line 2694: AFAICT this command will also fail if no feerate is specified - is that the behavior we want here?

To be clear, I think that's the behavior we have right now. Mentioning it here because that differs from what we do for asset minting or on-chain transfer, where a value of 0 feerate causes tapd to fetch a fee estimate and use that for the TX.

jharveyb (approved):
LGTM 👍🏽 

One related comment about how we handle this feerate arg, but it can be addressed in a separate PR.

itest/litd_custom_channels_test.go
gijswijs: I checked and your assumption is correct. Passing no feerate vs passing a feerate of zero yields the exact same result. We don't use fee estimation when funding channels. If we want that we could make a seperate issue as a feature request.

itest/litd_custom_channels_test.go
```diff
@@ -2637,3 +2638,87 @@ func testCustomChannelsOraclePricing(_ context.Context,
...
+		ctxb, &tchrpc.FundChannelRequest{
+			AssetAmount:        cents.Amount,
+			AssetId:            assetID,
+			PeerPubkey:         daveTap.node.PubKey[:],
+			FeeRateSatPerVbyte: tooLowFeeRate,
+			PushSat:            0,
+		},
+	)
+
+	errFeeRateTooLow := fmt.Sprintf("fee rate %s too low, "+
```
guggero, line 2721: This actually shows a problem with the sat/vByte fee rate... Because 1 sat/vByte is equal to 250 sat/kw but the absolute minimum relay fee there can ever be is defined as 253 sat/kw, this would mean that you could only fund channels with at least 2 sat/vBytes. Which is kind of unfortunate.
But I think we should fix this in tapd and unrelated to this PR. Just letting you know that this might change again in the future.

guggero (approved):

Issue #895, opened 2024-11-06
Host for Syncing Universe State is down since a while now. (universe.lightning.finance)


GTCLive:
Periodically filling the logs with _failed to connect_. Not sure if that's a good thing, neither know what is this fed universe thing is, but thought of reporting this for it to be removed or amended with a functional host. 

# lnd.log
`2024-11-06 11:59:57.432 [INF] UNIV: Synchronizing with 1 federation members`
`2024-11-06 11:59:57.437 [INF] UNIV: Syncing Universe state with server=(universe.ServerAddr) {`
` ID: (int64) 1,`
` addrStr: (string) (len=32) "universe.lightning.finance:10029",`
` addr: (net.Addr) <nil>`
`}`

`2024-11-06 11:59:57.438 [INF] UNIV: Attempting to sync universe: host=universe.lightning.finance:10029, sync_type=full, ids=([]universe.Identifier) <nil>`

`2024-11-06 11:59:57.647 [INF] UNIV: Fetching 0 roots`
`2024-11-06 11:59:57.647 [INF] UNIV: Obtained 0 roots from remote Universe server`
`2024-11-06 11:59:57.648 [WRN] GRPC: [core] [Channel #155 SubChannel #156]grpc: addrConn.createTransport failed to connect to {Addr: "35.82.128.37:10029", ServerName: "35.82.128.37:10029", }. Err: connection error: desc = "transport: Error while dialing: dial tcp 35.82.128.37:10029: operation was canceled"`

![Screenshot_2024-11-06_12-09-03](https://github.com/user-attachments/assets/01b1db2e-4c8d-4993-905f-b984d3b0dae9)

dstadulis:
Hmm unable to reproduce:
```shell
$ nc -zv universe.lightning.finance 10029
Connection to universe.lightning.finance port 10029 [tcp/*] succeeded!
$ nc -zv 35.82.128.37 10029
Connection to 35.82.128.37 port 10029 [tcp/*] succeeded!
```

GTCLive:
Mind sharing your dns provider so we can hook up to them? Prolly easier that way...

![Screenshot_2024-11-06_18-18-34](https://github.com/user-attachments/assets/890ea164-bd6a-4ebe-abe5-fb9281953b05)

guggero:
I don't think it's a DNS issue, as we seem to get the same IP address. I think the confusion comes from the server not answering to `ping`, which is just a firewall configuration.
[...]

PR #896, opened 2024-11-07
config: disable internal GRPC logger by default

guggero:
The GRPC connection logger is very verbose and normally not very useful. So it leads to more confusion and unnecessary log bloat than it actually helps to debug things.
So we disable it by default, meaning that if GRPC=<level> doesn't appear in the log level config string, we add GRPC=off.
That means we can still manually turn it on by adding ,GRPC=info to the log config (e.g. --lnd.debuglevel=debug,GRPC=info).

See https://github.com/lightninglabs/lightning-terminal/issues/895 for an example where the log is less than useful.


ViktorTigerstrom (approved):
Awesome, thanks. tACK LGTM :rocket:!

PR #899, opened 2024-11-07
config+terminal: allow configuring of lnd RPC timeout

guggero:
Allows the RPC timeout for the single `lndclient` instance that's given to all subservers to be configured.

Related to https://github.com/lightninglabs/taproot-assets/pull/1177.


ellemouton (approved):
Nice! this one I think makes sense to add to release notes since it's a new lit config flag 🙏

terminal.go
```diff
@@ -74,6 +74,7 @@ const (
 
 	defaultServerTimeout  = 10 * time.Second
 	defaultConnectTimeout = 15 * time.Second
+	defaultRPCTimeout     = 1 * time.Minute
```
ViktorTigerstrom, line 77: The actual default timeout used in `lndclient` `lnd_services.go` is 30 seconds:
https://github.com/lightninglabs/lndclient/blob/d89c44c7ef25fdf72c901876b6961ebee7d0c133/lnd_services.go#L30

Would we like to mimic that in `litd`, or would we specifically like to set it to 1 minute in `litd`?

terminal.go
```diff
@@ -873,6 +874,7 @@ func (g *LightningTerminal) setUpLNDClients(lndQuit chan struct{}) error {
 				BlockUntilUnlocked:    true,
 				CallerCtx:             ctxc,
 				CheckVersion:          minimalCompatibleVersion,
+				RPCTimeout:            g.cfg.LndRPCTimeout,
```
ViktorTigerstrom, line 877: Hmmmmm I want to also note that this modifies the timeout used in `lndclient` during the initial chain sync. If 

We've had many users which have experienced timeout issues during the initial chain sync, and from my experience i'd sometimes get a response after like 20 seconds when the syncing issues occur.
Here's an issue report for that specific issue:
https://github.com/lightningnetwork/lnd/issues/9108

I do believe that setting this value to a too low value would likely makes those syncing issues even worse when they actually occur.

Therefore: would we like to set a lower limit for this value, just so that some users don't set it to like 1 second? Or I'd at least mention in the config flag docs that it's recommended to not set this value to a too low value.

ViktorTigerstrom (approved):
Generally LGTM :rocket:! Thanks!

Posting 2 questions below that might be worth addressing.

terminal.go
guggero: Yes, this is an intentional bump. 30 seconds probably is too low already, especially looking at your other comment. Perhaps we should bump it even more?

terminal.go
guggero: Yeah, good point. I added a startup check and also mentioned it in the release notes.

terminal.go
ViktorTigerstrom: Thanks for clarifying! 

> Perhaps we should bump it even more?

Yeah maybe it'd make sense to bump it to 2 or 3 minutes even. Though I'd be ok with setting this default to 1 minute for now, and adjust it in future if needed. Feel free to set this to whatever you'd prefer of those alternatives :)!

terminal.go
ViktorTigerstrom: Awesome thanks :pray:!

ViktorTigerstrom (approved):
tACK LGTM :tada:! Thanks for the updates :pray:!

PR #906, opened 2024-11-22
itest: Extend custom channel liquidity itest for RFQ HTLC tracking

GeorgeTsagk:
## Descriptions

Adds an extra edge case to the custom channels itest which verifieds that the rfq htlc tracking mechanism works as expected.

Related tapd PR: https://github.com/lightninglabs/taproot-assets/pull/1186


ffranr:
@GeorgeTsagk do I only need to review the last three commits?

GeorgeTsagk:
@ffranr this was just rebased, it's just one commit now (if you skip the temp one)

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2050,166 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+	case <-done:
+		// If the payment completes before the slosh payment occurs then
+		// something went wrong, this is not the expected test case
+		// behavior.
+		t.Fatalf("payment completed before expected block period")
+	case <-timeoutChan:
+		// The expected time delay for the payment to complete has been
+		// passed, so now we wait for the payment to complete.
+		<-done
+	}
```
ffranr, line 2125: We have `BeforeTimeout` as a helper function. Would be good to have `AfterTimestamp` perhaps. But I think this sort of timing analysis strategy should be events driven instead.

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2050,166 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+	// payment is still trying to find a route, therefore the HTLC rejection
+	// works.
+	timeoutChan = time.After(PaymentTimeout / 2)
+	done = make(chan bool, 1)
+
+	ctxc, cancel := context.WithCancel(context.Background())
+
+	//nolint:lll
+	go func() {
+		// payInvoiceWithSatoshi(t.t, erin, iResp, lnrpc.Payment_FAILED)
```
ffranr, line 2189: needs to be removed?

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2050,166 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+		sendReq := &routerrpc.SendPaymentRequest{
+			PaymentRequest:   iResp.PaymentRequest,
+			TimeoutSeconds:   int32(PaymentTimeout.Seconds()),
+			MaxShardSizeMsat: 80_000_000,
+			FeeLimitMsat:     1_000_000,
+		}
+		stream, err := erin.RouterClient.SendPaymentV2(ctxc, sendReq)
+		if err == nil {
+			_, _ = getPaymentResult(stream)
+		}
```
ffranr, line 2199: I think you should use this patter here:
```
stream, err := erin.RouterClient.SendPaymentV2(ctxc, sendReq)
require.NoError(t.t, err)

_, _ = getPaymentResult(stream)
```

Need error check on `getPaymentResult` result?

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2050,166 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+
+	// Fabia creates an asset invoice of 100k assets. There is currently
+	// not enough asset liquidity on Erin's side to forward HTLCs to satisfy
+	// this invoice.
+	invoiceResp = createAssetInvoice(t.t, erin, fabia, 100_000, assetID)
+
+	// We set the waiting period for the slosh payment to occur. This is set
+	// to half of the default payment timeout, as we want it to occur half
+	// way through the in-flight payment.
+	sloshWait := PaymentTimeout / 3
```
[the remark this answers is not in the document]
GeorgeTsagk: ah yes, will either update comment or value in the next push

i think i made it `/ 3` just to wait a bit less overall on the test

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2050,166 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+		sendAssetKeySendPayment(
+			t.t, fabia, erin, 75_000, assetID,
+			fn.None[int64](), lnrpc.Payment_SUCCEEDED,
+			fn.None[lnrpc.PaymentFailureReason](),
+		)
+	}()
+
+	// To avoid goroutine uncertainty, we wait as much as the above routine
+	// minus a small delta. This is enough for us to be sure that the
+	// payment will complete due to the slosh.
```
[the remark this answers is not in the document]
GeorgeTsagk: This is a bit tricky. Under the hood we know that some HTLCs of some size are going to be forwarded and eventually cancelled. If we were to intercept (which adds some overhead in code complexity) we would also run into the issue of not knowing what to do with the HTLC we just intercepted. The reason the time based approach is nice here is because it's agnostic to the internal mechanisms of sender/router/receiver.

We don't know what's going on inside, but we want:
a) Sender to try to send payment and have their attempts failing
b) Slosh payment to complete ~halfway through the payment timeout
c) Sender to eventually complete payment

We could intercept HTLCs to guarantee that any attempt was made during a), but other than that could we use it any different way?

itest/litd_custom_channels_test.go
GeorgeTsagk: thanks, will look into `BeforeTimeout` and see if it can save us some lines

itest/litd_custom_channels_test.go
GeorgeTsagk: yeap leftover, ty

itest/litd_custom_channels_test.go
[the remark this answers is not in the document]
ffranr: You call into `payInvoiceWithAssets` below this point. And internally that handles a stream:
https://github.com/lightninglabs/lightning-terminal/blob/e52a8c563321f4fcc71c0e5bb22da56f0531a6dc/itest/assets_test.go#L888

Could you perhaps call tapd `SendPayment` in your test and then monitor the stream?

I think that the stream should have a payment result payment status of in-flight before sloshing. And then after sloshing you can listen on the stream for a payment result payment status success.

That way you can be sure that the send attempt is in-flight and then is successful after sloshing. And without timing complexity or perhaps any extra goroutines?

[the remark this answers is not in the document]
GeorgeTsagk: After latest foce-push I'm using a new approach, so PTAL

itest/litd_custom_channels_test.go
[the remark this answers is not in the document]
GeorgeTsagk: deprecated

itest/litd_custom_channels_test.go
GeorgeTsagk: not needed anymore with new approach

itest/litd_custom_channels_test.go
GeorgeTsagk: latest approach uses existing helpers, closing

itest/litd_custom_channels_test.go
```diff
@@ -1805,7 +1809,7 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
 			SatPerVByte: 5,
 		},
 	)
-	defer closeChannelAndAssert(t, net, dave, channelOp, false)
+	defer closeChannelAndAssert(t, net, dave, channelOp, true)
```
guggero, line 1812: Why do we need to force close here now?

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2051,176 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+
+	// Charlie tries to pay via Dave, by providing the scid that was
+	// manually created above.
+	scid := rfqmsg.SerialisedScid(resQ.GetAcceptedQuote().Scid)
+	payInvoiceWithAssets(
+		t.t, charlie, dave, hodlInv.payReq, assetID, true,
+		fn.Some(lnrpc.Payment_IN_FLIGHT), fn.Some(scid),
+	)
+
+	time.Sleep(time.Second)
```
guggero, line 2098: No need to manually sleep here I think. `assertNumHtlcs` will wait in a loop until the required number is reached.

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2051,176 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+	assertNumHtlcs(t.t, fabia, 0)
+
+	// Now Fabia creates the normal invoice.
+	invoiceResp = createAssetInvoice(
+		t.t, erin, fabia, 10_000, assetID,
+	)
+
+	// Now Charlie pays the invoice, again by using the manually created
+	// scid. This payment should succeed.
+	payInvoiceWithAssets(t.t, charlie, dave, invoiceResp.PaymentRequest,
```
guggero, line 2135: nit: formatting of multi-line function call.

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2051,176 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+	done = make(chan bool, 1)
+
+	ctxc, cancel := context.WithCancel(context.Background())
+
+	//nolint:lll
+	go func() {
+		// payInvoiceWithSatoshi(t.t, erin, iResp, lnrpc.Payment_FAILED)
+		sendReq := &routerrpc.SendPaymentRequest{
+			PaymentRequest:   iResp.PaymentRequest,
+			TimeoutSeconds:   int32(PaymentTimeout.Seconds()),
```
guggero, line 2203: Shouldn't we instead set this to a way shorter timeout and then just expect the payment to fail?
Then we don't need the extra goroutine.

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2051,176 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+	invoiceResp = createAssetInvoice(
+		t.t, erin, fabia, 10_000, assetID,
+	)
+
+	// Now Charlie pays the invoice, again by using the manually created
+	// scid. This payment should succeed.
+	payInvoiceWithAssets(t.t, charlie, dave, invoiceResp.PaymentRequest,
+		assetID, true, defaultPaymentStatusOpt, fn.Some(scid))
+
+	logBalance(t.t, nodes, assetID, "after manual rfq hodl")
```
guggero, line 2138: We don't actually create a HODL invoice here. So we only test adding HTLCs to the counter in the order manager. But we don't test that if we cancel a HODL invoice that we then can make a normal payment again, as the liquidity isn't used up yet.

guggero (commented):
Needs a couple of updates still, but definitely the better approach!

itest/litd_custom_channels_test.go
GeorgeTsagk: The very last edge case leaves a payment hanging (i.e waiting for timeout) so the channel would not coop close as there were still in-flight HTLCs. I'll see if there's a way to bypass this and wait for the htlcs to get cancelled

itest/litd_custom_channels_test.go
GeorgeTsagk: In the lines above we attempted to make a payment using multiple HTLCs (in the `assertNumHtlcs` calls above) using the same rfq scid

Since we asserted that HTLCs exist, then re-asserted that HTLCs do not exist anymore, that means the HTLCs got cancelled

am i missing smth?

guggero: Hmm, I must've looked at an old version of the code or didn't see the `CancelInvoice` above. Looks good to me the way it is now.

guggero (approved):
LGTM pending dependent PRs.

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2056,145 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+			PeerPubKey:     dave.PubKey[:],
+			TimeoutSeconds: 100,
+		},
+	)
+	require.NoError(t.t, err)
+
+	// We now create a hodl invoice on Fabia, for 10k assets.
+	hodlInv := createAssetHodlInvoice(t.t, erin, fabia, 10_000, assetID)
+
+	// Charlie tries to pay via Dave, by providing the scid that was
```
ffranr, line 2095: ... providing the RFQ quote ID...

itest/litd_custom_channels_test.go
```diff
@@ -2046,6 +2056,145 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
...
+	assertNumHtlcs(t.t, erin, 0)
+	assertNumHtlcs(t.t, fabia, 0)
+
+	// Now Fabia creates the normal invoice.
+	invoiceResp = createAssetInvoice(
+		t.t, erin, fabia, 10_000, assetID,
+	)
+
+	// Now Charlie pays the invoice, again by using the manually created
+	// scid. This payment should succeed.
```
ffranr, line 2138: ... manually specified RFQ quote ID.

ffranr (approved):

itest/assets_test.go
```diff
@@ -882,44 +896,62 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
 		sendReq.MaxShardSizeMsat = 80_000_000
 	}
 
+	var rfqBytes []byte
+	manualRfq.WhenSome(func(i rfqmsg.ID) {
+		rfq := manualRfq.UnsafeFromSome()
```
guggero, line 901: Just use `i` here. That's the whole point of `WhenSome()`, that you don't have to use `UnsafeFrom()`.

GeorgeTsagk:
Will wait for base PR to merge, then will update `go.mod` before merging this

itest/assets_test.go
```diff
@@ -882,44 +896,61 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
 		sendReq.MaxShardSizeMsat = 80_000_000
 	}
 
+	var rfqBytes []byte
+	manualRfq.WhenSome(func(i rfqmsg.ID) {
+		copy(rfqBytes, i[:])
```
guggero, line 901: `copy()` is tricky if the target slice is nil or empty, nothing happens. So you need to do `rfqBytes = make([]byte, len(i[:]))` first.

GeorgeTsagk: Nice catch! Interestingly enough the tests did not fail as the daemon would automatically acquire fresh quotes for the payment. Specifying an rfqID here was a effectively a noop.

The reason this wasn't caught by the next test case (which requires you use a specific rfq scid and none else) is because it is enforced by the receiver, while crafting the invoice.

PR #908, opened 2024-11-27
[custom channels]: Assert fix for duplicate script key issues in MPP payments

guggero:
Updates the force close HTLC sweep integration test to also do MPP payments, to make sure duplicate script keys due to identical MPP shard HTLCs don't cause an issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,14 +16,14 @@ require (
 	github.com/lightninglabs/lightning-node-connect v0.3.2-alpha.0.20240822142323-ee4e7ff52f83
 	github.com/lightninglabs/lightning-terminal/autopilotserverrpc v0.0.1
 	github.com/lightninglabs/lightning-terminal/litrpc v1.0.0
-	github.com/lightninglabs/lndclient v0.18.4-6
+	github.com/lightninglabs/lndclient v0.18.4-7
 	github.com/lightninglabs/loop v0.28.8-beta.0.20241022072406-1e8ae31ddc27
 	github.com/lightninglabs/loop/looprpc v1.0.0
 	github.com/lightninglabs/loop/swapserverrpc v1.0.10
 	github.com/lightninglabs/pool v0.6.5-beta.0.20241015105339-044cb451b5df
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.2
 	github.com/lightninglabs/pool/poolrpc v1.0.0
-	github.com/lightninglabs/taproot-assets v0.4.2-0.20241125202051-783fb1e5ab03
+	github.com/lightninglabs/taproot-assets v0.5.0-rc1
 	github.com/lightningnetwork/lnd v0.18.4-beta.rc1
 	github.com/lightningnetwork/lnd/cert v1.2.2
 	github.com/lightningnetwork/lnd/fn v1.2.3
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -2773,8 +2773,19 @@ func testCustomChannelsFee(_ context.Context,
 // testCustomChannelsHtlcForceClose tests that we can force close a channel
 // with HTLCs in both directions and that the HTLC outputs are correctly
 // swept.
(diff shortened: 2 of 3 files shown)


GeorgeTsagk (approved):
Lgtm!

itest/litd_custom_channels_test.go
```diff
@@ -2870,7 +2880,7 @@ func testCustomChannelsHtlcForceClose(_ context.Context, net *NetworkHarness,
 	var (
 		bobHodlInvoices   []assetHodlInvoice
 		aliceHodlInvoices []assetHodlInvoice
-		assetInvoiceAmt   = 100
+		assetInvoiceAmt   = 5_000
```
jharveyb, line 2883: nit: Bit confused on the math for this change causing MPP to get used.

AFAICT, setting `mpp = true` below limits the shard size to 80e6 msats / 80000 sats.

But the minted asset seems to be 1,000,000 `itest-asset-cents`. And it looks like `assetInvoiceAmt` is denominated in cents also. Not sure if decimal display is being set somewhere I can't find.

I also can't find the exchange rate being used here that would probably help clarify this math.

jharveyb: Update: I found the rate, from grepping after inspecting the logs, as the config value of `mockoracleassetsperbtc`. Would still be good to call out.

Also this log msg should be `got quote for %d msats`, not sats:

https://github.com/lightninglabs/lightning-terminal/blob/a72e388e5f5d3540d1c90e8b28664bbb7b1c9708/itest/assets_test.go#L1129

So, the quoted sat amount is ~85901 sats, which is over the MPP shard threshold of 80000 sats / 4656 asset units.

jharveyb (approved):
LGTM - confirmed MPP behavior from peeking at the logs 💯

itest/litd_custom_channels_test.go
```diff
@@ -2903,9 +2913,14 @@ func testCustomChannelsHtlcForceClose(_ context.Context, net *NetworkHarness,
...
 	}
 
-	// At this point, both sides should have 4 HTLCs active.
-	assertNumHtlcs(t.t, alice, 4)
-	assertNumHtlcs(t.t, bob, 4)
+	// At this point, both sides should have 4 (or +2 with MPP) HTLCs
+	// active.
+	numHtlcs := 4
+	if mpp {
+		numHtlcs += 2
```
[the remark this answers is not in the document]
guggero: Right, makes sense. That way we can also more easily increase the number of HTLCs later on.

itest/litd_custom_channels_test.go
guggero: Ah, yes, that's not super apparent. Added a comment to describe that, which then also makes the use of the `mpp` variable a bit more apparent.

go.mod
```diff
@@ -16,14 +16,14 @@ require (
...
-	github.com/lightninglabs/lndclient v0.18.4-6
+	github.com/lightninglabs/lndclient v0.18.4-7
 	github.com/lightninglabs/loop v0.28.8-beta.0.20241022072406-1e8ae31ddc27
 	github.com/lightninglabs/loop/looprpc v1.0.0
 	github.com/lightninglabs/loop/swapserverrpc v1.0.10
 	github.com/lightninglabs/pool v0.6.5-beta.0.20241015105339-044cb451b5df
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.2
 	github.com/lightninglabs/pool/poolrpc v1.0.0
-	github.com/lightninglabs/taproot-assets v0.4.2-0.20241125202051-783fb1e5ab03
+	github.com/lightninglabs/taproot-assets v0.4.2-0.20241127150634-c96f36961231
```
gijswijs, line 26: Could bump it now to RC5, right?

guggero: Yes, will do that once the tag is pushed in `tapd`.

itest/litd_custom_channels_test.go
gijswijs: I didn't grok this either. Thanks for explaining it @jharveyb.

PR #911, opened 2024-12-03
Fix litcli sendpayment panic

GeorgeTsagk:
## Description

Sometimes the asset units may not be sufficient to carry out a payment (depending on the granularity and asset price). This may be lead to calculating "zero" units for a payment. This pr fixes a small piece of code that would blindly divide by `numUnits` without checking whether it's zero. If it's zero we need to error out as this means the payment cannot be carried out.

## Solution

If we have `numUnits` equal to 0 we skip the division that would cause the panic, and instead return an error to the user which looks like this
```
Payment hash: 6dcf926e08d24a52c6d7aeec641eddc73127a99a9b0c2e086d3095af0a722ce6
Description: 
Amount (in satoshis): 10
Fee limit (in satoshis): 10
Destination: 0327e3031fa45faa56be7e702de23e3d8c2305a5d69b410e2f44fa8bbca6dfa5ef
Confirm payment (yes/no): yes
[litcli] asset rate 1000000 not sufficient to represent value of 10000 mSAT
```

### Related issues
Closes https://github.com/lightninglabs/taproot-assets/issues/1230


cmd/litcli/ln.go
```diff
@@ -216,6 +216,13 @@ func (w *resultStreamWrapper) Recv() (*lnrpc.Payment, error) {
 		amountMsat := lnwire.MilliSatoshi(w.amountMsat)
 		milliSatsFP := rfqmath.MilliSatoshiToUnits(amountMsat, *rate)
 		numUnits := milliSatsFP.ScaleTo(0).ToUint64()
+		if numUnits == 0 {
+			return nil, fmt.Errorf("asset rate %v not sufficient "+
```
Roasbeef, line 220: I think we should tune this error message slightly to make it clearer why the payment can't succeed. 

What about something like: 
> smallest payment with asset rate %v is %v, cannot send %v

Roasbeef: Also I thought the `min_amt` on the RFQ layer was meant to compute this value ahead of time to allow callers to exit sooner before even calling into `lnd`. cc @ffranr

Roasbeef (changes requested):

cmd/litcli/ln.go
GeorgeTsagk: should I just derive the smallest payment directly from the asset rate of the quote?

cmd/litcli/ln.go
```diff
@@ -216,6 +216,13 @@ func (w *resultStreamWrapper) Recv() (*lnrpc.Payment, error) {
...
+				"to represent value of %v",
+				rate.ToUint64(), amountMsat,
+			)
```
guggero, line 223: https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md

guggero (approved):
tACK

While testing I noticed that the payment actually stays in-flight in the background forever. But that's unrelated to this change and needs to be fixed in `lnd`. Going to create a PR for it.

Roasbeef (approved):
LGTM 👑

PR #912, opened 2024-12-04
[custom channels]: test forward traffic shaper

guggero:
~Depends on https://github.com/lightningnetwork/lnd/pull/9333.~
Depends on https://github.com/lightninglabs/taproot-assets/pull/1236.

Makes sure channels aren't force closed in certain edge cases.

### DIFF
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/lightninglabs/taproot-assets/tapchannel"
 	"github.com/lightninglabs/taproot-assets/taprpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/mintrpc"
+	oraclerpc "github.com/lightninglabs/taproot-assets/taprpc/priceoraclerpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/rfqrpc"
 	tchrpc "github.com/lightninglabs/taproot-assets/taprpc/tapchannelrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc/universerpc"
@@ -1730,7 +1731,7 @@ func testCustomChannelsBreach(_ context.Context, net *NetworkHarness,
 
 // testCustomChannelsLiquidityEdgeCases is a test that runs through some
 // taproot asset channel liquidity related edge cases.
-func testCustomChannelsLiquidityEdgeCases(_ context.Context,
+func testCustomChannelsLiquidityEdgeCases(ctxb context.Context,
 	net *NetworkHarness, t *harnessTest) {
 
 	lndArgs := slices.Clone(lndArgsTemplate)
@@ -2009,11 +2010,8 @@ func testCustomChannelsLiquidityEdgeCases(_ context.Context,
 	// satoshi, where we will check whether Dave's strict forwarding works
 	// as expected. Charlie is only used as a dummy RFQ peer in this case,
 	// Yara totally ignored the RFQ hint and pays agnostically with sats.
-	invoiceResp = createAssetInvoice(
-		t.t, charlie, dave, 1, assetID,
-	)
+	invoiceResp = createAssetInvoice(t.t, charlie, dave, 1, assetID)
 
-	ctxb := context.Background()
 	stream, err := dave.InvoicesClient.SubscribeSingleInvoice(
 		ctxb, &invoicesrpc.SubscribeSingleInvoiceRequest{
 			RHash: invoiceResp.RHash,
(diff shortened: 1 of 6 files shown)


itest/litd_custom_channels_test.go
```diff
@@ -3304,3 +3320,220 @@ func runCustomChannelsHtlcForceClose(ctxb context.Context, t *harnessTest,
...
+	)
+
+	// Test case 2: We cannot pay an invoice from Charlie to Fabia.
+	invoiceResp := createAssetInvoice(t.t, erin, fabia, 123, assetID)
+	payInvoiceWithSatoshi(
+		t.t, charlie, invoiceResp, lnrpc.Payment_FAILED,
+		fn.Some(errNoRoute),
+	)
+
+	// Test case 3: We now create an asset buy order for a normal amount of
```
GeorgeTsagk, line 3469: could incorporate everything so far into the `testCustomChannelsLiquidityEdgeCases` test?

not sure if we rather have many granular itests or very long itests, but the nature of these testcases generally fits what we do in the liquidity edge cases itest

itest/litd_custom_channels_test.go
```diff
@@ -3304,3 +3320,220 @@ func runCustomChannelsHtlcForceClose(ctxb context.Context, t *harnessTest,
...
+			}},
+		}},
+	})
+	require.NoError(t.t, err)
+
+	payInvoiceWithSatoshi(
+		t.t, dave, invoiceResp2, lnrpc.Payment_FAILED,
+		fn.Some(errNoRoute),
+	)
+
```
GeorgeTsagk, line 3532: could now do a successful payment, so that this itest is "complete"
(non-blocking: this is tested multiple times in the surrounding itests)

GeorgeTsagk (commented):
looks good, pending base PRs

itest/litd_custom_channels_test.go
guggero: I thought about that. But I think we have to be careful that we then don't run into false negatives. For example, we expect a payment to fail for a specific edge case reason, but it ends up failing because the mission control has recorded too many previous failures and gives up on the payment before even trying.
Sure, we could counter that by clearing mission control and/or checking for the correct failure.
But I think the extra overhead of 30 seconds per test is currently bearable in exchange for a clean test slate.

IMO we can do some cleanup and test case consolidation once we've released the current version.

GeorgeTsagk (approved):
LGTM
pending base PR

dstadulis:
Will need to highly prioritize for rc2

itest/litd_custom_channels_test.go
```diff
@@ -3434,3 +3432,222 @@ func runCustomChannelsHtlcForceClose(ctxb context.Context, t *harnessTest,
...
+		Coefficient: quote.AcceptedQuote.AskAssetRate.Coefficient,
+		Scale:       quote.AcceptedQuote.AskAssetRate.Scale,
+	})
+	require.NoError(t.t, err)
+
+	oneUnit := rfqmath.NewBigIntFixedPoint(1, 0)
+	oneUnitMilliSat := rfqmath.UnitsToMilliSatoshi(oneUnit, *rate)
+
+	t.Logf("Got quote for %v asset units per BTC", rate)
+	msatPerUnit := float64(oneUnitMilliSat) / float64(1)
```
ffranr, line 3619: this calculation is redundant and therefore so is the log message above it.

Perhaps consider setting the `1` to a variable. And then the code can be a bit more dynamic. Otherwise, no point in that calc.

itest/litd_custom_channels_test.go
```diff
@@ -3434,3 +3432,222 @@ func runCustomChannelsHtlcForceClose(ctxb context.Context, t *harnessTest,
...
+	t.Logf("Got quote for %v asset units at %3f msat/unit from peer %s "+
+		"with SCID %d", numUnits, msatPerUnit, erin.PubKeyStr,
+		quote.AcceptedQuote.Scid)
```
ffranr, line 3622: In general, I think it would be better if log messages were written with args at the end:
```
t.Logf("Got quote from peer (asset_units=%v, msat_to_asset_units=%3f, peer=%s, "+
		"scid=%d)", numUnits, msatPerUnit, erin.PubKeyStr,
		quote.AcceptedQuote.Scid)
```

ffranr (approved):

itest/litd_custom_channels_test.go
guggero: Yeah, I generally agree. And with the new `slog` based logging library we'll be able to do that in a more generalized manner (at least in `lnd`).
For now I'm trying to keep this aligned with other such messages in other tests.

PR #913, opened 2024-12-05
dev.Dockerfile: allow forcing a specific taproot-assets version

ZZiigguurraatt:
dev.Dockerfile: allow forcing a specific taproot-assets version through a build argument.

An example of usage can be found here: https://github.com/lightninglabs/tapdvalidation/blob/cb7c752e9b123dc99882a471afa04cc1066e5111/docker-compose.yaml#L66

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -23,8 +23,9 @@ COPY --from=nodejsbuilder /go/src/github.com/lightninglabs/lightning-terminal /g
 # queries required to connect to linked containers succeed.
 ENV GODEBUG netdns=cgo
 
-# allow forcing a specific taproot-assets version through a build argument
-# see https://go.dev/ref/mod#version-queries for the types of queries that can be used to define a version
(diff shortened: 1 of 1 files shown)


dev.Dockerfile
```diff
@@ -23,10 +23,18 @@ COPY --from=nodejsbuilder /go/src/github.com/lightninglabs/lightning-terminal /g
 # queries required to connect to linked containers succeed.
 ENV GODEBUG netdns=cgo
 
+# allow forcing a specific taproot-assets version through a build argument
```
guggero, line 26: nit: we use complete sentences including proper capitalization and punctuation for all comments in all our projects. And where possible (depending on the file format and content), we try to adhere to the 80 character limit outside of Go code as well.

guggero (approved):
tACK, LGTM :tada:

dev.Dockerfile
ZZiigguurraatt: fixed

PR #920, opened 2024-12-13
Add lastHop parameter for loop in quotes

### LINKED ISSUE #904: loop-in: quote does not use last hop pubkey
**Expected behavior** / **Actual behavior**
A loop cli quote for my WOS loop in of 350k sat @ 6 conf_target is giving 177 sats for service fee, while terminal is giving 5 sats. Looks like the discrepancy is that last hop pubkey isn't being set for the quote, which in turn means the subsequent loopIn will likely fail if that node is more expensive than a quote that doesn't specify a node to loop in to.

Here we would want to use `loopInLastHop`: https://github.com/lightninglabs/lightning-terminal/blob/9ad3f064a69b15aad4fa3bc3734c39d77057be33/app/src/store/views/buildSwapView.ts#L426-L428

Also would want a new param here for passing in that hop if it's specified: https://github.com/lightninglabs/lightning-terminal/blob/9ad3f064a69b15aad4fa3bc3734c39d77057be33/app/src/api/loop.ts#L57

itsrachelfish:
Fixes #904

Was able to reproduce this bug on regtest by changing the fee rate on one of my test nodes.

Screenshot from 2024-12-12 16-17-17

Fix was pretty simple because the `swapView` and loop protobuf definitons already had all of the necessary variables and functions defined :tada: 

Screenshot from 2024-12-12 17-07-12

### DIFF
--- a/app/src/api/loop.ts
+++ b/app/src/api/loop.ts
@@ -57,10 +57,12 @@ class LoopApi extends BaseApi<LoopEvents> {
   async getLoopInQuote(
     amount: Big,
     confTarget?: number,
+    lastHop?: string,
   ): Promise<LOOP.InQuoteResponse.AsObject> {
     const req = new LOOP.QuoteRequest();
     req.setAmt(amount.toString());
     if (confTarget) req.setConfTarget(confTarget);
+    if (lastHop) req.setLoopInLastHop(b64(lastHop));
     const res = await this._grpc.request(SwapClient.GetLoopInQuote, req, this._meta);
     return res.toObject();
   }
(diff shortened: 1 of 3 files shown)


ViktorTigerstrom (approved):
LGTM! Could be worth getting an ACK from someone that works with the LiT frontend code as well :)

jamaljsr:
The code LGTM but haven't tested. Tagging @jbrill for review since I am OOO this week.

jamaljsr (approved):
tACK LGTM 👌

jbrill (approved):
LGTM! Sorry about the delayed review

Might just need a rebase on the release notes!

PR #921, opened 2024-12-13
[custom channels]: enforce minimum amounts

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1252.

Adds test cases for the new minimum amount enforcement introduced in above PR.


itest/litd_custom_channels_test.go
```diff
@@ -2164,7 +2204,7 @@ func testCustomChannelsLiquidityEdgeCases(ctxb context.Context,
 	// Now Erin tries to pay the invoice. Since rfq quote cannot satisfy the
 	// total amount of the invoice this payment will fail.
 	payInvoiceWithSatoshi(
-		t.t, erin, iResp, withExpectTimeout(),
+		t.t, erin, iResp, withPayErrSubStr("context deadline exceeded"),
```
GeorgeTsagk, line 2207: 👍

GeorgeTsagk (approved):
lgtm! 🥼 

commits need reorder to compile individually

guggero:
> commits need reorder to compile individually

Yeah, I'll squash them into the existing commits of the `update-to-lnd-18-4` branch. So keeping them separately (go.mod changes and itest changes) will make that easier.

cmd/litcli/ln.go
```diff
@@ -182,6 +182,15 @@ var (
 			"set if there are multiple channels with the same " +
 			"asset ID present",
 	}
+
+	allowUnEconomicalFlag = cli.BoolFlag{
```
ffranr, line 186: How about `allowOverpay` or `allowAssetOverpay` or `allowTotalValueOverpay`

I think `allowTotalValueOverpay` is most clear.

cmd/litcli/ln.go
```diff
@@ -356,13 +356,12 @@ func sendPayment(ctx *cli.Context) error {
 			"%w", err)
 	}
 
-	// We use a constant amount of 500 to carry the asset HTLCs. In the
+	// We use a constant amount of 354 to carry the asset HTLCs. In the
```
ffranr, line 359: I don't think we should repeat constants in comments. Because if we change the constant then it's easy to forget to update the comment. Maybe:
```
// Use smallest HTLC sat amount to carry the asset HTLCs.
```

dstadulis: Or maybe call out where the `DefaultOnChainHtlcAmount` value is set?

cmd/litcli/ln.go
ffranr: Couldn't see this change in the latest commits. Are they incoming?

ffranr:
@guggero does this need a rebase before I review?

ffranr (approved):
There's a "check commit" lint error but otherwise ok.

PR #928, opened 2025-01-06
makefile: add unit test helpers

ellemouton:
Some helpers (lots of influence from the `taproot-assets` repo) to give devs more control over running unit tests. 

With this, we can now do things like:

```
# Run all tests in the accounts package.
make unit pkg=accounts

# Run a specific test in accounts package
make unit pkg=accounts case=TestAccountService

# Pass in additional flags (more useful in later PRs). 
make unit tags=test_db_postgres
```

### DIFF
--- a/make/testing_flags.mk
+++ b/make/testing_flags.mk
@@ -2,6 +2,7 @@ include make/compile_flags.mk
 
 ITEST_FLAGS =
 TEST_FLAGS =
+DEV_TAGS = dev
 
 # Define the integration test.run filter if the icase argument was provided.
 ifneq ($(icase),)
@@ -13,5 +14,10 @@ ifneq ($(case),)
 TEST_FLAGS += -test.run=$(case)
 endif
 
-UNIT := $(GOLIST) | $(XARGS) env $(GOTEST) -tags="$(COMPILE_TAGS)" $(TEST_FLAGS)
+# Add any additional tags that are passed in to make.
+ifneq ($(tags),)
+DEV_TAGS += ${tags}
+endif
+
+UNIT := $(GOLIST) | $(XARGS) env $(GOTEST) -tags="$(DEV_TAGS) $(COMPILE_TAGS)" $(TEST_FLAGS)
 UNIT_RACE := $(UNIT) -race
--- a/make/testing_flags.mk
+++ b/make/testing_flags.mk
@@ -1,11 +1,17 @@
 include make/compile_flags.mk
 
 ITEST_FLAGS =
+TEST_FLAGS =
 
 # Define the integration test.run filter if the icase argument was provided.
 ifneq ($(icase),)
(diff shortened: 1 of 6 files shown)


make/testing_flags.mk
```diff
@@ -1,6 +1,46 @@
...
+# targeted case. Otherwise, default to running all tests.
+ifeq ($(UNIT_TARGETED), yes)
+UNIT := $(GOTEST) -tags="$(DEV_TAGS) $(COMPILE_TAGS)" $(TEST_FLAGS) $(UNITPKG)
+UNIT_RACE := $(GOTEST) -tags="$(DEV_TAGS) $(COMPILE_TAGS)" $(TEST_FLAGS) -race $(UNITPKG)
+endif
+
+ifeq ($(UNIT_TARGETED), no)
+UNIT := $(GOLIST) | $(XARGS) env $(GOTEST) -tags="$(DEV_TAGS) $(COMPILE_TAGS)" $(TEST_FLAGS)
+UNIT_RACE := $(UNIT) -race
+endif
```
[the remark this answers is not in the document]
ellemouton: nice! thanks 🙏

make/testing_flags.mk
```diff
@@ -1,6 +1,46 @@
...
+
+# UNIT_TARGETED is undefined iff a specific package and/or unit test case is
+# not being targeted.
+UNIT_TARGETED ?= no
+
+# If a specific package/test case was requested, run the unit test for the
+# targeted case. Otherwise, default to running all tests.
+ifeq ($(UNIT_TARGETED), yes)
+UNIT := $(GOTEST) -tags="$(DEV_TAGS) $(COMPILE_TAGS)" $(TEST_FLAGS) $(UNITPKG)
```
bitromortac, line 39: `make unit case=TestObfuscateConfig` would not run, but only adding the package will add the test files in order for the test to be discoverable: `make unit case=TestObfuscateConfig pkg=firewall`. Do we want that, because it seems like an extra step that we could avoid?

ellemouton: note that this was mentioned in this commit message: https://github.com/lightninglabs/lightning-terminal/pull/928/commits/824a3c259c84fbde77e24febcf6c17e0fc5c9253

ellemouton: you have to specify package. Otherwise you have to search through all packages. Im sure there is a helper we can add later on but think this is fine for now?

bitromortac: > note that this was mentioned in this commit message: https://github.com/lightninglabs/lightning-terminal/commit/824a3c259c84fbde77e24febcf6c17e0fc5c9253

ah yes, now that makes sense! I just noticed that in lnd it's the same (usually I'm running tests differently), so it makes sense to keep things consistent across projects. it really takes a lot longer if we'd include all packages with 24s vs 2s :+1:

bitromortac (approved):

PR #932, opened 2025-01-13
multi: thread contexts through properly

ellemouton:
In this PR, we remove as many uses of `context.Background` as possible. By the end of 
this PR, the litd has a single parent context. We also let the CLI use the `getContext` helper that
we use in other repos that cancels the context used if an appropriate OS signal is caught.

### DIFF
--- a/cmd/litcli/ln.go
+++ b/cmd/litcli/ln.go
@@ -91,17 +91,17 @@ var fundChannelCommand = cli.Command{
 }
 
 func fundChannel(c *cli.Context) error {
-	tapdConn, cleanup, err := connectSuperMacClient(c)
+	ctx := getContext()
+	tapdConn, cleanup, err := connectSuperMacClient(ctx, c)
 	if err != nil {
 		return fmt.Errorf("error creating tapd connection: %w", err)
 	}
 
 	defer cleanup()
 
-	ctxb := context.Background()
 	tapdClient := taprpc.NewTaprootAssetsClient(tapdConn)
 	tchrpcClient := tchrpc.NewTaprootAssetChannelsClient(tapdConn)
-	assets, err := tapdClient.ListAssets(ctxb, &taprpc.ListAssetRequest{})
+	assets, err := tapdClient.ListAssets(ctx, &taprpc.ListAssetRequest{})
 	if err != nil {
 		return fmt.Errorf("error fetching assets: %w", err)
 	}
@@ -146,7 +146,7 @@ func fundChannel(c *cli.Context) error {
 	}
 
 	resp, err := tchrpcClient.FundChannel(
-		ctxb, &tchrpc.FundChannelRequest{
+		ctx, &tchrpc.FundChannelRequest{
 			AssetAmount:        requestedAmount,
 			AssetId:            assetIDBytes,
 			PeerPubkey:         nodePubBytes,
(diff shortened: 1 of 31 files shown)


session_rpcserver.go
```diff
@@ -135,7 +135,6 @@ func (s *sessionRpcServer) start() error {
 				continue
 			}
 
-			ctx := context.Background()
 			ctxc, cancel := context.WithTimeout(
```
bitromortac, line 138: it looks like not all `s.cfg.autopilot` queries have a timeout context, should we make this consistent?

cmd/litcli/ln.go
```diff
@@ -297,41 +297,46 @@ var sendPaymentCommand = cli.Command{
...
 
-	lndConn, cleanup, err := connectClient(ctx, false)
+	lndConn, cleanup, err := connectClient(cliCtx, false)
 	if err != nil {
 		return fmt.Errorf("unable to make rpc conn: %w", err)
 	}
 	defer cleanup()
 
-	tapdConn, cleanup, err := connectSuperMacClient(ctx)
+	// NOTE: we don't use `getContext()` here since it assigns the global
```
bitromortac, line 313: ah nice :+1:

cmd/litcli/ln.go
```diff
@@ -297,41 +297,46 @@ var sendPaymentCommand = cli.Command{
 	Action: sendPayment,
 }
 
-func sendPayment(ctx *cli.Context) error {
+func sendPayment(cliCtx *cli.Context) error {
```
bitromortac, line 300: nit: `cli` is maybe not an optimal parameter name for the pattern if we shadow the `cli` package (in the other commands)

terminal.go
```diff
@@ -235,15 +235,31 @@ func New() *LightningTerminal {
...
 	if err != nil {
 		return fmt.Errorf("could not intercept signals: %v", err)
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Make sure the context is canceled if the user requests shutdown and
+	// that the shutdown signal is requested if the context is canceled.
+	go func() {
```
bitromortac, line 250: nice to put this here :+1:

bitromortac (approved):
Nice work! LGTM :tada: (just have a suggestion and a nit)

terminal.go
```diff
@@ -235,13 +235,31 @@ func New() *LightningTerminal {
...
+		// Client requests shutdown, cancel the wait.
+		case <-shutdownInterceptor.ShutdownChannel():
+			cancel()
+
+		// The check was completed and the above defer canceled the
+		// context. We can just exit the goroutine, nothing more to do.
+		case <-ctx.Done():
+			shutdownInterceptor.RequestShutdown()
+		}
+	}()
```
ViktorTigerstrom, line 261: Niiiice :rocket:!

cmd/litcli/main.go
```diff
@@ -322,5 +322,21 @@ func connectSuperMacClient(ctx *cli.Context) (grpc.ClientConnInterface,
 		return nil, nil, fmt.Errorf("error decoding macaroon: %w", err)
 	}
 
-	return connectClientWithMac(ctx, macBytes)
+	return connectClientWithMac(cli, macBytes)
+}
+
+func getContext() context.Context {
```
ViktorTigerstrom, line 328: Awesome :fire:!

ViktorTigerstrom (approved):
Thanks a lot for this :rocket:! This will be really helpful for the future :pray:!

session_rpcserver.go
ellemouton: i thiiiink it is currently only the ones that are not directly called from an RPC method that dont do the timeout cause there is nothing else to cancel them. For rpc calls, if the rpc context is cancelled, the call is cancelled

cmd/litcli/ln.go
ellemouton: yeah tricky. But opted to prio having `ctx` be used for `context.Context` since that is standard.

Issue #917, opened 2024-12-11
db: use native SQL


ellemouton:
## Overview: 

We should follow the example set by LND and migrate our kvdb stores to SQL. Luckily, LiTs set of tables is quite small and so this is quite a contained task. We also have never supported the kv schema in other backend options like sql or postgres and don't currently have any tables in any sql backend. This is nice because it keeps set of possible starting points to a size of 1: Migrate bbolt kvstore data to native sql stores (either postgres or sqlite). I think that since currently we only store LNC data in LiT's DB, nothing is critical and so I think we should force the migration at the end of the day and not necessarily let uses opt to not run the migration. 

## Options: 

1) option 1 is to follow LNDs lead and properly migrate over all existing data to the new SQL tables. This is of course possible but will require a lot of careful work. We'll also need to retain all the current KVDB specific code in a migration folder since we will always want to remain backwards compatible. 
2) option 2: we just do a big breaking change release where we say "sorry, all previous LNC data is nuked" and then we nuke those DBs and start fresh with the new SQL backend. Downsides are: all existing LNC sessions would need to be re-created for things such as Zeus connections and Autopilot connections. Upside is: path to completion is much shorter & we dont need to keep around legacy code & spend a significant amount of time trying to get the migrations to be perfect. 

## Getting started:

Regardless of which option we go with above, we can get started by:

1) defining the new SQL tables (dont actually trigger them in anything other than tests)
2) defining the SQL queries & putting them behind an interface such that all current unit tests can be run against both the kvdb backends and new sql backends. 

The above can be done for both the `session` related tables and `firewalldb` related tables. 

We can get started on the work for the above and merge things well before we need to worry about the final switch to sql.

ellemouton:
cc @levmi for input re the two approach options

ellemouton:
TODO(elle): possibly use uints for timestamps instead of TIMESTAMP

Roasbeef:
>  possibly use uints for timestamps instead of TIMESTAMP

With the way it works to `sqlc`, if you use `TIMESTAMP`, then it [gets stored as a string in Go formatting (`time.Time`)](https://docs.sqlc.dev/en/stable/reference/datatypes.html#dates-and-times). You'd need to take care to ensure times are normalized to UTC timezone before inserting. Going with this approach over just an integer has a nice effect as you can do a bit more expressive queries as [you can use the `like` operator](https://www.sqlitetutorial.net/sqlite-like/).

PR #934, opened 2025-01-14
[sql-1]accounts: preparatory commits for  SQL-izing accounts

ellemouton:
This PR contains a couple of minor clean-up and prep commits. These set
up the work that will follow to start massaging the accounts DB Store to 
be ready for a SQL implementation. 

Each commit is ultra tiny :)

Issue: https://github.com/lightninglabs/lightning-terminal/issues/917


ellemouton:
updated commit messages to be more descriptive

accounts/checkers_test.go
```diff
@@ -523,7 +523,9 @@ func testSendPayment(t *testing.T, uri string) {
 	errFunc := func(err error) {
 		lndMock.mainErrChan <- err
 	}
-	service, err := NewService(t.TempDir(), errFunc)
+	store, err := NewBoltStore(t.TempDir(), DBFilename)
```
ViktorTigerstrom: Should we close the `store` at the end of the tests here? I see this is changed in the next commits, with addition of the interface, so maybe it'd be worth adding a `cleanup` return arg for that interface that closes the store?

accounts/service_test.go
```diff
@@ -829,7 +829,9 @@ func TestAccountService(t *testing.T) {
 			errFunc := func(err error) {
 				lndMock.mainErrChan <- err
 			}
-			service, err := NewService(t.TempDir(), errFunc)
+			store, err := NewBoltStore(t.TempDir(), DBFilename)
```
ViktorTigerstrom: Same here if not already closed in the `validate` func.

accounts/test_kvdb.go
```diff
@@ -0,0 +1,30 @@
+package accounts
```
ViktorTigerstrom, line 1: Nit: the commit message contains info regarding build flags logic, which isn't present in the actual commit (which i suspect will be added later?).

ViktorTigerstrom (approved):
Awesome initial push to SQL-izing accounts :rocket::fire:!!

accounts/service.go
```diff
@@ -193,7 +188,7 @@ func (s *InterceptorService) Start(ctx context.Context,
 					return
 				}
 
-				if err := s.invoiceUpdate(invoice); err != nil {
+				if err := s.invoiceUpdate(ctx, invoice); err != nil {
```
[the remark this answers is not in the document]
ellemouton: we should props try re-enable that linter at some point 🤓

accounts/checkers_test.go
ellemouton: in the commit that introduces the `NewTestDB` calls, it's taken care of in the clean up there. So I think your concerns are addressed by the end of the PR no? (let me know if i misunderstand).

accounts/service_test.go
ellemouton: same comment as above

accounts/test_kvdb.go
ellemouton: yes they will be added later. included them in the commit message so there is some context for this commit

bitromortac (approved):
Nice, LGTM 🗄️

PR #938, opened 2025-01-16
[sql-2] accounts: start replacing calls to UpdateAccount

ellemouton:
To prepare the `accounts.Store` to be more SQL compatible, we will want to replace the current
`UpdateAccount` method. Today it takes a full `OffChainAccount` struct, serialises it and replaces
any existing in the DB. This method is not very SQL appropriate as the better pattern is to instead
have specific update methods that read: "Update field X and account with ID Y". 

So this PR starts this process of adding more SQL friendly methods to the interface and thereby 
phasing out the use of `UpdateAccount`. By doing so, we also move very kvdb specific logic out
of the `accounts.InterceptorService`. 

I'm splitting this up over about 3 or 4 PRs since some of the replacements are more complicated than 
others. 

This PR focuses on just replacing the simple ones:
- `UpdateAccountBalanceAndExpiry`
- `AddAccountInvoice`
- `IncreaseAccountBalance`

### DIFF
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -230,6 +230,11 @@ type Store interface {
 	AddAccountInvoice(ctx context.Context, id AccountID,
 		hash lntypes.Hash) error
 
+	// IncreaseAccountBalance increases the balance of the account with the
+	// given ID by the given amount.
+	IncreaseAccountBalance(ctx context.Context, id AccountID,
+		amount lnwire.MilliSatoshi) error
+
 	// RemoveAccount finds an account by its ID and removes it from the¨
 	// store.
 	RemoveAccount(ctx context.Context, id AccountID) error
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -226,6 +226,10 @@ type Store interface {
 		newBalance fn.Option[lnwire.MilliSatoshi],
 		newExpiry fn.Option[time.Time]) error
 
+	// AddAccountInvoice adds an invoice hash to an account.
+	AddAccountInvoice(ctx context.Context, id AccountID,
+		hash lntypes.Hash) error
+
 	// RemoveAccount finds an account by its ID and removes it from the¨
 	// store.
 	RemoveAccount(ctx context.Context, id AccountID) error
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwire"
@@ -218,6 +219,12 @@ type Store interface {
 
 	// Accounts retrieves all accounts from the store and un-marshals them.
 	Accounts(ctx context.Context) ([]*OffChainBalanceAccount, error)
+
+	// UpdateAccountBalanceAndExpiry updates the balance and/or expiry of an
+	// account.
+	UpdateAccountBalanceAndExpiry(ctx context.Context, id AccountID,
+		newBalance fn.Option[lnwire.MilliSatoshi],
+		newExpiry fn.Option[time.Time]) error
 
 	// RemoveAccount finds an account by its ID and removes it from the¨
 	// store.
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/lightninglabs/lndclient"
 	"github.com/lightningnetwork/lnd/channeldb"
@@ -299,7 +300,7 @@ func (s *InterceptorService) NewAccount(ctx context.Context,
 // UpdateAccount writes an account to the database, overwriting the existing one
 // if it exists.
 func (s *InterceptorService) UpdateAccount(ctx context.Context,
-	accountID AccountID, accountBalance,
+	accountID AccountID, accountBalance btcutil.Amount,
 	expirationDate int64) (*OffChainBalanceAccount, error) {
 
 	s.Lock()
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -598,21 +598,13 @@ func (s *InterceptorService) invoiceUpdate(ctx context.Context,
 		return nil
 	}
 
-	account, err := s.store.Account(ctx, acctID)
-	if err != nil {
-		return s.disableAndErrorfUnsafe(
-			"error fetching account: %w", err,
-		)
-	}
-
 	// If we get here, the current account has the invoice associated with
 	// it that was just paid. Credit the amount to the account and update it
 	// in the DB.
-	account.CurrentBalance += int64(invoice.AmountPaid)
-	if err := s.store.UpdateAccount(ctx, account); err != nil {
-		return s.disableAndErrorfUnsafe(
-			"error updating account: %w", err,
-		)
+	err := s.store.IncreaseAccountBalance(ctx, acctID, invoice.AmountPaid)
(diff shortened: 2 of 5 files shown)


accounts/service.go
```diff
@@ -313,36 +313,35 @@ func (s *InterceptorService) UpdateAccount(ctx context.Context,
...
 	}
 
-	account, err := s.store.Account(ctx, accountID)
-	if err != nil {
-		return nil, fmt.Errorf("error fetching account: %w", err)
-	}
-
 	// If the expiration date was set, parse it as a unix time stamp. A
 	// value of -1 signals "don't update the expiration date".
+	var expiry fn.Option[time.Time]
```
bitromortac, line 318: probably nothing for this PR as we want minimal changes (just wanted to know your take on that), I think it would be better to convert to optional types one layer higher and use btcutil.Amount for the accountBalance?

accounts/store_test.go
```diff
@@ -105,6 +106,60 @@ func assertEqualAccounts(t *testing.T, expected,
...
+		require.NoError(t, err)
+
+		assertBalanceAndExpiry := func(balance int64,
+			expiry time.Time) {
+
+			dbAcct, err := store.Account(ctx, acct.ID)
+			require.NoError(t, err)
+			require.EqualValues(t, balance, dbAcct.CurrentBalance)
+			require.Equal(
+				t, expiry.Unix(), dbAcct.ExpirationDate.Unix(),
```
bitromortac, line 128: here we could use `WithinDuration` as well

accounts/store_kvdb.go
```diff
@@ -194,6 +195,51 @@ func (s *BoltStore) UpdateAccount(_ context.Context,
...
+
+		account, err := getAccount(bucket, id)
+		if err != nil {
+			return err
+		}
+
+		err = updateFn(account)
+		if err != nil {
+			return err
+		}
```
bitromortac, line 269: maybe we could wrap the errors to know what failed here?

accounts/store_test.go
```diff
@@ -105,6 +106,60 @@ func assertEqualAccounts(t *testing.T, expected,
...
+		assertBalanceAndExpiry(newBalance, newExpiry)
+
+		// Finally, update both the balance and expiry of the account.
+		newBalance = 456
+		newExpiry = time.Now().Add(2 * time.Hour)
+		err = store.UpdateAccountBalanceAndExpiry(
+			ctx, acct.ID, fn.Some(newBalance), fn.Some(newExpiry),
+		)
+		require.NoError(t, err)
+		assertBalanceAndExpiry(newBalance, newExpiry)
```
bitromortac, line 159: nit: we could add coverage for the None, None case

accounts/interface.go
```diff
@@ -226,6 +226,10 @@ type Store interface {
 		newBalance fn.Option[int64],
 		newExpiry fn.Option[time.Time]) error
 
+	// AddAccountInvoice adds and invoice hash to an account.
```
bitromortac, line 229: typo: `adds an invoice`

accounts/service.go
```diff
@@ -592,21 +592,13 @@ func (s *InterceptorService) invoiceUpdate(ctx context.Context,
...
 	// in the DB.
-	account.CurrentBalance += int64(invoice.AmountPaid)
-	if err := s.store.UpdateAccount(ctx, account); err != nil {
-		return s.disableAndErrorfUnsafe(
-			"error updating account: %w", err,
-		)
+	err := s.store.IncreaseAccountBalance(ctx, acctID, invoice.AmountPaid)
+	if err != nil {
+		return s.disableAndErrorfUnsafe("error updating account: %w",
+			err)
```
bitromortac, line 601: we could perhaps make the error message more specific

accounts/interface.go
```diff
@@ -230,6 +230,11 @@ type Store interface {
 	AddAccountInvoice(ctx context.Context, id AccountID,
 		hash lntypes.Hash) error
 
+	// IncreaseAccountBalance increases the balance of the account with the
+	// given ID by the given amount.
+	IncreaseAccountBalance(ctx context.Context, id AccountID,
+		amount lnwire.MilliSatoshi) error
```
bitromortac, line 236: I see you add the amount as `lnwire.MilliSatoshi`, so maybe we should also have a better type for the balance in `UpdateAccountBalanceAndExpiry` (see other comment)

accounts/store_kvdb.go
ellemouton: 👍

accounts/store_test.go
ellemouton: added 👍

accounts/interface.go
ellemouton: ah yes, great point! will update the store interface to at least use the correct types 👍

accounts/service.go
ellemouton: but re the optional types one layer above, yes I think let's leave that to a separate PR 👍

ellemouton: erg, actually - im just gonna add a commit to this pr for at least making the type for the service layer's UpdateAccount btcutil.Amount 👍

ellemouton: but let's leave the optional stuff to a diff PR

ellemouton: added commit at the end

bitromortac (approved):
Changes LGTM, the itest failure seems unrelated, not sure if this has been seen before.

ellemouton:
thanks @bitromortac , yeah this one has popped up a few times. I've made an issue for it now: https://github.com/lightninglabs/lightning-terminal/issues/940

accounts/service.go
```diff
@@ -439,15 +439,9 @@ func (s *InterceptorService) AssociateInvoice(ctx context.Context, id AccountID,
...
-	account, err := s.store.Account(ctx, id)
-	if err != nil {
-		return err
-	}
-
-	account.Invoices[hash] = struct{}{}
 	s.invoiceToAccount[hash] = id
 
-	return s.store.UpdateAccount(ctx, account)
+	return s.store.AddAccountInvoice(ctx, id, hash)
```
ViktorTigerstrom, line 444: Just noting that this changes the behaviour from the previous code, where we will now assign `s.invoiceToAccount[hash]` the `id` even if an account for that `id` doesn't exist in the db.
So if users are were to manipulate the account id with the context to an unknown id, that would occur.

I do think this should be fine even if that would somehow happen, despite the "memory leak" with the added entry to the map, as it must represent a real invoice which has added a lot of memory allocation either way, and the user wouldn't get credited if it was paid.

accounts/store_test.go
```diff
@@ -105,6 +107,140 @@ func assertEqualAccounts(t *testing.T, expected,
...
+			)
+		}
+
+		// Get the account from the store and check to see what its
+		// initial balance and expiry fields are set to.
+		assertBalanceAndExpiry(0, time.Time{})
+
+		// Now, update just the balance of the account.
+		newBalance := lnwire.MilliSatoshi(123)
+		err = store.UpdateAccountBalanceAndExpiry(
```
ViktorTigerstrom, line 139: This applies to all tests:
I think we should also test how these db functions behave if an unknown id is passed into the functions.

ViktorTigerstrom (approved):
Looks good :rocket:!

Agree with Tobias on idea of converting optional types one layer higher, so great to hear that there'll be a new PR for that :)

Also for clarification I've tested just for sanity sake that the update to use `btcutil.Amount` in the accounts `UpdateAccount` method doesn't affect the current `cli method` behaviour (as the actual rpc servers functions aren't covered by tests).

accounts/service.go
ellemouton: good catch! I've updated it to only assign to the in-mem map if the DB call succeeded. lemme know what you think

ellemouton (commented):
thanks @ViktorTigerstrom  🙏

accounts/store_test.go
ellemouton: added

PR #939, opened 2025-01-16
[sql-3] accounts: replace calls to UpdateAccount

ellemouton:
(similar PR description to #938 but they are not dependent on each other)

To prepare the `accounts.Store` to be more SQL compatible, we will want to replace the current
`UpdateAccount` method. Today it takes a full `OffChainAccount` struct, serialises it and replaces
any existing in the DB. This method is not very SQL appropriate as the better pattern is to instead
have specific update methods that read: "Update field X and account with ID Y".

So this PR starts this process of adding more SQL friendly methods to the interface and thereby
phasing out the use of `UpdateAccount`. By doing so, we also move very kvdb specific logic out
of the `accounts.InterceptorService`.

I'm splitting this up over about 3 or 4 PRs since some of the replacements are more complicated than
others.

This PR focuses on the calls that edit/add account payments:

- `UpsertAccountPayment`
- `DeleteAccountPayment`

### DIFF
--- a/accounts/errors.go
+++ b/accounts/errors.go
@@ -8,4 +8,9 @@ var (
 	ErrLabelAlreadyExists = errors.New(
 		"account label uniqueness constraint violation",
 	)
+
+	// ErrAlreadySucceeded is returned by the UpsertAccountPayment method
+	// if the WithErrAlreadySucceeded option is used and the payment has
+	// already succeeded.
+	ErrAlreadySucceeded = errors.New("payment has already succeeded")
 )
--- a/accounts/errors.go
+++ b/accounts/errors.go
@@ -13,4 +13,10 @@ var (
 	// if the WithErrAlreadySucceeded option is used and the payment has
 	// already succeeded.
 	ErrAlreadySucceeded = errors.New("payment has already succeeded")
+
+	// ErrPaymentNotAssociated indicate that the payment with the given hash
+	// has not yet been associated with the account in question.
+	ErrPaymentNotAssociated = errors.New(
+		"payment not associated with account",
+	)
 )
--- a/accounts/errors.go
+++ b/accounts/errors.go
@@ -15,7 +15,9 @@ var (
 	ErrAlreadySucceeded = errors.New("payment has already succeeded")
 
 	// ErrPaymentNotAssociated indicate that the payment with the given hash
-	// has not yet been associated with the account in question.
+	// has not yet been associated with the account in question. It is also
+	// returned when the WithErrIfUnknown option is used with
+	// UpsertAccountPayment if the payment is not yet known.
 	ErrPaymentNotAssociated = errors.New(
 		"payment not associated with account",
 	)
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -467,26 +467,7 @@ func (s *InterceptorService) PaymentErrored(ctx context.Context, id AccountID,
 			"has already started")
 	}
 
-	account, err := s.store.Account(ctx, id)
-	if err != nil {
-		return err
-	}
-
-	// Check that this payment is actually associated with this account.
-	_, ok = account.Payments[hash]
-	if !ok {
-		return fmt.Errorf("payment with hash %s is not associated "+
-			"with this account", hash)
-	}
-
-	// Delete the payment and update the persisted account.
-	delete(account.Payments, hash)
-
-	if err := s.store.UpdateAccount(ctx, account); err != nil {
-		return fmt.Errorf("error updating account: %w", err)
-	}
-
-	return nil
+	return s.store.DeleteAccountPayment(ctx, id, hash)
 }
 
 // AssociatePayment associates a payment (hash) with the given account,
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -510,10 +510,12 @@ func (s *InterceptorService) AssociatePayment(ctx context.Context, id AccountID,
 	// If the payment is already associated with the account but not in
 	// flight, we update the payment amount in case we have a zero-amount
 	// invoice that is retried.
-	return s.store.UpsertAccountPayment(
+	_, err := s.store.UpsertAccountPayment(
 		ctx, id, paymentHash, fullAmt, lnrpc.Payment_UNKNOWN,
 		WithErrIfAlreadyPending(),
 	)
+
+	return err
 }
 
 // invoiceUpdate credits the account an invoice was registered with, in case the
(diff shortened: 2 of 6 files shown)


accounts/service.go
```diff
@@ -637,34 +598,27 @@ func (s *InterceptorService) TrackPayment(ctx context.Context, id AccountID,
...
-	// track the payment again.
-	entry, ok := account.Payments[hash]
-	if ok && successState(entry.Status) {
-		return nil
+	// track the payment again. So we add the WithErrIfAlreadySucceeded
+	// option to ensure that we return an error if the payment has already
+	// succeeded. We can then match on the ErrAlreadySucceeded error and
+	// exit early if it is returned.
+	opts := []UpsertPaymentOption{
+		WithErrIfAlreadySucceeded(),
```
ViktorTigerstrom, line 607: We're not matching on `ErrAlreadySucceeded` below. This function should return nil if it is an `ErrAlreadySucceeded` error.

accounts/service.go
```diff
@@ -900,23 +845,24 @@ func (s *InterceptorService) removePayment(ctx context.Context,
 		return nil
 	}
 
-	account, err := s.store.Account(ctx, pendingPayment.accountID)
-	if err != nil {
-		return err
-	}
-
 	pendingPayment.cancel()
 	delete(s.pendingPayments, hash)
```
ViktorTigerstrom, line 849: Just noting that similar to the other PR, this will now run before we've checked if there's an account id for the pendingPayment. I do that should be unreachable though, and probably actually correct behaviour, so don't really think this needs to change.

accounts/store_kvdb.go
```diff
@@ -194,6 +196,131 @@ func (s *BoltStore) UpdateAccount(_ context.Context,
...
+
+				return fmt.Errorf("payment with hash %s is "+
+					"already in flight or succeeded "+
+					"(status %v)", paymentHash,
+					account.Payments[paymentHash].Status)
+			}
+
+			if opts.usePendingAmount {
+				fullAmount = entry.FullAmount
+			}
```
ViktorTigerstrom, line 242: Just noting that this is a difference from the current behaviour, where this was set regardless of if an entry existed or not, as long as `fullAmount` was zero. That feels pretty dangerous :sweat_smile:, so good fix :rocket:!

accounts/store_test.go
```diff
@@ -105,6 +105,188 @@ func assertEqualAccounts(t *testing.T, expected,
 	actual.LastUpdate = actualUpdate
 }
 
+// TestAccountUpdateMethods tests that all the Store methods that update an
+// account work correctly.
+func TestAccountUpdateMethods(t *testing.T) {
```
ViktorTigerstrom, line 110: Similar to the other PR, we should test the behaviour for an unknown account id :).

ViktorTigerstrom (commented):
Generally looks good! I've left a few comments below, where there's mainly one that needs to be fixed.

As also communicated offline, I do think the added options to the API `UpsertAccountPayment` makes this updated version harder to read and reason about than the current implementation. I think it's worth considering if it could be designed a bit differently as I feel there's a high risk of introducing bugs when the API is a bit complicated.

accounts/service.go
ellemouton: good catch - fixed 👍

accounts/service.go
ellemouton: can also move it below the db call yeah?

ellemouton (commented):
> I do think the added options to the API UpsertAccountPayment makes this updated version

Agreed. I do, however, want to keep this series of PRs focused on _just_ switching out the backends without changing the existing behaviour. If we want to make behaviour better, i think it should be done separately (ideas/PRs welcome!) to keep us from scope creeping on this project.

accounts/service.go
ViktorTigerstrom: Hmmm yes, specifically since these two marked functions cannot error, and we don't shutdown the `accounts` service if the public `RemovePayment` function errors, I think we should move this to after the db call.

Otherwise we can get a diff between the in-memory layer of the `service` and the `db` layer. For the other call sites that calls `removePayment`, it shouldn't matter as they'll shutdown the service, and it'll be resolved on the next startup.

ViktorTigerstrom (approved):
LGMT after the fixes :rocket:!

Just as a small note, since the latest push contained both merge changes + fixes for the PR, I think that would have been a great scenario to either push PR changes separately from merge changes, or include the changes as fixups :).

accounts/store_test.go
```diff
@@ -430,6 +430,15 @@ func TestAccountUpdateMethods(t *testing.T) {
...
 		require.ErrorContains(t, err, "is not associated")
+
+		// Try once more to insert a payment that is currently unknown
+		// but this time add the WithErrIfUnknown option. This should
+		// return the ErrPaymentUnknown error.
+		_, err = store.UpsertAccountPayment(
+			ctx, acct.ID, hash1, 600, lnrpc.Payment_SUCCEEDED,
+			WithErrIfUnknown(),
+		)
+		require.ErrorIs(t, err, ErrPaymentUnknown)
```
bitromortac, line 441: would it make sense to add a test that changes the status as well, since `removePayment` does that (perhaps using `WithErrIfUnknown` and `WithPendingAmount` together?

accounts/store_kvdb.go
```diff
@@ -317,6 +317,32 @@ func (s *BoltStore) UpsertAccountPayment(_ context.Context, id AccountID,
...
+func (s *BoltStore) DeleteAccountPayment(_ context.Context, id AccountID,
+	hash lntypes.Hash) error {
+
+	update := func(account *OffChainBalanceAccount) error {
+		// Check that this payment is actually associated with this
+		// account.
+		_, ok := account.Payments[hash]
+		if !ok {
+			return fmt.Errorf("payment with hash %s is not "+
+				"associated with this account", hash)
```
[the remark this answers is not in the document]
ellemouton: i've updated this method to return a new `ErrPaymentNotAssociated` error and then assert against that in the tests 👍 

i think the location of the errors in the errors.go is ok? what i've done instead is just indicate in the DeleteAccountPayment comment on the interface that this error is expected.

accounts/service.go
ellemouton: updated 👍

accounts/store_test.go
ellemouton: the tests already covers changing the status. See above: the initial status  for `hash2` is Unknown, then it gets changed to `succeeded`

ellemouton:
oops - fixing race now

bitromortac (commented):
Nice, looks almost good, have two non-blocking comments

accounts/errors.go
```diff
@@ -19,4 +19,8 @@ var (
 	ErrPaymentNotAssociated = errors.New(
 		"payment not associated with account",
 	)
+
+	// ErrPaymentUnknown is returned by the UpsertAccountPayment method if
+	// the WithErrIfUnknown option is used and the payment is not yet known.
+	ErrPaymentUnknown = errors.New("payment unknown")
```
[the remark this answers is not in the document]
ellemouton: indeed

accounts/store_test.go
ellemouton: gotcha 👍  - updated

bitromortac (approved):
LGTM :zap:

accounts/service.go
```diff
@@ -607,34 +609,30 @@ func (s *InterceptorService) TrackPayment(ctx context.Context, id AccountID,
...
+	// option to ensure that we return an error if the payment has already
+	// succeeded. We can then match on the ErrAlreadySucceeded error and
+	// exit early if it is returned.
+	opts := []UpsertPaymentOption{
+		WithErrIfAlreadySucceeded(),
 	}
 
 	// There is a case where the passed in fullAmt is zero but the pending
 	// amount is not. In that case, we should not overwrite the pending
 	// amount.
```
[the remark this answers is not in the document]
bitromortac: @ViktorTigerstrom, do you recall how when that can happen?

PR #941, opened 2025-01-19
[sql-4] accounts: remove last usages of `UpdateAccount`

ellemouton:
With this PR, we complete the `Prepare the accounts Store interface to be ready for a SQL impl` & `Prep unit tests to be ready to pass against a different backend` steps in #917 .

We do 2 things here:
1) we [revert one of the changes](https://github.com/lightninglabs/lightning-terminal/pull/938#discussion_r1918123976) from a previous PR where we change the balance type of `UpdateAccountBalanceAndExpiry` from an int64 to a more strict lnwire.Millisatoshi type. We revert that here though so that we can use the `UpdateAccountBalanceAndExpiry` in a unit test that uses `UpdateAccount` to set the balance to a negative value (which currently _is_ allowed). 

2) The rest of the PR removes the last couple of usages of `UpdateAccount` from tests.

### DIFF
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -207,11 +207,6 @@ type Store interface {
 		expirationDate time.Time, label string) (
 		*OffChainBalanceAccount, error)
 
-	// UpdateAccount writes an account to the database, overwriting the
-	// existing one if it exists.
-	UpdateAccount(ctx context.Context,
-		account *OffChainBalanceAccount) error
-
 	// Account retrieves an account from the Store and un-marshals it. If
 	// the account cannot be found, then ErrAccNotFound is returned.
 	Account(ctx context.Context, id AccountID) (*OffChainBalanceAccount,
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -223,7 +223,7 @@ type Store interface {
 	// UpdateAccountBalanceAndExpiry updates the balance and/or expiry of an
 	// account.
 	UpdateAccountBalanceAndExpiry(ctx context.Context, id AccountID,
-		newBalance fn.Option[lnwire.MilliSatoshi],
+		newBalance fn.Option[int64],
 		newExpiry fn.Option[time.Time]) error
 
 	// AddAccountInvoice adds an invoice hash to an account.
--- a/accounts/service_test.go
+++ b/accounts/service_test.go
@@ -22,8 +22,6 @@ var (
 	testExpiration = time.Now().Add(24 * time.Hour)
 	testTimeout    = time.Millisecond * 500
 	testInterval   = time.Millisecond * 20
-
-	testID2 = AccountID{22, 22, 22}
 )
 
 type mockLnd struct {
@@ -203,20 +201,22 @@ func TestAccountService(t *testing.T) {
 	testCases := []struct {
 		name  string
 		setup func(t *testing.T, lnd *mockLnd, r *mockRouter,
-			s *InterceptorService)
+			s *InterceptorService) []AccountID
 		startupErr string
 		validate   func(t *testing.T, lnd *mockLnd, r *mockRouter,
-			s *InterceptorService)
+			ids []AccountID, s *InterceptorService)
 	}{{
 		name: "startup err on invoice subscription",
 		setup: func(t *testing.T, lnd *mockLnd, r *mockRouter,
-			s *InterceptorService) {
+			s *InterceptorService) []AccountID {
 
 			lnd.invoiceSubscriptionErr = testErr
+
+			return nil
 		},
 		startupErr: testErr.Error(),
 		validate: func(t *testing.T, lnd *mockLnd, r *mockRouter,
-			s *InterceptorService) {
+			ids []AccountID, s *InterceptorService) {
 
 			lnd.assertNoInvoiceRequest(t)
 			require.False(t, s.IsRunning())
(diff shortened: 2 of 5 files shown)


accounts/interface.go
```diff
@@ -207,11 +207,6 @@ type Store interface {
 		expirationDate time.Time, label string) (
 		*OffChainBalanceAccount, error)
 
-	// UpdateAccount writes an account to the database, overwriting the
-	// existing one if it exists.
-	UpdateAccount(ctx context.Context,
```
bitromortac, line 212: nice to see this and how you changed to code to achieve that :tada:

accounts/service_test.go
```diff
@@ -202,43 +200,40 @@ func TestAccountService(t *testing.T) {
...
-				Invoices: AccountInvoices{
-					testHash: {},
-				},
-			}
+			s *InterceptorService) []AccountID {
 
-			err := s.store.UpdateAccount(ctx, acct)
+			acct, err := s.store.NewAccount(
+				ctx, 1234, testExpiration, "",
+			)
```
bitromortac, line 230: we don't want to add the invoice here, right (and in the next test)?

accounts/service_test.go
```diff
@@ -836,38 +813,46 @@ func TestAccountService(t *testing.T) {
...
 			}
 
 			// Any errors during startup expected?
 			err = service.Start(
 				ctx, lndMock, routerMock, chainParams,
 			)
 			if tc.startupErr != "" {
-				require.ErrorContains(tt, err, tc.startupErr)
+				require.ErrorContains(t, err,
+					tc.startupErr)
```
bitromortac, line 831: nit: line break not needed

accounts/service_test.go
```diff
@@ -836,38 +813,46 @@ func TestAccountService(t *testing.T) {
...
+						accountIDs, service,
 					)
 				}
 
 				return
 			}
 
-			// Any post execution validation that we need to run?
+			// Any post execution validation that we need to
+			// run?
```
bitromortac, line 846: nit: line break

accounts/service_test.go
ellemouton: didnt add that cause it doesnt affect the tests/isnt what is being tested. 

but re-added it for consistency and easier review 👍

ellemouton:
thanks @bitromortac ! gonna put it in "ready for review" since it is acked so that i can address all review comments in 1 go :)

accounts/store_test.go
```diff
@@ -37,9 +36,43 @@ func TestAccountStore(t *testing.T) {
...
+
+	// Add 2 invoices.
+	err = store.AddAccountInvoice(
+		ctx, acct1.ID, lntypes.Hash{12, 34, 56, 78},
+	)
+	require.NoError(t, err)
+	err = store.AddAccountInvoice(
+		ctx, acct1.ID, lntypes.Hash{34, 56, 78, 90},
+	)
+	require.NoError(t, err)
```
ViktorTigerstrom, line 70: nit: would personally put these after updating the in-memory `acct1`, as that makes more logical sense, and would also then change it to something like:

```
	for paymentHash, entry := range acct1.Payments {
		_, err = store.UpsertAccountPayment(
			ctx, acct1.ID, paymentHash, entry.FullAmount,
			entry.Status,
		)
		require.NoError(t, err)
	}

	for paymentHash := range acct1.Invoices {
		err = store.AddAccountInvoice(ctx, acct1.ID, paymentHash)
		require.NoError(t, err)
	}
```

To not have to repeat hashes and amounts.

ViktorTigerstrom (approved):
LGTM once feedback has been addressed :)!

accounts/service_test.go
```diff
@@ -836,25 +821,27 @@ func TestAccountService(t *testing.T) {
...
+			var accountIDs []AccountID
 			if tc.setup != nil {
-				tc.setup(t, lndMock, routerMock, service)
+				accountIDs = tc.setup(
+					t, lndMock, routerMock, service,
+				)
 			}
 
 			// Any errors during startup expected?
 			err = service.Start(
```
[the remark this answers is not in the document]
ellemouton: pre-existing but added a commit to address

accounts/store_test.go
ellemouton: > To not have to repeat hashes and amounts.

I kind of like keeping tests very explicit. Else a bug in one version can be transferred to the other version.

For example, if you are testing an `func Add(x, y int) int` function, then one should test explicitly like:

```
x := 1
y := 2
expected := 3
result := Add(x, y)
require.Equal(3, result)
```
 
instead of:

```
x := 1
y := 2
expected := x+y
result := Add(x, y)
require.Equal(expected, result)
```

accounts/store_test.go
```diff
@@ -96,8 +128,8 @@ func assertEqualAccounts(t *testing.T, expected,
 	actual.LastUpdate = time.Time{}
 
 	require.Equal(t, expected, actual)
-	require.Equal(t, expectedExpiry.UnixNano(), actualExpiry.UnixNano())
-	require.Equal(t, expectedUpdate.UnixNano(), actualUpdate.UnixNano())
+	require.Equal(t, expectedExpiry.Unix(), actualExpiry.Unix())
```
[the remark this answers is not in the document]
ellemouton: not sure what you mean by "this one" and "the other one"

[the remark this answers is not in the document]
ellemouton: reverted it for now. But i think we miiiight need to change it back to Unix() when we have the sqlite backend cause i dont think it does nanosecond precision. But will leave that for that PR 👍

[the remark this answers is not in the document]
bitromortac: > not sure what you mean by "this one" and "the other one"

for this, I referred to the marked line, because the `expectedExpiry` had the exact same timestamp, so `Equal` would have still worked, just the updated timestamp had a small diff

PR #944, opened 2025-01-21
[sql-5] accounts: use clock.Clock

ellemouton:
In this commit, we use the clock.Clock type instead of directly calling time.Now().
This is in preparation for cleaner testing against different SQL backends later on in the series.


bitromortac (approved):
LGTM :tada: this comment could be addressed here https://github.com/lightninglabs/lightning-terminal/pull/941#discussion_r1925079612

ViktorTigerstrom (approved):
:fire:

PR #950, opened 2025-01-22
Update tapd to v0.5.1-alpha.rc1, bump version to v0.14.1-alpha.rc1

guggero:

PR #952, opened 2025-01-23
Update faraday to v0.2.14-alpha

guggero:
Bumps `faraday` to the latest version and adds the version info for `litd v0.14.1-alpha` to the README (which was both forgotten in https://github.com/lightninglabs/lightning-terminal/pull/950).


ViktorTigerstrom (approved):
ACK on the faraday bump!

Regarding the second point:
> adds the version info for litd v0.14.1-alpha to the README (which was both forgotten in https://github.com/lightninglabs/lightning-terminal/pull/950).

I thought that was intentionally left out in the previous PR, as this is only an RC release, but this goes into the master branch README. If we want to include it, should we at least mention that those versions are for the RC version?

guggero:
> If we want to include it, should we at least mention that those versions are for the RC version?

Hmm, yeah, good point. It's the first time we do an RC from the `master` branch, so not sure what the procedure should look like.
IMO we have two options:
 - Don't update the README for RCs, only for final releases
 - Update the README for RCs but mention the `.rcX` suffix specifically, then for the final release remove the suffix

What do y'all prefer, cc @ellemouton?

ViktorTigerstrom:
I'm ok with both options, but with a slight lean torqrds:
> Don't update the README for RCs, only for final releases

But I don't have a strong opinion, so if you both prefer to add it to the README with the RC suffix, I'd be ok with that as well.

ellemouton (approved):
Yeah either one of those options sounds good to me. Perhaps more info is better, and ie, let's include the `-rc` in the README just so that it is up to date for folks running the rc tag.

guggero:
Hmm, two different preferences, so I decided to go with the one easiest to implement (e.g. add the `.rc1` suffix to the README).

PR #953, opened 2025-01-23
[sql-6] accounts: SQL boilerplate and accounts schemas and queries

ellemouton:
This PR addresses the next step in #917 

Here, we add the set of account tables and queries we will need to implement the SQL version of the accounts store. 

Since this is the first PR that makes use of sqlc, it also includes all the boilerplate code we need. The boiler plate code is copied from the `taproot-assets` repo.

While it would have been ideal to have a single PR that adds the boiler plate code and then a separate one for adding the accounts schemas, this is not possible due to the `make sqlc` command not working if no queries are defined. 

## TODO: i'll update this description with more details about the accounts schema in a bit (will leave in draft until then)


db/sqlc/migrations/000001_accounts.up.sql
```diff
@@ -0,0 +1,71 @@
...
+-- are associated with a particular account. These are used to determine
+-- when an account should be credited.
+CREATE TABLE IF NOT EXISTS account_invoices (
+    -- The account that this invoice is linked to.
+    account_id BIGINT NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
+
+    -- The payment hash of the invoice.
+    hash BLOB NOT NULL,
+
+    UNIQUE(account_id, hash)
```
bitromortac, line 60: these unique constraints are only for the individual tables, right? i.e. what happens if the same account tries to pay itself (invoice+payment), will it violate the constraint (if there would be only a single index)?

db/sqlc/migrations/000001_accounts.up.sql
```diff
@@ -0,0 +1,71 @@
...
+
+-- The account_indices table stores any string-to-integer mappings that are
+-- used by the accounts system.
+CREATE TABLE IF NOT EXISTS account_indicies (
+    -- The unique name of the index.
+    name TEXT NOT NULL UNIQUE,
+
+    -- The current value of the index.
+    value BIGINT NOT NULL
+);
```
bitromortac, line 71: is this correct? shouldn't this create an index instead of a table, to be the reverse of `DROP INDEX IF EXISTS idx_accounts_alias_prefix;`?

db/sqlc/migrations/000001_accounts.up.sql
```diff
@@ -0,0 +1,71 @@
...
+
+    -- An optional label to use for the account. If it is set, it must be
+    -- unique.
+    label TEXT UNIQUE,
+
+    -- The account type.
+    type SMALLINT NOT NULL,
+
+    -- The accounts initial balance. This is never updated.
+    intial_balance_msat BIGINT NOT NULL,
```
bitromortac, line 18: typo: `initial_balance_msat`

db/sqlc/migrations/000001_accounts.up.sql
```diff
@@ -0,0 +1,71 @@
+CREATE TABLE IF NOT EXISTS accounts (
+    -- The auto incrementing primary key.
+    id INTEGER PRIMARY KEY,
+
+    -- The ID that was used to identify the account in the legacy KVDB store.
+    -- In order to avoid breaking the API, we keep this field here so that
+    -- we can still look up accounts by this ID for the time being.
+    alias BLOB NOT NULL UNIQUE,
```
bitromortac, line 8: do we need to use blob here or can this be a more efficient type, maybe a bigint?

db/sqlc/migrations/000001_accounts.down.sql
```diff
@@ -0,0 +1,4 @@
+DROP INDEX IF EXISTS idx_accounts_alias_prefix;
```
bitromortac, line 1: `idx_accounts_alias_prefix` seems to be inconsistent in the naming

bitromortac (commented):
Did a first pass and only looked at the schemas. I've got a few questions to build up knowledge in the SQL area.

db/sqlc/migrations/000001_accounts.up.sql
ellemouton: yeah these are per table

db/sqlc/migrations/000001_accounts.up.sql
ellemouton: i think maybe you misunderstand the point of this table. It isnt for the alias thing. this is for the settle & add index. See the current schema and the sql diagram in the issue.

db/sqlc/migrations/000001_accounts.up.sql
ellemouton: good idea!

db/sqlc/migrations/000001_accounts.up.sql
bitromortac: ah right, thank you!

db/sqlc/migrations/000001_accounts.up.sql
ellemouton: updated

bitromortac (commented):
Looking good, nice initial additions! I've got some more SQL-related questions.

db/sqlc/models.go
```diff
@@ -0,0 +1,38 @@
...
+	Alias              int64
+	Label              sql.NullString
+	Type               int16
+	InitialBalanceMsat int64
+	CurrentBalanceMsat int64
+	LastUpdated        time.Time
+	Expiration         time.Time
+}
+
+type AccountIndicy struct {
```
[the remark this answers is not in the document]
ellemouton: it's the return type that would be returned if we were to have a query such as "select * from account_indices"
but we only have "select value from account_indices where name = x", so this doesnt get used. But it is still generated

db/sqlc/queries/accounts.sql
```diff
@@ -0,0 +1,100 @@
...
+UPDATE accounts
+SET last_updated = $1
+WHERE id = $2
+RETURNING id;
+
+-- name: AddAccountInvoice :exec
+INSERT INTO account_invoices (account_id, hash)
+SELECT id, $1
+FROM accounts
+WHERE id = $2;
```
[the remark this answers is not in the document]
ellemouton: ah yes good point - not sure what i was thinking. Same below. Will fix!

db/sqlc/queries/accounts.sql
```diff
@@ -0,0 +1,100 @@
...
+
+-- name: DeleteAccountPayment :exec
+DELETE FROM account_payments
+WHERE hash = $1
+  AND account_id = (SELECT id FROM accounts WHERE id = $2);
```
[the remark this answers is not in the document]
ellemouton: removed :)

db/sqlc/migrations/000001_accounts.up.sql
```diff
@@ -0,0 +1,71 @@
...
+    -- The time that the account will expire.
+    expiration TIMESTAMP NOT NULL
+);
+
+-- The account_payments table stores all the payment hashes of outgoing
+-- payments that are associated with a particular account. These are used to
+-- when an account should be debited.
+CREATE TABLE IF NOT EXISTS account_payments (
+    -- The account that this payment is linked to.
+    account_id BIGINT NOT NULL REFERENCES accounts(id) ON DELETE CASCADE,
```
ViktorTigerstrom, line 35: It feels a bit dangerous to do `ON DELETE CASCADE` rather than `RESTRICT` here and for `account_invoices`, as in my opinion we should be really careful when deleting an account with existing payments and invoices, and really ensure that those have been deleted prior to deleting the account.

But given our mindset of not changing the original logic, and that's how it's handled today, I guess that motivates having it as a `CASCADE`.

db/sqlc/queries/accounts.sql
```diff
@@ -0,0 +1,92 @@
...
+SELECT *
+FROM  account_invoices
+WHERE account_id = $1
+  AND hash = $2;
+
+-- name: SetAccountIndex :exec
+INSERT INTO account_indicies (name, value)
+VALUES ($1, $2)
+    ON CONFLICT (name)
+DO UPDATE SET value = $2;
```
ViktorTigerstrom, line 87: This might fall into the category of maintaining the current logic, which could justify not changing it. However, in my opinion, since this table will always have only two entries (`settle_index` and `add_index`), it might make sense to initialize the table with these two rows during setup, with `0` as the initial value for the `value` column.

If we do this, instead of using the `UPSERT` version, we could simplify it to just an `UPDATE` query.

ViktorTigerstrom (approved):
Niiiiice, uTACK LGTM :rocket::fire:! Leaving two comments which probably falls out of scope for this PR.

db/sqlc/migrations/000001_accounts.up.sql
ellemouton: ooo TIL about RESTRICT! 

very nice

but yeah, let's keep as is for now to keep the tests passing :) can update during account revamp fosho

db/sqlc/queries/accounts.sql
ellemouton: > since this table will always have only two entries

I dont think we know this for sure right? it's common to have a cursor table like this and could be useful for future I'd say.

But yeah, like you say, let's leave for project "revamp accounts" :)

bitromortac (approved):
LGTM :tada: nice to see this :zap: (one suggestion left)

db/sqlc/models.go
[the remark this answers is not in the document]
ellemouton: indeed! cool - updated :)

PR #954, opened 2025-01-23
[sql-7] accounts: add SQL implementation of the accounts store

ellemouton:
Part of #917 

In this commit, we add the SQLStore type which implements the
accounts.Store interface. To demonstrate that it works as expected, we
also plug this implementation into all the account unit tests to show
that they pass against the sqlite and postgres backends.

One can use `make unit pkg=accounts tags=test_db_postgres` or
`make unit pkg=accounts tags=test_db_sqlite` to test locally.

### DIFF
--- a/accounts/store_sql.go
+++ b/accounts/store_sql.go
@@ -1,0 +1,743 @@
+package accounts
+
+import (
+	"context"
+	"crypto/rand"
+	"database/sql"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/lightninglabs/lightning-terminal/db"
+	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightningnetwork/lnd/clock"
+	"github.com/lightningnetwork/lnd/fn"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+)
+
+const (
+	// addIndexName is the name of the key under which we store the last
+	// known invoice add index in the accounts_indices table.
+	addIndexName = "last_add_index"
+
+	// settleIndexName is the name of the key under which we store the
+	// last known invoice settle index in the accounts_indices table.
+	settleIndexName = "last_settle_index"
+)
+
+// SQLQueries is a subset of the sqlc.Queries interface that can be used
+// to interact with accounts related tables.
+//
+//nolint:lll
+type SQLQueries interface {
+	AddAccountInvoice(ctx context.Context, arg sqlc.AddAccountInvoiceParams) error
+	DeleteAccount(ctx context.Context, id int64) error
+	DeleteAccountPayment(ctx context.Context, arg sqlc.DeleteAccountPaymentParams) error
+	GetAccount(ctx context.Context, id int64) (sqlc.Account, error)
+	GetAccountByLabel(ctx context.Context, label sql.NullString) (sqlc.Account, error)
+	GetAccountIDByAlias(ctx context.Context, alias int64) (int64, error)
+	GetAccountIndex(ctx context.Context, name string) (int64, error)
+	GetAccountPayment(ctx context.Context, arg sqlc.GetAccountPaymentParams) (sqlc.AccountPayment, error)
+	InsertAccount(ctx context.Context, arg sqlc.InsertAccountParams) (int64, error)
+	ListAccountInvoices(ctx context.Context, id int64) ([]sqlc.AccountInvoice, error)
+	ListAccountPayments(ctx context.Context, id int64) ([]sqlc.AccountPayment, error)
+	ListAllAccounts(ctx context.Context) ([]sqlc.Account, error)
+	SetAccountIndex(ctx context.Context, arg sqlc.SetAccountIndexParams) error
+	UpdateAccountBalance(ctx context.Context, arg sqlc.UpdateAccountBalanceParams) (int64, error)
+	UpdateAccountExpiry(ctx context.Context, arg sqlc.UpdateAccountExpiryParams) (int64, error)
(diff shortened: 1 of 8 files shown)


accounts/store_sql.go
```diff
@@ -0,0 +1,704 @@
...
+	SQLQueries
+
+	db.BatchedTx[SQLQueries]
+}
+
+// SQLStore represents a storage backend.
+type SQLStore struct {
+	db BatchedSQLQueries
+
+	*sql.DB
```
ViktorTigerstrom, line 68: Nit:
It might be helpful to add docs for `db` and `DB` to clarify their differences in the IDE during development, as their field names are very similar.

accounts/store_sql.go
```diff
@@ -0,0 +1,704 @@
...
+
+		newAliasID, err := newAlias.ToInt64()
+		if err != nil {
+			return 0, err
+		}
+
+		aliasInt, err := newAlias.ToInt64()
+		if err != nil {
+			return 0, err
+		}
```
ViktorTigerstrom, line 253: Both of these calls newAlias.ToInt64(), so seems unnecessary to create two variables.

accounts/store_sql.go
```diff
@@ -0,0 +1,704 @@
...
+					ID:         id,
+					Expiration: t.UTC(),
+				},
+			)
+		})
+		if err != nil {
+			return err
+		}
+
+		return s.markAccountUpdated(ctx, db, id)
```
ViktorTigerstrom, line 383: Should maybe check that either `newBalance` or `newExpiry` is `Some` before `markAccountUpdated`?

accounts/store_sql.go
```diff
@@ -0,0 +1,704 @@
...
+			// If the errIfAlreadyPending option is set, we return
+			// an error if the payment is already in-flight or
+			// succeeded.
+			if opts.errIfAlreadyPending &&
+				currStatus != lnrpc.Payment_FAILED {
+
+				return fmt.Errorf("payment with hash %s is "+
+					"already in flight or succeeded "+
+					"(status %v)", hash, currStatus)
+			}
```
ViktorTigerstrom, line 552: Should probably have commented this in kvdb PR, but this will return an error if the `currStatus` is `lnrpc.Payment_UNKOWN`, which seems a bit incorrect?

accounts/store_sql.go
```diff
@@ -0,0 +1,704 @@
...
+			if errors.Is(err, sql.ErrNoRows) {
+				return ErrAccNotFound
+			} else if err != nil {
+				return err
+			}
+
+			_, err = db.UpdateAccountBalance(
+				ctx, sqlc.UpdateAccountBalanceParams{
+					ID: id,
+					CurrentBalanceMsat: acct.CurrentBalanceMsat -
```
ViktorTigerstrom, line 586: nit: Line length

ViktorTigerstrom (commented):
Generally looks great :rocket::fire:! Just a few minor comments.

accounts/store_sql.go
ellemouton: oversight 👍

accounts/store_sql.go
ellemouton: https://github.com/lightninglabs/lightning-terminal/blob/1ba09c3cf0701bd50b41462f652846151de9c73a/accounts/store_kvdb.go#L193-L208

accounts/store_sql.go
ellemouton: as discussed offline, seems like a parallel issue right?

accounts/store_sql.go
ViktorTigerstrom: Right, this is existing behaviour :)!
Also rechecked the old code prior to the addition of the `kvdb` refactors, and verified that it was existing behaviour prior to that as well.

accounts/store_sql.go
ViktorTigerstrom: I've rechecked this, and the current behavior is correct—it should return an error if `currStatus != lnrpc.Payment_FAILED`. In other words, it should also error on `lnrpc.Payment_UNKNOWN`. 

I believe the confusion comes from the doc comments, which state that this only fails if the payment is "in-flight or succeeded." However, we actually treat `lnrpc.Payment_UNKNOWN` as "in-flight" as well, which I think would be clearer if the docs mentioned that.

I can update the doc comments in a separate PR after this is merged if you'd like.

ViktorTigerstrom (approved):
LGTM :rocket:!

accounts/store_sql.go
ellemouton: I can update the docs on next review round :)

accounts/store_sql.go
```diff
@@ -0,0 +1,745 @@
...
+
+		// First check that this invoice does not already exist.
+		_, err = db.GetAccountInvoice(ctx, sqlc.GetAccountInvoiceParams{
+			AccountID: acctID,
+			Hash:      hash[:],
+		})
+		// If it does, there is nothing left to do.
+		if err == nil {
+			return nil
+		} else if err != nil && !errors.Is(err, sql.ErrNoRows) {
```
bitromortac, line 289: nit: I think `err != nil &&` could be removed?

bitromortac (approved):
Awesome to see that unit tests pass with this implementation, great work LGTM :tada::tada:! I compared all old/new methods, they should enforce the same constraints :+1:. Have only a few nits

accounts/store_sql.go
ellemouton: since the minimum compatible version of LND with LiT is v0.18.5-beta, and this UNKNOWN status will never be returned in that version since it is deprecated, i think it is fine to leave as is?

```
// Deprecated. This status will never be returned.
//
// Deprecated: Marked as deprecated in lightning.proto.
Payment_UNKNOWN Payment_PaymentStatus = 0
```

PR #955, opened 2025-01-23
[sql-8]: run itests against all DB types

ellemouton:
This completes the final step of the accounts section in phase 1 of https://github.com/lightninglabs/lightning-terminal/issues/917

Here we run our CI itests against all DB types. As of this PR, that means that the accounts store is spun up with different DB backends for the various iterations. 

(more detailed description to follow)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -275,6 +275,17 @@ jobs:
   itest:
     name: integration test
     runs-on: ubuntu-latest
+    strategy:
+      # Allow other tests in the matrix to continue if one fails.
+      fail-fast: false
+      matrix:
+        include:
+          - name: bbolt
+            args: dbbackend=bbolt
+          - name: sqlite
+            args: dbbackend=sqlite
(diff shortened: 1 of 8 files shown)


ViktorTigerstrom (approved):
Niiiiiiiice, looks great :rocket:!!

The itest CI check for all individual backends have passed, but the check hasn't reported back though. Is there something additional that needs to be added?

bitromortac (approved):
LGTM 🏁 :zap:

ellemouton:
> The itest CI check for all individual backends have passed, but the check hasn't reported back though. Is there something additional that needs to be added?

no - it's just an artifact from the previous name (since we now longer use that name). Might mean needing to clear a GH cache or something

ellemouton:
@ViktorTigerstrom  - turns out the reason was just that GH still had that old CI name listed under branch protection rules as a "Required" step. So just had to remove that and add the "Required" check for the new names :)

PR #962, opened 2025-02-05
accounts: add functionality to update balance by given amount 

### LINKED ISSUE #648: `accounts`: extend `update` command functionality

The functionality of the update command is that the command takes a id for an account, and then allows you to specify certain properties of the account to overwrite the account property value with.

One of the properties the command accepts, it to overwrite the `account_balance` with a specified value. This command will usually be used if you'd like to for example increase the account balance by 100k sats or similar.

However, using the command that way, is quite dangerous as the normal user flow would be as follows:
1. You'd check what the current account balance is by using the info command. Say this command returns that the balance is currently 200k.
2. You'd then sum the current balance with the amount you'd like to increase it with to get the new balance of the account, i.e.:
200k+100k=300k
3. You'd then run the `update` command to set the value to 300k.

Now the issue with this user flow, is that the the account user might have either received a payment to an invoice increasing the balance, or made a payment decreasing the balance, between step 1 and 3. Step 3 would then overwrite that account balance update. Say the accounts user makes a payment of 150k sats between step 1 and 3. The actual user account balance should be 150k after step 3, but will be 300k, meaning the account user profited 150k sats.

Another an issue we have, is that if a user receives an invoice update for a type of invoice we don't support (AMP invoices currently), or a payment that somehow errors, the entire account service will error be disabled for all accounts, and will be disabled even on next startup of litd when we try to process the invoice update or payment. 

It would be great if we could either update the `update` command, or make a new command to instead of overwriting the entire account balance, we'd have a command that either increases or decreases the user account balance by a certain amount, instead of setting the balance to a given amount. That way, any updates that happens between step 1 & 3 wouldn't be overwritten. I.e. the command would for example enable the node user to increase the user account balance by 100k, instead of actually setting it to exactly 300k.

To solve the issue of invoices and payments that make the account service crash, it would be great if the command would allow the node runner to deassociate a certain invoice or payment for a certain account id. That way, the account service would be able to run successfully on next startup.

**Additional context**

ViktorTigerstrom:
Fixes #648

Based on #954

This PR introduces functionality to update an account's balance by a given amount, rather than the current update method, which sets the balance to a specific value.

My plan is to deprecate the old functionality in a follow-up PR after this one is merged.


db/sqlc/queries/accounts.sql
```diff
@@ -9,6 +9,15 @@ SET current_balance_msat = $1
...
 
+-- name: AdjustAccountBalance :one
+UPDATE accounts
+SET current_balance_msat = current_balance_msat + CASE
+    WHEN sqlc.arg(is_addition)::BOOLEAN THEN sqlc.arg(amount)  -- If IsAddition is true, add the amount
+    ELSE -sqlc.arg(amount)         -- If IsAddition is false, subtract the amount
+END
+WHERE id = sqlc.arg(id)
+RETURNING id;
+
```
ViktorTigerstrom, line 20: Let me know if you prefer that I use the current functionality instead to fetch the current balance, and then set the balance to the `currentBalance +- amount`.
My motivation for introducing this was to discourage such functionality in the future, which directly sets the balance.

ellemouton:
@ViktorTigerstrom  - wont it be easier to get this in _before_ we merge #954? that way we dont have to go add edit migrations for tables we havent even created yet. Im happy to hold off on that PR and then adjust it accordingly once this is in

accounts/rpcserver.go
```diff
@@ -137,7 +137,46 @@ func (s *RPCServer) UpdateAccount(ctx context.Context,
...
 
-	return nil, fmt.Errorf("not implemented")
+	var (
+		isAddition bool
+		amount     lnwire.MilliSatoshi
+	)
+
+	switch reqType := req.GetUpdate().(type) {
+	case *litrpc.UpdateAccountBalanceRequest_Add:
+		log.Infof("[addbalance] id=%s, label=%v, amount=%d",
```
ellemouton, line 147: what do you think of using "credit/debit" instead of "add/deduct"?

accounts/store_test.go
```diff
@@ -71,6 +72,18 @@ func TestAccountStore(t *testing.T) {
 	)
 	require.NoError(t, err)
 
+	// Adjust the account balance by first adding 10000, and then deducting
+	// 5000.
```
ellemouton, line 76: for review it's nicer if unit tests are in the same commit as the new addition :)

cmd/litcli/accounts.go
```diff
@@ -26,6 +27,7 @@ var accountsCommands = []cli.Command{
 		Subcommands: []cli.Command{
 			createAccountCommand,
 			updateAccountCommand,
+			updateBalanceCommands,
```
ellemouton, line 30: wondering if we cant fit this under the existing `update` category somehow?

accounts/rpcserver.go
```diff
@@ -132,6 +132,14 @@ func (s *RPCServer) UpdateAccount(ctx context.Context,
 	return marshalAccount(account), nil
 }
 
+// UpdateBalance adds or deducts an amount from an existing account in the
```
ellemouton, line 135: just an idea for standard PR flow: we always try to (as much as possible at least) to keep commits ordered such that things are working commit by commit. For 2 reasons: 1) reverting back to any single commit isnt a problem and 2) review story telling. 

so the ideal flow for RPC additions is:
1) add DB/Service functionality along with any tests for those (in small PR world, that could honestly be the whole PR - since that is like 1 unit of code that is easy to reason about and also lets us continue the on the accounts SQL PR while we iterate on the second unit here which is proto/CLI design).
2) add proto definitions & call the new code from the rpc server code
3) add CLI calling code

accounts/rpcserver.go
```diff
@@ -137,7 +137,46 @@ func (s *RPCServer) UpdateAccount(ctx context.Context,
...
+			req.Id, req.Label, reqType.Deduct)
+
+		isAddition = false
+		amount = lnwire.MilliSatoshi(reqType.Deduct * 1000)
+	}
+
+	if amount <= 0 {
+		return nil, fmt.Errorf("amount %v must be greater than 0",
+			int64(amount/1000))
+	}
```
ellemouton, line 164: why not use uints in the proto definitions then?

ellemouton (commented):
took quick peek (before review was requested - hope you dont mind!) - just leaving some drive by comments in the mean time

db/sqlc/queries/accounts.sql
```diff
@@ -9,6 +9,15 @@ SET current_balance_msat = $1
...
 
+-- name: AdjustAccountBalance :one
+UPDATE accounts
+SET current_balance_msat = current_balance_msat + CASE
+    WHEN sqlc.arg(is_addition) THEN sqlc.arg(amount)  -- If IsAddition is true, add the amount
+    ELSE -sqlc.arg(amount)         -- If IsAddition is false, subtract the amount
+END
+WHERE id = sqlc.arg(id)
+RETURNING id;
+
```
ellemouton, line 20: i think let's keep things simple with just having
`CreditAccount` and `DebitAccount` queries?

db/sqlc/queries/accounts.sql
ellemouton: agreed re having specific queries for how we prefer to use it 👍 
see comment re splitting this out though

cmd/litcli/accounts.go
ellemouton: also, how about just having explicit credit/debit sub commands?

litrpc/lit-accounts.proto
```diff
@@ -148,6 +154,28 @@ message UpdateAccountRequest {
     string label = 4;
 }
 
+message UpdateAccountBalanceRequest {
```
ellemouton, line 157: so expanding on my other comment: it doesnt really make sense to starting off the PR with rpc definitions as we dont really know how this is going to be used yet etc.

litrpc/lit-accounts.proto
```diff
@@ -148,6 +154,28 @@ message UpdateAccountRequest {
...
+    /*
+    The balance update must either be an increase or a decrease of the balance
+    */
+    oneof update {
+        // Increase the account's balance by the given amount.
+        int64 add = 3;
+
+        // Deducts the given amount from the account balance.
+        int64 deduct = 4;
+    }
```
ellemouton, line 176: when defining enums like this, it's good to ask: is there a future where more than 2 options will actually be needed? if not, then this can either be a boolean. Or we can even just have separate `Credit`/`Debit` definitions

ellemouton:
> My plan is to deprecate the old functionality in a follow-up PR after this one is merged.

good idea. I think we always want to give notice of deprecated endpoints though. So i think in the proto definitions here, you can add a deprecated annotation along with the release where it will be completely removed. Then in the release notes, add a notice about this planned depreciation. Then our next release (or couple releases) will include supporting both to give folks a chance to update to the new endpoint

litrpc/lit-accounts.proto
ellemouton: so either:

```
Update {
       - ID
       - Amount
       - Credit/Debit bool
}
```

or 

```
Credit { ID, Amount }
_and_
Debit {ID, Amount}
```

ViktorTigerstrom: > redit { ID, Amount }
> _and_
> Debit {ID, Amount}

Thanks! Went with the second option :)

db/sqlc/queries/accounts.sql
ViktorTigerstrom: Updated in the new PR :)

litrpc/lit-accounts.proto
ViktorTigerstrom: Addressed in the new PR :)

accounts/rpcserver.go
ViktorTigerstrom: Good idea! Addressed in the new PR.

cmd/litcli/accounts.go
ViktorTigerstrom: > also, how about just having explicit credit/debit sub commands?

Went with this option!

accounts/rpcserver.go
ViktorTigerstrom: Ah yeah definitely! I was struggling with naming, and agree that credit & debit is better.

accounts/rpcserver.go
ViktorTigerstrom: Thanks a lot for the suggestion! I've tried to follow this with the introduction of the new #973 & #974 PRs.

Issue #966, opened 2025-02-09
Prepare the sessions `Store` interface to be ready for a SQL impl


ellemouton:
With the account store sorted, next up is the session store which once normalised will look as follows with an optional reference to the accounts table since a session _might_ be linked to an account:

![Image](https://github.com/user-attachments/assets/ea1a8b6c-9df8-446d-a8f0-e3aed915a45f)

## Context

A couple of points to keep in mind for this sub-project:

#### Session DB level ID vs Code/User level Alias:

- Similarly to `accounts` which now has a _DB_ level ID along with a _code level_ 8 byte _alias_ which is called `AccountID` in the code base, for sessions, we have a `session.ID` (4 bytes) in the code base but we will now have a DB level int ID for sessions too. And just like for accounts, we want to continue being able to reference sessions using the alias for backwards compatibility. 
            - Note: I'll probably open PRs to rename these variables to be `accounts.Alias` and `sessions.Alias` to remove confusion. 

#### The link from a session to an account:

- Today, an account can be tied to a session but this is done implicitly by embedding the account ID in the session's macaroon caveat. In SQL land, however, we would want this link to be explicit via a foreign key. We'll then want to have tests that cover that we cant add a session linked to an account if that account doesnt exist yet - but to do this, we first need to make sure that the existing KVDB code also does not allow this. So there will be some prep work to add these checks for the existing logic before adding SQL stuff. 

#### Session Creation flow 

- Today, a session is created as follows: 
          1) Lock Lit's `sessRegMu` mutex to ensure only one thread is trying to create a session at a time 
          2) call the session store's `db.GetUnusedIDAndKeyPair` method to find an unused alias & private key pair. 
          3) call the `NewSession` function to construct a new `Sessions` object (note: this does not touch the DB yet) .
          4) Finally, call the store's `db.CreateSession` method to actually go persist the new store.
          5) unlock `sessRegMu`. 
          6) (For linked sessions: we also use the `CheckSessionGroupPredicate` to ensure any linked sessions are inactive.) 

- How the above flow came about: For Autopilot sessions, we wanted to make sure we could successfully register a session with `Autopilot` before actually going and persisting it to our DB. But at the time of registration, we already want to know the ID/Local pub key pair we will use for the session. This is why we would do all this in-memory only work before actually persisting the session. This is also why we needed to lock the mutex to ensure only one thread calls `db.GetUnusedIDAndKeyPair` at a time. 
[...]

PR #967, opened 2025-02-09
[sql-9] sessions: small sql preparations

ellemouton:
This is a pure refactor that just does a bit of code-move and variable re-naming in preparation for adding `sql` variations of things. 
We also make sure to use the `session.Store` interface in the code-base instead of the raw DB pointer.

Part of https://github.com/lightninglabs/lightning-terminal/issues/966


jamaljsr (approved):
Overall, this looks great 👌. I was able to compile & run litd on regtest with no issues. I also confirmed the unit tests passed.

Regarding the code, the refactors seem pretty straight-forward. I'm still getting familiar with the codebase, so don't have any implementation specific feedback just yet.

session_rpcserver.go
```diff
@@ -58,7 +58,7 @@ type sessionRpcServer struct {
 // sessionRpcServerConfig holds the values used to configure the
 // sessionRpcServer.
 type sessionRpcServerConfig struct {
-	db                      *session.DB
+	db                      session.Store
```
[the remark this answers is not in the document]
ellemouton: so `session.Store` is an interface that `*session.DB` implements. So they are both pointers. `session.Store` will always have an underlying pointer

terminal.go
```diff
@@ -1457,6 +1457,11 @@ func (g *LightningTerminal) shutdownSubServers() error {
 		}
 	}
 
+	if err := g.sessionDB.Close(); err != nil {
```
ViktorTigerstrom, line 1460: Should also check that `g.sessionDB != nil`, incase `g.start` errors before setting the `sessionDB` field.

ViktorTigerstrom (commented):
Looks good :rocket:! Leaving a comment regarding one `nil` check that'd be needed with this refactor, but other than that, this looks good to go :fire:

terminal.go
ellemouton: def - good catch 🙏

PR #968, opened 2025-02-09
[sql-10] sessions: decouple super macaroon helpers from `sessions` package

ellemouton:
A pure refactor/code-move PR to move any helper functions and types from the `sessions`
package to the new `macaroons` package. This is to help avoid import cycles later on when 
the session package starts to import from the accounts package (for when a session links to an
account) which also makes use of these macaroon helpers/types.

Part of https://github.com/lightninglabs/lightning-terminal/issues/966


ViktorTigerstrom (approved):
uTACK LGTM 🚀:fire:!

jamaljsr (approved):
tACK LGTM 💪

Tested the build, unit tests, and running litd. All good

PR #969, opened 2025-02-09
[sql-11] sessions: interface massage 

ellemouton:
This is a tiny refactor PR that does various things to the session DB/Store
in preparation for making the interface ready for a SQL implementation. 

Part of https://github.com/lightninglabs/lightning-terminal/pull/966


session/tlv.go
```diff
@@ -237,8 +237,8 @@ func DeserializeSession(r io.Reader) (*Session, error) {
 	session.Label = string(label)
 	session.State = State(state)
 	session.Type = Type(typ)
-	session.Expiry = time.Unix(int64(expiry), 0)
-	session.CreatedAt = time.Unix(int64(createdAt), 0)
+	session.Expiry = time.Unix(int64(expiry), 0).UTC()
+	session.CreatedAt = time.Unix(int64(createdAt), 0).UTC()
```
ViktorTigerstrom, line 241: Hmmm can this somehow lead to some discrepancy between current sessions that were created without UTC, which are now deserialised with UTC?

ViktorTigerstrom (approved):
Looks good :rocket:! uTACK LGTM :fire:.

Leaving one questions regarding current sessions.

jamaljsr (approved):
tACK LGTM 🚀

The commits are very well structured. It's so easy to review the small diffs.

session/kvdb_store.go
```diff
@@ -77,13 +80,15 @@ const (
 // BoltStore is a bolt-backed persistent store.
 type BoltStore struct {
 	*bbolt.DB
+
+	clock clock.Clock
```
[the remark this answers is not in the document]
ellemouton: yeah it is mainly for deterministic tests so that we can guarantee stable (and incrementing) timestamps

session/tlv.go
ellemouton: did some testing just to make sure no issues.

ViktorTigerstrom: So I guess the worst case scenario would be that a user after upgrading has a very recently created session that's now deserialized with a timestamp that's a few hours in the future. Alternatively the same could occur for recently revoked sessions, that's seen as having been revoked in the future.

Other than that such scenarios would be pretty confusing (which I think would be fine), there's one edge case where that could lead to a recently created session being automatically revoked:
https://github.com/lightninglabs/lightning-terminal/blob/c0e85e0f243aad6360499d9b553beab4c7461f98/session_rpcserver.go#L442-L448

But given that it's such an edge case where the worst case scenario would jsut be that a session gets revoked, we deemed this to be an ok tradeoff offline.

PR #970, opened 2025-02-09
[sql-12] sessions: make ListSession methods SQL ready

ellemouton:
In this PR, we make various changes to the ListSessions method of the Store interface such that it
can better be implemented by a SQL backend. See commit messages for more details.

Part of #966

### DIFF
--- a/session/kvdb_store.go
+++ b/session/kvdb_store.go
@@ -371,6 +371,16 @@ func (db *BoltStore) ListSessions(filterFn func(s *Session) bool) ([]*Session, e
 	return db.listSessions(filterFn)
 }
 
+// ListSessionsByType returns all sessions currently known to the store that
+// have the given type.
+//
+// NOTE: this is part of the Store interface.
+func (db *BoltStore) ListSessionsByType(t Type) ([]*Session, error) {
+	return db.listSessions(func(s *Session) bool {
+		return s.Type == t
+	})
+}
+
 // listSessions returns all sessions currently known to the store that pass the
 // given filter function.
 func (db *BoltStore) listSessions(filterFn func(s *Session) bool) ([]*Session,
--- a/session/kvdb_store.go
+++ b/session/kvdb_store.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
@@ -367,6 +368,14 @@ func (db *BoltStore) GetSession(key *btcec.PublicKey) (*Session, error) {
 //
 // NOTE: this is part of the Store interface.
 func (db *BoltStore) ListSessions(filterFn func(s *Session) bool) ([]*Session, error) {
+	return db.listSessions(filterFn)
+}
+
+// listSessions returns all sessions currently known to the store that pass the
+// given filter function.
+func (db *BoltStore) listSessions(filterFn func(s *Session) bool) ([]*Session,
+	error) {
+
 	var sessions []*Session
 	err := db.View(func(tx *bbolt.Tx) error {
 		sessionBucket, err := getBucket(tx, sessionBucketKey)
@@ -398,6 +407,11 @@ func (db *BoltStore) ListSessions(filterFn func(s *Session) bool) ([]*Session, e
 	if err != nil {
 		return nil, err
 	}
+
+	// Make sure to sort the sessions by creation time.
+	sort.Slice(sessions, func(i, j int) bool {
+		return sessions[i].CreatedAt.Before(sessions[j].CreatedAt)
+	})
 
 	return sessions, nil
 }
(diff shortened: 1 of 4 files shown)


session_rpcserver.go
```diff
@@ -353,16 +350,6 @@ func (s *sessionRpcServer) resumeSession(ctx context.Context,
...
 
-	// We only start non-revoked, non-expired LiT sessions. Everything else
-	// we just skip.
-	if sess.State != session.StateInUse &&
-		sess.State != session.StateCreated {
-
-		log.Debugf("Not resuming session %x with state %d", pubKeyBytes,
-			sess.State)
-		return nil
-	}
```
jamaljsr, line 364: Is it safe to assume that the passed in session will not have an revoked/expired state? I see that all current calls to `resumeSession` ensure that the sesion is in a valid state, but I'm thinking about future code that may not have these safeguards in place.

If you think it's best to still remove this, then the function comment should be updated.

jamaljsr (approved):
tACK LGTM. I'm loving this micro PRs 🙌

I ran the TW e2e tests against a `litd` node running this branch to confirm that everything still works and it's all green. Great work 💪

I just have one questions, but otherwise approved.

session_rpcserver.go
ellemouton: cool yeah currently all the session passed to this method will already be in the Created/InUse state since it is either called from `start` which now only fetches sessions in these states or it is called on session creation which will defs have the CreatedState. 

GOod call - will update the comment 👍

session/kvdb_store.go
```diff
@@ -363,10 +364,39 @@ func (db *BoltStore) GetSession(key *btcec.PublicKey) (*Session, error) {
...
+// NOTE: this is part of the Store interface.
+func (db *BoltStore) ListSessionsByType(t Type) ([]*Session, error) {
+	return db.listSessions(func(s *Session) bool {
+		return s.Type == t
+	})
+}
+
+// listSessions returns all sessions currently known to the store that pass the
+// given filter function.
+func (db *BoltStore) listSessions(filterFn func(s *Session) bool) ([]*Session, error) {
```
ViktorTigerstrom, line 399: nit:
line length

session/store_test.go
```diff
@@ -23,11 +23,17 @@ func TestBasicSessionStore(t *testing.T) {
...
-	s4 := newSession(t, db, clock, "session 4", nil)
+	// Create a few sessions. We increment the time by one second between
+	// each session to ensure that the created at time is unique and hence
+	// that the ListSessions method returns the sessions in a deterministic
+	// order.
+	s1 := newSession(t, db, clock, "session 1")
+	clock.SetTime(testTime.Add(time.Second))
+	s2 := newSession(t, db, clock, "session 2")
+	clock.SetTime(testTime.Add(2 * time.Second))
+	s3 := newSession(t, db, clock, "session 3", withType(TypeAutopilot))
```
ViktorTigerstrom, line 34: Not sure if this really tests that the sessions are sorted by creation time, as one could argue they're just sorted by the order they were added here. So I would ensure that one of the latter sessions we add, say s3, has a creation time prior to s2.

session/kvdb_store.go
```diff
@@ -363,10 +364,39 @@ func (db *BoltStore) GetSession(key *btcec.PublicKey) (*Session, error) {
 	return session, nil
 }
 
-// ListSessions returns all sessions currently known to the store.
+// ListSessions returns all sessions currently known to the store that are in
+// the given states. If no states are provided, all sessions are returned.
 //
 // NOTE: this is part of the Store interface.
-func (db *BoltStore) ListSessions(filterFn func(s *Session) bool) ([]*Session, error) {
+func (db *BoltStore) ListSessions(states ...State) ([]*Session, error) {
```
ViktorTigerstrom, line 371: nit:

Given that we have a separate `ListSessionsByType` function, maybe it'd make sense to have a 
`ListSessionsByState` function separately that takes the `states ...State` arg, and then let the `ListSessions` itself take no args an have no filtering?
IMO it just feels a bit weird that the plain `ListSessions` func does take filtering `State(s)` args, while we have another separate function `ListSessionsByType` function.

ViktorTigerstrom (approved):
Niiiiice :rocket:! Very clean PR :fire:

Laaving a few non-blocking suggestions that I'll let you decide if you think they're worth implementing or not.

session/kvdb_store.go
ellemouton: sure yeah can do 👍

session/store_test.go
ellemouton: the main reason for incrementing the timestamps is to have deterministic results from ListSessions. We are just testing ListSessions as a whole, not that the results are ordered.

session/kvdb_store.go
ellemouton: updated as per suggestion

PR #974, opened 2025-02-11
accounts: Add credit and debit account functionality 

ViktorTigerstrom:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/648

This PR adds the actual RPC server implementation and exposes the functionality to credit or debit an account by a given amount.

This PR also introduces a depreciation warning for the `litcli accounts update --new_balance` flag.


litrpc/lit-accounts.proto
```diff
@@ -144,7 +144,10 @@ message UpdateAccountRequest {
     // The ID of the account to update. Either the ID or the label must be set.
     string id = 1;
 
-    // The new account balance to set. Set to -1 to not update the balance.
+    /*
+    Deprecated, use credit or debit command instead. The new account balance to
+    set. Set to -1 to not update the balance.
```
bitromortac, line 149: There is some deprecation functionality for proto files, which we could use in addition:
`int64 account_balance = 2 [deprecated = true];`

bitromortac (commented):
Nice work, this will be useful and safer than the current approach :+1:.

I tested a bit and found that the following works (although one is only able to credit and debit with the rpcs in the intended directions, so no sign change is possible).

No change (due to overflow):
`litcli.sh accounts credit account_id 9223372036854775808`

Credit X sat:
`litcli.sh accounts credit account_id 9223372036854775808 + X`

Debit X sat:
`litcli.sh accounts debit account_id 9223372036854775808 + X`

Maybe we want to prevent this on the RPC level?

ViktorTigerstrom:
Thanks a lot for the review @bitromortac!

> Edit: not sure it's possible to detect as that may depend on the flag library

As also discussed offline earlier: Yeah I'm not sure there's anything we could do to avoid that issue unfortunately :/. But thanks a lot for testing this and good spotting of the issue!

bitromortac (approved):
LGTM :zap:

litrpc/lit-accounts.proto
```diff
@@ -25,6 +25,18 @@ service Accounts {
     */
     rpc UpdateAccount (UpdateAccountRequest) returns (Account);
 
+    /* litcli: `accounts creditaccount`
```
ellemouton, line 28: think the call here is wrong? looks like it is `litcli accounts credit`

same for debit

litrpc/lit-accounts.proto
```diff
@@ -25,6 +25,18 @@ service Accounts {
...
+    CreditAccount increases the balance of an existing account in the account
+    database.
+    */
+    rpc CreditAccount (UpdateAccountBalanceRequest) returns (Account);
+
+    /* litcli: `accounts debitaccount`
+    DebitAccount decreases the balance of an existing account in the account
+    database.
+    */
+    rpc DebitAccount (UpdateAccountBalanceRequest) returns (Account);
```
ellemouton, line 38: I know the `UpdateAccount` method returns an `Account` only, but it is generally better practice to instead define new messages here: 
`CreditAccountRequest`, `CreditAccountResponse` etc etc

so that if how we handle the requests/responses for either of these ever changes/get added to, then it is easy to extend the individual messages rather than breaking the API

litrpc/lit-accounts.proto
```diff
@@ -148,6 +160,22 @@ message UpdateAccountRequest {
...
 
+message UpdateAccountBalanceRequest {
+    // The ID of the account to update. Either the ID or the label must be set.
+    string id = 1;
+
+    /*
+    The label of the account to update. If an account has no label, then the ID
+    must be used instead.
+    */
+    string label = 2;
```
ellemouton, line 174: should we have an AccountIdentifier type rather since we seem to need this for each account request?

litrpc/lit-accounts.yaml
```diff
@@ -15,3 +15,9 @@ http:
       get: "/v1/accounts"
     - selector: litrpc.Accounts.RemoveAccount
       delete: "/v1/accounts/{id}"
```
ellemouton, line 17: this should be done in the previous commit to keep the RPC check happy for each commit

accounts/rpcserver.go
```diff
@@ -130,6 +130,28 @@ func (s *RPCServer) UpdateAccount(ctx context.Context,
...
+// database, by the given amount.
+func (s *RPCServer) DebitAccount(ctx context.Context,
+	req *litrpc.UpdateAccountBalanceRequest) (*litrpc.Account, error) {
+
+	log.Infof("[debitaccount] id=%s, label=%v, amount=%d",
+		req.Id, req.Label, req.Amount)
+
+	return nil, fmt.Errorf("not implemented")
+}
+
```
ellemouton: this should not be needed in this commit. The RPC server embeds the UnimplementedAccountsServer

ellemouton (changes requested):

accounts/rpcserver.go
ViktorTigerstrom: Ah thanks! TIL

proto/lit-accounts.proto
```diff
@@ -148,6 +160,22 @@ message UpdateAccountRequest {
...
+    /*
+    The label of the account to update. If an account has no label, then the ID
+    must be used instead.
+    */
+    string label = 2;
+
+    /*
+    The amount by which the account's balance should be updated.
+    */
+    uint64 amount = 3 [jstype = JS_STRING];
```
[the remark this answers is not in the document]
ViktorTigerstrom: It got added automatically with the `make protos & rpc` commands.

litrpc/lit-accounts.proto
ViktorTigerstrom: Good idea, updated :)

litrpc/lit-accounts.proto
ViktorTigerstrom: Great point! Updated this to address your feedback.

ViktorTigerstrom:
Thanks for the review @ellemouton! I've addressed your feedback with the latest push :)

docs/release-notes/release-notes-0.14.2.md
```diff
@@ -32,6 +32,13 @@
   allow increasing or decreasing the balance of an off-chain account by a
   specified amount.
 
+
+* The [`accounts update` `--new_balance` flag has been
+  deprecated](https://github.com/lightninglabs/lightning-terminal/pull/974).
+  Use the `accounts credit` and `accounts debit` commands instead, to update
+  an account's balance. The flag will no longer be supported in Lightning
+  Terminal `v0.15.0-alpha`.
```
ViktorTigerstrom, line 40: Realized that we should push this to `v0.16.0`, as lnd `v0.19.0` will be `v0.15.0`. I'll update this a bit later today.

litrpc/lit-accounts.proto
```diff
@@ -186,3 +228,11 @@ message RemoveAccountRequest {
 
 message RemoveAccountResponse {
 }
+
+message AccountIdentifier {
+    // The ID of the account.
+    string id = 1;
+
+    // The label of the account.
```
ellemouton, line 239: "either the label or the id must be set" right? 

does that mean these can be a "oneof"?

accounts/rpcserver.go
```diff
@@ -130,6 +130,75 @@ func (s *RPCServer) UpdateAccount(ctx context.Context,
...
+	log.Infof("[creditaccount] id=%s, label=%v, amount=%d",
+		req.GetAccount().GetId(), req.GetAccount().GetLabel(),
+		req.Amount)
+
+	amount := lnwire.MilliSatoshi(req.Amount * 1000)
+
+	accountID, err := s.findAccount(
+		ctx, req.GetAccount().GetId(), req.GetAccount().GetLabel(),
+	)
```
ellemouton, line 151: you've already done the nil check at this point, so can just:

```
accountID, err := s.findAccount(ctx, req.Account.Id, req.Account.Label)
	if err != nil {
		return nil, err
	}
```

accounts/rpcserver.go
```diff
@@ -130,6 +130,75 @@ func (s *RPCServer) UpdateAccount(ctx context.Context,
...
+	if err != nil {
+		return nil, err
+	}
+
+	account, err := s.service.CreditAccount(
+		ctx, accountID, amount,
+	)
```
ellemouton, line 158: fits on one line

accounts/rpcserver.go
```diff
@@ -130,6 +130,75 @@ func (s *RPCServer) UpdateAccount(ctx context.Context,
...
+	if err != nil {
+		return nil, err
+	}
+
+	account, err := s.service.DebitAccount(
+		ctx, accountID, amount,
+	)
+	if err != nil {
+		return nil, err
+	}
```
ellemouton, line 195: same comments as above

cmd/litcli/accounts.go
```diff
@@ -232,6 +235,130 @@ func updateAccount(cli *cli.Context) error {
...
+
+		resp, err := client.DebitAccount(ctx, req)
+		if err != nil {
+			return err
+		}
+
+		respAccount = resp.Account
+	}
+
+	printRespJSON(respAccount)
```
ellemouton, line 358: i think just print the whole `resp`, not just `resp.Account` - that way nothing here needs to change if we add to either response

docs/release-notes/release-notes-0.14.2.md
ellemouton: still unaddressed i think 🙏

cmd/litcli/accounts.go
```diff
@@ -162,8 +162,8 @@ var updateAccountCommand = cli.Command{
 		},
 		cli.Int64Flag{
 			Name: "new_balance",
-			Usage: "The new balance of the account; -1 means do " +
-				"not update the balance.",
+			Usage: "(deprecated) The new balance of the account; " +
+				"-1 means do not update the balance.",
```
ellemouton, line 166: `Hidden: true,`

ellemouton (commented):
looking good! basically g2g

litrpc/lit-accounts.proto
ViktorTigerstrom: Good idea! Addressed in the latest push :)

cmd/litcli/accounts.go
ViktorTigerstrom: Ah cool, TIL :)

ViktorTigerstrom:
Thanks for the review @ellemouton! I've addressed your comments in the latest push :)

ellemouton (commented):
Looks good! Final suggestion I promise :) what do you think of moving the new commands to rather be sub-commands of the `update` command? that way we keep things organised. 

[litcli.patch](https://github.com/user-attachments/files/19503749/litcli.patch)

ViktorTigerstrom:
> Looks good! Final suggestion I promise :) what do you think of moving the new commands to rather be sub-commands of the `update` command? that way we keep things organised.
> 
> [litcli.patch](https://github.com/user-attachments/files/19503749/litcli.patch)

Cool thanks :)! I've updated the PR to address your feedback. Just a question of how we actually will implement the deprecation, as this will be my first time doing so:
In `v16.0`, will we just remote the `account_balance` field in the `UpdateAccountRequest` proto, and then also remove the `new_balance` argument in the `litcli update` command?

ellemouton (approved):
LGTM! Great work 🙏 

> Just a question of how we actually will implement the deprecation, as this will be my first time doing so:
In v16.0, will we just remote the account_balance field in the UpdateAccountRequest proto, and then also remove the new_balance argument in the litcli update command?

good question  - tbh I think it's sort of up to us here cause: in LND, i think we've only really ever marked things as deprecated but then never actually followed through with removing them. So I think it's up to us and it depends on how urgently we want it to be deprecated and also what we expect the usage of it to be.
For this, perhaps yes a 1 release cycle heads up is good enough. We can perhaps add an explicit `Deprecation notice` section in the release notes to make things way more clear. If users complain at some point that the 1 release cycle heads up was too short, then we can either revert the change temporarily or just take note of the feedback for future deprecations.

PR #976, opened 2025-02-12
multi: be consistent with UTC conversion of timestamp

ellemouton:
Since we convert to UTC timestamps at persistence read/write time, we gotta make sure to compare against a consistent `Local` when comparing against `time.Now()` in the code.

### DIFF
--- a/rules/rate_limit.go
+++ b/rules/rate_limit.go
@@ -134,7 +134,7 @@ func (r *RateLimitEnforcer) HandleRequest(ctx context.Context, uri string,
 	// Determine the start time of the actions window.
 	startTime := time.Now().Add(
 		-time.Duration(rateLim.NumHours) * time.Hour,
-	)
+	).UTC()
 
 	// Now count all relevant actions which have taken place after the
 	// start time.
--- a/session_rpcserver.go
+++ b/session_rpcserver.go
@@ -132,7 +132,7 @@ func (s *sessionRpcServer) start(ctx context.Context) error {
 				continue
 			}
 
-			if sess.Expiry.Before(time.Now()) {
+			if sess.Expiry.Before(time.Now().UTC()) {
 				continue
 			}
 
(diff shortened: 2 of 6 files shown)


session_rpcserver.go
```diff
@@ -190,7 +190,7 @@ func (s *sessionRpcServer) AddSession(ctx context.Context,
 	req *litrpc.AddSessionRequest) (*litrpc.AddSessionResponse, error) {
 
 	expiry := time.Unix(int64(req.ExpiryTimestampSeconds), 0)
```
ViktorTigerstrom, line 192: I think this one should also be:
 
```suggestion
	expiry := time.Unix(int64(req.ExpiryTimestampSeconds), 0).UTC()
```

Or we need to update protos to declare that the timestamp should explicitly be sent in through the request in UTC?

session_rpcserver.go
```diff
@@ -839,7 +839,7 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
 	}
 
 	expiry := time.Unix(int64(req.ExpiryTimestampSeconds), 0)
```
ViktorTigerstrom, line 841: Same here

rules/rate_limit.go
```diff
@@ -134,7 +134,7 @@ func (r *RateLimitEnforcer) HandleRequest(ctx context.Context, uri string,
 	// Determine the start time of the actions window.
 	startTime := time.Now().Add(
 		-time.Duration(rateLim.NumHours) * time.Hour,
-	)
+	).UTC()
```
ViktorTigerstrom, line 137: This should be:
```
	startTime := time.Now().UTC().Add(
		-time.Duration(rateLim.NumHours) * time.Hour,
	)
```

Else there can be an discrepancy when the local timezone conversion happens right at a boundary of `DST`.

ViktorTigerstrom (commented):
Leaving a few small comments.

ellemouton:
thank goodness we dont need this

PR #978, opened 2025-02-13
itest: ensure peer is connected before channel open

ellemouton:
This fixes a flake that has been occurring in our itests more and more frequently where we try
to open a second channel with one of our channel peers immediately after a restart. But we might
not yet have connected with them. So we should wait for the reconnect before we attempt to open 
the channel.


ViktorTigerstrom (approved):
Thanks!

jamaljsr (approved):
LGTM 🙏

Issue #979, opened 2025-02-17
[sqli-ze] sessions: atomic session creation


ellemouton:
- Today, a session is created as follows: 
          1) Lock Lit's `sessRegMu` mutex to ensure only one thread is trying to create a session at a time 
          2) call the session store's `db.GetUnusedIDAndKeyPair` method to find an unused alias & private key pair. 
          3) call the `NewSession` function to construct a new `Sessions` object (note: this does not touch the DB yet) .
          4) Finally, call the store's `db.CreateSession` method to actually go persist the new store.
          5) unlock `sessRegMu`. 
          6) (For linked sessions: we also use the `CheckSessionGroupPredicate` to ensure any linked sessions are inactive.) 

- How the above flow came about: For Autopilot sessions, we wanted to make sure we could successfully register a session with `Autopilot` before actually going and persisting it to our DB. But at the time of registration, we already want to know the ID/Local pub key pair we will use for the session. This is why we would do all this in-memory only work before actually persisting the session. This is also why we needed to lock the mutex to ensure only one thread calls `db.GetUnusedIDAndKeyPair` at a time. 
- For a clean interface though, all the above should really happen atomically under a single database read transaction & we should not need to expose things like `db.GetUnusedIDAndKeyPair` and `db.CheckSessionGroupPredicate`. 
- What we will instead do is:
          - Add a new `session.State`: `SessionReserved`, which will be the new _first_ state of a session. 
          - Then, we will change `NewSession` to instead be a method on the `Store` interface. This will
                    - take care of finding a unique ID/Local key pair (so we can remove that from the interface).
                    - insert the new session but under the `Reserved` state.
                    - if a group is linked, here is where we will check that all other sessions are revoked. So `CheckSessionGroupPredicate` can also be removed from the interface. 
           - `CreateSession` will then _only_ move a session from `Reserved` to `Created`. 
           - On DB startup, we make sure to delete all sessions in the `Reserved` state.

ellemouton:
@ViktorTigerstrom  & @bitromortac  for concept ACK here 🙏

ViktorTigerstrom:
concept ACK :rocket:

PR #980, opened 2025-02-17
[sql-14] sessions: atomic session creation

ellemouton:
See https://github.com/lightninglabs/lightning-terminal/issues/979 for more context. 

In this PR, we do the refactor to make session creation happen in 1 atomic DB transaction as described in the issue above. 

Fixes https://github.com/lightninglabs/lightning-terminal/issues/979

### DIFF
--- a/session/interface.go
+++ b/session/interface.go
@@ -216,12 +216,6 @@ type Store interface {
 	// GetSessionByID fetches the session with the given ID.
 	GetSessionByID(id ID) (*Session, error)
 
-	// CheckSessionGroupPredicate iterates over all the sessions in a group
-	// and checks if each one passes the given predicate function. True is
-	// returned if each session passes.
-	CheckSessionGroupPredicate(groupID ID,
-		fn func(s *Session) bool) (bool, error)
-
 	// DeleteReservedSessions deletes all sessions that are in the
 	// StateReserved state.
 	DeleteReservedSessions() error
--- a/session/interface.go
+++ b/session/interface.go
@@ -27,15 +27,15 @@ const (
 type State uint8
 
 /*
-		/---> StateExpired (terminal)
-StateCreated ---
-       		\---> StateRevoked (terminal)
+				   /---> StateExpired (terminal)
+StateReserved ---> StateCreated ---
+       				   \---> StateRevoked (terminal)
 */
 
 const (
 	// StateCreated is the state of a session once it has been fully
-	// committed to the Store and is ready to be used. This is the first
-	// state of a session.
+	// committed to the BoltStore and is ready to be used. This is the
+	// first state after StateReserved.
 	StateCreated State = 0
 
 	// StateInUse is the state of a session that is currently being used.
@@ -52,10 +52,10 @@ const (
 	// date.
 	StateExpired State = 3
 
-	// StateReserved is a temporary initial state of a session. On start-up,
-	// any sessions in this state should be cleaned up.
-	//
-	// NOTE: this isn't used yet.
+	// StateReserved is a temporary initial state of a session. This is used
+	// to reserve a unique ID and private key pair for a session before it
+	// is fully created. On start-up, any sessions in this state should be
+	// cleaned up.
 	StateReserved State = 4
 )
 
@@ -67,6 +67,9 @@ func (s State) Terminal() bool {
 // legalStateShifts is a map that defines the legal State transitions that a
 // Session can be put through.
 var legalStateShifts = map[State]map[State]bool{
+	StateReserved: {
+		StateCreated: true,
+	},
 	StateCreated: {
 		StateExpired: true,
 		StateRevoked: true,
@@ -141,7 +144,7 @@ func buildSession(id ID, localPrivKey *btcec.PrivateKey, label string, typ Type,
 	sess := &Session{
 		ID:                id,
 		Label:             label,
-		State:             StateCreated,
+		State:             StateReserved,
 		Type:              typ,
 		Expiry:            expiry.UTC(),
 		CreatedAt:         created.UTC(),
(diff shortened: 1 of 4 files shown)


session_rpcserver.go
```diff
@@ -313,24 +308,17 @@ func (s *sessionRpcServer) AddSession(ctx context.Context,
...
+		req.Label, typ, expiry, req.MailboxServerAddr,
 		req.DevServer, uniquePermissions, caveats, nil, false, nil,
 		session.PrivacyFlags{},
 	)
 	if err != nil {
 		return nil, fmt.Errorf("error creating new session: %v", err)
 	}
 
-	if err := s.cfg.db.CreateSession(sess); err != nil {
+	sess, err = s.cfg.db.CreateSession(sess.ID)
```
ellemouton, line 320: note: in [this](https://github.com/lightninglabs/lightning-terminal/pull/984) follow up, we add a way to do this in one go for normal (non-autopilot) sessions

session_rpcserver.go
```diff
@@ -1224,10 +1187,15 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
...
-	if err := s.cfg.db.CreateSession(sess); err != nil {
+	err = s.cfg.db.UpdateSessionRemotePubKey(sess.LocalPublicKey, remoteKey)
+	if err != nil {
+		return nil, fmt.Errorf("error setting remote pubkey: %v", err)
+	}
+
+	// We only activate the session if the Autopilot server registration
+	// was successful.
+	sess, err = s.cfg.db.CreateSession(sess.ID)
+	if err != nil {
```
ViktorTigerstrom, line 1198: There's nothing to change here, but just a reminder, as we discussed in the original session-linking PR, that we should allow session revocation based on `groupIDs` at the autopilot server level. 🤓 Otherwise, this could lead to a scenario where the autopilot server persists the reserved session, but `litd` removes it on the next startup if an error occurs here. That'd become problematic for linked sessions.

ViktorTigerstrom (commented):
Generally looks good 🚀! Leaving one comment regarding state enforcement during session revoking that I believe we should fix :)

session_rpcserver.go
ellemouton: ah yes, that's a great point

session/interface.go
```diff
@@ -27,15 +27,15 @@ const (
 type State uint8
 
 /*
-		/---> StateExpired
-StateCreated ---
-       		\---> StateRevoked
+				   /---> StateExpired
+StateReserved ---> StateCreated ---
+       				   \---> StateRevoked
```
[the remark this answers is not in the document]
ellemouton: what sort of errors would that create? 

but yes, can add a tight-map constraint (above DB layer as it isnt db related). But i think we can make it so that it's fine to always go from any state to "revoked"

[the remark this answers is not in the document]
ViktorTigerstrom: So I think mainly there's 2 errors that would occur that I can think of right now, but there might be more:

1. The issue described in the comment below (https://github.com/lightninglabs/lightning-terminal/pull/980#discussion_r1965315040) would occur, because calling `s.cfg.db.CreateSession(sess.ID)` after registering the session with the autopilot server results in an error in the `AddAutopilotSession` call, since the session is not in the `Reserved` state.

2. The actual `s.cfg.autopilot.SessionRevoked(ctx, pubKey)` call in `RevokeSession` would not actually revoke the session on the autopilot server, as the session has not yet been registered on the server. I don't know this could lead to unintended consequences that I can't think of right now.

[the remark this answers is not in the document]
ellemouton: have added the strict state shift stuff here: https://github.com/lightninglabs/lightning-terminal/pull/985

this PR now builds on top of that one

session_rpcserver.go
ellemouton: opened an issue on the appropriate repo - thanks for the flag 🙏

session/interface.go
```diff
@@ -67,6 +68,10 @@ func (s State) Terminal() bool {
 // legalStateShifts is a map that defines the legal State transitions that a
 // Session can be put through.
 var legalStateShifts = map[State]map[State]bool{
+	StateReserved: {
+		StateCreated: true,
+		StateRevoked: true,
```
ViktorTigerstrom, line 73: Hmmm wait, I thought we were not supposed to allow a session to move directly from `StateReserved` to `StateRevoked`, as per the comment above https://github.com/lightninglabs/lightning-terminal/pull/980#discussion_r1965308238?

Imo I see no need to allow that at all, since reversed sessions get wiped on next startup of `litd`, and due to the issues described in the linked comment thread? If you change this, this should also be reflected in the `State` "flow illustration" above.

ViktorTigerstrom (commented):
Looks good :rocket:! Still need to clarify regarding the same comment I left above though, but other than that, this looks good to go from my end :fire:

session/interface.go
ellemouton: is there a down side to just allowing this state transition? not sure it is a bad thing to allow a path to terminal state.

> The actual s.cfg.autopilot.SessionRevoked(ctx, pubKey) call in RevokeSession would not actually revoke the session on the autopilot server, as the session has not yet been registered on the server. It'll then actually get added to server, when the call to register the session to the autopilot server gets executed, after it has already been revoked on the client.

this i dont understand. Why would it get registered?

ellemouton: autopilot.RevokeSession will just do nothing right?

ellemouton: cool discussed offline. updating!

ViktorTigerstrom: Thanks for the updates :rocket:!

ViktorTigerstrom (approved):
Niiiiiiiiiiiiice! uTACK LGTM :fire:!

bitromortac (approved):
LGTM, nice clean-up :zap:

session/kvdb_store.go
```diff
@@ -250,7 +250,8 @@ func (db *BoltStore) NewSession(label string, typ Type, expiry time.Time,
 
 				// Ensure that the session is no longer active.
 				if sess.State == StateCreated ||
-					sess.State == StateInUse {
+					sess.State == StateInUse ||
+					sess.State == StateReserved {
```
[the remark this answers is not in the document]
ellemouton: ah yes, i forgot about the existing IsTerminal() method!

PR #981, opened 2025-02-18
accounts: Credit and debit accounts `kvdb` additions prep

ViktorTigerstrom:
Part of https://github.com/lightninglabs/lightning-terminal/issues/648

This PR adds DB preparation required for the KVDB store, in order to support functionality which allows crediting or debiting an account's balance by a given amount.

### DIFF
--- a/accounts/store_kvdb.go
+++ b/accounts/store_kvdb.go
@@ -223,11 +223,11 @@ func (s *BoltStore) AddAccountInvoice(_ context.Context, id AccountID,
 	return s.updateAccount(id, update)
 }
 
-// IncreaseAccountBalance increases the balance of the account with the given ID
+// CreditAccount increases the balance of the account with the given ID
 // by the given amount.
 //
 // NOTE: This is part of the Store interface.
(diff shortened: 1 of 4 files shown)


accounts/store_kvdb.go
```diff
@@ -244,6 +244,33 @@ func (s *BoltStore) CreditAccount(_ context.Context, id AccountID,
...
+//
+// NOTE: This is part of the Store interface.
+func (s *BoltStore) DebitAccount(_ context.Context, id AccountID,
+	amount lnwire.MilliSatoshi) error {
+
+	update := func(account *OffChainBalanceAccount) error {
+		if amount > math.MaxInt64 {
+			return fmt.Errorf("amount %v exceeds the maximum of %v",
+				amount, int64(math.MaxInt64))
+		}
```
ellemouton, line 258: think we can move this check out of the db transaction since it does not need access to the account

ellemouton (approved):
Very nice 🙏

accounts/store_kvdb.go
ViktorTigerstrom: Thanks :pray:! Addressed with the latest push.

PR #985, opened 2025-02-25
sessions: strict state shifts

ellemouton:
In this PR, we ensure all session state shifts are legal via a strict transition map. We then replace the
RevokeSession method with a new ShiftState method. Finally, for expired sessions, we use the Expired state 
instead of the Revoked state.


docs/release-notes/release-notes-0.14.1.md
```diff
@@ -28,6 +28,11 @@
   helpers](https://github.com/lightninglabs/lightning-terminal/pull/928) that 
   allow for more control over running unit tests. 
  
+* Correctly [move a session to the Expired
```
ViktorTigerstrom, line 31: v0.14.1 is already released :)! So this should be release-docs for v0.14.2.
You could copy the docs from:
https://github.com/lightninglabs/lightning-terminal/pull/974/
& specifically
https://github.com/lightninglabs/lightning-terminal/pull/974/files#diff-a414f99dddb12f8b527d87981dbb216b60f168b2c7bd360bd9bc876fa1e393b8 

If you'd prefer. I'll update that PR later today.

session/store_test.go
```diff
@@ -392,6 +396,49 @@ func TestCheckSessionGroupPredicate(t *testing.T) {
...
+	// "RevokedAt" time should be set.
+	s1, err = db.GetSession(s1.LocalPublicKey)
+	require.NoError(t, err)
+	require.Equal(t, StateRevoked, s1.State)
+	require.True(t, clock.Now().Equal(s1.RevokedAt))
+
+	// Trying to do the same state shift again should succeed since the
+	// session is already in the expected "dest" state.
+	err = db.ShiftState(s1.ID, StateRevoked)
+	require.NoError(t, err)
```
ViktorTigerstrom, line 434: Nit: maybe test that the `RevokedAt` time hasn't changed with this second `ShiftState` call.

session/store_test.go
```diff
@@ -392,6 +396,49 @@ func TestCheckSessionGroupPredicate(t *testing.T) {
...
+
+	// Trying to shift the state from a terminal state back to StateCreated
+	// should also fail since this is not a legal state transition.
+	err = db.ShiftState(s1.ID, StateCreated)
+	require.ErrorContains(t, err, "illegal session state transition")
```
ViktorTigerstrom, line 439: nit:
Would also be value in testing that shifting a `StateReserved` session to `StateRevoked` is not legal, as that's the concern in the comment that this PR is addressing :). Potentially also add coverage to show that `StateInUse` -> `StateRevoked` is legal.

ViktorTigerstrom (approved):
Niiiiiiiice, thanks for this :pray:!!

uTACK LGTM :rocket:!

Leaving some docs and test coverage comments.

session/store_test.go
ellemouton: sure

bitromortac (approved):
Nice, LGTM :tada:

PR #986, opened 2025-02-25
[sql-16] sessions: update Store interface methods to take a context

ellemouton:
In preparation for adding a SQL implementation of the Store interface, which will make calls that require a context, we update the store interface such that all methods take a context. 

Many commits but all just passing context through. So review load is small.

### DIFF
--- a/session/server.go
+++ b/session/server.go
@@ -1,6 +1,7 @@
 package session
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"sync"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/lightninglabs/lightning-node-connect/mailbox"
+	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/keychain"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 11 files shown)


ViktorTigerstrom (approved):
Cool :fire:!

session/server.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/lightninglabs/lightning-node-connect/mailbox"
+	"github.com/lightninglabs/taproot-assets/fn"
```
bitromortac, line 12: I'm not sure if we want to use this package or the lnd one?

session/server.go
```diff
@@ -43,10 +47,13 @@ func (m *mailboxSession) start(session *Session,
 
 	ecdh := &keychain.PrivKeyECDH{PrivKey: session.LocalPrivateKey}
 
+	ctx, cancel := context.WithCancel(context.Background())
```
bitromortac, line 50: why do we not derive this from a parent context?

bitromortac (approved):
LGTM :zap: just two questions

session/server.go
ellemouton: because this gets spun off into a go-routine. So we want it to be long lived passed the point of the initial call

session/server.go
ellemouton: yeah can switch the the lnd one 👍

session/server.go
ellemouton: we can eventually change things such that `start` is only ever called from a single event loop in which case then we can pass a long-lived context to it from it. 

but since this is a bit of a re-design, we can do it as a follow up

PR #988, opened 2025-02-27
[sql-17] sessions: test preparation 

ellemouton:
In preparation for having the unit tests pass against a different Store
implementation, we standardise some of the errors that get returned.

In preparation for adding helpers with the same names but that will
compile under different build flags, we add the helper DB constructors
to use when testing the session store logic against a KVDB backend.

Part of #917


ViktorTigerstrom (approved):
LGTM :fire:

bitromortac (approved):
LGTM :rocket:

PR #989, opened 2025-02-27
[sql-18] sessions: tightly couple sessions & accounts

ellemouton:
Depends on https://github.com/lightninglabs/lightning-terminal/pull/988

In this commit, we more tightly & explicitly link a session to an
account. At a persitance layer, we have always only linked a session to
an account by encoding the AccountID within the macaroon caveat that we
store with the session. We still keep this persistence the same but now
we first ensure that the account exists and we also add an AccountID
field to the Session struct.

This is in preparation for adding a SQL implementation of the session Store 
where this link will be done at a relational DB level.

### DIFF
--- a/accounts/interceptor.go
+++ b/accounts/interceptor.go
@@ -1,12 +1,15 @@
 package accounts
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"strings"
 
 	mid "github.com/lightninglabs/lightning-terminal/rpcmiddleware"
+	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/macaroons"
 	"google.golang.org/protobuf/proto"
@@ -21,6 +24,15 @@ const (
 	// accountMiddlewareName is the name that is used for the account system
 	// when registering it to lnd as an RPC middleware.
 	accountMiddlewareName = "lit-account"
+)
+
+var (
+	// caveatPrefix is the prefix that is used for custom caveats that are
+	// used by the account system. This prefix is used to identify the
+	// custom caveat and extract the condition (the AccountID) from it.
+	caveatPrefix = []byte(fmt.Sprintf(
+		"%s %s ", macaroons.CondLndCustom, CondAccount,
+	))
 )
 
 // Name returns the name of the interceptor.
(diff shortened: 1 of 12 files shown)


session/kvdb_store.go
```diff
@@ -211,6 +218,16 @@ func (db *BoltStore) NewSession(ctx context.Context, label string, typ Type,
 
 		sessionKey := getSessionKey(session)
 
+		// If an account is being linked, we first need to check that
+		// it exists.
+		session.AccountID.WhenSome(func(account accounts.AccountID) {
+			session.AccountID = fn.Some(account)
+			_, err = db.accounts.Account(ctx, account)
+		})
```
ViktorTigerstrom, line 226: Hmm, even though we've added a check to ensure that an account exists when adding a session, there's no guarantee that the account referenced by `session.AccountID` still exists at any given time.

This can happen in two cases:  
1. Sessions created before this change.  
2. The account was deleted from the accounts database using the accounts admin CLI after the session was added.  

I'm not entirely sure if this poses an issue for future PRs, but it's something we should keep in mind. To minimize potential problems, should we also add this check in `GetSession`? While that wouldn't provide a complete guarantee (since race conditions could still occur), it would significantly reduce the likelihood of this issue.

Additionally, we might want to remove sessions associated with a deleted `AccountID` when an account is removed via the accounts admin CLI.

Let me know if this is actually a concern and whether we should take any action on it!

session/test_kvdb.go
```diff
@@ -17,7 +18,24 @@ func NewTestDB(t *testing.T, clock clock.Clock) *BoltStore {
...
+
+	return store
+}
+
+// NewTestDBWithAccounts creates a new test session Store with access to an
+// existing accounts DB.
+func NewTestDBWithAccounts(t *testing.T, clock clock.Clock,
+	acctStore accounts.Store) *BoltStore {
+
+	store, err := NewDB(t.TempDir(), DBFilename, clock, acctStore)
```
ViktorTigerstrom, line 38: Nit:

Could potentially create a helper function that both of these use instead of both functions repeating the:
```
	store, err := NewDB(t.TempDir(), DBFilename, clock, acctStore)
	require.NoError(t, err)

	t.Cleanup(func() {
		require.NoError(t, store.DB.Close())
	})

	return store
```
part with slightly different args.

accounts/interceptor.go
```diff
@@ -199,22 +211,58 @@ func parseRPCMessage(msg *lnrpc.RPCMessage) (proto.Message, error) {
...
+	accountID.WhenSome(func(aID AccountID) {
+		id = &aID
+	})
+
+	return id, nil
+}
+
+// IDFromCaveats attempts to extract an AccountID from the given set of caveats
+// by looking for the custom caveat that binds a macaroon to a certain account.
+func IDFromCaveats(caveats []macaroon.Caveat) (fn.Option[AccountID], error) {
```
ViktorTigerstrom, line 234: Would be good with test coverage of this function if possible.

ViktorTigerstrom (commented):
Generally looks great 🚀 !

I'd like to get some clarifications on the first commit thought, before ACKing :).

The last to comments are optional and non-blocking.

accounts/interceptor.go
ellemouton: see commit message

session/kvdb_store.go
ellemouton: yep, all true but I think its good to keep in mind what the main goal is here: to completely move over to SQL backed DB where this relationship will be handled for us. In that world, if we try to delete an account but a session still exists, then we can let it do a cascading delete etc (either way, it will be handled for us). 

to do that same logic right now is I think not worth it as it would involve iterating through every single session and decoding it just to see if the account is in fact linked to a session. 

The only think, imo, we need to keep this in mind for is when we are writing the kvdb -> SQL migration. At that point in time we have to account for the fact that we may encounter sessions that link to accounts that no longer exist. In that case, we dont migrate that session - we just delete it. 

The reason for adding this check to the KVDB logic at all in the first place is just for unit test compatibility for when we add the SQL backed store where we do want things to error if we try to make a session that is linked to an account that does not exist

accounts/interceptor.go
ellemouton: what i mean in the commit message is that it is currently tested via itests since this is just extracting what is already done in the above function  (which now calls this). 

but sure, can add a unit tests too i guess

session/test_kvdb.go
ellemouton: sure

accounts/interceptor.go
ViktorTigerstrom: Awesome thanks 🎉

session/kvdb_store.go
ViktorTigerstrom: Ok thanks for clarifying the intent. Given that, I agree that we can ignore this 🙂

ViktorTigerstrom (approved):
uTACK LGTM 🚀🎉

session/store_test.go
```diff
@@ -371,13 +423,36 @@ func reserveSession(db Store, label string,
...
-		context.Background(), label, opts.sessType,
-		time.Date(99999, 1, 1, 0, 0, 0, 0, time.UTC),
-		"foo.bar.baz:1234",
+	options := []Option{
 		WithDevServer(),
 		WithPrivacy(PrivacyFlags{ClearPubkeys}),
 		WithLinkedGroupID(opts.groupID),
+	}
+
+	var caveats []macaroon.Caveat
```
bitromortac, line 432: nit: can be directly created via literal in inner scope

session_rpcserver.go
```diff
@@ -1518,6 +1524,7 @@ func (s *sessionRpcServer) marshalRPCSession(sess *session.Session) (
 		GroupId:                sess.GroupID[:],
 		FeatureConfigs:         clientConfig,
 		PrivacyFlags:           sess.PrivacyFlags.Serialize(),
+		AccountId:              accountID,
```
bitromortac, line 1527: :+1: nice catch

bitromortac (approved):
LGTM :tada:, have only a few nits

PR #993, opened 2025-03-04
[sql-20] sessions: more re-shaping

ellemouton:
Ok actual actual last moulding PR before plugging in the SQL backend :) 

See commit messages for more details.

PR #994, opened 2025-03-04
[sql-21] sessions: SQL Schemas & CRUD

ellemouton:
Depends on https://github.com/lightninglabs/lightning-terminal/pull/993


bitromortac (approved):
LGTM :tada: nice work!

db/sqlc/migrations/000002_sessions.up.sql
```diff
@@ -0,0 +1,128 @@
...
+    privacy BOOLEAN NOT NULL,
+
+    -- An optional account ID that this session is linked to.
+    account_id BIGINT REFERENCES accounts(id) ON DELETE CASCADE,
+
+    -- The session ID of the first session in this linked session group. This
+    -- is nullable for the case where the first session in the group is being
+    -- inserted, and so we first need to insert the session before we know the
+    -- ID to use for the group ID.
+    group_id BIGINT REFERENCES sessions(id) ON DELETE CASCADE
```
ViktorTigerstrom, line 66: Should we create an index for this as there's a queries using this as the identifier?

db/sqlc/migrations/000002_sessions.up.sql
```diff
@@ -0,0 +1,128 @@
...
+);
+
+CREATE INDEX IF NOT EXISTS sessions_type_idx ON sessions(type);
+
+-- The session_macaroon_permissions table contains the macaroon permissions
+-- that are associated with a session.
+CREATE TABLE IF NOT EXISTS session_macaroon_permissions (
```
ViktorTigerstrom, line 73: Hmm I think it makes sense to include an `id` field for this table, `session_feature_configs` & `session_privacy_flags` as it'll likely be useful in the future?

ViktorTigerstrom (approved):
LGTM :rocket:! Leaving a few suggestions that I think will be useful.

db/sqlc/migrations/000002_sessions.up.sql
ellemouton: ah, interesting - I thought foreign keys where automatically indexed. Turns out that is incorrect - so yes, will add 👍

db/sqlc/migrations/000002_sessions.up.sql
```diff
@@ -0,0 +1,128 @@
...
+CREATE TABLE IF NOT EXISTS session_feature_configs (
+    -- The ID of the session in the sessions table that this feature config is
+    -- associated with.
+    session_id BIGINT NOT NULL REFERENCES sessions(id) ON DELETE CASCADE,
+
+    -- The feature name.
+    feature_name TEXT NOT NULL,
+
+    -- The feature config blob.
+    config BLOB
```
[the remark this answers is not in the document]
ellemouton: yeah def 🙏

db/sqlc/migrations/000002_sessions.up.sql
ellemouton: ok so i think it makes sense to add the `id` for mac perms and caveats since those are lists and not necessarily unique but think for feature configs and priv flags, it does not make sense to add `id` since those will always have a unique <foreign-key:one-other-field> pair. 

sound good?

db/sqlc/migrations/000002_sessions.up.sql
ellemouton: so yeah great catch - i totally had the wrong assumption - It also means i need to add indices for the tables below on any foreign key

ellemouton (commented):
@ViktorTigerstrom  - could you take one last look at the last push (second last `compare`) 🙏

db/sqlc/migrations/000002_sessions.up.sql
```diff
@@ -0,0 +1,144 @@
...
+    caveat_id BLOB NOT NULL,
+
+    -- The verification ID. If this is not-null, it's a third party caveat.
+    verification_id BLOB,
+
+    -- The location hint for third party caveats.
+    location TEXT
+);
+
+CREATE INDEX IF NOT EXISTS sessions_mac_caveats_idx ON session_macaroon_caveats(session_id);
```
bitromortac, line 111: this is missed in `down.sql`

ellemouton: thanks 🙏  adding

ViktorTigerstrom (approved):
LGTM after addressing @bitromortac  last comment :) :rocket:!

PR #995, opened 2025-03-05
Custom channels itest: Use group keys on some payments & invoices

GeorgeTsagk:
## Description

We sprinkle around some group keys across the grouped asset custom channels test. Some of the invoices and payments in that test now use a group key instead of a specific asset ID.

Related tapd PR: https://github.com/lightninglabs/taproot-assets/pull/1423/commits

For the liquidity edge cases itest, we now support running it in assetID & groupKey mode, meaning we can go through the full suite of edge cases by using either a single assetID or a groupKey.

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -741,6 +741,12 @@ func sendAssetKeySendPayment(t *testing.T, src, dst *HarnessNode, amt uint64,
 		opt(cfg)
 	}
 
+	// Nullify assetID if group key is set. RPC methods won't accept both so
+	// let's prioritize the group key if set.
+	if len(cfg.groupKey) > 0 {
+		assetID = nil
+	}
+
 	ctxb := context.Background()
 	ctxt, cancel := context.WithTimeout(ctxb, defaultTimeout)
 	defer cancel()
@@ -770,6 +776,7 @@ func sendAssetKeySendPayment(t *testing.T, src, dst *HarnessNode, amt uint64,
 	stream, err := srcTapd.SendPayment(ctxt, &tchrpc.SendPaymentRequest{
 		AssetId:        assetID,
 		AssetAmount:    amt,
+		GroupKey:       cfg.groupKey,
 		PaymentRequest: sendReq,
 	})
 	require.NoError(t, err)
@@ -942,6 +949,7 @@ type payConfig struct {
 	payStatus         lnrpc.Payment_PaymentStatus
 	failureReason     lnrpc.PaymentFailureReason
 	rfq               fn.Option[rfqmsg.ID]
+	groupKey          []byte
 }
 
 func defaultPayConfig() *payConfig {
@@ -955,6 +963,12 @@ func defaultPayConfig() *payConfig {
 }
 
 type payOpt func(*payConfig)
+
+func withGroupKey(groupKey []byte) payOpt {
+	return func(c *payConfig) {
+		c.groupKey = groupKey
+	}
+}
 
 func withSmallShards() payOpt {
 	return func(c *payConfig) {
@@ -1010,6 +1024,12 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
 		opt(cfg)
 	}
 
+	// Nullify assetID if group key is set. RPC methods won't accept both so
+	// let's prioritize the group key if set.
+	if len(cfg.groupKey) > 0 {
+		assetID = []byte{}
+	}
+
 	ctxb := context.Background()
 	ctxt, cancel := context.WithTimeout(ctxb, defaultTimeout)
 	defer cancel()
@@ -1041,6 +1061,7 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
 	stream, err := payerTapd.SendPayment(ctxt, &tchrpc.SendPaymentRequest{
 		AssetId:        assetID,
 		PeerPubkey:     rfqPeer.PubKey[:],
+		GroupKey:       cfg.groupKey,
 		PaymentRequest: sendReq,
 		RfqId:          rfqBytes,
 		AllowOverpay:   cfg.allowOverpay,
@@ -1102,6 +1123,7 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
 
 type invoiceConfig struct {
 	errSubStr string
+	groupKey  []byte
 }
 
 func defaultInvoiceConfig() *invoiceConfig {
@@ -1118,6 +1140,12 @@ func withInvoiceErrSubStr(errSubStr string) invoiceOpt {
 	}
 }
 
+func withInvGroupKey(groupKey []byte) invoiceOpt {
+	return func(c *invoiceConfig) {
+		c.groupKey = groupKey
+	}
+}
+
 func createAssetInvoice(t *testing.T, dstRfqPeer, dst *HarnessNode,
 	assetAmount uint64, assetID []byte,
 	opts ...invoiceOpt) *lnrpc.AddInvoiceResponse {
@@ -1125,6 +1153,12 @@ func createAssetInvoice(t *testing.T, dstRfqPeer, dst *HarnessNode,
 	cfg := defaultInvoiceConfig()
 	for _, opt := range opts {
 		opt(cfg)
+	}
+
+	// Nullify assetID if group key is set. RPC methods won't accept both so
+	// let's prioritize the group key if set.
+	if len(cfg.groupKey) > 0 {
+		assetID = []byte{}
 	}
 
 	ctxb := context.Background()
@@ -1141,6 +1175,7 @@ func createAssetInvoice(t *testing.T, dstRfqPeer, dst *HarnessNode,
 
 	resp, err := dstTapd.AddInvoice(ctxt, &tchrpc.AddInvoiceRequest{
 		AssetId:     assetID,
+		GroupKey:    cfg.groupKey,
 		AssetAmount: assetAmount,
 		PeerPubkey:  dstRfqPeer.PubKey[:],
 		InvoiceRequest: &lnrpc.Invoice{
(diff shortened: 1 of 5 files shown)


itest/assets_test.go
```diff
@@ -1198,7 +1216,15 @@ func assertInvoiceHtlcAssets(t *testing.T, node *HarnessNode,
...
 
-	targetID := hex.EncodeToString(assetID)
+	var targetID string
+	switch {
+	case len(assetID) > 0:
+		targetID = hex.EncodeToString(assetID)
+
+	case len(groupID) > 0:
+		groupHash := sha256.Sum256(groupID)
+		targetID = hex.EncodeToString(groupHash[:])
```
Roasbeef, line 1226: Hmm, do we need to modify what we return as `JsonHtlc` to start to include a distinct group key field? IIRC, it's all TLV within the db/lnd, so shouldn't be a big lift there, and should be backwards compat.

Roasbeef (commented):
I think this is a good start as a sanity check for some of the new logic. I think we'll want to more or less run _each_ test w/ both asset ID _and_ group key, but that might eventually be a bit unwieldy when it comes to execution time. 

Ultimately, we _do_ expect most of the primary use cases to use group keys, so pehrhaps it's fine to lean towards coverage there?

itest/assets_test.go
guggero: Hmm, good idea. We can definitely add that to make certain things easier. Even if it's just informative to the user and useful in itests (but not explicitly expected in the RFQ logic).

guggero (commented):
Concept ACK

itest/assets_test.go
```diff
@@ -1189,7 +1207,15 @@ func assertInvoiceHtlcAssets(t *testing.T, node *HarnessNode,
...
 	t.Logf("Asset invoice: %v", toProtoJSON(t, invoice))
 
-	targetID := hex.EncodeToString(assetID)
+	var targetID string
+	switch {
+	case len(assetID) > 0:
+		targetID = hex.EncodeToString(assetID)
+
+	case len(groupID) > 0:
+		groupHash := sha256.Sum256(groupID)
```
guggero: This should be just the x-coordinate now.

itest/assets_test.go
```diff
@@ -1237,7 +1263,15 @@ func assertPaymentHtlcAssets(t *testing.T, node *HarnessNode, payHash []byte,
...
 	t.Logf("Asset payment: %v", toProtoJSON(t, payment))
 
-	targetID := hex.EncodeToString(assetID)
+	var targetID string
+	switch {
+	case len(assetID) > 0:
+		targetID = hex.EncodeToString(assetID)
+
+	case len(groupID) > 0:
+		groupHash := sha256.Sum256(groupID)
```
guggero: Same here, re x-coordinate. I wonder why this didn't cause things to fail? Or did this still point to the old commit in `tapd`?

itest/litd_custom_channels_test.go
```diff
@@ -936,22 +940,24 @@ func testCustomChannelsGroupedAsset(ctx context.Context, net *NetworkHarness,
...
 	invoiceResp := createAssetInvoice(
-		t.t, charlie, dave, daveInvoiceAssetAmount, assetID,
+		t.t, charlie, dave, daveInvoiceAssetAmount, nil,
+		withInvGroupKey(groupID),
 	)
 	payInvoiceWithAssets(
-		t.t, charlie, dave, invoiceResp.PaymentRequest, assetID,
+		t.t, charlie, dave, invoiceResp.PaymentRequest, nil,
 		withSmallShards(),
+		withGroupKey(groupID),
```
guggero, line 950: nit: fits on previous line.

itest/assets_test.go
```diff
@@ -727,6 +726,12 @@ func sendAssetKeySendPayment(t *testing.T, src, dst *HarnessNode, amt uint64,
 		opt(cfg)
 	}
 
+	// Nullify assetID if group key is set. RPC methods won't accept both so
+	// let's prioritize the group key if set.
+	if len(cfg.groupKey) > 0 {
+		assetID = []byte{}
```
guggero, line 732: nit: set to `nil` instead? A bit more easy to parse visually IMO.

itest/assets_test.go
```diff
@@ -1209,12 +1226,11 @@ func assertInvoiceHtlcAssets(t *testing.T, node *HarnessNode,
 
 	var targetID string
 	switch {
+	case len(groupID) > 0:
```
guggero, line 1229: Ah, I see. Move this part to the previous commit?

itest/litd_custom_channels_test.go
```diff
@@ -1896,10 +1918,13 @@ func testCustomChannelsLiquidityEdgeCases(ctx context.Context,
 
 	logBalance(t.t, nodes, assetID, "10k sats")
 
+	opts = []payOpt{}
```
guggero, line 1921: nit: `var opts []payOpt` to match empty initialization style above?

itest/assets_test.go
```diff
@@ -1363,6 +1391,22 @@ func createAssetHodlInvoice(t *testing.T, dstRfqPeer, dst *HarnessNode,
 	}
 }
 
+// addGroupModeOpt may add a group key option to the opts array, if the group
+// mode boolean is true.
+func addGroupModeOpt(opts *[]payOpt, groupMode bool, groupID []byte) {
+	if groupMode {
+		*opts = append(*opts, withGroupKey(groupID))
```
guggero, line 1398: How about just a new functional option:
```
func withMaybeGroupKey(groupMode bool, groupKey []byte) payOpt {
	return func(c *payConfig) {
		if !groupMode {
			return
		}

		c.groupKey = groupKey
	}
}
```

And the same for `withInvGroupKey`? Should reduce the diff quite a bit below.

itest/litd_custom_channels_test.go
```diff
@@ -2162,16 +2268,41 @@ func testCustomChannelsLiquidityEdgeCases(ctx context.Context,
...
+}
+
+// testCustomChannelsLiquidityEdgeCasesGroup is a test that runs through some
+// taproot asset channel liquidity related edge cases using group keys.
+func testCustomChannelsLiquidityEdgeCasesGroup(ctx context.Context,
+	net *NetworkHarness, t *harnessTest) {
+
+	// Run liquidity edge cases and only use group keys for invoices and
+	// payments.
+	testCustomChannelsLiquidtyEdgeCasesCore(ctx, net, t, true)
```
guggero, line 2303: Can't we run this directly in `testCustomChannelsLiquidityEdgeCases`? Meaning to call the same sub test function there once with true once with false.
Each new test has quite a bit of overhead, so the more we can combine, the better.

guggero (commented):
Looks pretty good! Have some ideas to reduce the size of the diff though.

itest/litd_custom_channels_test.go
```diff
@@ -1975,9 +2020,11 @@ func testCustomChannelsLiquidityEdgeCases(ctx context.Context,
...
 	// amount of 354 sats.
 	createAssetInvoice(t.t, dave, charlie, 1, assetID, withInvoiceErrSubStr(
-		"cannot create invoice over 1 asset units, as the minimal "+
-			"transportable amount",
-	))
+		"cannot create invoice over 1 asset units, as the "+
+			"minimal transportable amount",
+	),
+		withMaybeInvGroupKey(groupMode, groupID),
+	)
```
guggero, line 2027: nit: should be formatted as:
```
	createAssetInvoice(
		t.t, dave, charlie, 1, assetID, withInvoiceErrSubStr(
			"cannot create invoice over 1 asset units, as the "+
				"minimal transportable amount",
		),
		withMaybeInvGroupKey(groupMode, groupID),
	)
```

itest/litd_custom_channels_test.go
```diff
@@ -1992,17 +2039,18 @@ func testCustomChannelsLiquidityEdgeCases(ctx context.Context,
...
 	})
 	require.NoError(t.t, err)
+
 	payInvoiceWithAssets(
 		t.t, charlie, dave, btcInvoiceResp.PaymentRequest, assetID,
-		withFeeLimit(2_000), withPayErrSubStr(
-			"rejecting payment of 20000 mSAT",
-		),
+		withFeeLimit(2_000), withPayErrSubStr("rejecting payment of "+
+			"20000 mSAT"), withMaybeGroupKey(groupMode, groupID),
```
guggero, line 2046: nit: I'd probably format this as:
```
	payInvoiceWithAssets(
		t.t, charlie, dave, btcInvoiceResp.PaymentRequest, assetID,
		withFeeLimit(2_000), withPayErrSubStr(
			"rejecting payment of 20000 mSAT",
		), withMaybeGroupKey(groupMode, groupID),
	)
```

itest/litd_custom_channels_test.go
guggero: Yeah, I've actually come full circle now and started taking apart the HTLC force close test in https://github.com/lightninglabs/lightning-terminal/pull/1017 and making two out of them. That helps with starting with a clean slate.
And the performance optimizations I have in https://github.com/lightninglabs/lightning-terminal/pull/987 should make it so that there is no additional overhead for running things in the same test vs. in separate ones.
So this is okay the way it is now.

guggero (approved):
LGTM pending nits and update to the actual merged commit of `tapd`.

itest/assets_test.go
```diff
@@ -956,6 +964,16 @@ func defaultPayConfig() *payConfig {
...
 
+func withMaybeGroupKey(groupMode bool, groupKey []byte) payOpt {
+	return func(c *payConfig) {
+		if !groupMode {
+			return
+		}
+
+		c.groupKey = groupKey
+	}
+}
```
ffranr, line 975: `c.groupKey` defaults to `nil`, so we could just drop the "Maybe" and `groupMode` arg and just pass in `groupKey == nil` when not in group mode?

ffranr: No problem!

I think you also have the "Maybe" term still in the function name, I think you should remove that because the "maybe" aspect is up to the caller.

ffranr (approved):

PR #996, opened 2025-03-06
[sql-22] session: sessions CRUD

ellemouton:
In this PR, the SQL implementation of the `sessions.Store` is added. 

part of https://github.com/lightninglabs/lightning-terminal/issues/917

### DIFF
--- a/session/sql_store.go
+++ b/session/sql_store.go
@@ -1,0 +1,844 @@
+package session
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/lightninglabs/lightning-node-connect/mailbox"
+	"github.com/lightninglabs/lightning-terminal/accounts"
+	"github.com/lightninglabs/lightning-terminal/db"
+	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightningnetwork/lnd/clock"
+	"github.com/lightningnetwork/lnd/fn"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+	"gopkg.in/macaroon.v2"
+)
+
+// SQLQueries is a subset of the sqlc.Queries interface that can be used to
+// interact with session related tables.
+type SQLQueries interface {
+	GetAliasBySessionID(ctx context.Context, id int64) ([]byte, error)
+	GetSessionByID(ctx context.Context, id int64) (sqlc.Session, error)
+	GetSessionsInGroup(ctx context.Context, groupID sql.NullInt64) ([]sqlc.Session, error)
+	GetSessionAliasesInGroup(ctx context.Context, groupID sql.NullInt64) ([][]byte, error)
+	GetSessionByAlias(ctx context.Context, legacyID []byte) (sqlc.Session, error)
+	GetSessionByLocalPublicKey(ctx context.Context, localPublicKey []byte) (sqlc.Session, error)
+	GetSessionFeatureConfigs(ctx context.Context, sessionID int64) ([]sqlc.SessionFeatureConfig, error)
+	GetSessionMacaroonCaveats(ctx context.Context, sessionID int64) ([]sqlc.SessionMacaroonCaveat, error)
+	GetSessionIDByAlias(ctx context.Context, legacyID []byte) (int64, error)
+	GetSessionMacaroonPermissions(ctx context.Context, sessionID int64) ([]sqlc.SessionMacaroonPermission, error)
+	GetSessionPrivacyFlags(ctx context.Context, sessionID int64) ([]sqlc.SessionPrivacyFlag, error)
+	InsertSessionFeatureConfig(ctx context.Context, arg sqlc.InsertSessionFeatureConfigParams) error
+	SetSessionRevokedAt(ctx context.Context, arg sqlc.SetSessionRevokedAtParams) error
+	InsertSessionMacaroonCaveat(ctx context.Context, arg sqlc.InsertSessionMacaroonCaveatParams) error
+	InsertSessionMacaroonPermission(ctx context.Context, arg sqlc.InsertSessionMacaroonPermissionParams) error
+	InsertSessionPrivacyFlag(ctx context.Context, arg sqlc.InsertSessionPrivacyFlagParams) error
+	InsertSession(ctx context.Context, arg sqlc.InsertSessionParams) (int64, error)
+	ListSessions(ctx context.Context) ([]sqlc.Session, error)
+	ListSessionsByType(ctx context.Context, sessionType int16) ([]sqlc.Session, error)
+	ListSessionsByState(ctx context.Context, state int16) ([]sqlc.Session, error)
+	SetSessionRemotePublicKey(ctx context.Context, arg sqlc.SetSessionRemotePublicKeyParams) error
+	SetSessionGroupID(ctx context.Context, arg sqlc.SetSessionGroupIDParams) error
+	UpdateSessionState(ctx context.Context, arg sqlc.UpdateSessionStateParams) error
+	DeleteSessionsWithState(ctx context.Context, state int16) error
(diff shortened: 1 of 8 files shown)


session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+// state.
+//
+// NOTE: this is part of the Store interface.
+func (s *SQLStore) ShiftState(ctx context.Context, alias ID, dest State) error {
+	var writeTxOpts db.QueriesTxOptions
+	return s.db.ExecTx(ctx, &writeTxOpts, func(db SQLQueries) error {
+		dbSession, err := db.GetSessionByAlias(ctx, alias[:])
+		if err != nil {
+			return fmt.Errorf("unable to get session: %w", err)
+		}
```
ViktorTigerstrom, line 380: This function returns:
`ErrSessionNotFound` 

For the current kvdb logic when no session for the alias exists, so I think we'd want to mimic that rather then returning the sqlc `NoRows` error to outside the API?

session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+		readTxOpts = db.NewQueryReadTx()
+		sess       *Session
+	)
+	err := s.db.ExecTx(ctx, &readTxOpts, func(db SQLQueries) error {
+		dbSess, err := db.GetSessionByLocalPublicKey(
+			ctx, key.SerializeCompressed(),
+		)
+		if err != nil {
+			return fmt.Errorf("unable to get session: %w", err)
+		}
```
ViktorTigerstrom, line 451: Same as above.

session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+	var (
+		writeTxOpts db.QueriesTxOptions
+		remoteKey   = remotePubKey.SerializeCompressed()
+	)
+	return s.db.ExecTx(ctx, &writeTxOpts, func(db SQLQueries) error {
+		id, err := db.GetSessionIDByAlias(ctx, alias[:])
+		if err != nil {
+			return fmt.Errorf("unable to get session Alias: %v",
+				err)
+		}
```
ViktorTigerstrom, line 514: Same as above.

session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+	var (
+		readTxOpts    = db.NewQueryReadTx()
+		legacyGroupID ID
+	)
+	err := s.db.ExecTx(ctx, &readTxOpts, func(db SQLQueries) error {
+		// Get the session using the legacy Alias.
+		sess, err := db.GetSessionByAlias(ctx, sessionID[:])
+		if err != nil {
+			return ErrUnknownGroup
+		}
```
ViktorTigerstrom, line 600: Maybe return `ErrUnknownGroup` on no sql rows only?

session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+			return fmt.Errorf("session does not have a group Alias")
+		}
+
+		// Get the legacy group Alias using the session group Alias.
+		legacyGroupIDB, err := db.GetAliasBySessionID(
+			ctx, sess.GroupID.Int64,
+		)
+		if err != nil {
+			return err
+		}
```
ViktorTigerstrom, line 612: Maybe return `ErrSessionNotFound` on no rows, but I think that'd be unreachable.

session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+
+		sessIDs, err := db.GetSessionAliasesInGroup(
+			ctx, sql.NullInt64{
+				Int64: groupID,
+				Valid: true,
+			},
+		)
+		if err != nil {
+			return fmt.Errorf("unable to get session IDs: %v", err)
+		}
```
ViktorTigerstrom, line 651: Just noting that original logic actually seems to error if no sessions for the legacyGroupID is found, while this would return an emprty list. But given that I think the new logic is actually more correct, and I don't see a scenario where we'd hit that anyways, I think we should probably keep as is.

terminal.go
```diff
@@ -236,6 +236,15 @@ func New() *LightningTerminal {
 	}
 }
 
+// stores holds a collection of the DB stores that are used by LiT.
+type stores struct {
+	accounts accounts.Store
+	sessions session.Store
+
+	// close is a callback that can be used to close all the stores in the
```
ViktorTigerstrom, line 244: nit: Unfinished sentence.

ViktorTigerstrom (commented):
Generally looks great :rocket:!! 

Just leaving some comments mainly regarding differences in results between the original kvdb logic and the new sql implementation :).

session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+		if err != nil {
+			return fmt.Errorf("unable to unmarshal session: %w",
+				err)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
ellemouton: sure - not a big deal

session/sql_store.go
ellemouton: oops! good catch thanks

session/sql_store.go
```diff
@@ -0,0 +1,839 @@
...
+
+	// Get the macaroon caveats if they exist.
+	caveats, err := db.GetSessionMacaroonCaveats(ctx, dbSess.ID)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get macaroon "+
+			"caveats: %v", err)
+	}
+
+	var macRecipe *MacaroonRecipe
+	if perms != nil || caveats != nil {
```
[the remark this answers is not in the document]
ellemouton: yes - look at the calls that follow

ellemouton (commented):
thanks yall 🙏

session/sql_store.go
```diff
@@ -0,0 +1,841 @@
...
+		writeTxOpts db.QueriesTxOptions
+		remoteKey   = remotePubKey.SerializeCompressed()
+	)
+	return s.db.ExecTx(ctx, &writeTxOpts, func(db SQLQueries) error {
+		id, err := db.GetSessionIDByAlias(ctx, alias[:])
+		if err != nil && errors.Is(err, sql.ErrNoRows) {
+			return fmt.Errorf("%w: %w", ErrSessionNotFound, err)
+		} else if err != nil {
+			return fmt.Errorf("%w: unable to get session: %w",
+				ErrSessionNotFound, err)
```
ViktorTigerstrom, line 514: nit: The second one here shouldn't be a `ErrSessionNotFound` error?

session/sql_store.go
```diff
@@ -0,0 +1,841 @@
...
+	var (
+		readTxOpts = db.NewQueryReadTx()
+		sessionIDs []ID
+	)
+	err := s.db.ExecTx(ctx, &readTxOpts, func(db SQLQueries) error {
+		groupID, err := db.GetSessionIDByAlias(ctx, legacyGroupID[:])
+		if err != nil {
+			return fmt.Errorf("unable to get session Alias: %v",
+				err)
+		}
```
ViktorTigerstrom, line 643: nit: potentially add:
```
if errors.Is(err, sql.ErrNoRows) {
	return ErrUnknownGroup
} else if err != nil {
```

session/sql_store.go
```diff
@@ -0,0 +1,841 @@
...
+//
+// NOTE: this is part of the Store interface.
+func (s *SQLStore) ShiftState(ctx context.Context, alias ID, dest State) error {
+	var writeTxOpts db.QueriesTxOptions
+	return s.db.ExecTx(ctx, &writeTxOpts, func(db SQLQueries) error {
+		dbSession, err := db.GetSessionByAlias(ctx, alias[:])
+		if err != nil {
+			return fmt.Errorf("%w: unable to get session: %w",
+				ErrSessionNotFound, err)
+		}
```
ViktorTigerstrom, line 377: nit:
Only return `ErrSessionNotFound` if no `sql.NoRows`.

ellemouton (commented):
addressed, thanks @ViktorTigerstrom 🙏

PR #997, opened 2025-03-07
multi: update LiT and all deps to use LND 0.19

ellemouton:


ellemouton:
soz, this needs to be re-opened via an upstream branch. gonna close and re-open

PR #1000, opened 2025-03-10
[sql-23] db/sqlc: only use named UNIQUE indices

ellemouton:
Replace any named UNIQUE indices in the accounts tables and replace with named ones. This is for a future where we want to be able to edit these indices without having to drop the table.
It is fine for us to just edit these tables since these are not available in "prod" yet.


bitromortac (approved):
LGTM, what would be a reason to edit such indices?

ellemouton:
> what would be a reason to edit such indices?

Just in case we want to add to the index in future maybe or even just to be consistent incase we want to add another one. But I think also just hard to predict the future use-cases.

ViktorTigerstrom (approved):
Nice, thanks :fire:

PR #1001, opened 2025-03-11
[sql-23] firewalldb: thread contexts through for kv-store interfaces

ellemouton:
This PR does 1 main things:

1) updates various of the rule db kv-store interfaces to take contexts and then threads contexts through where necessary. 
2) creates a generic `DBExecutor` interface and use init the KVStores interface. Later on we will use it to replace the PrivacyMap interface too.

part of https://github.com/lightninglabs/lightning-terminal/issues/917


ViktorTigerstrom (approved):
Nice :fire:!

PR #1004, opened 2025-03-13
litd_custom_channels_test.go: Better commenting of itest cases

ZZiigguurraatt:
This PR is focused on making comments surrounding itest cases easier to understand with more details and a visual of the route that the payment is expected to take.

There are a few cases that I have adjusted and would like feedback on this approach. If it is desired, I will continue to do the same thing to all test cases.


guggero (approved):
Thank you!

PR #998, opened 2025-03-07
multi: update LiT and all deps to use LND 0.19

ellemouton:

PR #1006, opened 2025-03-15
multi: go version and linter bump

ellemouton:
This PR siphons of a few commits from https://github.com/lightninglabs/lightning-terminal/pull/998 which we can get in without the LND dep bump. 

The main purpose being so that other PRs that introduce more generic concepts can pass the linter check (the updated linter dep works better with generics).


.golangci.yml
```diff
@@ -31,8 +31,10 @@ linters-settings:
     excludes:
       - G402 # Look for bad TLS connection settings.
       - G306 # Poor file permissions used when writing to a new file.
+      - G601 # Implicit memory aliasing in for loop.
+      - G115 # Integer overflow in conversion.
```
bitromortac, line 35: maybe we could fix those in the future

bitromortac (approved):
LGTM :zap:

ViktorTigerstrom (approved):
LGTM ⚡

PR #1008, opened 2025-03-21
re-use `createNormalInvoice` in `createAndPayNormalInvoiceWithBtc`

ZZiigguurraatt:
small elimination of unnecessary redundant code

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -805,17 +805,7 @@ func sendKeySendPayment(t *testing.T, src, dst *HarnessNode,
 func createAndPayNormalInvoiceWithBtc(t *testing.T, src, dst *HarnessNode,
 	amountSat btcutil.Amount) {
 
-	ctxb := context.Background()
-	ctxt, cancel := context.WithTimeout(ctxb, defaultTimeout)
-	defer cancel()
-
-	expirySeconds := 10
-	invoiceResp, err := dst.AddInvoice(ctxt, &lnrpc.Invoice{
-		Value:  int64(amountSat),
-		Memo:   "normal invoice",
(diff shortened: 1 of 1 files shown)


ViktorTigerstrom (approved):

ellemouton (approved):

bitromortac (approved):

ellemouton:
@ZZiigguurraatt  - could you just update the commit message to use the standard structure pls? 🙏  thanks

ZZiigguurraatt:
> @ZZiigguurraatt - could you just update the commit message to use the standard structure pls? 🙏 thanks

Where do I find a definition of the standard structure so that I can do that properly?

guggero:
It's not linked specifically in all our projects but we try to follow these guidelines everywhere: https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md

ZZiigguurraatt:
OK. how is the new commit message I've just changed to?

ViktorTigerstrom:
> OK. how is the new commit message I've just changed to?

I recommend you to also read how commit messages should be prefixed according to the golang convention docs, which is also linked in our contribution guidelines:  https://go.dev/doc/contribute#commit_messages

In practise, your commit message here should be prefixed by the primary affected package (`itest` in this case). Also, you shouldn't capitalize the first word in the commit title. Other than that, your commit message looks good IMO :). 
If anything, the title would potentially be even clearer if phrased as:
itest: utilise `createNormalInvoice` helper

But I'll let you decide on that as that's very nit-picky :)

ZZiigguurraatt:
OK, looks good now?

Issue #940, opened 2025-01-17
[itest flake]: custom channels `test_custom_channels_htlc_force_close` flake


ellemouton:
The following 2 flakes in the `test_custom_channels_htlc_force_close` test occur quite frequently in our CI:

[Flake 1](https://github.com/lightninglabs/lightning-terminal/actions/runs/12824017132/job/35762460522?pr=938)

```
    litd_custom_channels_test.go:3204: 
        	Error Trace:	/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_custom_channels_test.go:3204
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_custom_channels_test.go:2924
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/test_harness.go:103
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_test.go:102
        	Error:      	Received unexpected error:
        	            	wanted 1, found 0 txs in mempool: []
        	Test:       	TestLightningTerminal/test_custom_channels_htlc_force_close
```

[Flake 2](https://github.com/lightninglabs/lightning-terminal/actions/runs/12805233549/job/35702160888)

```
    assets_test.go:350: 
        	Error Trace:	/home/runner/work/lightning-terminal/lightning-terminal/itest/assets_test.go:350
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_custom_channels_test.go:3311
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_custom_channels_test.go:2923
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/test_harness.go:103
        	            				/home/runner/work/lightning-terminal/lightning-terminal/itest/litd_test.go:102
        	Error:      	Received unexpected error:
        	            	missing anchor block hash, transfer not confirmed
        	Test:       	TestLightningTerminal/test_custom_channels_htlc_force_close
```

PR #1017, opened 2025-03-27
multi: bump to latest lnd/tapd, assert close chan data

guggero:
~Depends on https://github.com/lightningnetwork/lnd/pull/9504.~
~Depends on https://github.com/lightninglabs/taproot-assets/pull/1441.~

Serves as validation code for the changes in https://github.com/lightningnetwork/lnd/pull/9504.

Also fixes https://github.com/lightninglabs/lightning-terminal/issues/940.

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -473,9 +473,9 @@ func assertPendingChannels(t *testing.T, node *HarnessNode,
 		pendingChan.Channel.CustomChannelData, &pendingJSON,
 	)
 	require.NoError(t, err)
-	require.Len(t, pendingJSON.Assets, 1)
+	require.Len(t, pendingJSON.FundingAssets, 1)
 
-	require.NotZero(t, pendingJSON.Assets[0].Capacity)
+	require.NotZero(t, pendingJSON.Capacity)
 
 	// Check the decimal display of the channel funding blob. If no explicit
 	// value was set, we assume and expect the value of 0.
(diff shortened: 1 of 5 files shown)


itest/assets_test.go
```diff
@@ -501,21 +501,36 @@ func assertPendingChannels(t *testing.T, node *HarnessNode,
...
 		a, err := getChannelCustomData(src, dst)
 		if err != nil {
 			return err
 		}
 
-		if a.AssetInfo.AssetGenesis.AssetID != assetIDStr {
-			return fmt.Errorf("expected asset ID %s, got %s",
-				assetIDStr, a.AssetInfo.AssetGenesis.AssetID)
+		assetID := mintedAsset.AssetGenesis.AssetId
+		if !haveFundingAsset(a, assetID) {
```
gijswijs, line 529: For my understanding:
In the old situation there could only be one type of AssetID as the funding asset of a custom channel. But in the new situation we can have funding assets with different AssetID's and we check if _one_ of them is the assetID we expect.

itest/assets_test.go
```diff
@@ -2200,3 +2221,108 @@ func assertInvoiceState(t *testing.T, hn *HarnessNode, payAddr []byte,
...
+	}
+	require.NotNil(t, waitingCloseChan)
+	require.NotEmpty(t, waitingCloseChan.CustomChannelData)
+
+	var closeData rfqmsg.JsonAssetChannel
+	err = json.Unmarshal(waitingCloseChan.CustomChannelData, &closeData)
+	require.NoError(t, err)
+
+	require.GreaterOrEqual(t, len(closeData.FundingAssets), 1)
+}
```
gijswijs, line 2328: I tried to come up with a clever way for a helper function, so that these three funcs don't have so much similar code, but I fell short. It's just slightly different at different spots such that 3 separate funcs are warranted. :+1:

gijswijs (commented):
Only a few questions for my understanding.
The itests affected by this PR do currently fail.

gijswijs (approved):
lgtm :game_die:

guggero:
Yesss, itests are stable now. cc @GeorgeTsagk

PR #1020, opened 2025-03-31
Asset Invoice with sats value itest

GeorgeTsagk:
## Description

This itest tests the behavior exposed in https://github.com/lightninglabs/taproot-assets/pull/1448

It is very minimal, and currently only to be used as a proof-of-concept sanity check for the above feature.

### DIFF
--- a/itest/assets_test.go
+++ b/itest/assets_test.go
@@ -1427,6 +1427,7 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
 type invoiceConfig struct {
 	errSubStr  string
 	groupKey   []byte
+	msats      lnwire.MilliSatoshi
 	routeHints []*lnrpc.RouteHint
 }
 
@@ -1447,6 +1448,12 @@ func withInvoiceErrSubStr(errSubStr string) invoiceOpt {
 func withInvGroupKey(groupKey []byte) invoiceOpt {
 	return func(c *invoiceConfig) {
 		c.groupKey = groupKey
+	}
+}
+
+func withMsatAmount(amt uint64) invoiceOpt {
+	return func(c *invoiceConfig) {
+		c.msats = lnwire.MilliSatoshi(amt)
 	}
 }
 
@@ -1478,7 +1485,8 @@ func createAssetInvoice(t *testing.T, dstRfqPeer, dst *HarnessNode,
 		InvoiceRequest: &lnrpc.Invoice{
 			Memo: fmt.Sprintf("this is an asset invoice for "+
 				"%d units", assetAmount),
-			Expiry: timeoutSeconds,
+			Expiry:    timeoutSeconds,
+			ValueMsat: int64(cfg.msats),
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


itest/litd_custom_channels_test.go
```diff
@@ -1972,18 +1972,6 @@ func testCustomChannelsLiquidityEdgeCases(ctx context.Context,
 	logBalance(t.t, nodes, assetID, "after big asset payment (asset "+
 		"invoice, multi-hop)")
 
-	// Edge case: Now Charlie creates a tiny asset invoice to be paid for by
-	// Yara with satoshi. This is a multi-hop payment going over 2 asset
-	// channels, where the total asset value is less than the default anchor
-	// amount of 354 sats.
-	createAssetInvoice(t.t, dave, charlie, 1, assetID, withInvoiceErrSubStr(
```
guggero, line 1979: Why are we removing this edge case test? This should still apply IMO. The error message has changed a bit though.

itest/litd_custom_channels_test.go
```diff
@@ -2098,7 +2086,7 @@ func testCustomChannelsLiquidityEdgeCases(ctx context.Context,
 
 	// Now Fabia creates the normal invoice.
 	invoiceResp = createAssetInvoice(
-		t.t, erin, fabia, 10_000, assetID,
+		t.t, erin, fabia, 0, assetID, withMsatAmount(170_000_000),
```
guggero, line 2089: Should we add a new case for this new functionality? Or at least update the comment above this to point out that we're testing a specific case?

itest/assets_test.go
```diff
@@ -1168,11 +1176,20 @@ func createAssetInvoice(t *testing.T, dstRfqPeer, dst *HarnessNode,
 
 	t.Logf("Got quote for %v asset units per BTC", rate)
 
-	assetUnits := rfqmath.NewBigIntFixedPoint(assetAmount, 0)
-	numMSats := rfqmath.UnitsToMilliSatoshi(assetUnits, *rate)
-	mSatPerUnit := float64(decodedInvoice.NumMsat) / float64(assetAmount)
+	var mSatPerUnit float64
+
+	if cfg.msats > 0 {
+		require.EqualValues(t, decodedInvoice.NumMsat, cfg.msats)
```
guggero, line 1182: I don't think this will always be an exact amount? Since we're doing a full msat -> asset -> msat round trip. So I think we should allow for at least a full asset unit (expressed in msat) worth of difference.
And to validate that, we should probably choose an msat amount in the test that is not evenly divisible by the rate.

guggero (commented):
I know this wasn't marked for review yet, but did a drive-by anyway.

guggero (approved):
LGTM :tada: 

But needs to be based on top of https://github.com/lightninglabs/lightning-terminal/pull/987 as well to fully satisfy the CI. Then the first commit can be dropped.

guggero:
Rebased and ready for review, @ffranr.

ZZiigguurraatt:
I get

```
e563e22bdadf:/$ litcli ln addinvoice --asset_id 5f2e3073ef4d23ac2b9482ad9988fb6ae7d3cc48d260ed00b0f9f655d4a7e56e --amt=1000
[litcli] asset_amount argument missing
e563e22bdadf:/$ 
```

is that something you want to fix in this PR, or a separate one?

ffranr (approved):

ffranr:
My last forced push on this PR was a rebase.

ffranr:
> I get
> 
> ```
> e563e22bdadf:/$ litcli ln addinvoice --asset_id 5f2e3073ef4d23ac2b9482ad9988fb6ae7d3cc48d260ed00b0f9f655d4a7e56e --amt=1000
> [litcli] asset_amount argument missing
> e563e22bdadf:/$ 
> ```
> 
> is that something you want to fix in this PR, or a separate one?

Separate PR.

PR #1029, opened 2025-04-07
[sql-28] firewall+sessions: clean up in preparation for SQL backend

ellemouton:
In this PR, we do the last bit of clean up of the bbolt kvstores impl along with the tests of the kvstores. 

The main thing this PR does is to assert strong coupling between kvstores and the session it is referencing. We do this in the bbolt implementation so that our unit tests can assert the correct errors. After this, we should be able to just plug in the SQL impl without making any further changes to the kvstore unit tests. 

See commit messages for more details.

Part of #917

### DIFF
--- a/firewalldb/kvstores_test.go
+++ b/firewalldb/kvstores_test.go
@@ -5,8 +5,10 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/lightninglabs/lightning-terminal/session"
+	"github.com/lightningnetwork/lnd/clock"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


firewalldb/kvstores_test.go
```diff
@@ -376,6 +392,81 @@ func TestKVStoreNameSpaces(t *testing.T) {
...
+func TestKVStoreSessionCoupling(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+
+	sessions := session.NewTestDB(t, clock.NewDefaultClock())
+	db := NewTestDBWithSessions(t, sessions)
+
+	// Get a kvstore namespaced by a session ID for a session that does
+	// not exist.
+	store := db.GetKVStores("AutoFees", [4]byte{1, 1, 1, 1}, "auto-fees")
```
bitromortac, line 406: would it make sense to already error here in case the session is not found, or does this prevent some functionality?

firewalldb/kvstores_test.go
```diff
@@ -376,6 +392,81 @@ func TestKVStoreNameSpaces(t *testing.T) {
 	require.True(t, bytes.Equal(v, []byte("thing 3")))
 }
 
+// TestKVStoreSessionCoupling tests if we attempt to write to a kvstore that
+// is namespaced by a session that does not exist, then we should get an error.
+func TestKVStoreSessionCoupling(t *testing.T) {
```
ViktorTigerstrom, line 397: :fire:

ViktorTigerstrom (approved):
Niiiice uTACK LGTM :fire:!

bitromortac (approved):
LGTM :zap:

PR #1030, opened 2025-04-07
[sql-29]: firewalldb: implement SQL version of RulesDB (kvstores)

ellemouton:
This PR adds the SQL version of the RulesDb

### DIFF
--- a/firewalldb/kvstores_sql.go
+++ b/firewalldb/kvstores_sql.go
@@ -1,0 +1,451 @@
+package firewalldb
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/lightninglabs/lightning-terminal/db"
+	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightninglabs/lightning-terminal/session"
+	"github.com/lightningnetwork/lnd/fn"
+)
+
+// SQLSessionQueries is a subset of the sqlc.Queries interface that can be used
+// to interact with the session table.
+type SQLSessionQueries interface {
+	GetSessionIDByAlias(ctx context.Context, legacyID []byte) (int64, error)
+}
+
+// SQLKVStoreQueries is a subset of the sqlc.Queries interface that can be
+// used to interact with the kvstore tables.
+//
+//nolint:lll
+type SQLKVStoreQueries interface {
+	SQLSessionQueries
+
+	DeleteFeatureKVStoreRecord(ctx context.Context, arg sqlc.DeleteFeatureKVStoreRecordParams) error
+	DeleteGlobalKVStoreRecord(ctx context.Context, arg sqlc.DeleteGlobalKVStoreRecordParams) error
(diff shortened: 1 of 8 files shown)


terminal.go
```diff
@@ -238,6 +238,7 @@ func New() *LightningTerminal {
 type stores struct {
 	accounts accounts.Store
 	sessions session.Store
+	rules    firewalldb.RulesDB
 
 	firewall     *firewalldb.DB
```
bitromortac, line 243: there is some redundancy between rules and firewall, is it possible to remove one of them, or how will that look in the end? will we have a single firewall db here or the individual RulesDB/PrivacyDB?

ellemouton (commented):
@bitromortac  - addressed nits

ellemouton:
note: will fix linter after next review round

firewalldb/kvstores_sql.go
```diff
@@ -0,0 +1,452 @@
...
+//
+// NOTE: part of the KVStore interface.
+func (s *sqlKVStore) Del(ctx context.Context, key string) error {
+	// Note: we pass in true here for "read-only" since because this is a
+	// Delete, if the record does not exist, we don't need to create one.
+	// But no need to error out if it doesn't exist.
+	ruleID, sessionID, featureID, err := s.genNamespaceFields(ctx, true)
+	if errors.Is(err, sql.ErrNoRows) ||
+		errors.Is(err, session.ErrUnknownGroup) {
+		fmt.Println(err)
```
ViktorTigerstrom, line 290: Nit: Other than the linting issue here, is this an intentional `fmt.Println` and not a `log`?

ViktorTigerstrom (approved):
LGTM after nits have been addressed! Great work :rocket::fire:

firewalldb/kvstores_sql.go
ellemouton: this is the linting issue :) 

anycase, fixed now!

terminal.go
bitromortac: right, I think it's still possible and rules it not needed, unless I misunderstood you, see the patch
[changes.txt](https://github.com/user-attachments/files/19758562/changes.txt)

bitromortac (approved):
LGTM :tada: (one comment left and a nit)

firewalldb/sql_store.go
```diff
@@ -0,0 +1,86 @@
...
+// BatchedSQLQueries is a version of the SQLQueries that's capable of batched
+// database operations.
+type BatchedSQLQueries interface {
+	SQLQueries
+
+	db.BatchedTx[SQLQueries]
+}
+
+// SQLDB represents a storage backend.
+type SQLDB struct {
```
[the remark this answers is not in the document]
ellemouton: think let's address in a follow up

terminal.go
ellemouton: ah ok yes indeed - you're totally right. Thanks for the call out 🙏

bitromortac: thank you for the update!

ellemouton: ah, the diff needed one more thing i think. fixed now i hope 🤞

PR #1034, opened 2025-04-14
Add `commit_hash`  field to `GetInfo` response

ViktorTigerstrom:
Mimics the behaviour of loop pull 914 (https://github.com/lightninglabs/loop/pull/914), but for `litd`.

This change adds a new field `commit_hash` field to the `GetInfo` response, which will contain the full commit hash of the commit that LiT release binary build was based on. If the build had uncommitted changes, the `commit_hash` field will contain the most recent commit hash, suffixed by "-dirty".

The contents of `version` field of the `GetInfo` response is also will also be changed, so that the it'll only include the `litd` `version`, and not the commit_hash/tag.

A `commit_hash` field is also added to the `--version` output, so which will always contains the full commit hash of the latest commit (with no suffix even on dirty builds).

Note to reviewers:
Just to ensure that this doesn't break the reproducible build across different hosts:
Could you please create a local tag on top of this:

1. `git tag -s v123456789.123456789.123456789-alpha`
2. And build the release with `make docker-release tag=v123456789.123456789.123456789-alpha` if you're on mac, or `make release tag=v123456789.123456789.123456789-alpha` with the correct GO version if you're on Linux.
3. Then verify that the `sha256` output for the `./lightning-terminal-v123456789.123456789.123456789-alpha/manifest-v123456789.123456789.123456789-alpha.txt` file is: `e9bc6a43642d7ae679f1f2bd809c3a517cefc0bddb32beee5e990fc60bd75783`


litrpc/proxy.proto
```diff
@@ -54,4 +54,7 @@ message GetInfoRequest {
 message GetInfoResponse {
     // The version of the LiTd software that the node is running.
     string version = 1;
+
+    // The git commit hash of the litd binary.
+    string commit_hash = 2;
```
ViktorTigerstrom, line 59: Let me know if you think `commit` would be a more suitable name for this, as the contents may be the `tag` rather than a commit hash. But opted for this naming as it then mimics the `loop` getinfo response :)

bitromortac (approved):
LGTM, nice :zap:

ellemouton (commented):
I guess I dont love the "it's sometimes this but then sometimes that" part. 

Cant we just leave "Commit" as it is (to not change behaviour there) but then _add_ "Commit_hash" which is always the revision hash no matter what?

ViktorTigerstrom:
Thanks for the reviews! 

I've updated the PR to address the reviews, by instead adding a `tag` and a `commit_hash` field to the `GetInfo` response.
The `tag` field will be set to the Git tag that the LiT release binary build was based on. If the build was not based on a clean tagged commit, this field will contain the most recent tag suffixed by the commit hash, and/or a "-dirty" suffix (i.e. the same behaviour as the old "-Commit=" part in the old `GetInfo` response).

The `commit_hash` field will contain the full commit hash of the commit that LiT release binary build was based on.

The contents of the `version` field in the `GetInfo` response has also been updated to only include the semantic version number of the LiT release, following the semantic versioning 2.0.0 spec (http://semver.org/).

@bitromortac & @ellemouton, can you please check the updates, and check if you think the outputs of the `litcli getinfo` and `litcli --version` commands makes sense? If you do, I'll proceed to communicate the changes to the `loop` team and coordinate to and ensure that we try to follow the same standard for all daemons.

ellemouton (commented):
Left some comments, let me know your thoughts :)

ViktorTigerstrom:
Updated the PR to address feedback we discussed offline:

In general, this PR has been updated so that:
A `commit_hash` field has been added to the `GetInfo` response. The `commit_hash` field will contain the full commit hash of the commit that LiT release binary build was based on. If the build had uncommitted changes, the `commit_hash` field will contain the most recent commit hash, suffixed by "-dirty".

The contents of the `version` field in the `GetInfo` response has also been updated to only include the semantic version number of the LiT release, following the semantic versioning 2.0.0 spec (http://semver.org/).

Additionally, a `commit_hash` field has been added to the `--version` output. The `commit_hash` field will contain the full commit hash of the commit that LiT release binary build was based on and will never be suffixed by "-dirty", even if the build was dirty.

Makefile
```diff
@@ -69,6 +70,8 @@ make_ldflags = $(2) -X $(LND_PKG)/build.Commit=lightning-terminal-$(COMMIT) \
 	-X $(LND_PKG)/build.RawTags=$(shell echo $(1) | sed -e 's/ /,/g') \
```
ViktorTigerstrom, line 70: Note that we do not change the:
`$(LND_PKG)/build.Commit=lightning-terminal-$(COMMIT)`

above here, so that the `Commit` in response through `lncli` remains the same. I think it doesn't make sense to change this until we also change the output of `getinfo` in `lnd`.

ellemouton (approved):
lgtm!

bitromortac (approved):

PR #1051, opened 2025-04-29
[sql-33] session: add migration code from kvdb to SQL

ViktorTigerstrom:
This PR introduces the migration logic for transitioning the sessions store from kvdb to SQL.


session/sql_store.go
```diff
@@ -745,9 +745,12 @@ func unmarshalSession(ctx context.Context, db SQLQueries,
...
 	if perms != nil || caveats != nil {
-		macRecipe = &MacaroonRecipe{
-			Permissions: unmarshalMacPerms(perms),
-			Caveats:     unmarshalMacCaveats(caveats),
+		macRecipe = &MacaroonRecipe{}
+		if perms != nil {
+			macRecipe.Permissions = unmarshalMacPerms(perms)
+		}
+		if caveats != nil {
+			macRecipe.Caveats = unmarshalMacCaveats(caveats)
```
ViktorTigerstrom, line 753: Let me know if you think we should rather fix the kvdb store implementation rather than the sql one. But as we've previously opted for not updating the store behaviour, and make the sql one mimic the behaviour until we've completed the migration, I opted for this version here.

session/sql_migration.go
```diff
@@ -0,0 +1,403 @@
...
+	// has been created.
+	if !session.RevokedAt.IsZero() {
+		err = tx.SetSessionRevokedAt(
+			ctx,
+			makeSetRevokedAtParams(sqlId, session.RevokedAt),
+		)
+		if err != nil {
+			return err
+		}
+	}
```
ViktorTigerstrom, line 201: Could potentially update the insert session sql query to also take an optional `RevokedAt sql.NullTime` param, to not have to do this as a separate step. I opted for this version to keep things as is as much as possible, but if you think that'd be better, I'll update to do that instead :)

session/sql_migration_test.go
```diff
@@ -0,0 +1,738 @@
...
+					time.Unix(1000, 0), "",
+					WithLinkedGroupID(&sess1.ID),
+				)
+				require.NoError(t, err)
+			},
+		},
+		{
+			"randomized sessions",
+			randomizedSessions,
+		},
```
ViktorTigerstrom, line 309: Opted for a test that uses the "standard" `rand` lib here, as @bitromortac had a preference for that in https://github.com/lightninglabs/lightning-terminal/pull/1047#discussion_r2059882334. But can also look into adding a test that uses `rapid` if you'd prefer.

session/sql_migration_test.go
```diff
@@ -0,0 +1,738 @@
...
+
+	loopLen := rand.Intn(10) + 1
+
+	if rand.Intn(2) == 0 {
+		for i := 0; i < loopLen; i++ {
+			var macCaveat macaroon.Caveat
+
+			// We always have a caveat.Id, but the rest are
+			// randomized if they exist or not.
+			macCaveat.Id = randomBytes(rand.Intn(10) + 1)
```
ViktorTigerstrom, line 598: Let me know if this **is not true** in for real world data, as migrations WILL fail if we don't set a `caveat.Id` when the `caveats` param of `WithMacaroonRecipe` is not `nil`.

session/sql_migration_test.go
```diff
@@ -0,0 +1,738 @@
...
+		b[i] = letterBytes[rand.Intn(len(letterBytes))]
+	}
+	return string(b)
+}
+
+// shiftStateUnsafe updates the state of the session with the given ID to the
+// "dest" state, without checking if the state transition is legal.
+//
+// NOTE: this function should only be used for testing purposes.
+func (db *BoltStore) shiftStateUnsafe(_ context.Context, id ID,
```
ViktorTigerstrom, line 708: Not fully sure if we add test functions like this one that's a `struct` member to the `_test` file, or if we add it in the actual struct file?

ViktorTigerstrom (commented):
Adding some review comments where input would be extra helpful :)!

session/sql_migration_test.go
bitromortac: right, not too great, you could change to `func shiftStateUnsafe(_ context.Context, db *BoltStore, id ID, dest State) error {`

bitromortac (commented):
Great work, very clean on first pass! :fire:

ellemouton (commented):
holding off on review here until the prep PR discussed in the accounts mig PR is in (since that will be used to test this one)

ViktorTigerstrom:
Now that #1047 has been merged, I'll update this PR to adhere to the feedback brought up in that PR, and then re-request reviews afterwards.

ViktorTigerstrom:
Updated the PR to also address feedback raised in #1047! Will address the small linting issue after next review round :)

session/sql_migration_test.go
```diff
@@ -0,0 +1,368 @@
...
+	"github.com/lightningnetwork/lnd/clock"
+	"github.com/lightningnetwork/lnd/macaroons"
+	"github.com/lightningnetwork/lnd/sqldb"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/macaroon-bakery.v2/bakery/checkers"
+	"gopkg.in/macaroon.v2"
+)
+
+// TestSessionsStoreMigration tests the migration of session store from a bolt
+// backed to a SQL database. Note that this test does not attempt to be a
```
bitromortac, line 25: nit: `backend`

bitromortac (approved):
LGTM, utACK :tada:

session/sql_migration_test.go
```diff
@@ -0,0 +1,776 @@
...
+				_, err := store.NewSession(
+					ctx, "test", TypeMacaroonAdmin,
+					time.Unix(1000, 0), "foo.bar.baz:1234",
+					WithMacaroonRecipe(caveats, nil),
+				)
+				require.NoError(t, err)
+			},
+		},
+		{
+			name: "macaroon recipe with nil perms and caveats",
```
ViktorTigerstrom, line 241: Note that this test is new, as this can actually happen for real world sessions.

session/sql_store.go
ViktorTigerstrom: Alternatively, I could also override this in newly added `overrideMacaroonRecipe` function before the migration deep equals check. But I kept this as is, as I do think it's correct behaviour that we don't set the `Permissions` or `Caveats` and keep them at `nil`, if no value actually exists for them.

session/sql_migration_test.go
ViktorTigerstrom: Could potentially override this as well in the new `overrideMacaroonRecipe` as well. But I do not think this scenario can happen for real world data. But open to implementing that as well if you think it's worth doing it just to be sure.

ViktorTigerstrom (commented):
Thanks a lot for the review @bitromortac :tada:!! I noticed that we can have real world data that sets a non-nil macaroon recipe, but with `nil` `perms`& `caveats`. I updated the code to be able to handle that case as well.

session/sql_migration.go
```diff
@@ -0,0 +1,364 @@
...
+
+	kvSessions, err := kvStore.ListAllSessions(ctx)
+	if err != nil {
+		return err
+	}
+
+	// If sessions are linked to a group, we must insert the initial session
+	// of each group before the other sessions in that group. This ensures
+	// we can retrieve the SQL group ID when inserting the remaining
+	// sessions. Therefore, we first insert all initial group sessions,
```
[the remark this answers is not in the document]
ellemouton: ie, i think we can just:

```
for s := sessions {
       if s.groupID  = s.ID {
                    <insert new session. no coupled group>
                   continue
       }
       
       groupID := db.GetGroupIDByAlias(s.groupID)
       
        <insert new session. couple to group>
}  
```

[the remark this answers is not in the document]
ellemouton: or did you actually run into a scenario where these were not in order?

session/sql_migration.go
```diff
@@ -0,0 +1,364 @@
...
+		var acctDBID int64
+		acctDBID, err = tx.GetAccountIDByAlias(ctx, acctAlias)
+		if errors.Is(err, sql.ErrNoRows) {
+			err = accounts.ErrAccNotFound
+			return
+		} else if err != nil {
+			return
+		}
+
+		acctID = sql.NullInt64{
```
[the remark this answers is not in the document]
ViktorT-11: Thanks, addressed these and others :pray:!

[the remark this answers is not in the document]
ViktorT-11: Thanks, addressed these and others :pray:!

session/sql_migration.go
```diff
@@ -0,0 +1,364 @@
...
+			Int64: acctDBID,
+			Valid: true,
+		}
+	})
+	if err != nil {
+		return err
+	}
+
+	// The remote public key is currently only set for autopilot sessions,
```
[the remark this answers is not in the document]
ViktorT-11: Ah yeah of course 😅... My bad, thanks :pray:!

session/sql_migration.go
```diff
@@ -0,0 +1,364 @@
...
+	// Finally we insert the privacy flags.
+	for _, privacyFlag := range session.PrivacyFlags {
+		privacyFlagParams := sqlc.InsertSessionPrivacyFlagParams{
+			SessionID: sqlId,
+			Flag:      int32(privacyFlag),
+		}
+
+		err = tx.InsertSessionPrivacyFlag(ctx, privacyFlagParams)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
ViktorT-11: Addressed this and others :)

session/sql_migration.go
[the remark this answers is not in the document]
ViktorT-11: Hmmm so the `listSessions` for the kvdb returns the sessions sorted by the `CreatedAt` field. So given that this field actually contains timestamps in the correct order, your suggestion should work. But I see 2 potential scenarios where that wouldn't be true:
1. The user has managed to change the used time, so that linked sessions gets a `CreatedAt` timestamp prior to the initial session they are linking to.
2. Even though it's a real edge case, I believe that when we changed to using `clock.Clock` for `litd`, that we realised that this could have the side effect that the new timestamps would potentially get changed by +- a few hours unless I'm remembering this incorrectly? In that scenario, for some edge case users, they could have created the initial session before updating to the new `litd` version that added the `clock.Clock` implementation, and then linked a session to it right after which could then have an earlier `CreatedAt` timestamp than the initial session.

As I don't see any real cons with my current approach, other than it adding some more code complexity, I think it's a worth tradeoff than risking the migrations failing for such edge cases. Do you agree?

ViktorTigerstrom:
I added a commit 6b8c58f9c9168471714adbba78446047b8614137 at the end of the PR to address the feedback of https://github.com/lightninglabs/lightning-terminal/pull/1051#discussion_r2111840035, as well as a fixup commit I will apply if you think the approach of added commit is good @ellemouton. Would therefore appreciate feedback on if this approach is what you had in mind when you suggested that I should change the kvdb implementation? Note that this actually changes how the `MacaroonRecipe` is persisted in `kvdb`, and it will set the field to `null` (which will also be represented when listing the session with `litcli`). 

Note that I will fix the failing unit test tomorrow, that's due to my update of how kvdb implementation stores the `MacaroonRecipe`.

ViktorTigerstrom:
Updated the PR by ensuring that the `overrideMacaroonRecipe` by itself handles the differences in how the `MacaroonRecipe` is represented between the kvdb & the SQL store. Therefore no changes are needed for how we'd persist it for neither the `kvdb` nor the `sql` stores.

@ellemouton, please let me know if you prefer that approach, or the other approach I've also implemented that changes how the kvdb persists the `MacaroonRecipe`. 
Personally I'm in strong favour of letting the `overrideMacaroonRecipe` func handle the differences, and i.e. drop the last 2 commits that currently exists in this PR.

session/sql_store.go
ViktorT-11: Please see https://github.com/lightninglabs/lightning-terminal/pull/1051#issuecomment-2934569218

session/interface.go
```diff
@@ -329,5 +329,8 @@ type Store interface {
 	// "dest" state.
 	ShiftState(ctx context.Context, id ID, dest State) error
 
+	// Close closes the underlying store.
+	Close() error
+
```
ellemouton: i think you can drop this commit. 

The interface should not expose details about the underlying store if possible. This exposes that the underlying store is something that needs to be closed. It also gives things we pass the interface to access to this Close method which we dont want.

session/sql_migration.go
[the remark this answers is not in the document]
ellemouton: > Hmmm so the listSessions for the kvdb returns the sessions sorted by the CreatedAt field

I'd opt for not using public methods like ListAllSessions for the migration. Rather let the migration manually get to the buckets/sub-buckets that it needs. We want to be able to change how ListAllSessions works without affecting the migration.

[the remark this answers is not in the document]
ellemouton: >  it adding some more code complexity

yeah i think it's simpler and we have to worry about fewer things if we just directly iterate over the buckets.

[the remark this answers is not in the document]
ellemouton: now looking back, i think we should also change the accounts migration to not use the public Accounts method. Rather iterate directly over buckets.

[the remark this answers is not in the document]
ellemouton: in other words, rather than passing in a BoltStore, just pass in a *bbolt.DB instead (ie, the equivalent to SQLQueries)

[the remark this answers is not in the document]
ellemouton: notice that the LND invoice migration does the same: takes a raw kvdb.Backend and *sqlc.Queries

session/sql_migration.go
```diff
@@ -0,0 +1,361 @@
...
+			return fmt.Errorf("unable to get migrated session "+
+				"from sql store: %w", err)
+		}
+
+		migratedSession, err := unmarshalSession(ctx, tx, sqlSess)
+		if err != nil {
+			return fmt.Errorf("unable to unmarshal migrated "+
+				"session: %w", err)
+		}
+
```
ellemouton, line 146: perhaps it was on the other PR that i commented this but i think we should first complete the whole migration before doing the validation no?

ellemouton: ah, i see invoices was done like this too... ok all good

session/sql_migration.go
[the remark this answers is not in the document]
ellemouton: im also more in favour of doing it this way cause then the IDs that the sessions have will be more intuitive as then (as long as the user didnt do anything strange) , the created at times will mostly be increasing at the IDs increase

ellemouton (commented):
shweet! a few more comments. 

Agreed with you that the override macaroon recipe approach is good 👍  i think i didnt understand correctly before that for bbolt we could still have the TLV actually be written with out any contents (right?) so yes, the overwrite makes sense rather than the "un-set"

session/interface.go
ellemouton: ah, i think you're doing it cause of test cleanup - rather just call t.Cleanup sooner?

session/sql_migration.go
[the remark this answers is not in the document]
ViktorT-11: Cool thanks! I updated this PR to instead pass a `*bbolt.DB` and directly iterate over the buckets, rather than using the public functions of the `BoltStore`. I've done the same for the accounts migration as well now with:
https://github.com/lightninglabs/lightning-terminal/pull/1084

I've also looked through the previous migrations of the sessions kvdb as discussed offline, and none of them should be able to effect the order of the sessions entires in the kvdb store. However as also discussed offline, I didn't update the logic in the migrations, so that non-linked sessions are still migrated prior to linked sessions. I'm open to changing this though if you believe it's worth updating.

session/interface.go
ViktorT-11: Cool thanks for the idea :pray:! Updated with a commit that calls `t.Cleanup` sooner instead.

ViktorTigerstrom:
Also, re-requested a review from @bitromortac, as quite much has changed since your last review. Hope that's ok!

accounts/sql_migration_test.go
```diff
@@ -39,9 +39,6 @@ func TestAccountStoreMigration(t *testing.T) {
 		*db.TransactionExecutor[SQLQueries]) {
```
ellemouton: fwiw for future, I dont think these 2 account commits needed to be in this PR as i dont think there is a dependency here. These can be reviewed and merged in parallel in whichever order

session/sql_migration.go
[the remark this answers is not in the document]
ellemouton: > However as also discussed offline, I didn't update the logic in the migrations, so that non-linked sessions are still migrated prior to linked sessions. I'm open to changing this though if you believe it's worth updating.

ah, i thought the offline conclusion was the opposite 😅 cause i think it is just nice to have that the order in the DB is (at least on a best effort basis) the same as the order of creation.
Not completely blocking but i'd defs prefer it.

ellemouton (approved):
lgtm! 
great work

session/sql_migration.go
```diff
@@ -0,0 +1,396 @@
...
+// as nil in the bbolt store. Therefore, we also override the permissions
+// or caveats to nil for the migrated session in that scenario, so that the
+// deep equals check does not fail in this scenario either.
+func overrideMacaroonRecipe(kvSession *Session, migratedSession *Session) {
+	if kvSession.MacaroonRecipe != nil {
+		kvPerms := kvSession.MacaroonRecipe.Permissions
+		kvCaveats := kvSession.MacaroonRecipe.Caveats
+
+		if kvPerms == nil && kvCaveats == nil {
+			migratedSession.MacaroonRecipe = &MacaroonRecipe{}
```
bitromortac, line 389: In `GetSession` we use
```
	var macRecipe *MacaroonRecipe
	if perms != nil || caveats != nil {
		macRecipe = &MacaroonRecipe{
			Permissions: unmarshalMacPerms(perms),
			Caveats:     unmarshalMacCaveats(caveats),
		}
	}
```
but couldn't we just leave away the nil checks and `unmarshalX` nevertheless? this at least would give a bit more consistency (besides the empty slice vs nil slice behavior)

session/sql_migration_test.go
```diff
@@ -379,3 +387,390 @@ func TestSessionsStoreMigration(t *testing.T) {
...
+	var (
+		// numberOfSessions is set to 100 to add enough sessions to get
+		// enough variation between randomized sessions, but kept low
+		// enough for the test not take too long to run, as the test
+		// time increases drastically by the number of sessions we
+		// migrate.
+		numberOfSessions = 100
+	)
+
+	for i := 0; i < numberOfSessions; i++ {
```
bitromortac: nit: there are a few occurrences here and afterwards where the new loop syntax could be used: `	for i := range numberOfSessions {`

bitromortac (approved):
Nice, LGTM :tada:, just a question and a nit.

session/sql_migration.go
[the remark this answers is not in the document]
ViktorT-11: Sorry I should have been aware of this earlier! Since we insert the sessions in the kvdb store like this into the bucket:


```
if err := SerializeSession(&buf, session); err != nil {
	return err
}

return bucket.Put(getSessionKey(session), buf.Bytes())
```

Where the bucket is the bucket for the `sessionBucketKey` and the result of the `getSessionKey` will be `session.LocalPublicKey.SerializeCompressed()`, the order the sessions when iterating over the `sessionBucketKey` will actually be the lexicographical order of the serialized `LocalPublicKey`, and not in insertion order of the sessions.

Therefore there's no way to iterate over them in "insertion order", unless you know some way I'm unaware of :)? The closest we can get is to do what the kvdb store `listSessions` function does, which sorts the sessions by the `CreatedAt` timestamp, but that has the issue of what I brought up earlier with: https://github.com/lightninglabs/lightning-terminal/pull/1051#discussion_r2122411183

Therefore I think our best option that doesn't over complicate the code is to keep the migration code as what I've implemented today, i.e. first insert non-linked sessions, followed by inserting linked sessions?

session/sql_migration_test.go
ViktorT-11: Thanks addressed :)!

session/sql_migration.go
ViktorT-11: I think that would come down to a decision of how we'd want to represent that there is no `MacaroonRecipe` stored for the session in the SQL db. If we went with your suggestion, we'd get the behaviour that the session would always have a `MacaroonRecipe` set, even if had no `perms` & `caveats` set for the SQL version of the `session`, as those would then be empty lists. Note that we'd still need the `overrideMacaroonRecipe` with that version, but with the specific nil check you've marked removed.

Given this comment though, I think we wanted to get the SQL version as `neat` and understandable as possible, and IMO, i think the most understandable version is to not have any `MacaroonRecipe` set for a session, if none is actually set during the creation of the session:
https://github.com/lightninglabs/lightning-terminal/pull/1051#discussion_r2111840035 

Open to change this though if you both feel differently, but I suggest we'd discuss this and do it in a follow-up PR if so, to not make this PR stall on that discussion :)?

accounts/sql_migration_test.go
ViktorT-11: Did it that way so this wouldn't get merged without closing the stores in the tests, but I agree that it wouldn't be strictly necessary for merging :)

ViktorTigerstrom:
Thanks for the reviews :tada::zap:!

PR #1052, opened 2025-04-30
cli: add group key support to channel related calls.

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1498.

Fixes https://github.com/lightninglabs/taproot-assets/issues/1497.
Fixes https://github.com/lightninglabs/taproot-assets/issues/1183.

@ZZiigguurraatt can you help test this PR please?

### DIFF
--- a/cmd/litcli/ln.go
+++ b/cmd/litcli/ln.go
@@ -7,9 +7,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"strconv"
+	"time"
 
-	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/rfq"
 	"github.com/lightninglabs/taproot-assets/rfqmath"
 	"github.com/lightninglabs/taproot-assets/rpcutils"
@@ -35,8 +34,9 @@ const (
 
 var lnCommands = []cli.Command{
 	{
-		Name:     "ln",
-		Usage:    "Interact with the Lightning Network.",
+		Name: "ln",
+		Usage: "Interact with Taproot Assets on the Lightning " +
+			"Network.",
 		Category: "Taproot Assets on LN",
 		Subcommands: []cli.Command{
 			fundChannelCommand,
@@ -83,8 +83,16 @@ var fundChannelCommand = cli.Command{
 				"channel.",
 		},
 		cli.StringFlag{
-			Name:  "asset_id",
-			Usage: "The asset ID to commit to the channel.",
+			Name: "asset_id",
+			Usage: "The asset ID to commit to the channel; " +
+				"cannot be used at the same time as " +
+				"--group_key",
+		},
+		cli.StringFlag{
+			Name: "group_key",
+			Usage: "The group key to use for selecting assets to " +
+				"commit to the channel; cannot be used at " +
+				"the same time as --asset_id",
 		},
 	},
 	Action: fundChannel,
@@ -106,9 +114,9 @@ func fundChannel(c *cli.Context) error {
 		return fmt.Errorf("error fetching assets: %w", err)
 	}
 
-	assetIDBytes, err := hex.DecodeString(c.String("asset_id"))
+	assetIDBytes, groupKeyBytes, err := parseAssetIdentifier(c)
 	if err != nil {
-		return fmt.Errorf("error hex decoding asset ID: %w", err)
+		return fmt.Errorf("unable to parse asset identifier: %w", err)
 	}
 
 	requestedAmount := c.Uint64("asset_amount")
@@ -125,7 +133,21 @@ func fundChannel(c *cli.Context) error {
 	assetFound := false
 	totalAmount := uint64(0)
 	for _, rpcAsset := range assets.Assets {
-		if !bytes.Equal(rpcAsset.AssetGenesis.AssetId, assetIDBytes) {
+		if len(assetIDBytes) > 0 && !bytes.Equal(
+			rpcAsset.AssetGenesis.AssetId, assetIDBytes,
+		) {
+
+			continue
+		}
+
+		if len(groupKeyBytes) > 0 && rpcAsset.AssetGroup == nil {
+			continue
+		}
+
+		if len(groupKeyBytes) > 0 && !bytes.Equal(
+			rpcAsset.AssetGroup.TweakedGroupKey, groupKeyBytes,
+		) {
+
 			continue
 		}
 
@@ -149,6 +171,7 @@ func fundChannel(c *cli.Context) error {
 		ctx, &tchrpc.FundChannelRequest{
 			AssetAmount:        requestedAmount,
 			AssetId:            assetIDBytes,
+			GroupKey:           groupKeyBytes,
 			PeerPubkey:         nodePubBytes,
 			FeeRateSatPerVbyte: uint32(c.Uint64("sat_per_vbyte")),
 			PushSat:            c.Int64("push_amt"),
@@ -167,7 +190,15 @@ var (
 	assetIDFlag = cli.StringFlag{
 		Name: "asset_id",
 		Usage: "the asset ID of the asset to use when sending " +
-			"payments with assets",
+			"payments with assets; cannot be used at the same " +
+			"time as --group_key",
+	}
+
+	groupKeyFlag = cli.StringFlag{
+		Name: "group_key",
+		Usage: "the group key of the asset to use when sending " +
+			"payments with assets; cannot be used at the same " +
+			"time as --asset_id",
 	}
 
 	assetAmountFlag = cli.Uint64Flag{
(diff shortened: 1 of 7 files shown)


ZZiigguurraatt:
> @ZZiigguurraatt can you help test this PR please?

Working on it.

ZZiigguurraatt:
For `litcli ln sendpayment`, why do we have the requirement `Note that this will only work in concert with the --keysend argument.`? This conflicts with the statement

```
One can either specify the full
  parameters of the payment, or just use a payment request which encodes
  all the payment details.

  If payment isn't manually specified, then only a payment request needs
  to be passed using the --pay_req argument.
```
.

Confused if this comment belongs here or in https://github.com/lightninglabs/taproot-assets/pull/1498 ?

ZZiigguurraatt:
`litcli ln sendpayment - Send a payment over Lightning, potentially using a mulit-asset channel as the first hop` needs multi spelled right.

ZZiigguurraatt:
Here we also need to mention group_key

https://github.com/lightninglabs/lightning-terminal/pull/1052/files#diff-f9c05a4a36f7d2dcc59718bda9351aeab75a35de48e367bcd08d8cbf1f4be8f7R666-R667

ZZiigguurraatt:
At https://github.com/lightninglabs/lightning-terminal/pull/1052/files#diff-f9c05a4a36f7d2dcc59718bda9351aeab75a35de48e367bcd08d8cbf1f4be8f7R37 you say `Interact with the Lightning Network.`, but I think we should be more specific so that when running `litcli ln` we know under `NAME` that this is about taproot assets without having to read through the description of each sub-command in detail.

ZZiigguurraatt:
In https://github.com/lightninglabs/lightning-terminal/pull/1052/files#diff-f9c05a4a36f7d2dcc59718bda9351aeab75a35de48e367bcd08d8cbf1f4be8f7R301-R302 you say `To send an multi-asset LN payment to a single hop, the --asset_id=X argument should be used`. This does not make any mention of `group_key`. Also, I'm a bit confused because with this command, I must specify either a `asset_id` or `group_key`, there is no other option, but the word choice leads me to believe that maybe I can send a non multi-asset payment with the command.

ZZiigguurraatt:
If all these things can be fixed, I think we can also close https://github.com/lightninglabs/taproot-assets/issues/1183 .

guggero:
@ZZiigguurraatt thanks for your feedback! I think I've addressed all your comments in the latest push.

ZZiigguurraatt:
> As an aside: if you go to the "Files changed" tab in a PR, then you can add comments to a specific line of a PR.

this did not work for lines out of range from the edits you made

ZZiigguurraatt:
> @ZZiigguurraatt thanks for your feedback! I think I've addressed all your comments in the latest push.

Okay, what do you think about this comment I made above "If I try `sendpayment --pay_req=` instead, I get `[litcli] the --keysend flag must be set.`", but `litcli ln payinvoice -h` says `This command is a shortcut for 'sendpayment --pay_req='`, so I feel like we are getting some conflicting information.

guggero:
> > @ZZiigguurraatt thanks for your feedback! I think I've addressed all your comments in the latest push.
> 
> Okay, what do you think about this comment I made above "If I try `sendpayment --pay_req=` instead, I get `[litcli] the --keysend flag must be set.`", but `litcli ln payinvoice -h` says `This command is a shortcut for 'sendpayment --pay_req='`, so I feel like we are getting some conflicting information. This is also related to [#1052 (comment)](https://github.com/lightninglabs/lightning-terminal/pull/1052#issuecomment-2847413523)

I didn't see that comment anywhere before...
But I'll look into it.

ZZiigguurraatt (changes requested):

ZZiigguurraatt:
> I didn't see that comment anywhere before...

actually, maybe you don't get them unless I click "submit review"? I thought that meant I was satisfied with changes, but looks like I actually have a few options when I submit.

cmd/litcli/ln.go
```diff
@@ -437,17 +453,18 @@ var payInvoiceCommand = cli.Command{
 	This command attempts to pay an invoice using an asset channel as the
 	source of the payment. The asset ID of the channel must be specified
 	using the --asset_id flag.
+
+	This command is a shortcut for 'sendpayment --pay_req='.
```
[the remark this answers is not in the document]
guggero: Yeah, that was an incorrect check. Fixed that.

guggero:
> > I didn't see that comment anywhere before...
> 
> actually, maybe you don't get them unless I click "submit review"? I thought that meant I was satisfied with changes, but looks like I actually have a few options when I submit.

Yeah, as long as it says "pending", only you see the comment.

cmd/litcli/ln.go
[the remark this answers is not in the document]
guggero: Okay, that didn't work. See main comment at the bottom.

guggero:
Okay, so the attempt at making `This command is a shortcut for 'sendpayment --pay_req=` work made me realize that we never intended the `sendpayment` command to work in any other way than with `--keysend`.

This whole confusion came out of the attempt to re-use a lot of the description and also validation logic from `lnd`. But that just made it very hard to understand.

So I did this instead:
 - Remove the `This command is a shortcut for 'sendpayment --pay_req=` part from the description of `payinvoice`.
 - Updated the description of `sendpayment` to make it very explicit the command is _only_ for keysend payments.
 - I considered renaming the command to `keysend` and removing the need for the `--keysend` flag. But that would break existing users/scripts and would lead to a bunch of duplicated code. So I decided to leave things this way. It's a bit verbose but at least the description should be more clear now.

I think with that things should be way more clear. I hope you agree with the changes, @ZZiigguurraatt.

cmd/litcli/ln.go
```diff
@@ -280,20 +297,94 @@ func (w *resultStreamWrapper) Recv() (*lnrpc.Payment, error) {
...
 
-	Note that this will only work in concert with the --keysend argument.
+	Asset keysend payments are only supported to be sent to direct peers.
+	Multi-hop asset payments must be sent using invoices and the
+	corresponding 'ln payinvoice' subcommand.
+
+	To send a multi-asset LN keysend payment, the --asset_id=X or
+	--group_key=X argument can be used to specify the asset to use.
+
+	Note that this command will only work with the --keysend argument set.
```
gijswijs, line 328: :+1: for the B1-level writing.

cmd/litcli/ln.go
```diff
@@ -322,23 +413,15 @@ func sendPayment(cliCtx *cli.Context) error {
...
 
-	assetIDStr := cliCtx.String(assetIDFlag.Name)
-	assetIDBytes, err := hex.DecodeString(assetIDStr)
+	assetIDBytes, groupKeyBytes, err := parseAssetIdentifier(cliCtx)
 	if err != nil {
-		return fmt.Errorf("unable to decode assetID: %v", err)
-	}
-
-	assetAmountToSend := cliCtx.Uint64(assetAmountFlag.Name)
-	if assetAmountToSend == 0 {
```
gijswijs, line 340: We don't need to check for zero amounts? Or is this offloaded to tapd?

guggero:
> `[litcli] asset with ID  not found or no combined UTXOs with at least amount 100000 is available`

Oops, my bad, sorry. Fixed that issue and tested the flow manually as well this time.

itest/litd_custom_channels_test.go
```diff
@@ -4320,12 +4320,10 @@ func testCustomChannelsDecodeAssetInvoice(ctx context.Context,
 
 	// Now that we have our payment request, we'll call into the new decode
 	// asset pay req call.
-	decodeResp, err := aliceTap.DecodeAssetPayReq(
-		ctx, &tchrpc.AssetPayReq{
-			AssetId:      assetID,
-			PayReqString: payReq,
-		},
-	)
+	decodeResp, err := aliceTap.DecodeAssetPayReq(ctx, &tchrpc.AssetPayReq{
```
gijswijs, line 4323: This doesn't follow our dev guidelines. When wrapping long function calls, all arguments should begin in a new line after the open parenthesis and the close parenthesis should be placed on its own line. This is also true for the `alice.AddInvoice` function call above, and the `aliceTap.DecodeAssetPayReq` call below.

gijswijs (commented):
Made some comments throughout. The most important one is this one: https://github.com/lightninglabs/lightning-terminal/pull/1052#discussion_r2079565909

The others are all nits.

I tested the itest locally.

cmd/litcli/ln.go
```diff
@@ -726,3 +786,40 @@ func decodeAssetInvoice(cli *cli.Context) error {
...
+
+	if cli.IsSet("group_key") {
+		groupKeyBytes, err = hex.DecodeString(cli.String("group_key"))
+		if err != nil {
+			return nil, nil, fmt.Errorf("error hex decoding group "+
+				"key: %w", err)
+		}
+	}
+
+	if len(assetIDBytes) == 0 && len(groupKeyBytes) == 0 {
```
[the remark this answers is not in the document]
guggero: Ah yes, great catch! Fixed.

cmd/litcli/ln.go
guggero: Hmm, you're right. The removal of this was left over from a version where I thought we'd also want the `sendpayment` command to work with invoices, which I've rolled back in the meantime (see https://github.com/lightninglabs/lightning-terminal/pull/1052#issuecomment-2858052649). So this check is still needed, added it back.

itest/litd_custom_channels_test.go
guggero: Hmm, you're right, it's not documented in https://github.com/lightningnetwork/lnd/blob/master/docs/development_guidelines.md#wrapping-long-function-calls.
But it's an implicit style that we're using a lot in `lnd`, with the rule being that if you start a new code block with either `{}` or `()`, then you still get the visual guidance of where the call ends from the closing parentheses.
And we can avoid a level of indentation that way.

For example:
```
	invoiceResp2, err := fabia.AddInvoice(ctx, &lnrpc.Invoice{
		Memo:      "too small invoice",
		ValueMsat: int64(oneUnitMilliSat - 1),
		RouteHints: []*lnrpc.RouteHint{{
			HopHints: []*lnrpc.HopHint{hopHint},
		}},
	})
```

or
```
	payInvoiceWithSatoshi(t.t, dave, invoiceResp2, withFailure(
		lnrpc.Payment_FAILED, failureNoRoute,
	))
```

But yeah, as the person always ranting about code style, I can't use undocumented styles.
What would you prefer in this case? Revert the changes or add this "exception" to the `lnd` formatting guidelines?

gijswijs: I'm in favor of avoiding indentation, so I would prefer it to be added to the `lnd` formatting guidelines.

I was just so happy to catch you on a coding style nit. :smile_cat:

gijswijs (approved):
tACK

ZZiigguurraatt:
OK, things took longer on my end with some stupid mistakes managing multiple in flight PR that need to be coordinated on top of each other. I'm now able to get it to work, but I get the following error if I don't specify a fee rate

```
4bb544a08a1e:/$ litcli ln fundchannel --node_key  03d344679dc9a285183618a3eb1be0c897c9e6833f5f8c80513dd6534706e9ed59 --asset_amount 1000000 --group_key  02220e67ea8260deefdda3d514de83e2634ac0611e8133af21855ce43981dc598c
[litcli] error funding channel: rpc error: code = Unknown desc = fee rate must be specified
4bb544a08a1e:/$
```

however,

```
4bb544a08a1e:/$ litcli ln fundchannel -h
NAME:
   litcli ln fundchannel - Open a Taproot Asset channel with a node on the Lightning Network.

USAGE:
   litcli ln fundchannel [command options] [arguments...]

CATEGORY:
   Channels

OPTIONS:
   --node_key value       the identity public key of the target node/peer serialized in compressed format, must already be connected to
   --sat_per_vbyte value  (optional) a manual fee expressed in sat/vByte that should be used when crafting the transaction (default: 0)
   --push_amt value       the number of satoshis to give the remote side as part of the initial commitment state, this is equivalent to first opening a channel and then sending the remote party funds, but done all in one step; therefore, this is equivalent to a donation to the remote party, unless they reimburse the funds in another way (outside the protocol) (default: 0)
   --asset_amount value   The amount of the asset to commit to the channel. (default: 0)
   --asset_id value       The asset ID to commit to the channel; cannot be used at the same time as --group_key
   --group_key value      The group key to use for selecting assets to commit to the channel; cannot be used at the same time as --asset_id
   
4bb544a08a1e:/$ 
```
tells me the fee rate should be optional.

I don't remember having to do this last time I ran `lncli openchannel`.

Also, is `(default: 0)` supposed to actually be `0` or some auto fee selection (both in `litcli ln fundchannel` and `lncli openchannel`?


I did try

```
4bb544a08a1e:/$ litcli ln fundchannel --node_key  03d344679dc9a285183618a3eb1be0c897c9e6833f5f8c80513dd6534706e9ed59 --asset_amount 1000000 --group_key  02220e67ea8260deefdda3d514de83e2634ac0611e8133af21855ce43981dc598c --sat_per_vbyte 1
[litcli] error funding channel: rpc error: code = Unknown desc = error funding channel: fee rate 250 sat/kw too low, min_relay_fee: 253 sat/kw
4bb544a08a1e:/$ 
```

and

```
4bb544a08a1e:/$ litcli ln fundchannel --node_key  03d344679dc9a285183618a3eb1be0c897c9e6833f5f8c80513dd6534706e9ed59 --asset_amount 1000000 --group_key  02220e67ea8260deefdda3d514de83e2634ac0611e8133af21855ce43981dc598c --sat_per_vbyte 0
[litcli] error funding channel: rpc error: code = Unknown desc = fee rate must be specified
4bb544a08a1e:/$ 
```

and then finally what worked was 

```
4bb544a08a1e:/$ litcli ln fundchannel --node_key  03d344679dc9a285183618a3eb1be0c897c9e6833f5f8c80513dd6534706e9ed59 --asset_amount 1000000 --group_key  02220e67ea8260deefdda3d514de83e2634ac0611e8133af21855ce43981dc598c --sat_per_vbyte 3
{
	"txid": "17522c5c0384ed7f8ea7ebc76c402f1df768a003f97ca7a3fe5545af042584f2"
}
4bb544a08a1e:/$ 
```

ZZiigguurraatt:
I am still going through some other tests, so I have not approved this yet.

ZZiigguurraatt:
> I don't remember having to do this last time I ran `lncli openchannel`.

I just re-tested and this works:

```
4bb544a08a1e:/$ lncli openchannel --node_key 03d344679dc9a285183618a3eb1be0c897c9e6833f5f8c80513dd6534706e9ed59 --local_amt 5000000
{
    "funding_txid": "29ab2aebc847708433b1b0d879fa2308b6e81a6c6d7c12ea97642e60bea357d3"
}
4bb544a08a1e:/$ 
```

itest/litd_custom_channels_test.go
guggero: Cool, see https://github.com/lightningnetwork/lnd/pull/9799.

guggero:
> I don't remember having to do this last time I ran lncli openchannel.

I think, starting with lnd 18 or 19 it's now required in `lnd` as well. Depending on a (potentially faulty) estimated fee has given us too many user support issues.

> Also, is (default: 0) supposed to actually be 0

This is an unfortunate way the CLI library displays things. A numeric value of 0 just means "by default this flag is not set". So no, there is no auto fee estimation, you have to specify a value.

> `error funding channel: fee rate 250 sat/kw too low, min_relay_fee: 253 sat/kw`

I did fix this issue with an additional commit in https://github.com/lightninglabs/taproot-assets/pull/1517, so you should be able to specify `--sat_per_vbyte 1` now.

ZZiigguurraatt (approved):

bitcoin-coder-bob:
This line: https://github.com/lightninglabs/lightning-terminal/pull/1052/files#diff-f9c05a4a36f7d2dcc59718bda9351aeab75a35de48e367bcd08d8cbf1f4be8f7L294

It removes some of the useful payment flags, for example the `data` flag. Was this deliberate? I would like to be able to still attach custom data to these asset payments.

guggero:
> This line: https://github.com/lightninglabs/lightning-terminal/pull/1052/files#diff-f9c05a4a36f7d2dcc59718bda9351aeab75a35de48e367bcd08d8cbf1f4be8f7L294
> 
> It removes some of the useful payment flags, for example the `data` flag. Was this deliberate? I would like to be able to still attach custom data to these asset payments.

Not on purpose, really. Wanted to remove some flags that didn't make sense in the keysend scenario, but apparently removed a few too much. Added `--allow_self_payment` and `--data` back in an upcoming PR.
Any other flag you feel is missing?

bitcoin-coder-bob:
@guggero those 2 are the only ones I would personally need that I can think of. thank you!

PR #1057, opened 2025-05-02
Refactor litclient and perms to reduce LNC dependencies

ffranr:
Context: https://github.com/lightninglabs/lightning-node-connect/issues/113

---

- Extracted `perms` into a standalone module to limit `lnc` dependencies.  
- Moved `jsoncallbacks.go` into `litrpc` to reduce `lnc` import scope.
- Bumped `lnd` version in `litrpc` to match parent module.


bhandras (approved):
LGTM, very nice! 🎉 (needs release notes to pass CI)

perms/go.mod
```diff
@@ -0,0 +1,190 @@
+module github.com/lightninglabs/lightning-terminal/perms
```
guggero, line 1: Another option would be to just move this into the `litrpc` package, to avoid needing yet another submodule. But not sure if it fits there 100% content/topic wise. So just a thought.

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/lightninglabs/lightning-node-connect v0.3.3-alpha.0.20250306111457-cad4234830cc
 	github.com/lightninglabs/lightning-terminal/autopilotserverrpc v0.0.2
 	github.com/lightninglabs/lightning-terminal/litrpc v1.0.1
+	github.com/lightninglabs/lightning-terminal/perms v1.0.1
```
guggero, line 24: This should probably be `v1.0.0`, as it's the first version. I only needed to bump `litrpc` to `v1.0.1` because the go sum proxy somehow cached that the `v1.0.0` didn't exist yet (probably just needed to wait a bit but was impatient).

guggero (approved):
Nice, LGTM :tada:

perms/go.mod
guggero: Okay, chatted with @ellemouton offline to gauge opinions on this. And the decision was to keep `perms` as its separate module, so you can ignore my comment above :+1:

PR #1065, opened 2025-05-07
[sql-34]: actions: some basic clean-up of the firewall actions code

ellemouton:
This PR adds some initial clean up of the actions code in preparation for adding a SQL implementation. 

Part of #917


rules/config.go
```diff
@@ -48,7 +48,7 @@ type ConfigImpl struct {
 
 	// ActionsDB can be used by rules to list any past actions that were
 	// made for the specific session or feature.
-	ActionsDB firewalldb.ActionsDB
+	ActionsDB firewalldb.ActionsListDB
```
ViktorTigerstrom, line 51: nit: Could potentially rename the `ActionsDB` field in the config as well as the `GetActionsDB` in the interface as it's publicly exposed? I don't know what your upcoming commit that will use the `ActionsDB` name to represent the full Actions DB looks like though.

ViktorTigerstrom (approved):
Nice, LGTM :fire::rocket:!

rules/config.go
bitromortac: Naming it `ActionsLister` could also be an idiomatic way to name it (but I'm fine with the name).

bitromortac (approved):

rules/config.go
ellemouton: cool - gonna leave rename till the end of the series if that's ok! just realised it causes quite a lot of merge conflicts if i change this now

PR #1066, opened 2025-05-07
mod+itest: make custom channel itest flakes less annoying

guggero:
Runs the litd/terminal only itests separately from the custom channel itests, so flakes can be seen more clearly and ignored if not relevant.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -279,11 +279,14 @@ jobs:
       matrix:
         include:
           - name: bbolt
-            args: dbbackend=bbolt
+            args: itest icase=terminal dbbackend=bbolt
           - name: sqlite
-            args: dbbackend=sqlite
+            args: itest icase=terminal dbbackend=sqlite
           - name: postgres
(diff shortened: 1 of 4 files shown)


ellemouton (approved):
nice! thank you!

ViktorTigerstrom (approved):
Awesome thanks, LGTM code wise :rocket::fire:!

Though, I think we need to adjust the "Required Status Checks" before merging this, so that it reflects the new CI jobs. Currently we still get extra pending CI "duplicate jobs":
```
integration test (bbolt, dbbackend=bbolt)
integration test (postgres, dbbackend=postgres)
integration test (sqlite, dbbackend=sqlite)
```

As pending checks in the CI job list, as the required checks don't reflect the new updates. I think I lack the required permissions to change that though, so maybe you could do that before merging this @guggero :)?

guggero:
@ViktorTigerstrom yes, that's correct. I'll need to update them. But only jobs on `master` can be chosen, so this has to be bypass-merged and then I can update the rules.

PR #1070, opened 2025-05-13
[sql-37] firewalldb: more preparations for SQL actions store

ellemouton:
This PR mainly does some refactors in preparation for adding the SQL version of the Actions store. 

- We adjust the `AddActions` method so that it takes an `AddActionsReq` instead of an `Action`. This is more idiomatic since `Action` includes fields that should not be required by the caller to set. 
- We rename the `Action.SessionID` field to `MacaroonIdentifier` to more clearly describe what it actually is and then add an optional `Session.ID` which is meant to represent an actual session the action is linked to. For our bbolt impl: we continue to only read from the `action.MacaroonIdentifier` at persistence time (as we did before) and then at read time, we do a best effort attempt at populating the new `SessionID` field. 

Part of #917


firewalldb/actions.go
```diff
@@ -32,10 +33,17 @@ const (
...
+	// bytes of the root key ID of the macaroon used to perform the action.
+	MacaroonIdentifier [4]byte
+
+	// SessionID holds the optional session ID of the session that this
+	// action was performed with.
+	//
+	// NOTE: for our BoltDB impl, this is not persisted in any way, and we
+	// populate it by casting the macaroon ID to a session.ID and so is not
+	// guaranteed to be linked to an existing session.
+	SessionID fn.Option[session.ID]
```
bitromortac, line 46: I wonder if there's a test anywhere for an intercepted action, where we don't set this (i.e. it's not related to a session request).

firewalldb/actions_test.go
```diff
@@ -171,7 +174,7 @@ func TestListActions(t *testing.T) {
 		actionIds++
 
 		actionReq := &AddActionReq{
-			SessionID:          sessionID,
+			MacaroonIdentifier: sessionID,
```
bitromortac, line 180: should we set `SessionID` here as well?

firewalldb/actions_test.go
```diff
@@ -14,11 +14,23 @@ import (
 var (
 	testTime1 = time.Unix(32100, 0)
 	testTime2 = time.Unix(12300, 0)
+)
+
+// TestActionStorage tests that the ActionsListDB CRUD logic.
```
bitromortac, line 19: thank you for the nice diffs :+1:

bitromortac (approved):
LGTM :zap:

firewalldb/actions_kvdb.go
```diff
@@ -53,8 +54,14 @@ var (
...
 
 // AddAction serialises and adds an Action to the DB under the given sessionID.
-func (db *BoltDB) AddAction(_ context.Context, action *Action) (ActionLocator,
-	error) {
+func (db *BoltDB) AddAction(_ context.Context,
+	req *AddActionReq) (ActionLocator, error) {
+
+	action := &Action{
+		AddActionReq: *req,
+		AttemptedAt:  db.clock.Now().UTC(),
```
ViktorTigerstrom, line 62: Just noting that the time is set to the clocks `.Now().UTC()`, while the previous version set it to `time.Now()` (i.e. not UTC). I think the new implementation is probably more correct, but wanted to note that. Could this cause any issues, anddo we need to update previous inserted actions to represent that in some way?

ViktorTigerstrom (approved):
LGMT given that my comment isn't an issue :rocket:!

firewalldb/actions_kvdb.go
ellemouton: yeah - similarly to sessions/accounts - we're going with "it may just be off by a couple of hours and that's ok" since mostly these timestamps are just metadata for display to the user - we dont use them (much) in logic.,

firewalldb/actions.go
ellemouton: yeah i'll expand on the tests for SQL (where it actually will matter if it is set or not) once we've added the sql impl here 👍

firewalldb/actions.go
```diff
@@ -32,10 +33,17 @@ const (
 // It contains all the information that is needed to create a new Action in the
 // ActionStateInit State.
 type AddActionReq struct {
-	// SessionID is the ID of the session that this action belongs to.
-	// Note that this is not serialized on persistence since the action is
-	// already stored under a bucket identified by the session ID.
-	SessionID session.ID
+	// MacaroonIdentifier is a 4 byte identifier created from the last 4
+	// bytes of the root key ID of the macaroon used to perform the action.
+	MacaroonIdentifier [4]byte
```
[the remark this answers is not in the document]
ellemouton: currently we dont use this in other places - so i think it's ok for now

firewalldb/actions_test.go
ellemouton: want to keep the tests as close to how they are today as possible - then will expand the tests to cover more cases once the SQL impl is added 👍

PR #1071, opened 2025-05-13
[sql-38] multi: couple sessions and accounts to actions

ellemouton:
In this PR, we add best effort logic to ensure tight coupling between actions and sessions&accounts at the
time of action creation. This forces us to update our tests to expect certain errors if linked sessions/accounts
do not yet exist at the time of action creation. This sets us up nicely for the SQL impl of the actions DB which will
enforce strict coupling.

Part of #917

### DIFF
--- a/firewalldb/actions_kvdb.go
+++ b/firewalldb/actions_kvdb.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/lightninglabs/lightning-terminal/accounts"
 	"github.com/lightninglabs/lightning-terminal/session"
 	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/tlv"
@@ -64,6 +65,17 @@ func (db *BoltDB) AddAction(ctx context.Context,
 	var err error
 	req.SessionID.WhenSome(func(id session.ID) {
 		_, err = db.sessionIDIndex.GetSession(ctx, id)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// If the new action links to an account, the account must exist.
+	// For the bbolt impl of the store, this is our best effort attempt
+	// at ensuring each action links to an account. If the account is
+	// deleted later on, however, then the action will still exist.
+	req.AccountID.WhenSome(func(id accounts.AccountID) {
+		_, err = db.accountsDB.Account(ctx, id)
 	})
 	if err != nil {
 		return nil, err
--- a/firewalldb/actions_kvdb.go
+++ b/firewalldb/actions_kvdb.go
@@ -54,8 +54,20 @@ var (
 )
 
 // AddAction serialises and adds an Action to the DB under the given sessionID.
-func (db *BoltDB) AddAction(_ context.Context,
+func (db *BoltDB) AddAction(ctx context.Context,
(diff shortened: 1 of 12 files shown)


firewalldb/actions_kvdb.go
```diff
@@ -54,9 +55,32 @@ var (
...
+	// For the bbolt impl of the store, this is our best effort attempt
+	// at ensuring each action links to a session. If the session is
+	// deleted later on, however, then the action will still exist.
+	var err error
+	req.SessionID.WhenSome(func(id session.ID) {
+		_, err = db.sessionIDIndex.GetSession(ctx, id)
+	})
+	if err != nil {
+		return nil, err
+	}
```
ViktorTigerstrom, line 71: nit: could potentially wrap the error on `session.ErrSessionNotFound` errors to make it clear that was a linked session for an action that wasn't found.

firewalldb/actions_kvdb.go
```diff
@@ -54,9 +55,32 @@ var (
...
+	// If the new action links to an account, the account must exist.
+	// For the bbolt impl of the store, this is our best effort attempt
+	// at ensuring each action links to an account. If the account is
+	// deleted later on, however, then the action will still exist.
+	req.AccountID.WhenSome(func(id accounts.AccountID) {
+		_, err = db.accountsDB.Account(ctx, id)
+	})
+	if err != nil {
+		return nil, err
+	}
```
ViktorTigerstrom, line 82: nit:
Same here but for `accounts.ErrAccNotFound`.

firewalldb/actions_test.go
```diff
@@ -440,6 +504,9 @@ func TestListGroupActions(t *testing.T) {
 }
 
 func assertEqualActions(t *testing.T, expected, got *Action) {
+	// Accounts are not explicitly linked in our bbolt DB implementation.
+	got.AccountID = expected.AccountID
```
ViktorTigerstrom, line 508: nit: Should we remove the overwritten value from `got` at the end of this function as this is a pointer we're passing?

ViktorTigerstrom (approved):
Nice, LGTM :rocket:! Only leaving a few nits which are optional.

Also: Could potentially add test coverage under "kvdb" builds that show that the `sessions` + `accounts` aren't tightly coupled. I think that'll be a quite nice coverage to have once the tightly coupling under "sql" builds exists, so that you can show that it's coupled there. Just to prove that this is expected behaviour. Potentially though, that's something that would make more sense to add coverage of in that PR.

bitromortac (approved):
LGTM :zap:

firewalldb/actions_kvdb.go
ellemouton: `GetSession` already handles that - no need to wrap again i'd say! This is proved by the unit tests asserting the error

firewalldb/actions_kvdb.go
ellemouton: already handled - see unit tests that assert this

firewalldb/actions_test.go
ellemouton: sure

ellemouton (commented):
> Potentially though, that's something that would make more sense to add coverage of in that PR.

yeah can add later

PR #1072, opened 2025-05-13
[sql39]: Actions schemas, queries and SQL store impl

ellemouton:
This is the final PR for part 1 of #917 🎉 

In this PR, we introduce the schema for the `actions` table. See the diagram in #917  for how this fits in with regard
to our other tables (namely sessions & accounts). 

We then implement the SQL version of the Actions store using the new queries. 

Finally, our `dev` build is updated to use the SQL version of the actions table.


db/sqlc/migrations/000005_actions.up.sql
```diff
@@ -0,0 +1,52 @@
...
+    macaroon_identifier BLOB,
+
+    -- The name of the entity who performed the action.
+    actor_name TEXT,
+
+    -- The name of the feature that the action is being performed by.
+    feature_name TEXT,
+
+    -- Meta info detailing what caused this action to be executed.
+    trigger TEXT,
```
ViktorTigerstrom, line 23: nit: trigger is a reserved keyword in SQL, so could potentially consider using another name for this column.

db/sqlc/queries/actions.sql
```diff
@@ -0,0 +1,81 @@
...
+    ) OR sqlc.narg('group_id') IS NULL
+  )
+ORDER BY
+    CASE WHEN @reversed = TRUE THEN a.created_at END DESC,
+    CASE WHEN NOT @reversed = TRUE THEN a.created_at END ASC
+    LIMIT @num_limit
+OFFSET @num_offset;
+
+-- name: ListActions :many
+SELECT a.*
```
ViktorTigerstrom, line 42: Hmmm I'm not sure that the indexes will end up getting with these queries, due to the `"OR IS NULL"` clauses for all of the index filter params. It requires some more investigation if it's the case or not, but if this is an issue, we should consider if it's worth creating separate queries that's being used when for example the `sessionID` is `not nil` at the call site. That way, we'd ensure that the indexes actually gets used.

Just for some reference, I copied the table + this query into GPT and asked if the indexes will gets used. I'm posting the response below:

---
Because each filter is written as

```sql
(column = :param OR :param IS NULL)
```

PostgreSQL’s planner can only turn that into an index‐based lookup when it can prove, at plan‐time, that `:param` is non‐NULL and that the condition simplifies to a simple `column = constant`.  As soon as you wrap it in an `OR … IS NULL`, you introduce uncertainty:

1. **For a single invocation with literal parameters**

   * If you prepare the statement and bind, say, `session_id = 42`, then

     ```sql
     (a.session_id = 42 OR 42 IS NULL)
     ```

     *is* semantically equivalent to `a.session_id = 42`, so **in principle** the `actions_session_id_idx` can be used for that clause.

2. **When the planner builds a “generic” plan**

   * PostgreSQL distinguishes *custom* plans (where it knows your parameter values at plan‐time) from *generic* plans (where it has to pick one plan for *all* future binds).
   * By default, after a few executions it switches to a generic plan, where it *doesn’t* know whether each `:param IS NULL` will turn on or off the filter.  In that case it almost always chooses a **sequential scan**, because it can’t assume you’ll eliminate most rows.

3. **Multiple optional filters multiply the uncertainty**

   * You have six of these `(… OR … IS NULL)` filters all ANDed together.  Even if one of them would by itself use an index, the planner sees the whole WHERE clause as “maybe trivial, maybe selective” and still tends to fall back to seq‐scan.

4. **Dynamic ordering makes it worse**

   * The `CASE WHEN :reversed … END DESC, CASE WHEN NOT :reversed … END ASC` trick also prevents an “index‐only scan in index‐order” on `created_at_idx`; PostgreSQL won’t match that to a simple `ORDER BY created_at ASC/DESC`.

---

### Bottom line

* **In custom plans** (i.e. if the planner *does* re-optimize on each bind, or if you force `plan_cache_mode = force_custom_plan`), then *when you supply non-NULL parameters* each `(col = constant OR constant IS NULL)` really does reduce to `col = constant`, and the corresponding single-column index **can** be used.

* **In generic plans** (the default after a few executions of a prepared statement), the planner assumes the worst (that your filters might all be turned off), so it chooses a **sequential scan** and then a regular sort.

---

## How to guarantee index usage

1. **Generate the SQL dynamically** so you only include `WHERE session_id = …` clauses for the parameters you actually passed.  No ORs, no uncertainty — the planner sees plain equality and happily uses the index.

2. **Use a stored procedure with `EXECUTE format()`**.  Inside PL/pgSQL you can build the WHERE list piecewise and then `EXECUTE` the resulting text; the final plan is always custom.

3. **Split into two code‐paths**:

   * One static SQL with *all* parameters NOT NULL (index usage guaranteed).
   * One static SQL with *no* filters (trivial `SELECT * … ORDER BY …`; index usage isn’t needed).

4. **Partial indexes or multicolumn indexes** won’t help here unless you know in advance which combinations of columns you’ll filter on.

---

### TL;DR

> **Will the indexes actually end up getting used?**
> **Not reliably**.  With the `(col = :param OR :param IS NULL)` pattern inside a prepared statement, Postgres will usually fall back to a sequential scan and a regular sort—unless you force it to rebuild a custom plan each time or rewrite the query to eliminate the `OR … IS NULL` out of your WHERE clause.

db/sqlc/migrations/000005_actions.up.sql
```diff
@@ -0,0 +1,52 @@
...
+    -- The session ID of the session that this action is associated with.
+    -- This may be null for actions that are not coupled to a session.
+    session_id BIGINT REFERENCES sessions(id) ON DELETE CASCADE,
+
+    -- The account ID of the account that this action is associated with.
+    -- This may be null for actions that are not coupled to an account.
+    account_id BIGINT REFERENCES accounts(id) ON DELETE CASCADE,
+
+    -- An ID derived from the macaroon used to perform the action.
+    macaroon_identifier BLOB,
```
ViktorTigerstrom, line 14: This could potentially be `BLOB NOT NULL` instead. We could potentially also add a constraint for the length of this column as well, since the unmarshaling of this requires exactly 4 bytes.
Note that even for actions where we don't set any macaroon, we pass a [4]byte array with each byte set to `0` bytes which will get written to this column. If this is not intentional, we should change this so that this gets set an empty array when being persisted, and then also handle that case during the unmarshaling.

firewalldb/actions_sql.go
```diff
@@ -0,0 +1,429 @@
...
+
+		actor          = sql.NullString{String: req.ActorName}
+		feature        = sql.NullString{String: req.FeatureName}
+		trigger        = sql.NullString{String: req.Trigger}
+		intent         = sql.NullString{String: req.Intent}
+		structuredJson = sql.NullString{String: req.StructuredJsonData}
+	)
+	if req.ActorName != "" {
+		actor.Valid = true
+	}
```
ViktorTigerstrom, line 83: Instead of these `if` cases here and below, we can instead just set:
`actor.Valid = opts.actorName != ""`

We can actually also just set during the `var` initialization, so that it's instead set as:
```
actor = sql.NullString{
	String: req.ActorName
	Valid: opts.actorName != ""
}
```

firewalldb/actions_sql.go
```diff
@@ -0,0 +1,429 @@
...
+		}
+		if opts.state != 0 {
+			actionState.Valid = true
+		}
+		if !opts.startTime.IsZero() {
+			startTime.Valid = true
+		}
+		if !opts.endTime.IsZero() {
+			endTime.Valid = true
+		}
```
ViktorTigerstrom, line 256: Same as above.

ViktorTigerstrom (commented):
Generally this looks great, really awesome work :fire::rocket:!!

Leaving some comments below, where I think the most substantial comment is in regards to the actual usage of the indexes.

db/sqlc/queries/actions.sql
ellemouton: ok so turns out: postgres correctly uses the index here, but sqlite does not - so good catch!

went with creating these queries manually instead

db/sqlc/migrations/000005_actions.up.sql
ellemouton: (edit: opening a prep PR)

ellemouton (commented):
Thanks @ViktorTigerstrom 🙏

bitromortac (commented):
Very cool and impressive work! So happy to see all schemas implemented and running against unit tests :tada:.

firewalldb/actions_sql.go
```diff
@@ -0,0 +1,418 @@
...
+	"github.com/lightninglabs/lightning-terminal/accounts"
+	"github.com/lightninglabs/lightning-terminal/db"
+	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightninglabs/lightning-terminal/session"
+	"github.com/lightningnetwork/lnd/fn"
+	"github.com/lightningnetwork/lnd/sqldb"
+)
+
+// SQLAccountQueries is a subset of the sqlc.Queries interface that can be used
+// to interact with the actions table.
```
bitromortac, line 19: should be `accounts table`

firewalldb/actions_sql.go
```diff
@@ -0,0 +1,418 @@
...
+		// Do best effort to see if this action is linked to a session,
+		// and/or an action or none.
+		var (
+			sessionID sql.NullInt64
+			accountID sql.NullInt64
+		)
+
+		// First check session DB.
+		var sessErr error
+		req.SessionID.WhenSome(func(id session.ID) {
```
bitromortac, line 102: nit: we could name the `id` `alias` here as well

firewalldb/actions_sql.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+
+	var writeTxOpts db.QueriesTxOptions
+	return s.db.ExecTx(ctx, &writeTxOpts, func(db SQLQueries) error {
+		return db.SetActionState(ctx, sqlc.SetActionStateParams{
+			ID:          locator.id,
+			ActionState: int16(state),
+			ErrorReason: sql.NullString{
+				String: errReason,
+				Valid:  errReason != "",
```
bitromortac, line 202: nit: we could set it `true` here, since that must be the case

bitromortac (approved):
LGTM :rocket:

db/sqlc/actions_custom.go
```diff
@@ -0,0 +1,212 @@
...
+		args = append(args, params.NumLimit, params.NumOffset)
+	}
+
+	return query, args
+}
+
+// fillPlaceHolders replaces all '?' placeholders in the SQL query with
+// positional placeholders like $1, $2, etc. This is necessary for
+// compatibility with Postgres.
+func fillPlaceHolders(query string) string {
```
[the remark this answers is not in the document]
bitromortac: yes true, was thinking that as well, not a blocker

db/sqlc/actions_custom.go
```diff
@@ -0,0 +1,212 @@
...
+type ActionQueryParams struct {
+	SessionID   sql.NullInt64
+	AccountID   sql.NullInt64
+	FeatureName sql.NullString
+	ActorName   sql.NullString
+	RpcMethod   sql.NullString
+	State       sql.NullInt16
+	EndTime     sql.NullTime
+	StartTime   sql.NullTime
+	GroupID     sql.NullInt64
```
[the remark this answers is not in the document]
bitromortac: I'm not sure every field is covered in unit tests, sqlc does all the renaming automatically if a field's name changes, but we would have to manually change the representations in the query and would only fail once we add a unit test. Was just an idea, but not a blocker :)

db/sqlc/actions_custom.go
```diff
@@ -0,0 +1,212 @@
...
+// sqlite.
+//
+// The 'count' param indicates whether the query should return a count of
+// actions that match the criteria or the actions themselves.
+func buildActionsQuery(params ActionQueryParams, count bool) (string,
+	[]interface{}) {
+
+	var (
+		conditions []string
+		args       []interface{}
```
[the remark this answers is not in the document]
ellemouton: def def

db/sqlc/actions_custom.go
```diff
@@ -0,0 +1,212 @@
...
+// fillPlaceHolders replaces all '?' placeholders in the SQL query with
+// positional placeholders like $1, $2, etc. This is necessary for
+// compatibility with Postgres.
+func fillPlaceHolders(query string) string {
+	var (
+		sb     strings.Builder
+		argNum = 1
+	)
+
+	for i := 0; i < len(query); i++ {
```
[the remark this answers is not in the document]
ellemouton: ooooo nice!

firewalldb/actions_sql.go
ellemouton: not true right? it is defs possible for SetAction to be called with an empty `errReason`

db/sqlc/actions_custom.go
```diff
@@ -0,0 +1,210 @@
+package sqlc
```
ViktorT-11, line 1: Very nice approach with the addition of this file :tada::fire:!

firewalldb/actions_sql.go
```diff
@@ -0,0 +1,418 @@
...
+// state to the given state.
+//
+// NOTE: This is a part of the ActionDB interface.
+func (s *SQLDB) SetActionState(ctx context.Context, al ActionLocator,
+	state ActionState, errReason string) error {
+
+	if errReason != "" && state != ActionStateError {
+		return fmt.Errorf("error reason should only be set for " +
+			"ActionStateError")
+	}
```
ViktorT-11, line 188: Nit: As a follow-up to @bitromortac's feedback below, we could potentially also check that an `errReason` is actually set when `state == ActionStateError`, though that may be a bit dangerous to require that the middleware interceptor sets an error for all responses marked as "errors". So therefore I think it's probably safer to leave it as is.

ViktorTigerstrom (approved):
Very nice changes!! Awesome job :tada::fire:

PR #1074, opened 2025-05-15
custom channels: add itest to demonstrate that backward compatibility with v0.14.1-alpha is restored

guggero:
Shows that https://github.com/lightninglabs/taproot-assets/pull/1537 restores backward compatibility.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,12 @@ PUBLIC_URL :=
 # versions are checked against this version.
 GO_VERSION = 1.23.6
 
+# LITD_COMPAT_VERSIONS is a space-separated list of litd versions that are
+# installed before running the integration tests which include backward
+# compatibility tests. The list of versions must be in sync with any version
+# used in the backwardCompat map in itest/litd_test_list_on_test.go.
+LITD_COMPAT_VERSIONS = v0.14.1-alpha
+
 LOOP_COMMIT := $(shell cat go.mod | \
 		grep $(LOOP_PKG) | \
 		head -n1 | \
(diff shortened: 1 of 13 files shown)


scripts/install-backward-compat-versions.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -ex
+
```
ffranr, line 4: Consider adding a top level doc comment to describe what this does, why it exists.

(I should have done that for the go version sh files.)

ffranr: Example:
```bash
#!/bin/bash

# Build and install older versions of `litd` for backward compatibility
# integration tests.
#
# Usage:
#   ./scripts/install-backward-compat-versions.sh "<version1> <version2> ..."
#
# Example:
#   ./scripts/install-backward-compat-versions.sh "v0.12.0-beta v0.14.1-alpha"
#
# Environment Variables:
#   VERSIONS  Space-separated list of `litd` versions to install (required).
#
# Notes:
#   - Each version will be installed under a version-specific path.
```

ffranr (approved):

Makefile
```diff
@@ -233,6 +233,9 @@ build-itest:
 	CGO_ENABLED=0 $(GOBUILD) -tags="$(ITEST_TAGS)" -o itest/btcd-itest -ldflags "$(ITEST_LDFLAGS)" $(BTCD_PKG)
 	CGO_ENABLED=0 $(GOBUILD) -tags="$(ITEST_TAGS)" -o itest/lnd-itest -ldflags "$(ITEST_LDFLAGS)" $(LND_PKG)/cmd/lnd
 
+	@$(call print, "Installing old versions of litd for backward compatibility tests.")
+	scripts/install-backward-compat-versions.sh 'v0.14.1-alpha'
```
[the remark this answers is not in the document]
guggero: Good idea, added. Also made a reference from the Makefile to the itest list and vice versa.

scripts/install-backward-compat-versions.sh
guggero: Great idea, added. Thanks!

scripts/install-backward-compat-versions.sh
```diff
@@ -0,0 +1,75 @@
...
+for VERSION in ${VERSIONS}; do
+  VERSION_DIR="${TARGET_DIR}/litd-${VERSION}"
+
+  echo "Building litd ${VERSION}..."
+  cd "${VERSION_DIR}"
+  
+  # The Makefile was changed a bit, but the "itest-only" target that skips the
+  # frontend build has existed for a while. We use it to build the itest binary
+  # (but don't actually run the tests).
+  make itest-only icase=nothing
```
gijswijs, line 67: Shouldn't the build below be skipped if we already have the binary in our folder?
Something like
```
 if [[ -f "${ITEST_DIR}/litd-itest-${VERSION}" ]]; then
    echo "Binary ${VERSION} is already available."
    continue
 fi
```

The build cache speeds up a build, but it's probably better to prevent them from happening if they are unneeded.

gijswijs (approved):
I'm surprised by how small this commit is! Great work.
Made some small remarks, but otherwise LGTM!

scripts/install-backward-compat-versions.sh
```diff
@@ -0,0 +1,75 @@
...
+# Environment Variables:
+#   REPO_URL  URL of the litd repository (defaults to:
+#             https://github.com/lightninglabs/lightning-terminal.git)
+#
+# Notes:
+#   - Each version will be installed under a version-specific path.
+
+set -ex
+
+VERSIONS=$1
```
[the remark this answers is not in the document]
guggero: Good point, fixed.

ellemouton:
WOW this is very cool!!!
cc @ViktorTigerstrom  - we could potentially use this in itests for testing the new migrations!

Issue #1073, opened 2025-05-13
[bug]: `litcli ln addinvoice --amt 1000` does not work


ZZiigguurraatt:
I'm getting 

```
c97deaf39c89:/$ litcli ln addinvoice --amt 1000
[litcli] asset_amount argument missing
c97deaf39c89:/$ 
```

but I should expect this to work based on https://github.com/lightninglabs/taproot-assets/pull/1448 .

PR #1075, opened 2025-05-16
Use sat/msat arguments of `ln addinvoice`

GeorgeTsagk:
## Description

Uses the existing LND arguments for sat/msat amount for the `litcli ln addinvoice` command. This new argument exposes the behavior introduced in https://github.com/lightninglabs/taproot-assets/pull/1448

Closes https://github.com/lightninglabs/lightning-terminal/issues/1073

### DIFF
--- a/cmd/litcli/ln.go
+++ b/cmd/litcli/ln.go
@@ -679,12 +679,23 @@ func addInvoice(cli *cli.Context) error {
 
 	var (
 		assetAmount = cli.Uint64("asset_amount")
+		msatAmount  = cli.Int64("amt_msat")
+		satAmount   = cli.Int64("amt")
 		preimage    []byte
 		descHash    []byte
 		err         error
 	)
-	if assetAmount == 0 {
-		return fmt.Errorf("asset_amount argument missing")
+	if assetAmount == 0 && msatAmount == 0 && satAmount == 0 {
+		return fmt.Errorf("must set asset amount or sat/msat amount")
+	}
+
+	if assetAmount != 0 && (msatAmount != 0 || satAmount != 0) {
+		return fmt.Errorf("must only set one of asset amount or " +
+			"sat/msat amount")
+	}
+
+	if msatAmount != 0 && satAmount != 0 {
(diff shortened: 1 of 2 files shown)


cmd/litcli/ln.go
```diff
@@ -659,8 +659,14 @@ var addInvoiceCommand = cli.Command{
...
 		},
 		cli.Uint64Flag{
-			Name:  "asset_amount",
-			Usage: "the amount of assets to receive",
+			Name: "asset_amount",
+			Usage: "the amount of assets to receive, mutually " +
+				"exclusive to msat_amount",
+		},
+		cli.Uint64Flag{
+			Name: "msat_amount",
```
guggero, line 667: If you look at `litcli ln addinvoice`, you see that we already have all flags from `commands.AddInvoiceCommand.Flags`:

```
   --amt value                      the amt of satoshis in this invoice (default: 0)
   --amt_msat value                 the amt of millisatoshis in this invoice (default: 0)
```

So we should use those.

guggero (changes requested):

cmd/litcli/ln.go
```diff
@@ -659,8 +659,14 @@ var addInvoiceCommand = cli.Command{
 				"cannot be used at the same time as --asset_id",
 		},
 		cli.Uint64Flag{
-			Name:  "asset_amount",
-			Usage: "the amount of assets to receive",
+			Name: "asset_amount",
+			Usage: "the amount of assets to receive, mutually " +
+				"exclusive to msat_amount",
+		},
+		cli.Uint64Flag{
```
guggero, line 666: The flag isn't needed.

guggero (approved):
tACK. While testing I found out that `--amt` doesn't work, only `--amt_msat`. But that needs to be fixed in `tapd`, which I'm going to do now.

guggero:
Needs a rebase to get the itests running.

guggero:
LOL, apparently if you use the "fixes" keyword and then link to another PR, that one is closed... Sorry, my bad.

PR #1077, opened 2025-05-21
Prepare for release of v0.15.0-alpha.rc1

guggero:


go.mod
```diff
@@ -245,4 +245,6 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 // automatically, so we need to add it manually.
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
 
+replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.19.0-beta
+
```
ellemouton, line 249: why the replace instead of direct update?

ellemouton (approved):
lgtm! just one question

ellemouton (changes requested):
oh wait! almost forgot - i think we need to please update the minimum LND version here

go.mod
guggero: Because most of the projects still reference the RC3. If I don't use the replace, `go mod tidy` thinks RC3 is equally good and uses that instead. We could do a PR to each project and bump from RC3 to final to avoid this. But I thought about doing this between RC1 and the final version.

guggero:
> oh wait! almost forgot - i think we need to please update the minimum LND version here to v0.19.0.

Ah, I did that in the README but forgot to also bump it in the code. Will do that asap.

go.mod
ellemouton: gotcha! thanks

ellemouton (approved):
awesome, thanks!

go.mod
```diff
@@ -22,23 +22,23 @@ require (
...
 	github.com/lightninglabs/lightning-terminal/perms v1.0.0
-	github.com/lightninglabs/lndclient v0.19.0-4
-	github.com/lightninglabs/loop v0.31.0-beta.0.20250502080242-8ef7584f9f56
+	github.com/lightninglabs/lndclient v0.19.0-5
+	github.com/lightninglabs/loop v0.31.1-beta
 	github.com/lightninglabs/loop/looprpc v1.0.7
 	github.com/lightninglabs/loop/swapserverrpc v1.0.14
 	github.com/lightninglabs/pool v0.6.5-beta.0.20250305125211-4e860ec4e77f
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.3-0.20250305125211-4e860ec4e77f
 	github.com/lightninglabs/pool/poolrpc v1.0.1-0.20250305125211-4e860ec4e77f
```
ViktorTigerstrom, line 31: Should we update this to use the poolrpc `v1.0.1` tag?

go.mod
```diff
@@ -22,23 +22,23 @@ require (
...
-	github.com/lightninglabs/taproot-assets v0.6.0-rc1.0.20250520102102-f6baa2c6f739
-	github.com/lightninglabs/taproot-assets/taprpc v1.0.3
+	github.com/lightninglabs/taproot-assets v0.6.0-rc2
+	github.com/lightninglabs/taproot-assets/taprpc v1.0.4
```
ViktorTigerstrom, line 33: Could be worth adding this `taprpc` to the `litrpc` package as well.

go.mod
```diff
@@ -245,4 +245,6 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 // automatically, so we need to add it manually.
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
 
-go 1.23.6
+replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.19.0-beta
+
+go 1.23.9
```
ViktorTigerstrom, line 250: Could bump golang in the `litrpc`, `autopilotserverrpc`, `perms` & `tools` packages as well.

go.mod
```diff
@@ -245,4 +245,6 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 // automatically, so we need to add it manually.
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
 
-go 1.23.6
+replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.19.0-beta
```
ViktorTigerstrom, line 248: Could add this replace for the `litrpc` package as well :).

ViktorTigerstrom (commented):
tACK :rocket:! Thanks a lot :pray:.

Leaving a few suggestions, especially around also updating the `litrpc` package as well. If you'd prefer to do that in a separate PR and push a tag for that separately, let me know and I'll proceed to approve this :)

go.mod
guggero: It won't do anything for us. Since replaces aren't transitive (meaning it wouldn't affect any consumers of the module) and the `litrpc` itself doesn't need the latest version by itself. But I'll fix this for the final release anyway (by bumping the lnd version in each project).

ViktorTigerstrom: Ah good point! Makes sense, thanks :pray:!

ViktorTigerstrom (approved):
Thanks for the changes, LGTM :tada:!

PR #1079, opened 2025-05-27
[sql-41] firewalldb: add migration code for kvstores from kvdb to SQL

ViktorTigerstrom:
Based on #1051 

This PR introduces the migration logic for transitioning the kv stores from kvdb to SQL.

Once #1051 is merged, I will rebase this PR and request reviews.

### DIFF
--- a/db/sqlc/queries/kvstores.sql
+++ b/db/sqlc/queries/kvstores.sql
@@ -21,7 +21,7 @@ FROM features
 WHERE name = sqlc.arg('name');
 
 -- name: InsertKVStoreRecord :exec
-INSERT INTO kvstores (perm, rule_id, session_id, feature_id, entry_key, value)
+INSERT INTO kvstores (perm, rule_id, group_id, feature_id, entry_key, value)
 VALUES ($1, $2, $3, $4, $5, $6);
 
 -- name: DeleteAllTempKVStores :exec
@@ -38,16 +38,16 @@ FROM kvstores
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id IS NULL
+  AND group_id IS NULL
   AND feature_id IS NULL;
 
--- name: GetSessionKVStoreRecord :one
+-- name: GetGroupKVStoreRecord :one
 SELECT value
 FROM kvstores
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id = sqlc.arg('session_id')
+  AND group_id = sqlc.arg('group_id')
   AND feature_id IS NULL;
 
 -- name: GetFeatureKVStoreRecord :one
@@ -56,7 +56,7 @@ FROM kvstores
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id = sqlc.arg('session_id')
+  AND group_id = sqlc.arg('group_id')
   AND feature_id = sqlc.arg('feature_id');
 
 -- name: DeleteGlobalKVStoreRecord :exec
@@ -64,15 +64,15 @@ DELETE FROM kvstores
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id IS NULL
+  AND group_id IS NULL
   AND feature_id IS NULL;
 
--- name: DeleteSessionKVStoreRecord :exec
+-- name: DeleteGroupKVStoreRecord :exec
 DELETE FROM kvstores
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id = sqlc.arg('session_id')
+  AND group_id = sqlc.arg('group_id')
   AND feature_id IS NULL;
 
 -- name: DeleteFeatureKVStoreRecord :exec
@@ -80,7 +80,7 @@ DELETE FROM kvstores
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id = sqlc.arg('session_id')
+  AND group_id = sqlc.arg('group_id')
   AND feature_id = sqlc.arg('feature_id');
 
 -- name: UpdateGlobalKVStoreRecord :exec
@@ -89,16 +89,16 @@ SET value = $1
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id IS NULL
+  AND group_id IS NULL
   AND feature_id IS NULL;
 
--- name: UpdateSessionKVStoreRecord :exec
+-- name: UpdateGroupKVStoreRecord :exec
 UPDATE kvstores
 SET value = $1
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id = sqlc.arg('session_id')
+  AND group_id = sqlc.arg('group_id')
   AND feature_id IS NULL;
 
 -- name: UpdateFeatureKVStoreRecord :exec
@@ -107,5 +107,5 @@ SET value = $1
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id = sqlc.arg('session_id')
+  AND group_id = sqlc.arg('group_id')
   AND feature_id = sqlc.arg('feature_id');
--- a/db/sqlc/queries/kvstores.sql
+++ b/db/sqlc/queries/kvstores.sql
@@ -28,6 +28,10 @@ VALUES ($1, $2, $3, $4, $5, $6);
 DELETE FROM kvstores
 WHERE perm = false;
 
+-- name: ListAllKVStoresRecords :many
+SELECT *
+FROM kvstores;
+
 -- name: GetGlobalKVStoreRecord :one
 SELECT value
 FROM kvstores
--- a/firewalldb/db.go
+++ b/firewalldb/db.go
@@ -14,29 +14,21 @@ var (
 	ErrNoSuchKeyFound = fmt.Errorf("no such key found")
 )
 
-// firewallDBs is an interface that groups the RulesDB and PrivacyMapper
-// interfaces.
-type firewallDBs interface {
-	RulesDB
-	PrivacyMapper
-	ActionDB
-}
-
 // DB manages the firewall rules database.
 type DB struct {
 	started sync.Once
 	stopped sync.Once
 
-	firewallDBs
+	FirewallDBs
 
 	cancel fn.Option[context.CancelFunc]
 }
 
 // NewDB creates a new firewall database. For now, it only contains the
 // underlying rules' and privacy mapper databases.
-func NewDB(dbs firewallDBs) *DB {
+func NewDB(dbs FirewallDBs) *DB {
 	return &DB{
-		firewallDBs: dbs,
+		FirewallDBs: dbs,
 	}
 }
 
--- a/firewalldb/interface.go
+++ b/firewalldb/interface.go
@@ -134,3 +134,11 @@ type ActionDB interface {
 	// and feature name.
 	GetActionsReadDB(groupID session.ID, featureName string) ActionsReadDB
 }
+
+// FirewallDBs is an interface that groups the RulesDB, PrivacyMapper and
+// ActionDB interfaces.
+type FirewallDBs interface {
+	RulesDB
+	PrivacyMapper
+	ActionDB
+}
--- a/firewalldb/kvstores_kvdb.go
+++ b/firewalldb/kvstores_kvdb.go
@@ -16,13 +16,13 @@ the temporary store changes instead of just keeping an in-memory store is that
 we can then guarantee atomicity if changes are made to both the permanent and
 temporary stores.
 
-rules -> perm -> rule-name -> global   -> {k:v}
-              -> sessions -> group ID  -> session-kv-store  -> {k:v}
-			               -> feature-kv-stores -> feature-name -> {k:v}
+"rules" -> "perm" -> <rule-name> -> "global" -> {k:v}
+                                 -> "session-kv-store" -> <group-ID> -> {k:v}
+                                                                     -> "feature-kv-stores" -> <feature-name> -> {k:v}
 
-      -> temp -> rule-name -> global   -> {k:v}
-	      -> sessions -> group ID  -> session-kv-store  -> {k:v}
-				       -> feature-kv-stores -> feature-name -> {k:v}
+        -> "temp" -> <rule-name> -> "global" -> {k:v}
+                                 -> "session-kv-store" -> <group-ID> -> {k:v}
+                                                                     -> "feature-kv-stores" -> <feature-name> -> {k:v}
 */
 
 var (
--- a/firewalldb/sql_migration.go
+++ b/firewalldb/sql_migration.go
@@ -1,0 +1,492 @@
+package firewalldb
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightningnetwork/lnd/fn"
+	"github.com/lightningnetwork/lnd/sqldb"
+	"go.etcd.io/bbolt"
+)
+
+// kvEntry represents a single KV entry inserted into the BoltDB.
+type kvEntry struct {
+	perm     bool
+	ruleName string
+	key      string
+	value    []byte
+
+	// groupAlias is the legacy session group alias that the entry is
+	// associated with. For global entries, this will be fn.None[[]byte].
+	groupAlias fn.Option[[]byte]
+
+	// featureName is the name of the feature that the entry is associated
+	// with. If the entry is not feature specific, this will be
+	// fn.None[string].
+	featureName fn.Option[string]
+}
+
+// sqlKvEntry represents a single KV entry inserted into the SQL DB, containing
+// the same fields as the kvEntry, but with additional fields that represent the
+// SQL IDs of the rule, session group, and feature.
+type sqlKvEntry struct {
+	*kvEntry
+
+	ruleID int64
+
+	// groupID is the sql session group ID that the entry is associated
+	// with. For global entries, this will be Valid=false.
+	groupID sql.NullInt64
+
+	// featureID is the sql feature ID that the entry is associated with.
+	// This is only set if the entry is feature specific, and will be
+	// Valid=false for other types entries. If this is set, then groupID
+	// will also be set.
+	featureID sql.NullInt64
+}
+
+// namespacedKey returns a string representation of the kvEntry purely used for
+// logging purposes.
+func (e *kvEntry) namespacedKey() string {
+	ns := fmt.Sprintf("perm: %t, rule: %s", e.perm, e.ruleName)
+
+	e.groupAlias.WhenSome(func(alias []byte) {
+		ns += fmt.Sprintf(", group: %s", alias)
+	})
+
+	e.featureName.WhenSome(func(feature string) {
+		ns += fmt.Sprintf(", feature: %s", feature)
+	})
+
+	ns += fmt.Sprintf(", key: %s", e.key)
+
+	return ns
+}
+
+// MigrateFirewallDBToSQL runs the migration of the firwalldb stores from the
+// bbolt database to a SQL database. The migration is done in a single
+// transaction to ensure that all rows in the stores are migrated or none at
+// all.
+//
+// Note that this migration currently only migrates the kvstores, but will be
+// extended in the future to also migrate the privacy mapper and action stores.
+//
+// NOTE: As sessions may contain linked sessions and accounts, the sessions and
+// accounts sql migration MUST be run prior to this migration.
+func MigrateFirewallDBToSQL(ctx context.Context, kvStore *bbolt.DB,
+	sqlTx SQLQueries) error {
+
+	log.Infof("Starting migration of the rules DB to SQL")
+
+	err := migrateKVStoresDBToSQL(ctx, kvStore, sqlTx)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("The rules DB has been migrated from KV to SQL.")
+
+	// TODO(viktor): Add migration for the privacy mapper and the action
+	// stores.
+
+	return nil
+}
+
+// migrateKVStoresDBToSQL runs the migration of all KV stores from the KV
+// database to the SQL database. The function also asserts that the
+// migrated values match the original values in the KV store.
+func migrateKVStoresDBToSQL(ctx context.Context, kvStore *bbolt.DB,
+	sqlTx SQLQueries) error {
+
+	log.Infof("Starting migration of the KV stores to SQL")
+
+	var pairs []*kvEntry
+
+	// 1) Collect all key-value pairs from the KV store.
+	err := kvStore.View(func(tx *bbolt.Tx) error {
+		var err error
+		pairs, err = collectAllPairs(tx)
+		return err
+	})
+	if err != nil {
+		return fmt.Errorf("collecting all kv pairs failed: %w", err)
+	}
+
+	var insertedPairs []*sqlKvEntry
+
+	// 2) Insert all collected key-value pairs into the SQL database.
+	for _, entry := range pairs {
+		insertedPair, err := insertPair(ctx, sqlTx, entry)
+		if err != nil {
+			return fmt.Errorf("inserting kv pair %v failed: %w",
+				entry.key, err)
+		}
+
+		insertedPairs = append(insertedPairs, insertedPair)
+	}
+
+	// 3) Validate the migrated values against the original values.
+	for _, insertedPair := range insertedPairs {
+		// Fetch the appropriate SQL entry's value.
+		migratedValue, err := getSQLValue(ctx, sqlTx, insertedPair)
+		if err != nil {
+			return fmt.Errorf("getting SQL value for key %s "+
+				"failed: %w", insertedPair.namespacedKey(), err)
+		}
+
+		// Compare the value of the migrated entry with the original
+		// value from the KV store.
+		// NOTE: if the insert a []byte{} value into the sqldb as the
+		// entry value, and then retrieve it, the value will be
+		// returned as nil. The bytes.Equal will pass in that case,
+		// and therefore such cases won't error out. The kvdb instance
+		// can store []byte{} values.
+		if !bytes.Equal(migratedValue, insertedPair.value) {
+			return fmt.Errorf("migrated value for key %s "+
+				"does not match original value: "+
+				"migrated %x, original %x",
+				insertedPair.namespacedKey(), migratedValue,
+				insertedPair.value)
+		}
+	}
+
+	log.Infof("Migration of the KV stores to SQL completed. Total number "+
+		"of rows migrated: %d", len(pairs))
+
+	return nil
+}
+
+// collectAllPairs collects all key-value pairs from the KV store, and returns
+// them as a slice of kvEntry structs. The function expects the KV store to be
+// stuctured as described in the comment in the firewalldb/kvstores_kvdb.go
+// file. Any other structure will result in an error.
+// Note that this function and the subsequent functions are intentionally
+// designed to iterate over all buckets and values that exist in the KV store.
+// That ensures that we find all stores and values that exist in the KV store,
+// and can be sure that the kv store actually follows the expected structure.
+func collectAllPairs(tx *bbolt.Tx) ([]*kvEntry, error) {
+	var entries []*kvEntry
+	for _, perm := range []bool{true, false} {
+		mainBucket, err := getMainBucket(tx, false, perm)
+		if err != nil {
+			return nil, err
+		}
+
+		if mainBucket == nil {
+			// If the mainBucket doesn't exist, there are no entries
+			// to migrate under that bucket, therefore we don't
+			// error, and just proceed to not migrate any entries
+			// under that bucket.
+			continue
+		}
+
+		// Loop over each rule-name bucket.
+		err = mainBucket.ForEach(func(rule, v []byte) error {
+			if v != nil {
+				return errors.New("expected only " +
+					"buckets under main bucket")
+			}
+
+			ruleBucket := mainBucket.Bucket(rule)
+			if ruleBucket == nil {
+				return fmt.Errorf("rule bucket %s not found",
+					rule)
+			}
+
+			pairs, err := collectRulePairs(
+				ruleBucket, perm, string(rule),
+			)
+			if err != nil {
+				return err
+			}
+
+			entries = append(entries, pairs...)
+
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return entries, nil
+}
+
+// collectRulePairs processes a single rule bucket, which should contain the
+// global and session-kv-store key buckets.
+func collectRulePairs(bkt *bbolt.Bucket, perm bool, rule string) ([]*kvEntry,
+	error) {
+
+	var params []*kvEntry
+
+	err := verifyBktKeys(
+		bkt, true, globalKVStoreBucketKey, sessKVStoreBucketKey,
+	)
+	if err != nil {
+		return params, fmt.Errorf("verifying rule bucket %s keys "+
+			"failed: %w", rule, err)
+	}
+
+	if globalBkt := bkt.Bucket(globalKVStoreBucketKey); globalBkt != nil {
+		p, err := collectKVPairs(
+			globalBkt, true, perm, rule,
+			fn.None[[]byte](), fn.None[string](),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("collecting global kv pairs "+
+				"failed: %w", err)
+		}
+
+		params = append(params, p...)
+	}
+
+	if sessBkt := bkt.Bucket(sessKVStoreBucketKey); sessBkt != nil {
+		err := sessBkt.ForEach(func(groupAlias, v []byte) error {
+			if v != nil {
+				return fmt.Errorf("expected only buckets "+
+					"under %s bucket", sessKVStoreBucketKey)
+			}
+
+			groupBucket := sessBkt.Bucket(groupAlias)
+			if groupBucket == nil {
+				return fmt.Errorf("group bucket for group "+
+					"alias %s not found", groupAlias)
+			}
+
+			kvPairs, err := collectKVPairs(
+				groupBucket, false, perm, rule,
+				fn.Some(groupAlias), fn.None[string](),
+			)
+			if err != nil {
+				return fmt.Errorf("collecting group kv "+
+					"pairs failed: %w", err)
+			}
+
+			params = append(params, kvPairs...)
+
+			err = verifyBktKeys(
+				groupBucket, false, featureKVStoreBucketKey,
+			)
+			if err != nil {
+				return fmt.Errorf("verification of group "+
+					"bucket %s keys failed: %w", groupAlias,
+					err)
+			}
+
+			ftBkt := groupBucket.Bucket(featureKVStoreBucketKey)
+			if ftBkt == nil {
+				return nil
+			}
+
+			return ftBkt.ForEach(func(ftName, v []byte) error {
+				if v != nil {
+					return fmt.Errorf("expected only "+
+						"buckets under %s bucket",
+						featureKVStoreBucketKey)
+				}
+
+				// The feature name should exist, as per the
+				// verification above.
+				featureBucket := ftBkt.Bucket(ftName)
+				if featureBucket == nil {
+					return fmt.Errorf("feature bucket "+
+						"%s not found", ftName)
+				}
+
+				featurePairs, err := collectKVPairs(
+					featureBucket, true, perm, rule,
+					fn.Some(groupAlias),
+					fn.Some(string(ftName)),
+				)
+				if err != nil {
+					return fmt.Errorf("collecting "+
+						"feature kv pairs failed: %w",
+						err)
+				}
+
+				params = append(params, featurePairs...)
+
+				return nil
+			})
+		})
+		if err != nil {
+			return nil, fmt.Errorf("collecting session kv pairs "+
+				"failed: %w", err)
+		}
+	}
+
+	return params, nil
+}
+
+// collectKVPairs collects all key-value pairs from the given bucket, and
+// returns them as a slice of kvEntry structs. If the errorOnBuckets parameter
+// is set to true, then the function will return an error if the bucket
+// contains any sub-buckets. Note that when the errorOnBuckets parameter is
+// set to false, the function will not collect any key-value pairs from the
+// sub-buckets, and will just ignore them.
+func collectKVPairs(bkt *bbolt.Bucket, errorOnBuckets, perm bool,
+	ruleName string, groupAlias fn.Option[[]byte],
+	featureName fn.Option[string]) ([]*kvEntry, error) {
+
+	var params []*kvEntry
+
+	return params, bkt.ForEach(func(key, value []byte) error {
+		// If the value is nil, then this is a bucket, which we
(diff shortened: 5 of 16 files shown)


firewalldb/sql_migration.go
```diff
@@ -0,0 +1,405 @@
...
+		migratedValue, err := sqlTx.GetGlobalKVStoreRecord(
+			ctx,
+			sqlc.GetGlobalKVStoreRecordParams{
+				Key:    globalInsertParams.EntryKey,
+				Perm:   globalInsertParams.Perm,
+				RuleID: globalInsertParams.RuleID,
+			},
+		)
+		if err != nil {
+			return fmt.Errorf("retreiving of migrated global kv "+
```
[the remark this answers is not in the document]
ViktorT-11: Ah yeah, good idea! Addressed this in the latest push.

ViktorTigerstrom:
Thanks for the swift review @ellemouton :pray:! I've rebased this to be based on master and not the sessions migration PR. Note that I've kept 2 commits from the sessions migration PR in this PR (the 2 first), as they are required. So feel free to ignore reviewing the first to commits of this PR.
Also note that since this is not based on the sessions migrations PR, we can't test this migration locally as we need the sessions to be migrated prior to the kvstores (so that we can fetch the sessions SQL IDs). Let me know if you feel a need to test this prior to merging the sessions PR though, and I'll send you a branch which adds local testing that's based on both PRs.

firewalldb/interface.go
```diff
@@ -141,4 +141,7 @@ type FirewallDBs interface {
 	RulesDB
 	PrivacyMapper
 	ActionDB
+
+	// Close closes the underlying store.
+	Close() error
```
ellemouton: dont think we need this

ellemouton: just call t.Cleanup sooner

firewalldb/db.go
```diff
@@ -14,29 +14,21 @@ var (
 	ErrNoSuchKeyFound = fmt.Errorf("no such key found")
 )
```
ellemouton, line 16: your commit messages could be fleshed out a bit more pls. it isnt clear why this needs to be exported (same for other commits in this PR)

firewalldb/interface.go
ellemouton: (again commit message could explain more as it took me a sec to figure out why you're adding this)

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,478 @@
...
+// bbolt database to a SQL database. The migration is done in a single
+// transaction to ensure that all rows in the stores are migrated or none at
+// all.
+//
+// Note that this migration currently only migrates the kvstores, but will be
+// extended in the future to also migrate the privacy mapper and action stores.
+//
+// NOTE: As sessions may contain linked sessions and accounts, the sessions and
+// accounts sql migration MUST be run prior to this migration.
+func MigrateFirewallDBToSQL(ctx context.Context, kvStore *BoltDB,
```
ellemouton: same as for sessions PR, rather pass raw kvdb pointer in here & not the BoltDB

ellemouton: note that you are already using the raw pointer only & not actually using the BboltDb at all

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,478 @@
...
+func migrateKVStoresDBToSQL(ctx context.Context, kvStore *BoltDB,
+	sqlTx SQLQueries) error {
+
+	log.Infof("Starting migration of the KV stores to SQL")
+
+	// allParams will hold all the kvParams that are inserted into the
+	// SQL database during the migration.
+	var allParams []kvParams
+
+	err := kvStore.Update(func(kvTx *bbolt.Tx) error {
```
ellemouton: why `Update`?

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,478 @@
...
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// After the migration is done, we validate that all inserted kvParams
```
ellemouton: cool yeah thanks for separating this 👍  i think it is especially important for the kv stores since there is a higher chance of overwriting 👍

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,527 @@
...
+			name:       "session specific records",
+			populateDB: sessionSpecificRecords,
+		},
+		{
+			name:       "feature specific records",
+			populateDB: featureSpecificRecords,
+		},
+		{
+			name:       "random records",
+			populateDB: randomKVRecords,
```
ellemouton: i think it's worth having a case that does some explicit entries that write to all the store types. currently we only do that for the random one.

ellemouton (commented):
looking great! a couple of comments. I'll do a final round once the session PR is in 👍

firewalldb/interface.go
ViktorT-11: Thanks :pray:! Addressed with a new commit in the latest push :)

firewalldb/db.go
ViktorT-11: Sorry, completely forgot updating the commit messages in the latest push. I've updated them in the latest push though, so I hope the commits should be understandable now!

firewalldb/sql_migration.go
ViktorT-11: It was due to the migration utilising the `getMainBucket` helper function, which with the previous version created the main bucket if it didn't exist. However I realised that if the bucket doesn't exist, it just means that we can skip migrating the records, so I updated the PR to do that instead with the latest push.

firewalldb/sql_migration_test.go
ViktorT-11: Good idea! Added an extra test with the latest push.

firewalldb/kvstores_kvdb.go
```diff
@@ -16,13 +16,13 @@ the temporary store changes instead of just keeping an in-memory store is that
 we can then guarantee atomicity if changes are made to both the permanent and
 temporary stores.
 
-rules -> perm -> rule-name -> global   -> {k:v}
-              -> sessions -> group ID  -> session-kv-store  -> {k:v}
-			               -> feature-kv-stores -> feature-name -> {k:v}
+"rules" -> "perm" -> rule-name -> "global" -> {k:v}
```
bitromortac: thanks for the update here :+1:

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+			return fmt.Errorf("unexpected bucket %s under "+
+				"rule-name bucket", string(k))
+		}
+	})
+}
+
+// processGlobalRuleBucket processes the global bucket under a rule bucket,
+// which contains the global key-value store records for the rule.
+// It inserts the records into the SQL database and asserts that
+// the migrated values match the original values in the KV store.
```
bitromortac: nit: there are no assertions in here, same in `processGroupBucket`

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+
+// processGroupBucket processes a single group bucket, which contains the
+// session-wide kv records and as well as the feature-kv-stores key bucket for
+// that group. For the session-wide kv records, it inserts the records into the
+// SQL database and asserts that the migrated values match the original values.
+func processGroupBucket(ctx context.Context, sqlTx SQLQueries, perm bool,
+	ruleSqlId int64, groupAlias []byte,
+	groupBucket *bbolt.Bucket) ([]kvParams, error) {
+
+	groupSqlId, err := sqlTx.GetSessionIDByAlias(
```
bitromortac: nit: contract to line

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+				if v != nil {
+					return errors.New("expected only " +
+						"buckets under main bucket")
+				}
+
+				ruleName := k
+				ruleNameBucket := mainBucket.Bucket(k)
+				if ruleNameBucket == nil {
+					return fmt.Errorf("rule bucket %s "+
+						"not found", string(k))
```
bitromortac: I think you don't need those casts as the %s directive already does it

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,537 @@
...
+	"golang.org/x/exp/rand"
+)
+
+// kvStoreRecord represents a single KV entry inserted into the BoltDB.
+type kvStoreRecord struct {
+	Perm        bool
+	RuleName    string
+	EntryKey    string
+	Global      bool
+	GroupID     *session.ID
```
bitromortac: you don't use the nilness of GroupID, right? could you use `fn.Option[GroupID]` to encode that this is a global entry?

firewalldb/kvstores_kvdb.go
```diff
@@ -16,13 +16,13 @@ the temporary store changes instead of just keeping an in-memory store is that
...
+				  "session-kv-store" -> group ID -> {k:v}
```
ellemouton: missing the `->` here

firewalldb/kvstores_kvdb.go
ellemouton: while we are here, can we add `<` & `>` around keys that vary? like `<rule-name>`

ellemouton: and `<group ID>`& `<feature name>`

ellemouton: yeah thanks for the fix!! 🙏

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+
+			if mainBucket == nil {
+				// If the mainBucket doesn't exist, there are no
+				// records to migrate under that bucket,
+				// therefore we don't error, and just proceed
+				// to not migrate any records under that bucket.
+				continue
+			}
+
+			err = mainBucket.ForEach(func(k, v []byte) error {
```
ellemouton: suggest adding a short comment: "Loop over each rule-name in the bucket"

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+				if v != nil {
+					return errors.New("expected only " +
+						"buckets under main bucket")
+				}
+
+				ruleName := k
+				ruleNameBucket := mainBucket.Bucket(k)
```
ellemouton: ruleNameBucket := mainBucket.Bucket(k) -> ruleNameBucket := mainBucket.Bucket(ruleName)

ellemouton: same below, let's not continue to use `k`

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+		}
+
+		groupBucket := mainSessionBucket.Bucket(groupId)
+		if groupBucket == nil {
+			return fmt.Errorf("group bucket for group id %s"+
+				"not found", string(groupId))
+		}
+
+		p, err := processGroupBucket(
+			ctx, sqlTx, perm, ruleSqlId, groupId, groupBucket,
```
ellemouton: makes me think we should maybe take this opportunity to rename the `session_id` column in the `kvstores` table to `group_id`?

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+	log.Infof("Migration of the KV stores to SQL completed. Total number "+
+		"of rows migrated: %d", len(allParams))
+
+	return nil
+}
+
+// processRuleBucket processes a single rule bucket, which contains the
+// global and session-kv-store key buckets.
+func processRuleBucket(ctx context.Context, sqlTx SQLQueries, perm bool,
+	ruleSqlId int64, ruleBucket *bbolt.Bucket) ([]kvParams, error) {
```
ellemouton: style nit: usually best practice to put like cfg/resource params at the start, then followed by variable things: 

so ie, the bucket should be before ruleSQLID

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+
+	var params []kvParams
+
+	return params, ruleBucket.ForEach(func(k, v []byte) error {
```
ellemouton: why use `ForEach` here? i think better to just be strict and just process the two buckets that we expect to find.

firewalldb/sql_migration.go
ellemouton: same with other `process*` functions in this file

firewalldb/sql_migration.go
ellemouton: it would also remove a level of indentation

firewalldb/sql_migration.go
ellemouton: also , can just call it "bucket" since the method is already called "processRuleBucket"

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+// which contains the global key-value store records for the rule.
+// It inserts the records into the SQL database and asserts that
+// the migrated values match the original values in the KV store.
+func processGlobalRuleBucket(ctx context.Context, sqlTx SQLQueries, perm bool,
+	ruleSqlId int64, globalBucket *bbolt.Bucket) ([]kvParams, error) {
+
+	var params []kvParams
+
+	return params, globalBucket.ForEach(func(k, v []byte) error {
+		if v == nil {
```
ellemouton: still getting through the diff, but just noting that we should defs test that things dont break if we _do_ insert a key-value pair where the value is `nil`

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+			Value:    v,
+			Perm:     perm,
+			RuleID:   ruleSqlId,
+		}
+
+		err := sqlTx.InsertKVStoreRecord(ctx, globalInsertParams)
+		if err != nil {
+			return fmt.Errorf("inserting global kv store "+
+				"record failed %w", err)
+		}
```
ellemouton: since we are already separating out the migration & validation steps, I think we can also split this out: 

1) collect all params
2) call the "insert" for each param
3) validate.

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+		params = append(params, globalInsertParams)
+
+		return nil
+	})
+}
+
+// processSessionBucket processes the session-kv-store bucket under a rule
+// bucket, which contains the group-id buckets for that rule.
+func processSessionBucket(ctx context.Context, sqlTx SQLQueries, perm bool,
+	ruleSqlId int64, mainSessionBucket *bbolt.Bucket) ([]kvParams, error) {
```
ellemouton: mainSessionBucket -> bucket

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+
+	var params []kvParams
+
+	return params, mainSessionBucket.ForEach(func(groupId, v []byte) error {
+		if v != nil {
+			return fmt.Errorf("expected only buckets under "+
+				"%s bucket", string(sessKVStoreBucketKey))
+		}
+
+		groupBucket := mainSessionBucket.Bucket(groupId)
```
ellemouton: perhaps call it `groupAlias` to not confuse it with groupID later on

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+			"not found in sql db", groupAlias)
+	} else if err != nil {
+		return nil, err
+	}
+
+	var params []kvParams
+
+	return params, groupBucket.ForEach(func(k, v []byte) error {
+		switch {
+		case v != nil:
```
ellemouton: same here: dont think the FOrEach is needed. Just index into the buckets directly

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+// transaction to ensure that all rows in the stores are migrated or none at
+// all.
+//
+// Note that this migration currently only migrates the kvstores, but will be
+// extended in the future to also migrate the privacy mapper and action stores.
+//
+// NOTE: As sessions may contain linked sessions and accounts, the sessions and
+// accounts sql migration MUST be run prior to this migration.
+func MigrateFirewallDBToSQL(ctx context.Context, kvStore *bbolt.DB,
+	tx SQLQueries) error {
```
ellemouton: something i've just noticed is that we are passing in a single SQL tx here so that we know all SQL queries will happen in the same tx. But the same is not true for our kv store - here we are passing in the DB object meaning we are then creating txs for it within this migration. 
we should probably think of the consequences of this - especially if we decide to do all the LiT migrations in one go - in that case we might want all the bbolt queries to happen in the same tx

firewalldb/sql_migration_test.go
ellemouton: it still doesnt have a deterministic test that stores in 2 feature tables/ 2 rule tables etc right?

like, currently `globalRecords` only inserts under a single rule name, for example

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,537 @@
...
+		{
+			name: "empty",
+			populateDB: func(t *testing.T, ctx context.Context,
+				boltDB *BoltDB,
+				sessionStore session.Store) []kvStoreRecord {
+
+				// Don't populate the DB.
+				return make([]kvStoreRecord, 0)
+			},
+		},
```
ellemouton, line 235: are we covering nil/empty byte array values somewhere?

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,537 @@
...
+
+	permKvRecord := kvStoreRecord{
+		RuleName:    ruleName,
+		GroupID:     &sess.GroupID,
+		FeatureName: featureNameOpt,
+		EntryKey:    entryKey,
+		Value:       entryVal,
+		Perm:        true,
+		Global:      global,
+	}
```
ellemouton: if you are inserting the same key-value pair in both, then things are not properly being tested since you are not testing that the one namespace doesnt overwrite the other

firewalldb/sql_migration_test.go
ellemouton: also, looks like for this new test, all the key-value pairs inserted at all the levels are the same... so again it doesnt properly test that the correct things mapping is taking place

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,537 @@
...
+		entryVal = []byte{1, 2, 3}
+	)
+
+	// Create a session that we can reference.
+	sess, err := sessionStore.NewSession(
+		ctx, "test", session.TypeAutopilot,
+		time.Unix(1000, 0), "something",
+	)
+	require.NoError(t, err)
+
```
ellemouton: only need to create session if !global

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,537 @@
...
+		t.Fatalf("cannot set both global and feature specific at the " +
+			"same time")
+	}
+
+	kvStores := boltDB.GetKVStores(
+		record.RuleName, *record.GroupID,
+		record.FeatureName.UnwrapOr(""),
+	)
+
+	err := kvStores.Update(ctx, func(ctx context.Context,
```
ellemouton, line 520: suggest:

```
	err := kvStores.Update(ctx, func(ctx context.Context,
		tx KVStoreTx) error {

		store := tx.Global()
		switch {
		case record.Global && !record.Perm:
			store = tx.GlobalTemp()
		case !record.Global && !record.Perm:
			store = tx.LocalTemp()
		case !record.Global && record.Perm:
			store = tx.Local()
		}

		return store.Set(ctx, record.EntryKey, record.Value)
	})
	require.NoError(t, err)
```

ellemouton: (ie, reduce repetition where possible)

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
+package firewalldb
```
ellemouton, line 1: there is a lot of nesting happening here with various operations happening a couple layers deep. There is also quite a lot of repetition. 

Let me know what you think of this diff which follows a more step-by-step approach with fewer layers & more code re-use. Notice how `migrateKVStoresDBToSQL` follows an easy-to-ready step-by-step flow: 
1) collect pairs (kvdb only)
2) migrate pairs (sql only)
3) validate migrated pairs. (compare kvdb & sql)

(just delete `.txt` from the name - GH wont let me upload the `.go` file)
[sql_migration.go.txt](https://github.com/user-attachments/files/20724625/sql_migration.go.txt)

firewalldb/kvstores_kvdb.go
ViktorT-11: Good idea! done :)

firewalldb/sql_migration.go
ViktorT-11: Agree, added a new commit for this, which does this in a braking change matter, as the SQL db isn't used in production by anyone yet. Let me know if you disagree with the motivation to adding this a breaking change, and would prefer a separate migration.

firewalldb/sql_migration.go
ViktorT-11: As discussed offline, this was done to ensure that the migration really catches all kv entries, and doesn't assume that kvstore is only structured in one way without erroring.
In order to keep the benefits of directly indexing into the buckets, while still not adding such assumptions, I added a new `verifyBktKeys` function. I hope you agree with the addition of that function.

firewalldb/sql_migration.go
ViktorT-11: Good idea. Updated this throughout the PR.

firewalldb/sql_migration.go
ViktorT-11: See https://github.com/lightninglabs/lightning-terminal/pull/1079#discussion_r2178724972

firewalldb/sql_migration_test.go
ViktorT-11: Thanks for pointing that out. This surfaced an issue in the migration_test code when inserting empty byte array values, as the test assertion would fail while the migration assertion would not. The reason for the issue was that when inserting empty byte array value into the SQL db, and then fetching is, the returned value will be `nil` and not an empty byte array.
The migration assertion code uses `byte.Equals` which will pass if the expected value is `[]byte{}` but the actual value is `nil`, while the test code used `require.Equals` which will error in that case.
I therefore updated the test code to also use `byte.Equals`, as there's no other way that I can think of that mitigates the issue.

firewalldb/sql_migration.go
ViktorT-11: See https://github.com/lightninglabs/lightning-terminal/pull/1079#discussion_r2178730669

firewalldb/sql_migration.go
ViktorT-11: Thanks a lot for the file 🎉🙏! I used that as the base for the migration code, and updated it with additional changes mentioned in the above comments, plus some additional cleanups. Hope you agree with my changes :)

firewalldb/sql_migration.go
ViktorT-11: Ah nice, thanks :pray:!

firewalldb/sql_migration_test.go
ViktorT-11: Good idea, added :)

firewalldb/sql_migration_test.go
ViktorT-11: Ah yeah, good point! I updated this test quite a bit (now renamed to `allEntryCombinations`) to hopefully cover all different types of entries in a deterministic way. I think it should now cover all of those cases.

firewalldb/sql_migration_test.go
ViktorT-11: I'm not fully sure that I got exactly what this feedback was referring to here, but I added additional inserts to the `allEntryCombinations` as mentioned in https://github.com/lightninglabs/lightning-terminal/pull/1079#discussion_r2178737696. I hope that covers this feedback, else let me know!

firewalldb/sql_migration.go
ViktorT-11: Good point! IMO though, I do think it should be ok to do multiple txes for the Bbolt DB, as long as it's only `View` transactions right? That's given that we ensure that we never add functionality that inserts entries into the Bbolt DB in between those `View` transaction. I think we should be able to guarantee that behaviour though, as we are planning on deprecating the kvdb store implementation and therefore won't add additional functionality to it.
Or do you see any scenario where this would be an issue?

ViktorTigerstrom:
Thanks a lot for the reviews and the great suggestions @ellemouton & @bitromortac :pray::tada:!!! I've overhauled this PR quite a bit since the last push to address that feedback just FYI, so expect the PR to contain quite a lot of new changes.

Also, apologies for the linting error. I'll address that one after the next review round, in order to not waste extra CI resources just for that.

db/sqlc/queries/kvstores.sql
```diff
@@ -38,16 +38,16 @@ FROM kvstores
 WHERE entry_key = sqlc.arg('key')
   AND rule_id = sqlc.arg('rule_id')
   AND perm = sqlc.arg('perm')
-  AND session_id IS NULL
+  AND group_id IS NULL
   AND feature_id IS NULL;
 
--- name: GetSessionKVStoreRecord :one
+-- name: GetSessionGroupKVStoreRecord :one
```
bitromortac: nit: we could shorten `SessionGroup` to just `Group` as I think it isn't that ambiguous

firewalldb/kvstores_kvdb.go
```diff
@@ -16,13 +16,13 @@ the temporary store changes instead of just keeping an in-memory store is that
 we can then guarantee atomicity if changes are made to both the permanent and
 temporary stores.
 
-rules -> perm -> rule-name -> global   -> {k:v}
-              -> sessions -> group ID  -> session-kv-store  -> {k:v}
-			               -> feature-kv-stores -> feature-name -> {k:v}
+"rules" -> "perm" -> <rule-name> -> "global" -> {k:v}
+				 -> "session-kv-store" -> <group-ID> -> {k:v}
+							             -> "feature-kv-stores" -> <feature-name> -> {k:v}
```
bitromortac: tiny nit: I would prefer to use spaces only instead of tabs

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+// kvEntry represents a single KV entry inserted into the BoltDB.
+type kvEntry struct {
+	perm     bool
+	ruleName string
+	key      string
+	value    []byte
+
+	// GroupID is the legacy session group alias that the entry is
+	// associated with. For global entries, this will be fn.None[[]byte].
+	groupAlias fn.Option[[]byte]
```
bitromortac, line 25: nit: doc mismatch, same in `sqlKvEntry`

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+// database to the SQL database. The function also asserts that the
+// migrated values match the original values in the KV store.
+func migrateKVStoresDBToSQL(ctx context.Context, kvStore *bbolt.DB,
+	sqlTx SQLQueries) error {
+
+	log.Infof("Starting migration of the KV stores to SQL")
+
+	var pairs []*kvEntry
+
+	// 1) Collect all key-value pairs from the KV store.
```
bitromortac, line 108: We don't anticipate any memory concerns here I guess since value objects seem to be quite small right? I think for actions we may need to consider a chunked approach

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+				"failed: %w", insertedPair.namespacedKey(), err)
+		}
+
+		// Compare the value of the migrated entry with the original
+		// value from the KV store.
+		// NOTE: if the insert a []byte{} value into the sqldb as the
+		// entry value, and then retrieve it, the value will be
+		// returned as nil. The bytes.Equal will pass in that case,
+		// and therefore such cases won't error out. The kvdb instance
+		// can store []byte{} values.
```
bitromortac, line 146: I'd favor the store to return `[]byte{}` if it internally stores it as `nil` when passing in `[]byte{}`, with a conversion at the lowest layer. Can the kv database currently have `[]byte(nil)` stored via some code path? If that's the case we may want to remove `NOT NULL` from `value BLOB`.

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+
+	return bkt.ForEach(func(key, v []byte) error {
+		if v != nil {
+			// If we allow key-values, then we can just continue
+			// to the next key. Else we need to error out, as we
+			// only expect buckets under the passed bucket.
+			if errorOnKeyValues {
+				return fmt.Errorf("unexpected key-value pair "+
+					"found: key=%s, value=%x", key, v)
+			} else {
```
bitromortac: nit: since it's exiting, you can spare the else clause

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,666 @@
...
+	testEntryKey2    = "test-entry-key-2"
+	testEntryKey3    = "test-entry-key-3"
+)
+
+var (
+	testEntryValue = []byte{1, 2, 3}
+)
+
+// TestFirewallDBMigration tests the migration of firewalldb from a bolt
+// backed to a SQL database. Note that this test does not attempt to be a
```
bitromortac: nit: `backend`

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,666 @@
...
+			ruleIDs    = make(map[string]int64)
+			groupIDs   = make(map[string]int64)
+			featureIDs = make(map[string]int64)
+			err        error
+		)
+
+		getRuleID := func(ruleName string) int64 {
+			ruleID, ok := ruleIDs[ruleName]
+			if !ok {
+				ruleID, err = sqlStore.GetRuleID(
```
bitromortac: nit: contract

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,666 @@
...
+		time.Unix(1000, 0), "something",
+	)
+	require.NoError(t, err)
+
+	return fn.Some(sess.GroupID[:])
+}
+
+// insertTempAndPermEntry populates the kv store with one entry for the local
+// temp store, and one entry for the local store. Both of the entries will be
+// inserted with the same groupAlias, ruleName, entryKey and entryValue.
```
bitromortac: nit: this acts on global and local stores, so I think `local` could be removed

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,666 @@
...
+// combinations of entries in the kvstores, including nil and empty entry
+// values. That therefore ensures that the migrations don't overwrite or miss
+// any entries when the entry set is more complex than just a single entry at
+// each level.
+func allEntryCombinations(t *testing.T, ctx context.Context, boltDB *BoltDB,
+	sessionStore session.Store) []*kvEntry {
+
+	// First lets create standard entries at all levels, which represents
+	// the entries added by other tests.
+	gEntries1 := globalEntries(t, ctx, boltDB, sessionStore)
```
bitromortac: nit: I would favor the following pattern, as it makes the test more robust with respect to changes
```golang
	var result []*kvEntry
	add := func(entry []*kvEntry) {
		result = append(result, entry...)
	}
	add(globalEntries(t, ctx, boltDB, sessionStore))
```

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,666 @@
...
+		ruleName        = "initial-rule"
+		groupAlias      []byte
+		featureName     = "initial-feature"
+	)
+
+	// Create a random session that we can reference for the initial group
+	// ID.
+	sess, err := sessionStore.NewSession(
+		ctx, "initial-session", session.Type(uint8(rand.Intn(5))),
+		time.Unix(1000, 0), randomString(rand.Intn(10)+1),
```
bitromortac: nit: I suggest to remove randomness here, since we are not testing session functionality, to have clearer code

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,666 @@
...
+			// value
+			if rand.Intn(2) == 0 {
+				entry.value = nil
+			} else {
+				entry.value = []byte{}
+			}
+		} else {
+			// Else generate a random value for all entries,
+			entry.value = []byte(randomString(rand.Intn(100) + 1))
+		}
```
bitromortac, line 645: this was already done above

firewalldb/kvstores_kvdb.go
ellemouton: @ViktorTigerstrom  - you should be able to set your IDE up to auto replace tabs as spaces

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,666 @@
...
+		switch {
+		case entry.groupAlias.IsNone() && entry.perm:
+			store = tx.Global()
+		case entry.groupAlias.IsNone() && !entry.perm:
+			store = tx.GlobalTemp()
+		case entry.groupAlias.IsSome() && !entry.perm:
+			store = tx.LocalTemp()
+		case entry.groupAlias.IsSome() && entry.perm:
+			store = tx.Local()
+		}
```
ellemouton: note that in the example i posted before, we could not possibly end up with a nil store. So if you want to have these 4 cases and not pre assigne a default, the rather have a default case to catch that none of the cases match

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+	groupID sql.NullInt64
+
+	// featureID is the sql feature ID that the entry is associated with.
+	// This is only set if the entry is feature specific, and will be
+	// Valid=false for other types entries. If this is set, then groupID
+	// will also be set.
+	featureID sql.NullInt64
+}
+
+func (e *kvEntry) namespacedKey() string {
```
ellemouton, line 54: nit: missing godoc. Mention that it is purely for logging

firewalldb/sql_migration.go
ellemouton: yeah this is the sort of thing we should write benchmarks for. perhaps a good candidate for a follow up: vitkor, you can create a TODO in the epic description & then it can be done in a follow up i think (for all the migrations)

firewalldb/sql_migration.go
ellemouton: > Can the kv database currently have []byte(nil) stored via some code path? If

Should make sure to cover in a unit test

firewalldb/sql_migration.go
```diff
@@ -0,0 +1,486 @@
...
+	entry.featureName.WhenSome(func(feature string) {
+		id, err := tx.GetOrInsertFeatureID(ctx, feature)
+		if err != nil {
+			err = fmt.Errorf("getting/inserting feature id for %s "+
+				"failed: %w", feature, err)
+			return
+		}
+
+		p.FeatureID = sqldb.SQLInt64(id)
+	})
```
ellemouton: error not checked!!

ellemouton (commented):
great work!

I think g2g on next push 🛩️

db/sqlc/queries/kvstores.sql
ViktorT-11: Addressed :)

firewalldb/kvstores_kvdb.go
ViktorT-11: Thanks!

firewalldb/sql_migration.go
ViktorT-11: > I'd favor the store to return []byte{} if it internally stores it as nil when passing in []byte{}

I agree that this would be the ideal solution, but I've tested and unfortunately I can't get the SQLLite DB to return it as a `[]byte{}`, even when removing the `NOT NULL` constraint. For Postgres DBs, I can get it to work. So I believe this is a differentiation of how the DB implementations handles that type of data, and therefore IMO the `bytes.equals` solution is the best we can do.

> Can the kv database currently have []byte(nil) stored via some code path?

In a prod env, there should be no code path that leads to a `[]byte(nil)` value being inserted from my research.

> Should make sure to cover in a unit test

The `allEntryCombinations` & `randomKVEntries` unit tests already covered that by inserting `nil` values, which are equivalent, but just to explicitly also insert a value set to exactly `[]byte(nil)`, I added an extra `nilSliceValue = []byte(nil)` to the `allEntryCombinations` unit test :)

firewalldb/sql_migration_test.go
ViktorT-11: Ah nice, agree :). Addressed!

firewalldb/sql_migration_test.go
ViktorT-11: Thanks!

firewalldb/sql_migration_test.go
ViktorT-11: Ok I changed to your previous example :). The reason I went for the above, was that I liked that it was very explicit what each possible logical combination would lead to, and that this was in a test file (which is why I avoided the extra unreachable `default` case).

But addressed this with your previous suggestion :)

firewalldb/sql_migration.go
ViktorT-11: Ah snap, good catch :pray:! This shows that linters can be really useful for more than just formatting sometimes 😃!

firewalldb/sql_migration.go
ViktorT-11: Good idea with the benchmarking! Added that to the epic.

> We don't anticipate any memory concerns here I guess since value objects seem to be quite small right? I think for actions we may need to consider a chunked approach

Yeah agree, we discussed this a bit offline previously, and for kvstores we deemed that this shouldn't be an issue as there shouldn't be much data stored in the kvstores for our users. But definitely agree that this becomes much more important for the actions migration, and I'd argue that it might be worth it to insert and validate the migration  of one action at a time there just to ensure that this can't be an issue there either.

ViktorTigerstrom:
Thanks a lot for the in depth reviews @bitromortac & @ellemouton :pray::fire:! Addressed your feedback with the latest push.

firewalldb/sql_migration.go
bitromortac: > I can't get the SQLLite DB to return it as a []byte{}

Right, I tested that as well, and it doesn't seem to work. I meant another solution, to update `sqlKVStore.Get` to convert `[]byte(nil)` to `[]byte{}`, I think that could resolve this.

ViktorT-11: Do you mean that we'd implement a conversion in the actual sql store layer (the `SQLDB` struct, i.e. the layer between the `SQLQueries` and the rest of the application), which converts `nil` values fetched from the db to `[]byte{}`?
We could potentially get that to work in the `sql_migration_test.go` file, but not in the `sql_migration.go` file, as we need to use the pure `SQLQueries` object in the `sql_migration.go` file (this becomes very important when implementing migrations using the `sqldb/v2` library). Eventually I think we may want to go for the using the pure `SQLQueries` object in the test assertions as well, but I think that's out of scope for this PR.

Potentially we could manually update the query implementation to always return `[]byte{}` on `nil` values, but that would mean that we wouldn't be able to use `sqlc` to generate this query, which I'm not sure we'd want to do due to the downsides? Would be appreciate your opinions on that as well :)!

bitromortac: I see, one workaround could be to apply a patch with `gen_sqlc_docker.sh` like [here](https://github.com/lightninglabs/taproot-assets/blob/a62b68d6fa7bbc875fb636d00c77dc6cbf52eab0/scripts/gen_sqlc_docker.sh#L58) to the respective queries.

ViktorT-11: Hmmm yeah cool idea! I can look into if it'd be possible to replace `nil` results with `[]byte{}` for all queries in the sql db by a adding some kind of script there, if you think it's worth it. Do you think we should do that in this PR, or as a follow-up?

ViktorTigerstrom:
Thanks for the review @ellemouton 🎉! Added that + the sqldb v2 follow-up to the epic.

firewalldb/sql_migration_test.go
```diff
@@ -0,0 +1,664 @@
...
+		t, ctx, boltDB, testRuleName2, groupAlias,
+		fn.Some(testFeatureName2), testEntryKey, testEntryValue,
+	))
+
+	// Finally, lets add a few entries with nil and empty values set for the
+	// actual key value, at all different levels, to ensure that tests don't
+	// break if the value is nil or empty.
+	var (
+		nilValue      []byte = nil
+		nilSliceValue        = []byte(nil)
```
bitromortac, line 406: nit: wanted to point out that those are equivalent (can be addressed later)

firewalldb/kvstores_kvdb.go
bitromortac: just wanted to note that the whitespace still consists of tabs, but it's ok, as it was just small nit

firewalldb/sql_migration_test.go
ViktorT-11: Yeah, I included this to address https://github.com/lightninglabs/lightning-terminal/pull/1079#discussion_r2194485826, to clarify that this was really covered by the unit tests. I added an extra comment in the tests just to make that clearer :)

firewalldb/kvstores_kvdb.go
ViktorT-11: Ah snap! Thanks again, this latest push should really have addressed this now :)

PR #1084, opened 2025-06-09
[sql-42] accounts: Update SQL migration to iterate over buckets + ensure proper closure of SQL test stores 

ViktorT-11:
This PR addresses review comment https://github.com/lightninglabs/lightning-terminal/pull/1051#discussion_r2126419970, and changes the implementation of the accounts SQL migration to instead iterate over the Bbolt Buckets, rather than using the public `BoltStore` struct members.

Additionally, we add also add a `createStore` helper function for the test SQL db creation methods, that ensures that the SQL db is properly closed during the test cleanup. We will add a similar function for the `sessions` + `firewalldb` packages in their respective PRs, and also build upon that function in the upcoming PR that adds support for the `sqldb/v2` implementation, which is why that function may feel a bit too verbose for just this PR alone.

### DIFF
--- a/accounts/sql_migration_test.go
+++ b/accounts/sql_migration_test.go
@@ -39,9 +39,6 @@ func TestAccountStoreMigration(t *testing.T) {
 		*db.TransactionExecutor[SQLQueries]) {
 
 		testDBStore := NewTestDB(t, clock)
-		t.Cleanup(func() {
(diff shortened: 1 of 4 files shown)


bitromortac (approved):
LGTM :heavy_check_mark:

PR #1085, opened 2025-06-10
[sql-43] Add support for `sqldb/v2` in litd

ViktorT-11:
Currently based on #1079 

Implements step 1. of "Phase 3" in #917.

This is a draft PR to updates litd to use the new `sqldd/v2` package. Note that this with just this PR, `litd` will not utilize the capabilities of sqldb v2 to run specific post migrations steps (such as migrating the kvdb to SQL). That functionality will be added in upcoming PRs.

Instead, this PR just focuses on adding support for the new sqldb v2 package, and the functionality of the SQL stores are expected to remain the same as prior to this commit.

Note that this is only a draft PR for now, as the `sqldb/v2` is not released, and we therefore need to add a forked dependency in order to run this. Once we have an official release of the `sqldb/v2` package, I'll update this PR to use that dependency.

### DIFF
--- a/accounts/test_postgres.go
+++ b/accounts/test_postgres.go
@@ -16,7 +16,7 @@ var ErrDBClosed = errors.New("database is closed")
 
 // NewTestDB is a helper function that creates an SQLStore database for testing.
 func NewTestDB(t *testing.T, clock clock.Clock) Store {
-	return createStore(t, db.NewTestPostgresV2DB(t).BaseDB, clock)
+	return createStore(t, db.NewTestPostgresDB(t).BaseDB, clock)
 }
 
 // NewTestDBFromPath is a helper function that creates a new SQLStore with a
@@ -24,5 +24,5 @@ func NewTestDB(t *testing.T, clock clock.Clock) Store {
 func NewTestDBFromPath(t *testing.T, dbPath string,
 	clock clock.Clock) Store {
 
-	return createStore(t, db.NewTestPostgresV2DB(t).BaseDB, clock)
+	return createStore(t, db.NewTestPostgresDB(t).BaseDB, clock)
 }
--- a/db/postgres.go
+++ b/db/postgres.go
@@ -56,9 +56,9 @@ func (s *PostgresConfig) DSN(hidePassword bool) string {
 		s.DBName, sslMode)
 }
 
-// NewTestPostgresV2DB is a helper function that creates a Postgres database for
-// testing, using the sqldb v2 package's definition of the PostgresStore.
-func NewTestPostgresV2DB(t *testing.T) *sqldb.PostgresStore {
+// NewTestPostgresDB is a helper function that creates a Postgres database for
+// testing.
+func NewTestPostgresDB(t *testing.T) *sqldb.PostgresStore {
 	t.Helper()
 
 	t.Logf("Creating new Postgres DB for testing")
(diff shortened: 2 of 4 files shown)


ViktorTigerstrom:
@bitromortac & @ellemouton, I just requested reviews from both of you as I think we're ready to start reviewing this PR now.
Note though that I'll keep this in draft until we have https://github.com/lightningnetwork/lnd/pull/10175 merged into a separate branch in `lnd`.

Therefore we can't merge this PR yet, as the dependency link will change when the above has merged. The actual code in this PR is final though, so it is worth reviewing IMO :)

db/interfaces.go
```diff
@@ -1,311 +1,5 @@
 package db
 
-import (
```
ellemouton, line 3: 🤩

ellemouton (commented):
wow, this PR is absolutely impeccable: really great descriptions, super modular commits, very easy to review - great work! 🔥

session/sql_store.go
```diff
@@ -66,19 +70,37 @@ type SQLStore struct {
...
-	executor := db.NewTransactionExecutor(
-		sqlDB, func(tx *sql.Tx) SQLQueries {
-			return sqlDB.WithTx(tx)
+type SQLQueriesExecutor[T sqldb.BaseQuerier] struct {
+	*sqldb.TransactionExecutor[T]
+
+	SQLQueries
+}
+
+func NewSQLQueriesExecutor(baseDB *sqldb.BaseDB,
```
bitromortac: It seems like we don't need to export those?

bitromortac (commented):
> wow, this PR is absolutely impeccable: really great descriptions, super modular commits, very easy to review - great work! 🔥

100%, only a first pass for me (and I don't have a good view of the upstream work as well). Thanks for the great commit structure! :rocket:

go.mod
```diff
@@ -32,9 +32,9 @@ require (
 	github.com/lightninglabs/pool v0.6.6-beta
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.3
 	github.com/lightninglabs/pool/poolrpc v1.0.1
-	github.com/lightninglabs/taproot-assets v0.7.0
-	github.com/lightninglabs/taproot-assets/taprpc v1.0.11
-	github.com/lightningnetwork/lnd v0.20.0-beta
+	github.com/lightninglabs/taproot-assets v0.7.1-0.20260127120359-89b1ea3a7888
+	github.com/lightninglabs/taproot-assets/taprpc v1.0.12-0.20260121192339-ac2ec3419cec
+	github.com/lightningnetwork/lnd v0.20.0-beta.rc4.0.20251127014118-f8b5cb0e8918
```
ellemouton, line 37: these would all be bumped to tagged versions before releasing lit right?

ViktorT-11: Correct 🙂!

db/sql_migrations.go
```diff
@@ -0,0 +1,29 @@
+package db
+
+import (
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database/pgx/v5"
+	"github.com/lightningnetwork/lnd/sqldb/v2"
+)
+
+var (
+	LitdMigrationStream = sqldb.MigrationStream{
```
bitromortac, line 12: nit: maybe add godocs

PR #1089, opened 2025-06-16
channels: create test cases for two new bug fixes

guggero:
Depends on https://github.com/lightninglabs/taproot-assets/pull/1601, creates test cases for the two bug fixes included there.

### DIFF
--- a/docs/release-notes/release-notes-0.15.0.md
+++ b/docs/release-notes/release-notes-0.15.0.md
@@ -95,7 +95,7 @@
 ### Taproot Assets
 
 * Updated [`tapd` to
-  `v0.6.0-rc3`](https://github.com/lightninglabs/lightning-terminal/pull/1082).
+  `v0.6.0`](https://github.com/lightninglabs/lightning-terminal/pull/1089).
 * All Taproot Asset Channel related commands (`litcli ln ...`) can now [use
   a new `--group_key` flag to interact with grouped asset
   channels](https://github.com/lightninglabs/lightning-terminal/pull/1052).


itest/litd_custom_channels_test.go
```diff
@@ -2407,6 +2407,26 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
...
+		CltvExpiryDelta:           80,
+		FeeBaseMsat:               1000,
+		FeeProportionalMillionths: 1,
+	}
+	sendKeySendPayment(
+		t.t, dave, fabia, 1, withPayRouteHints([]*lnrpc.RouteHint{{
+			HopHints: []*lnrpc.HopHint{hinEF},
+		}}),
+	)
+	logBalance(t.t, nodes, assetID, "after satogram keysend")
```
GeorgeTsagk: would feel nice to also have the negative version of this test case present, i.e what would happen if
```
// This is not allowed as Fabia has assets on her side.
```

Now that I'm thinking about this, we can't really end up in a state where a party has assets plus below dust balance, as it is an invalid state in itself

Curious to see if there's a way where we can expect the keysend to error out

non blocking

itest/litd_custom_channels_test.go
```diff
@@ -2870,6 +2890,47 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
...
+		PayReq: invoiceResp.PaymentRequest,
+	})
+	require.NoError(t.t, err)
+
+	require.Len(t.t, req.RouteHints, 1)
+	require.Len(t.t, req.RouteHints[0].HopHints, 1)
+	invoiceHint := req.RouteHints[0].HopHints[0]
+	require.EqualValues(t.t, 31337, invoiceHint.FeeBaseMsat)
+	require.EqualValues(t.t, 443322, invoiceHint.FeeProportionalMillionths)
+	require.EqualValues(t.t, 19, invoiceHint.CltvExpiryDelta)
```
GeorgeTsagk, line 2945: why not finish things up by paying this invoice, to make sure the policies were correctly met by the sender?

go.mod
```diff
@@ -32,8 +32,8 @@ require (
 	github.com/lightninglabs/pool v0.6.6-beta
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.3
 	github.com/lightninglabs/pool/poolrpc v1.0.1
-	github.com/lightninglabs/taproot-assets v0.6.0-rc3
-	github.com/lightninglabs/taproot-assets/taprpc v1.0.6
+	github.com/lightninglabs/taproot-assets v0.6.0
+	github.com/lightninglabs/taproot-assets/taprpc v1.0.7
```
ViktorT-11, line 36: nit: While we're at this, we could update this in `litrpc` as well :).

ViktorTigerstrom (approved):
LGTM! Don't have too much context in regards to the `tapd` itests, but from what I can see, they look good :rocket:

go.mod
guggero: Good catch, done.

PR #1092, opened 2025-06-17
[sql-44] firewalldb: add migration code for privacy mapper from kvdb to SQL

ViktorTigerstrom:
Based on #1079.

This PR introduces the migration logic for transitioning the privacy mapper from kvdb to SQL.

I will rebase this on the new version of #1079, once the current feedback has been addressed after we agree on the approach there. 

As this PR is heavily dependent on #1079, this is blocked on the dependent PR until it has been merged. This is therefore not ready for review.
 
### DIFF
--- a/firewalldb/sql_migration.go
+++ b/firewalldb/sql_migration.go
@@ -15,6 +15,11 @@ import (
 // line length in the migration code.
 type kvParams = sqlc.InsertKVStoreRecordParams
 
+// privacyPairs is a type alias for a map that holds the privacy pairs, where
+// the outer key is the group ID, and the value is a map of real to pseudo
+// values.
+type privacyPairs = map[int64]map[string]string
+
 // MigrateFirewallDBToSQL runs the migration of the firwalldb stores from the
 // bbolt database to a SQL database. The migration is done in a single
 // transaction to ensure that all rows in the stores are migrated or none at
@@ -35,10 +40,14 @@ func MigrateFirewallDBToSQL(ctx context.Context, kvStore *bbolt.DB,
 		return err
 	}
 
+	err = migratePrivacyMapperDBToSQL(ctx, kvStore, tx)
+	if err != nil {
+		return err
+	}
+
 	log.Infof("The rules DB has been migrated from KV to SQL.")
 
-	// TODO(viktor): Add migration for the privacy mapper and the action
-	// stores.
+	// TODO(viktor): Add migration for the action stores.
 
 	return nil
 }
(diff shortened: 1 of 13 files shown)


ellemouton:
@ViktorTigerstrom  - feel free to push a base branch to upstream so that this pr can just contain the commits that are relevant

ViktorTigerstrom:
Rebased this on the latest version of #1079, as that PR feels good to go, and this PR is therefore close to not being blocked by that PR (and therefore ready to review).

ViktorTigerstrom:
Requested a review by both of you @ellemouton & @bitromortac, even though #1079 isn't merged yet, just so you know that it is ready for review once that's been merged.

firewalldb/sql_migration_test.go
```diff
@@ -34,6 +34,11 @@ var (
 	testEntryValue = []byte{1, 2, 3}
 )
 
+// expectedResult represents the expected result of a migration test.
+type expectedResult struct {
```
ellemouton, line 39: very nice 🙏

firewalldb/sql_migration.go
```diff
@@ -490,3 +499,299 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+	// 1) Collect all privacy pairs from the KV store.
+	privPairs, err := collectPrivacyPairs(ctx, kvStore, sqlTx)
+	if err != nil {
+		return fmt.Errorf("error migrating privacy mapper store: %w",
+			err)
+	}
+
+	// 2) Insert all collected privacy pairs into the SQL database.
+	err = insertPrivacyPairs(ctx, sqlTx, privPairs)
+	if err != nil {
```
ellemouton, line 517: we may want to tests the performance of this one as i think unlike the kv-store db, this one could be quite populated. 

So for this one it _might_ make sense to migrate on the fly instead of collecting everything in memory first.

firewalldb/sql_migration.go
```diff
@@ -490,3 +499,299 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+		}
+
+		err = sqlTx.InsertPrivacyPair(
+			ctx, sqlc.InsertPrivacyPairParams{
+				GroupID:   groupID,
+				RealVal:   realVal,
+				PseudoVal: pseudoVal,
+			},
+		)
+		if err != nil {
```
ellemouton: can we not potentially do this in one go with a single query that has a conflict rule?

firewalldb/sql_migration.go
```diff
@@ -490,3 +499,299 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+func collectGroupPairs(bkt *bbolt.Bucket) (map[string]string, error) {
+	var (
+		realToPseudoRes map[string]string
+		pseudoToRealRes map[string]string
+		err             error
+		missMatchErr    = errors.New("privacy mapper pairs mismatch")
+	)
+
+	if realBkt := bkt.Bucket(realToPseudoKey); realBkt != nil {
+		realToPseudoRes, err = collectPairs(realBkt)
```
ellemouton, line 604: just checking: do we have a test somewhere that makes sure that things dont fail on an empty kvdb?

firewalldb/sql_migration_test.go
```diff
@@ -680,6 +733,136 @@ func randomKVEntries(t *testing.T, ctx context.Context,
 
 	return &expectedResult{
 		kvEntries: fn.Some(insertedEntries),
+		// No privacy pairs are inserted in this test.
+		privPairs: fn.None[privacyPairs](),
```
ellemouton: we should have a test that includes both

firewalldb/sql_migration_test.go
```diff
@@ -34,6 +34,11 @@ var (
 	testEntryValue = []byte{1, 2, 3}
 )
 
+// expectedResult represents the expected result of a migration test.
+type expectedResult struct {
+	kvEntries fn.Option[[]*kvEntry]
```
bitromortac: is it necessary to have an option type here, as an empty `[]*kvEntry` already signals that, not? (also for the pair data)

firewalldb/sql_migration_test.go
```diff
@@ -213,6 +218,20 @@ func TestFirewallDBMigration(t *testing.T) {
 		}
 	}
 
+	// The assertMigrationResults function will currently assert that
```
bitromortac: nit: I'd rather not put comments of what will happen in the future, because it may add confusion for somebody who's not involved in the process and it also may never happen, or we forget to update the comment

firewalldb/sql_migration.go
```diff
@@ -87,10 +92,14 @@ func MigrateFirewallDBToSQL(ctx context.Context, kvStore *bbolt.DB,
 		return err
```
bitromortac, line 89:
```
// Note that this migration currently only migrates the kvstores, but will be
// extended in the future to also migrate the privacy mapper and action stores.
```
this is also an example of a "future" comment, where the comment wasn't updated

firewalldb/sql_migration.go
```diff
@@ -490,3 +499,299 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+			"failed: %w", err)
+	}
+
+	log.Infof("Migration of the privacy mapper stores to SQL completed. "+
+		"Total number of rows migrated: %d", len(privPairs))
+	return nil
+}
+
+// collectPrivacyPairs collects all privacy pairs from the KV store and
+// returns them as the privacyPairs type alias.
```
bitromortac: nitty: `and returns them as the privacyPairs type alias.` I'd leave that away, as it's obvious from the function signature

firewalldb/sql_migration.go
```diff
@@ -490,3 +499,299 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+				GroupID:   groupID,
+				PseudoVal: pseudoVal,
+			},
+		)
+		if err == nil {
+			return fmt.Errorf("duplicate privacy pair %s:%s: %w",
+				realVal, pseudoVal, ErrDuplicatePseudoValue)
+		} else if !errors.Is(err, sql.ErrNoRows) {
+			return err
+		}
```
bitromortac: are those checks necessary as they look to be covered by `collectGroupPairs`?

firewalldb/sql_migration_test.go
```diff
@@ -680,6 +733,136 @@ func randomKVEntries(t *testing.T, ctx context.Context,
...
 	return &expectedResult{
 		kvEntries: fn.Some(insertedEntries),
+		// No privacy pairs are inserted in this test.
+		privPairs: fn.None[privacyPairs](),
+	}
+}
+
+// multiSessionsOnePrivPairs inserts 1 session with 1 privacy pair into the
+// boltDB.
+func oneSessionAndPrivPair(t *testing.T, ctx context.Context,
```
bitromortac, line 766: nit: we could use `createPrivacyPairs` directly

firewalldb/sql_migration_test.go
```diff
@@ -680,6 +733,136 @@ func randomKVEntries(t *testing.T, ctx context.Context,
...
+		sqlGroupID, err := sessSQLStore.GetSessionIDByAlias(
+			ctx, groupID[:],
+		)
+		require.NoError(t, err)
+
+		groupPairs := make(map[string]string)
+
+		for j := 0; j < numPairsPerSession; j++ {
+			realKey := fmt.Sprintf("real-%d-%d", i, j)
+			pseudoKey := fmt.Sprintf("pseudo-%d-%d", i, j)
```
bitromortac: would it add anything if we'd let real/pseudo keys be the same for the sessions, to leave `i` away?

firewalldb/sql_migration.go
ViktorT-11: Good idea, I will benchmark this as a follow-up and see if it makes sense to process every privacy pair individually, or if this version works as well. I've prepared a separate branch that changes the code here to processes each pair one by one here:
https://github.com/ViktorTigerstrom/lightning-terminal/tree/2025-06-migrate-privacy-mapper-individual-pair-processing

That branch adds one additional commit that changes this code. If you think it already makes sense to change the code to that version already before benchmarking, let me know and I'll fixup that last commit into this PR.

firewalldb/sql_migration.go
ViktorT-11: Actually, I noticed that since the `privacy_mapper` SQL table already has unique constraints, those additional duplication checks becomes redundant. I included those as they were present in the `sql_store.go` insertion function. As those checks therefore also are redundant there, I included an extra commit 550637205f70a4a70cf1e519705a57ef6dd5cdb8 that removes them there as well. If you think we shouldn't remove them there as well, let me know and I'll drop that commit :)

firewalldb/sql_migration.go
ViktorT-11: I think the "empty" test in every migration does exactly that? Note that the real kvdb to SQL migration will also initiate the actual kvdb stores by calling `accounts.NewBoltStore`, `session.NewDB` & `firewalldb.NewBoltDB` before the migration starts, which initiates the base buckets.

firewalldb/sql_migration_test.go
ViktorT-11: Ah good idea! I added a `randomFirewallDBEntries` that combines the results of `randomKVEntries` & `randomPrivacyPairs` into one test :)

firewalldb/sql_migration_test.go
ViktorT-11: Ah true, agree :pray:! I added a new commit 4d2a4233c872443d68c826374936660bb4ab88e6 that removes (hopefully) all of those comments

firewalldb/sql_migration.go
ViktorT-11: Addressed with https://github.com/lightninglabs/lightning-terminal/pull/1092#discussion_r2226880547

firewalldb/sql_migration_test.go
ViktorT-11: I added these functions to avoid having to change the signature of the `populateDB` function, and to not to not have to define the `populateDB` function within the actual `tests` list definition. If you have an idea of how to do it in a cleaner way than having these separate functions, let me know :)

firewalldb/sql_migration_test.go
ViktorT-11: Hmm interesting idea! I changed this so that the `realKey` values will be the same across sessions (i.e. removed the `i` part for the `realKey`), but kept the `pseudoKey` value as is. The motivation behind this is that in real world data, there will often be cases where the `real` value in the privacy pairs will be the same across sessions, while the `pseudo` value should differ. If you disagree with that reasoning and think we should also change the `pseudoKey` values to be the same across sessions, let me know :)

firewalldb/sql_migration_test.go
ViktorT-11: Hmm I think it's a bit cleaner to use an option type here to make it explicitly clear if there's any expected `kvEntries` or `privPairs` set. I noticed that we have a few other cases where the option type is used in combination with a `[]` in `litd` & `lnd`. I therefore kept this as is for now. I don't have super strong preference though, so if you do have strong preference for not using the option type here, I'll update to remove it.

firewalldb/sql_migration.go
ViktorT-11: >  I included an extra commit https://github.com/lightninglabs/lightning-terminal/commit/550637205f70a4a70cf1e519705a57ef6dd5cdb8

Actually, I noticed that we have those checks explicitly in the `kvdb` stores as well, so I think this falls back to what we've discussed before: Don't change or optimize & change the definition in the SQL store before we've migrated, and instead mimic the `kvdb` store behaviour. Therefore I updated the PR again to remove the extra commit, as I believe we need to wait until after the migration has been shipped before we can change that.

firewalldb/sql_migration_test.go
```diff
@@ -676,6 +730,156 @@ func randomKVEntries(t *testing.T, ctx context.Context,
...
+func createPrivacyPairs(t *testing.T, ctx context.Context,
+	boltDB *BoltDB, sessionStore session.Store, numSessions int,
+	numPairsPerSession int) *expectedResult {
+
+	pairs := make(privacyPairs)
+
+	sessSQLStore, ok := sessionStore.(*session.SQLStore)
+	require.True(t, ok)
+
+	for i := 0; i < numSessions; i++ {
```
bitromortac: nit: this and a case below can be simplified to the range int syntax, same in `randomPrivacyPairs`

we could also remove the random identifiers from the `NewSession` calls in these tests similarly like we did in the other PR, to remove a bit of noise

firewalldb/sql_migration_test.go
```diff
@@ -272,6 +303,26 @@ func TestFirewallDBMigration(t *testing.T) {
 			name:       "random kv entries",
 			populateDB: randomKVEntries,
 		},
+		{
+			name:       "one session and privacy pair",
+			populateDB: oneSessionAndPrivPair,
+		},
+		{
+			name:       "multiple sessions with one privacy pair",
+			populateDB: multiSessionsOnePrivPairs,
```
bitromortac: the naming is inconsistent with the action of the function

firewalldb/sql_migration_test.go
ellemouton: nit: can remove "currently"?

ellemouton: ah, i see it's done in next commit... could probs have just been squashed here. but all good! nbd

firewalldb/sql_migration.go
```diff
@@ -487,3 +496,272 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+	// after they were inserted.
+	err = validatePrivacyPairsMigration(ctx, sqlTx, privPairs)
+	if err != nil {
+		return fmt.Errorf("migration validation of privacy pairs "+
+			"failed: %w", err)
+	}
+
+	log.Infof("Migration of the privacy mapper stores to SQL completed. "+
+		"Total number of rows migrated: %d", len(privPairs))
+	return nil
```
ellemouton, line 534: nit: nl above return

firewalldb/sql_migration_test.go
bitromortac: ah, I see, using anonymous functions like in the empty case would do as well, would have made it easier to spot the discrepancy in `multiSessionsOnePrivPairs`, but fine like it is

firewalldb/sql_migration_test.go
bitromortac: cool, thanks!

firewalldb/sql_migration_test.go
bitromortac: we can leave it like it is, I think

bitromortac (approved):
LGTM nice :tada:

firewalldb/sql_migration.go
```diff
@@ -490,3 +496,272 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+	}
+
+	if len(realToPseudoRes) != len(pseudoToRealRes) {
+		return nil, missMatchErr
+	}
+
+	for realVal, pseudoVal := range realToPseudoRes {
+		if rv, ok := pseudoToRealRes[pseudoVal]; !ok || rv != realVal {
+			return nil, missMatchErr
+		}
```
ellemouton: worth adding flavour to these error strings

firewalldb/sql_migration.go
```diff
@@ -490,3 +496,272 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+	return nil
+}
+
+// insertGroupPairs inserts the privacy pairs for a specific group into
+// the SQL database. It checks for duplicates before inserting, and returns
+// an error if a duplicate pair is found. The function takes a map of real
+// to pseudo values, where the key is the real value and the value is the
+// corresponding pseudo value.
+func insertGroupPairs(ctx context.Context, sqlTx SQLQueries,
+	pairs map[string]string, groupID int64) error {
```
ellemouton: nit: paramater order. makes more sense to have static/dependent things first and then the variable thing. ie, have groupID before pairs

firewalldb/sql_migration.go
```diff
@@ -490,3 +496,272 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
...
+	return nil
+}
+
+// validateGroupPairsMigration validates that the migrated privacy pairs for
+// a specific group match the original values in the KV store. It checks that
+// for each real value, the pseudo value in the SQL database matches the
+// original pseudo value, and vice versa. If any mismatch is found, it returns
+// an error indicating the mismatch.
+func validateGroupPairsMigration(ctx context.Context, sqlTx SQLQueries,
+	pairs map[string]string, groupID int64) error {
```
ellemouton: same here with ordering

firewalldb/sql_migration_test.go
```diff
@@ -213,46 +216,113 @@ func TestFirewallDBMigration(t *testing.T) {
...
+	// original entries in the BoltDB.
+	assertMigrationResults := func(t *testing.T, sqlStore *SQLDB,
+		expRes *expectedResult) {
+
+		// If the expected result contains kv records, then we
+		// assert that the kv store migration results match
+		// the expected results.
+		expRes.kvEntries.WhenSome(func(kvEntries []*kvEntry) {
+			assertKvStoreMigrationResults(t, sqlStore, kvEntries)
+		})
```
ellemouton: and when not some? should we not do an assertion then too?

firewalldb/sql_migration.go
```diff
@@ -490,3 +496,272 @@ func verifyBktKeys(bkt *bbolt.Bucket, errorOnKeyValues bool,
 		return fmt.Errorf("unexpected key found: %s", key)
 	})
 }
+
+func migratePrivacyMapperDBToSQL(ctx context.Context, kvStore *bbolt.DB,
```
ellemouton, line 503: nit: missing godoc

ellemouton (approved):
non-blocking nits only from my side!
Really beautiful piece of code 🤩

firewalldb/sql_migration.go
ViktorT-11: Addressed :)

firewalldb/sql_migration.go
ViktorT-11: Ah yeah, addressed :)

firewalldb/sql_migration_test.go
ViktorT-11: I changed this, and by doing so I also changed to address @bitromortac's old comment here:
https://github.com/lightninglabs/lightning-terminal/pull/1092#discussion_r2213122438

Doing so also required me to update the `assertPrivacyMapperMigrationResults` slightly to make sure no privacy pairs exists in the SQL db which we do not expect to be there.

firewalldb/sql_migration_test.go
ViktorT-11: Done :)

firewalldb/sql_migration_test.go
ViktorT-11: Ah yeah, good catch!

PR #1106, opened 2025-06-30
Custom channels liquidity edge case: AddInvoice with existing route hints

GeorgeTsagk:
## Description

Adds a new edge case to the custom channels test suite, where we call the `taprpc.AddInvoice` with existing route hints, which include a valid accepted quote.

Related tapd PR: https://github.com/lightninglabs/taproot-assets/pull/1627

### DIFF
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -2959,6 +2959,42 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
 	payInvoiceWithSatoshi(
 		t.t, dave, invoiceResp, withFeeLimit(100_000_000),
 	)
+
+	logBalance(t.t, nodes, assetID, "after policy checks")
+
+	resBuy, err := daveTap.RfqClient.AddAssetBuyOrder(
+		ctx, &rfqrpc.AddAssetBuyOrderRequest{
+			AssetSpecifier: &assetSpecifier,
+			AssetMaxAmt:    1_000,
+			Expiry:         uint64(inOneHour.Unix()),
+			PeerPubKey:     charlie.PubKey[:],
+			TimeoutSeconds: 100,
+		},
+	)
+	require.NoError(t.t, err)
+
+	scid := resBuy.GetAcceptedQuote().Scid
+
+	invResp := createAssetInvoice(
+		t.t, charlie, dave, 1_000, assetID,
+		withInvGroupKey(groupID), withRouteHints([]*lnrpc.RouteHint{
+			{
+				HopHints: []*lnrpc.HopHint{
+					{
+						NodeId: charlie.PubKeyStr,
+						ChanId: scid,
+					},
+				},
+			},
+		}),
+	)
+
+	payInvoiceWithAssets(
+		t.t, charlie, dave, invResp.PaymentRequest, assetID,
+		withGroupKey(groupID),
+	)
+
+	logBalance(t.t, nodes, assetID, "after invoice with route hints")
 }
 
 // testCustomChannelsLiquidityEdgeCases is a test that runs through some


guggero (approved):
Thanks, LGTM 🎉 

I've increased the linter timeout in the `tapd-main-branch` branch. If you rebase the failures should go away.

go.mod
```diff
@@ -248,6 +248,4 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 // automatically, so we need to add it manually.
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
 
-// tapd wants v0.19.0-12, but loop can't handle that yet. So we'll just use the
-// previous version for now.
-replace github.com/lightninglabs/lndclient => github.com/lightninglabs/lndclient v0.19.0-11
+replace github.com/lightninglabs/taproot-assets => github.com/lightninglabs/taproot-assets v0.6.1-0.20250725201706-0e736bb0d7b1
```
ffranr: why do we have this replace, why not just bump the dep directly

itest/assets_test.go
```diff
@@ -1687,14 +1687,18 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
...
-	var rfqBytes []byte
+	var rfqBytes, peerPubKey []byte
 	cfg.rfq.WhenSome(func(i rfqmsg.ID) {
 		rfqBytes = make([]byte, len(i[:]))
 		copy(rfqBytes, i[:])
 	})
 
+	if rfqPeer != nil {
+		peerPubKey = rfqPeer.PubKey[:]
+	}
```
ffranr, line 1769: `var peerPubKey []byte` just above this code block?

itest/assets_test.go
```diff
@@ -1719,7 +1719,13 @@ func payInvoiceWithAssets(t *testing.T, payer, rfqPeer *HarnessNode,
 	// was established, no network or auth error), we expect the error to be
 	// returned on the first read on the stream.
 	if cfg.errSubStr != "" {
-		_, err := stream.Recv()
+		msg, err := stream.Recv()
+
+		// On errors we still get an empty set of RFQs as a response.
+		if msg.GetAcceptedSellOrders() != nil {
+			_, err = stream.Recv()
+		}
```
ffranr, line 1798: I think the comment above is stale?
```
we expect the error to be
	// returned on the first read on the stream.
```

itest/assets_test.go
```diff
@@ -1355,7 +1355,7 @@ func sendAssetKeySendPayment(t *testing.T, src, dst *HarnessNode, amt uint64,
 		return
 	}
 
-	result, err := getAssetPaymentResult(stream, false)
+	result, _, err := getAssetPaymentResult(t, stream, false)
```
ffranr: I think we should split this commit up a bit. The additional return value in `getAssetPaymentResult` could be a separate commit perhaps.

And then extending `testCustomChannelsMultiRFQ` could be another.

itest/litd_accounts_test.go
```diff
@@ -443,43 +445,75 @@ func getPaymentResult(stream routerrpc.Router_SendPaymentV2Client,
 	}
 }
 
-func getAssetPaymentResult(
+func getAssetPaymentResult(t *testing.T,
 	s tapchannelrpc.TaprootAssetChannels_SendPaymentClient,
-	isHodl bool) (*lnrpc.Payment, error) {
+	isHodl bool) (*lnrpc.Payment, rfqmath.FixedPoint[rfqmath.BigInt],
+	error) {
```
ffranr: I think this should return a new struct, something like `TapPaymentT`, which we can expand over time.

That way we could include the `lnrpc.Payment` instance as well as both potential quotes (options), for example.

go.mod
GeorgeTsagk: This was part of the stale / non-rebased code, using latest refs now as defined in `tapd-main-branch`

itest/assets_test.go
GeorgeTsagk: After rebasing I realized that the commit which updates `itest/assets_test.go` has been already merged in `tapd-main-branch`: https://github.com/lightninglabs/lightning-terminal/commit/505aa236a7d9522d29382258309712b9900ca074

itest/assets_test.go
GeorgeTsagk: https://github.com/lightninglabs/lightning-terminal/pull/1106#discussion_r2354841663

itest/assets_test.go
GeorgeTsagk: Will create a new commit to update the nits

itest/assets_test.go
GeorgeTsagk: Was already extracted as a separate commit, see https://github.com/lightninglabs/lightning-terminal/pull/1106#discussion_r2354841663

itest/litd_accounts_test.go
```diff
@@ -445,10 +445,24 @@ func getPaymentResult(stream routerrpc.Router_SendPaymentV2Client,
...
+	// lndResult contains the lnd part of the payment result.
+	lndResult *lnrpc.Payment
+
+	// assetRate contains the asset rate that was used to convert the assets
+	// to sats.
+	assetRate rfqmath.FixedPoint[rfqmath.BigInt]
+
+	// err contains the returned error from the payment.
+	err error
+}
```
ffranr: you could use `fn.Result` here to handle the `error`. Then`TapPaymentResult` can be `TapPayment`.

or just return `(TapPayment, error)` like usual?

GeorgeTsagk: Went with the `(TapPayment, error)` approach, my bad I thought you wanted to wrap everything under a single type

ffranr: error is usually a special case

PR #1113, opened 2025-07-11
[sql-45] package queries & models present at the time of the kvdb to sql migration

ViktorT-11:
Based on and blocked by #1085

Implements step 2. of "Phase 3" in https://github.com/lightninglabs/lightning-terminal/issues/917.

This is a draft PR that introduces a new sqlc package called `sqlcmig6`, which packages the exact SQL queries & models present at the time of the kvdb to sql migration. This PR then also updates the kvdb to sql migrations to use the queries & models from that package, instead of the standard `sqlc` queries. This is required, as the `sqlc` package, queries & models may change in such a way in the future, that they are no longer compatible with the kvdb to sql migrations, and we therefore need to enable functionality that runs the migrations with the exact queries & models present at the time of when the kvdb to sql migration was introduced. 

Note that this is only a draft PR for now, as the sqldb/v2 is not released, and we therefore need to add a forked dependency in order to run this. Once we have an official release of the sqldb/v2 package, I'll update this PR to use that dependency.

### DIFF
--- a/accounts/store_sql.go
+++ b/accounts/store_sql.go
@@ -52,9 +52,6 @@ type SQLQueries interface {
 	UpdateAccountBalance(ctx context.Context, arg sqlc.UpdateAccountBalanceParams) (int64, error)
 	UpdateAccountExpiry(ctx context.Context, arg sqlc.UpdateAccountExpiryParams) (int64, error)
 	UpdateAccountLastUpdate(ctx context.Context, arg sqlc.UpdateAccountLastUpdateParams) (int64, error)
-	// UpdateAccountAliasForTests is a query intended only for testing
-	// purposes, to change the account alias.
-	UpdateAccountAliasForTests(ctx context.Context, arg sqlc.UpdateAccountAliasForTestsParams) (int64, error)
 	UpsertAccountPayment(ctx context.Context, arg sqlc.UpsertAccountPaymentParams) error
 	GetAccountInvoice(ctx context.Context, arg sqlc.GetAccountInvoiceParams) (sqlc.AccountInvoice, error)
 }
--- a/db/sqlc/accounts.sql.go
+++ b/db/sqlc/accounts.sql.go
@@ -313,26 +313,6 @@ func (q *Queries) SetAccountIndex(ctx context.Context, arg SetAccountIndexParams
 	return err
 }
 
-const updateAccountAliasForTests = `-- name: UpdateAccountAliasForTests :one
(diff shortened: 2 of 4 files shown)


db/sqlcmig6/privacy_pairs.sql.go
```diff
@@ -0,0 +1,91 @@
+package sqlcmig6
```
bitromortac, line 1: nit: file name, but preexisting :)

db/sqlcmig6/db.go
```diff
@@ -25,3 +26,23 @@ func (q *Queries) WithTx(tx *sql.Tx) *Queries {
...
+	executor := sqldb.NewTransactionExecutor(
+		baseDB, func(tx *sql.Tx) *Queries {
+			return queries.WithTx(tx)
+		},
+	)
+	return &TxExecutor[*Queries]{
+		TransactionExecutor: executor,
+		Queries:             queries,
+	}
+}
```
bitromortac: I got the suggestion to use this instead, to force transaction usage and not accidentally use auto-commit (needs updates to remove generics at call sites). Could mabye be useful in db code as well?
```golang
type TxExecutor struct {
	*sqldb.TransactionExecutor[*Queries]
}

func NewTxExecutor(baseDB *sqldb.BaseDB,
	queries *Queries) *TxExecutor {

	executor := sqldb.NewTransactionExecutor(
		baseDB, func(tx *sql.Tx) *Queries {
			return queries.WithTx(tx)
		},
	)
	return &TxExecutor{
		TransactionExecutor: executor,
	}
}
```

accounts/sql_migration.go
```diff
@@ -151,17 +154,79 @@ func getBBoltAccounts(db kvdb.Backend) ([]*OffChainBalanceAccount, error) {
 	return accounts, nil
 }
 
+// getAndMarshalAccount retrieves the account with the given ID. If the account
```
bitromortac: nit: docstring

bitromortac (approved):
LGTM :rocket:, thank you for the nice structure!

db/sqlcmig6/db.go
ViktorT-11: Sure, addressed 😃

db/sqlcmig6/privacy_pairs.sql.go
ViktorT-11: Updated :)

PR #1114, opened 2025-07-11
[sql-46] trigger kvdb to sql migration using `sqldb/v2`

ViktorT-11:
Based on and blocked by https://github.com/lightninglabs/lightning-terminal/pull/1113

Implements step 3. of "Phase 3" in https://github.com/lightninglabs/lightning-terminal/issues/917.

This PR implements the functionality to use `sqldb/v2` to trigger the kvdb to sql migration under the `dev` build-tagged `litd` builds, using code migrations which `sqldb/v2` adds support for. This PR also adds support for having separate `dev` sql migrations, which are only run under build-tagged `litd` builds.

Note that this is only a draft PR for now, as the sqldb/v2 is not released, and we therefore need to add forked dependencies in order to run this. Once we have an official release of the sqldb/v2 package, I'll update this PR to use that dependency.


db/migrationstreams/log.go
```diff
@@ -0,0 +1,25 @@
+package migrationstreams
```
bitromortac: bit: maybe a bit long, `migstreams`?

accounts/store_kvdb.go
```diff
@@ -24,7 +24,7 @@ import (
 const (
 	// DBFilename is the filename within the data directory which contains
 	// the macaroon stores.
-	DBFilename = "accounts.db"
+	DBFilename = "accounts.DB"
```
bitromortac: I assume this rename went wrong? Also the other comment renames are not needed.

db/sql_migrations.go
```diff
@@ -3,16 +3,24 @@
...
-var (
-	// LitdMigrationStream defines the SQL migration stream used to create
+// MakeMigrationStreams creates the migration streams for production
+// environments.
+func MakeMigrationStreams(_ context.Context, _ lnrpc.LightningClient, _ string,
+	_ clock.Clock) []sqldb.MigrationStream {
+
+	// migStream defines the SQL migration stream used to create
 	// and upgrade LiT's SQL schema.
-	LitdMigrationStream = sqldb.MigrationStream{
```
bitromortac, line 12: nit: you could have introduced MakeMigrationStreams before

db/migstreams/post_migration_callbacks_dev.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func kvdbToSqlMigrationCallback(ctx context.Context,
+	basicClient lnrpc.LightningClient, macPath string, _ *sqldb.BaseDB,
+	clock clock.Clock, q *sqlcmig6.Queries) error {
+
+	start := time.Now()
+	log.Infof("Starting KVDB to SQL migration for all stores")
+
+	accountStore, err := accounts.NewBoltStore(
```
bitromortac, line 74: do we need to close the database(s)?

db/sql_migrations.go
ViktorT-11: Kept this one specifically as is, as I personally think it makes more sense to only introduce it when it's actually needed, especially given the functions parameters that do not make sense before the commit where the function is introduced.

db/migrationstreams/log.go
ViktorT-11: Addressed :)

accounts/store_kvdb.go
ViktorT-11: Good catch!

db/migstreams/post_migration_callbacks_dev.go
ViktorT-11: Yeah probably a good idea! Added closing of the different stores created here :)

ellemouton (approved):
lgtm!

PR #1119, opened 2025-07-22
dev.Dockerfile: allow forcing lnd/tapd/taprpc/loop repo - rebased

ZZiigguurraatt:
rebased version of https://github.com/lightninglabs/lightning-terminal/pull/1118

We can now use the LND_REPO, TAPROOT_ASSETS_REPO, TAPRPC_REPO, and
LOOP_RPC build arguments to force a specific repo to be used so that
commits referenced by LND_VERSION, TAPROOT_ASSETS_VERSION,
TAPRPC_VERSION, and LOOP_VERSION don't have to exist in the default
repository. If any of these build arguments are not defined, the build
continues using the default repository for that module.

NOTE: If these arguments ARE defined then the corresponding
_VERSION argument MUST also be defined, otherwise the build continues
using the default repository defined for that module.

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -30,13 +30,35 @@ COPY --from=nodejsbuilder /go/src/github.com/lightninglabs/lightning-terminal /g
 # queries required to connect to linked containers succeed.
 ENV GODEBUG netdns=cgo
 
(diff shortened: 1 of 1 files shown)


guggero (approved):

ViktorTigerstrom (approved):
Thanks, uTACK LGTM :rocket:!

PR #1125, opened 2025-07-31
Update litcli for multi-rfq send

GeorgeTsagk:
## Description

With the introduction of [multi-rfq send](https://github.com/lightninglabs/taproot-assets/pull/1613) we added a new response item in the send payment stream which includes all of the acquired quotes in an array.

This PR updates the RPC client that handles the stream in the context of litcli to correctly read & parse all the possible response items.

### DIFF
--- a/docs/release-notes/release-notes-0.16.0.md
+++ b/docs/release-notes/release-notes-0.16.0.md
@@ -20,6 +20,11 @@
 ### Technical and Architectural Updates
 
 ## RPC Updates
+- [Updated litcli](https://github.com/lightninglabs/lightning-terminal/pull/1125)
+to support sending an asset payment without specifying a single peer. This will
+use the new multi-rfq feature of taproot-assets which will automatically pick
+the rfq peers for you. The field `rfq_peer_pubkey` is now optional for both
+adding invoices and sending payments.
 
 ## Integrated Binary Updates
 


cmd/litcli/ln.go
```diff
@@ -237,74 +236,73 @@ type resultStreamWrapper struct {
...
-			// We derive the minimum amount.
-			minAmt := rfqmath.UnitsToMilliSatoshi(unit, *rate)
-
-			// We return the error to the user.
-			return nil, fmt.Errorf("smallest payment with asset "+
-				"rate %v is %v, cannot send %v",
-				rate.ToUint64(), minAmt, amountMsat)
-		}
-
 		msatPerUnit := uint64(w.amountMsat) / numUnits
```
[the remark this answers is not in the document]
GeorgeTsagk: Reason I removed this check was that this is done within the SendPayment RPC endpoint anyway, so felt redundant to make the same checks in the litcli context

Could just skip logging here if that is the case, we'll get an error if our sanity checks were triggered

cmd/litcli/ln.go
```diff
@@ -237,74 +236,73 @@
...
+			" peer %s with SCID %d\n", numUnits, msatPerUnit,
 			quote.Peer, quote.Scid)
 
-		resp, err = w.stream.Recv()
+		return nil
+	}
+
+	// A boolean to indicate whether the first quote was printed via the
+	// legacy single-rfq response field.
+	legacyFirstPrint := false
```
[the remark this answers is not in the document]
GeorgeTsagk: We know that the single quote response arrives first on the stream

cmd/litcli/ln.go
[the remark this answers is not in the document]
GeorgeTsagk: Changed to skip the logging if produced num units is 0

cmd/litcli/ln.go
```diff
@@ -210,9 +210,8 @@ var (
 	rfqPeerPubKeyFlag = cli.StringFlag{
 		Name: "rfq_peer_pubkey",
 		Usage: "(optional) the public key of the peer to ask for a " +
-			"quote when converting from assets to sats; must be " +
-			"set if there are multiple channels with the same " +
-			"asset ID present",
+			"quote when converting from assets to sats; if left " +
+			"unset then rfq peers will be picked automatically",
```
ViktorT-11, line 214: hmmm not seeing that the rest of the changes in this PR actually changes this? If this is due to some other change that's already been merged previously, consider splitting this change into a separate commit and explain why this is now added.

cmd/litcli/ln.go
```diff
@@ -237,74 +236,80 @@ type resultStreamWrapper struct {
...
+				if legacyFirstPrint {
+					legacyFirstPrint = false
+					continue
+				}
+
+				err := printQuote(quote)
+				if err != nil {
+					return nil, err
+				}
+			}
```
ViktorT-11, line 302: Just checking that this is correct behaviour:
There is no `return` statement for the `*tchrpc.SendPaymentResponse_AcceptedSellOrder` & `*tchrpc.SendPaymentResponse_AcceptedSellOrders` cases here unless there's no error, meaning that they won't break the for loop without any errors being thrown somewhere in the next loop iteration. Is that correct behaviour?

ViktorT-11 (commented):
The code generally locks good :rocket:! Though I lack quite a bit of context to fully understand this change, so generally just reviewing from a code perspective.

Leaving a few questions that could or could not be helpful. 
Other then the suggestions above, I think it would also be helpful for reviewers if you add more details to the commit message description.

ViktorT-11:
Also, this would need release notes updates, whenever this gets merged to `master`.

cmd/litcli/ln.go
GeorgeTsagk: Yeah great observation. That's actually how the RPC endpoint behaves.

We'll first return any quote-related information, then we'll follow up with the payment result.

cmd/litcli/ln.go
```diff
@@ -237,74 +236,80 @@ type resultStreamWrapper struct {
...
+				err)
 		}
 
 		amountMsat := lnwire.MilliSatoshi(w.amountMsat)
 		milliSatsFP := rfqmath.MilliSatoshiToUnits(amountMsat, *rate)
 		numUnits := milliSatsFP.ScaleTo(0).ToUint64()
 
-		// If the calculated number of units is 0 then the asset rate
-		// was not sufficient to represent the value of this payment.
+		// The purpose of this function is just to print, so let's avoid
```
darioAnongba: This should not be a comment in the code (since it references deleted code), it's more like a remark on the PR or a commit description.

cmd/litcli/ln.go
```diff
@@ -237,74 +236,80 @@ type resultStreamWrapper struct {
...
-				"rate %v is %v, cannot send %v",
-				rate.ToUint64(), minAmt, amountMsat)
+			return nil
 		}
 
 		msatPerUnit := uint64(w.amountMsat) / numUnits
 
 		fmt.Printf("Got quote for %v asset units at %v msat/unit from "+
-			"peer %s with SCID %d\n", numUnits, msatPerUnit,
+			" peer %s with SCID %d\n", numUnits, msatPerUnit,
```
darioAnongba: double space?
```suggestion
			"peer %s with SCID %d\n", numUnits, msatPerUnit,
```

cmd/litcli/ln.go
```diff
@@ -237,74 +236,80 @@ type resultStreamWrapper struct {
...
 			return nil, err
 		}
 
-		if resp == nil || resp.Result == nil ||
-			resp.GetPaymentResult() == nil {
+		res := resp.Result
 
-			return nil, errors.New("unexpected nil result")
-		}
+		switch r := res.(type) {
```
darioAnongba, line 277: Switching over `type` feels strange to me but I have little context on `litd`.
I'm surprised we need to differentiate between deprecated `AcceptSellOrder` and the new `AcceptedSellOrders`. If the `litd` version supporting multi-rfq is released after the `tapd` version that adds the new `AcceptedSellOrders`, it seems reasonable that `litd` should simply not support the deprecated type. This is backwards compatibility squared.

Otherwise might be worth adding a comment explaining when the support for legacy can be dropped  because there is no direct correlation between dropping support in `tapd` and `litd`.

docs/release-notes/release-notes-0.16.0.md
```diff
@@ -20,6 +20,10 @@
 ### Technical and Architectural Updates
 
 ## RPC Updates
+- [Updated litcli](https://github.com/lightninglabs/lightning-terminal/pull/1125)
```
darioAnongba, line 23: The actual impact in `litcli` could be clearer? Like:
> "`rfq_peer_pubkey` is now optional in all contexts"

or something.

cmd/litcli/ln.go
ViktorT-11: Agree that switching over the `type` feels a bit strange, but it's commonly used throughout our daemons, so totally ok IMO.

Good point in regards to the supporting of the deprecated type! Something to note is that for our other cli's we do allow `remote` mode, and can therefore not know exactly which `daemon` version the user is running. But for `tapd`, remote mode is not yet supported.

darioAnongba: Yes it's totally fine but my point was that `litd` should decide what is the minimum `tapd` version it supports. There is no reason IMO for `litd` to keep backwards compatibility on `tapd` for too long. If using `remote` mode is not supported then I see no reason to keep backwards compatibility since we control the `tapd` version?

GeorgeTsagk: Some more context on the new RPC field / deprecation of the old one:
https://github.com/lightninglabs/taproot-assets/pull/1613#discussion_r2224642966

PR #1126, opened 2025-08-04
[sql-47] Actions migration prep

ViktorTigerstrom:
This PR prepares litd's startup process in preparation for the actions migration. 

In the upcoming actions migration, we need fetch LNDs macaroons prior to initializing the stores, as the macaroons will be required during the migration.

This requires that we setup the connection to LND before we initialize the stores, as we can only fetch the macaroons after the LND connection is established.

This PR refactors the litd startup process, so that the stores are initialized after the LND connection is established. This also requires that we refactor the RPC server setup, as litd's RPC servers require the stores to be initialized prior to being created.

Pushing this in draft until I've completed some final testing.


accounts/rpcserver.go
```diff
@@ -16,8 +18,17 @@ import (
...
+var (
+	// ErrServerNotActive indicates that the server has started but hasn't
+	// fully finished the startup process.
+	ErrServerNotActive = errors.New("accounts server is still in the " +
+		"process of starting")
+)
+
 // RPCServer is the main server that implements the Accounts gRPC service.
 type RPCServer struct {
+	active int32 // atomic
```
bitromortac: I think `atomic.Bool` would be a better type here

session_rpcserver.go
```diff
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
```
bitromortac: nit: I think the commit message doesn't need to be repeated, it could just say that we do a similar separation like in the previous commit

session_rpcserver.go
```diff
@@ -40,8 +41,17 @@ import (
...
+var (
+	// ErrServerNotActive indicates that the server has started but hasn't
+	// fully finished the startup process.
+	ErrServerNotActive = errors.New("session server is still in the " +
+		"process of starting")
+)
+
 // sessionRpcServer is the gRPC server for the Session RPC interface.
 type sessionRpcServer struct {
+	active int32 // atomic
```
bitromortac: same `atomic.Bool` as in the other commit

terminal.go
```diff
@@ -471,9 +471,7 @@ func (g *LightningTerminal) start(ctx context.Context) error {
 		)
 	}
 
-	g.accountRpcServer = accounts.NewRPCServer(
-		g.accountService, superMacBaker,
-	)
+	g.accountRpcServer = accounts.NewRPCServer()
```
bitromortac, line 458: we make this change instead of also shifting it after the lnd connection creation in order to error out on RPC calls, right?

terminal.go
```diff
@@ -881,12 +897,34 @@ func (g *LightningTerminal) setUpLNDClients(ctx context.Context,
...
-	// Now we know that the connection itself is ready. But we also need to
-	// wait for two things: The chain notifier to be ready and the lnd
-	// wallet being fully synced to its chain backend. The chain notifier
-	// will always be ready first so if we instruct the lndclient to wait
-	// for the wallet sync, we should be fully ready to start all our
-	// subservers. This will just block until lnd signals readiness.
+	return nil
+}
+
+// setupFullNDClient connects a up a full LND client to LND. Note that the setup
```
bitromortac: `setupFullLNDClient`

accounts/rpcserver.go
ViktorT-11: Ah nice, I see that `lnd` mostly have switched to that as well. Thanks :tada:!

session_rpcserver.go
ViktorT-11: Ok updated to not repeat the message and just mention the previous commit.

The main I see though for repetition, is that I do think it's much better from future perspective where someone needs to `git blame` a specific row from that commit. It makes more sense if the actual commit message mentions why the change was needed, rather than referring to a previous commit message as that requires someone to look up the exact ordering in the commit tree.

terminal.go
ViktorT-11: Thanks :pray:!

terminal.go
ViktorT-11: Not exactly:
This isn't intuitive when just looking at the `litd` code, but when the `lnd` connection is setup for an `integrated` `lnd` instance, `lnd` will call the `RegisterGrpcSubserver` in `litd` during the setup of the `lnd` connection.
If we kept the `registerSubDaemonGrpcServers` as is while changing the behaviour so that the `g.sessionRpcServer` & `g.accountRpcServer` objects where initialised after the `lnd` connection was setup, you'd then register `nil` pointers for those RPC servers. Later when `litd` has been fully started and request gets passed to those RPC servers, they would then `nil` pointer panic.

We could circumvent this by refactoring the `registerSubDaemonGrpcServers` and manually register those servers with the `g.rpcProxy.grpcServer` after the `lnd` connection has been setup and we've initialised the RPC servers, but that would require updates to the `itests` and be harder to reason about.

ellemouton: > manually register those servers with the g.rpcProxy.grpcServer after the lnd connection has been setup and we've initialised the RPC servers

I think that isnt possible right? i dont think you can register grpc servers after the server has been started

ellemouton: > lnd will call the RegisterGrpcSubserver in litd during the setup of the lnd connection.
If we kept the registerSubDaemonGrpcServers as is while changing the behaviour so that the g.sessionRpcServer & g.accountRpcServer 

Ok that makes sense - this would have been helpful in the PR description btw :) i only understood after reading this comment 👍

accounts/rpcserver.go
```diff
@@ -26,13 +37,29 @@ type RPCServer struct {
...
-		service:       service,
-		superMacBaker: superMacBaker,
+// Start adds the necessary dependencies for the RPCServer to be able to process
+// requests, and starts the RPCServer.
+func (s *RPCServer) Start(service *InterceptorService,
+	superMacBaker litmac.Baker) error {
+
+	if s.active.Swap(true) {
+		return errors.New("accounts rpc server is already started")
 	}
```
ellemouton: i think only set it after setting the internal variables - else can result in nil pointer deref if a call comes through after `active` is set but before the members have been assigned

accounts/rpcserver.go
```diff
@@ -110,6 +141,10 @@ func (s *RPCServer) CreateAccount(ctx context.Context,
 func (s *RPCServer) UpdateAccount(ctx context.Context,
 	req *litrpc.UpdateAccountRequest) (*litrpc.Account, error) {
 
+	if !s.started() {
+		return nil, ErrServerNotActive
+	}
```
ellemouton: i think if this is a pattern we use everywhere, then use a grpc interceptor pattern instead. Else we have to remember to check this each time.

ellemouton: but actually also: isnt this _already_ checked by the rpc proxy's interceptor (`checkSubSystemStarted`)?

session_rpcserver.go
```diff
@@ -1349,6 +1405,10 @@ func (s *sessionRpcServer) RevokeAutopilotSession(ctx context.Context,
 	req *litrpc.RevokeAutopilotSessionRequest) (
 	*litrpc.RevokeAutopilotSessionResponse, error) {
 
+	if !s.started() {
+		return nil, ErrServerNotActive
+	}
```
ellemouton: similar to previous comment: can probs just add a `subservers.SESSIONS` registration & then the rpc proxy should already check this via it's interceptor yeah?

terminal.go
ellemouton: perhaps  it would be useful to add a comment before the LND set-up just to make this super clear since it is quite hard to see by just reading the lit code & has always been a bit confusing imo

accounts/rpcserver.go
ViktorT-11: Good point! Addressed.

accounts/rpcserver.go
ViktorT-11: > but actually also: isnt this _already_ checked by the rpc proxy's interceptor (`checkSubSystemStarted`)?

Yes, this is already checked with the `rpcProxy.checkSubSystemStarted` function. I had added a comment for this in the commit message:
https://github.com/lightninglabs/lightning-terminal/pull/1126/commits/3d859440524c4891df564e6412faafa43e935815

I have now also added docs for the actual `active` variables to make that very clear.

My reasoning for this is that I did it for a defensive programming approach to ensure that we don't end up adding future updates which breaks this `rpcProxy` interceptor logic and leads to `nil` pointer panics. The main scenarios I see for that would be if we start white-listening some `account` or `session` RPCs, or if we add some custom status to the `lit` sub-server that makes it possible to send some account` or `session` RPC requests prior to the `lit` sub-server being marked as running.

This may just be over-engineering though, so I'd be fine to just remove these `active` booleans if you think we shouldn't try to optimize for that already :)!

session_rpcserver.go
ViktorT-11: See https://github.com/lightninglabs/lightning-terminal/pull/1126#discussion_r2273698882

terminal.go
ViktorT-11: Added a commit https://github.com/lightninglabs/lightning-terminal/pull/1126/commits/604c1095c051cdd1a1a18ef1b3a8e6dd98b0c905 that adds some docs for that :)! I also updated 3d859440524c4891df564e6412faafa43e935815 & 4a6f6278cce01295732538a774a0d0177e639542 to clarify that during the init of the accounts and session RPC server variables

accounts/rpcserver.go
ellemouton: i think for now it is over-engineering imo. i just done love that we are basically doing the job of an interceptor & doing so by repeating the same check everywhere - imo that makes it easy to forget to add that check to any new method. 

So id then rather have some more generic way of doing it: like [this](https://go.dev/play/p/Ic6dO-S6cxD) for example. And then a comment saying that any Registrar of the service must plug in the interceptor.

Also, just re nil pointers - it is very standard to assume that a subserver's methods are not ready unless Start has actually been called.

terminal.go
```diff
@@ -555,6 +555,9 @@ func (g *LightningTerminal) start(ctx context.Context) error {
 		go func() {
 			defer g.wg.Done()
 
+			// Note that LND will call litd's RegisterGrpcSubserver
+			// function during the execution of this, as `g` is
+			// referenced in the passed `implCfg`
```
ellemouton, line 556: 🙏

accounts/rpcserver.go
ViktorT-11: Thanks a lot for the suggestion @ellemouton! After discussing this offline, we decided that we're just going to drop the checks of if the `account` & `session` servers are started. Instead we'll keep the logic as is, and let the `rpcProxy` `checkSubSystemStarted` function that checks if the `lit` sub-server is started or not which will ensure that no requests are passed to the `account` & `session` servers before they've been started.

bitromortac: Thanks for the explanation, I think I understand that better now. Something that's not clear to me is why we are registering the sub-services with the lnd-created grpc server and not only with lit's proxy server?

accounts/rpcserver.go
```diff
@@ -26,13 +34,19 @@ type RPCServer struct {
...
+}
 
-	return &RPCServer{
-		service:       service,
-		superMacBaker: superMacBaker,
-	}
+// Start adds the necessary dependencies for the RPCServer to be able to process
+// requests, and starts the RPCServer.
+func (s *RPCServer) Start(service *InterceptorService,
+	superMacBaker litmac.Baker) error {
```
bitromortac: the returned error is not needed

session_rpcserver.go
```diff
@@ -70,41 +77,10 @@ type sessionRpcServerConfig struct {
...
-			// Register various grpc servers with the LNC session
-			// server.
-			cfg.registerGrpcServers(grpcServer)
-
-			return grpcServer
-		},
-	)
-
+// newSessionRPCServer creates a new sessionRpcServer.
+func newSessionRPCServer() (*sessionRpcServer, error) {
```
bitromortac: the returned error is not needed

terminal.go
bitromortac: Thanks for the explanation, I think I understand that better now. Something that's not clear to me is why we are registering the sub-services with the lnd-created grpc server and not only with lit's proxy server?

bitromortac (approved):
LGTM, nice investigation making it work to establish the LND connection first :tada:

ViktorTigerstrom:
> @ViktorTigerstrom - the build seems broken

Oops, my bad :pray:! Addressed with the latest push

PR #1132, opened 2025-08-15
[sql-49] account + session: Add determintistic sorting of maps during the KVDB to SQL migration.

ViktorT-11:
Addresses comment https://github.com/lightninglabs/lightning-terminal/pull/1129#pullrequestreview-3104677577.

Based on #1129.

This PR adds deterministic sorting of the maps in the accounts and sessions during the KVDB to SQL migration.
Will remove from draft once I've done more testing.

### DIFF
--- a/accounts/sql_migration.go
+++ b/accounts/sql_migration.go
@@ -6,13 +6,16 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


ViktorT-11:
Closing this one as I've determined it to not be necessary. Comparisons of maps in Go should be safe, as they have no particular order. We've also not seen any issues with from these compares in migrations so far.

PR #1133, opened 2025-08-19
[sql-50] session migration duplicate ID fix

ViktorTigerstrom:
The logic to avoid duplicate session IDs was first added with the session linking linking functionality. That means that there may be old legacy sessions that were added prior to that which may have duplicate IDs. Before this commit, such sessions would cause the kvdb to SQL migration to error, as there is a uniqueness constraint on the legacy session alias in the SQL sessions table.

We want to keep that constraint, so in such rare cases, we update the kvdb to SQL migration logic to drop all but the session with the latest CreatedAt time if for sessions that share the same ID.
That follows the logic in c8b78bd10dd5fee9cb250edc91229aa05d824877 as closely as possible, as that migration ensured that all sessions but the latest one was revoked if multiple sessions shared the same ID.

### DIFF
--- a/session/sql_migration.go
+++ b/session/sql_migration.go
@@ -40,35 +40,16 @@ func MigrateSessionStoreToSQL(ctx context.Context, kvStore *bbolt.DB,
 		return err
 	}
 
-	// If sessions are linked to a group, we must insert the initial session
-	// of each group before the other sessions in that group. This ensures
-	// we can retrieve the SQL group ID when inserting the remaining
-	// sessions. Therefore, we first insert all initial group sessions,
-	// allowing us to fetch the group IDs and insert the rest of the
-	// sessions afterward.
-	// We therefore filter out the initial sessions first, and then migrate
-	// them prior to the rest of the sessions.
-	var (
-		initialGroupSessions []*Session
-		linkedSessions       []*Session
-	)
+	initialGroupSessions, linkedSessions := filterSessions(kvSessions)
 
-	for _, kvSession := range kvSessions {
-		if kvSession.GroupID == kvSession.ID {
-			initialGroupSessions = append(
-				initialGroupSessions, kvSession,
-			)
-		} else {
-			linkedSessions = append(linkedSessions, kvSession)
-		}
-	}
-
+	// Migrate the non-linked sessions first.
 	err = migrateSessionsToSQLAndValidate(ctx, tx, initialGroupSessions)
 	if err != nil {
 		return fmt.Errorf("migration of non-linked session failed: %w",
 			err)
 	}
 
+	// Then migrate the linked sessions.
 	err = migrateSessionsToSQLAndValidate(ctx, tx, linkedSessions)
 	if err != nil {
 		return fmt.Errorf("migration of linked session failed: %w", err)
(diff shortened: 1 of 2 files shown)


session/sql_migration.go
```diff
@@ -81,6 +63,70 @@ func MigrateSessionStoreToSQL(ctx context.Context, kvStore *bbolt.DB,
...
+					sessions[j].CreatedAt,
+				)
+			})
+			sessionToKeep = sessions[0]
+
+			// Log the sessions that will be dropped.
+			for _, s := range sessions[1:] {
+				log.Warnf("Dropping duplicate session with ID "+
+					"%x created at %v", id, s.CreatedAt)
+			}
```
bitromortac: very nitty (I think the current approach is fine as well): we could just determine the max timestamp, it also has the benefit of being stable for testing if two sessions are present with the same timestamp, which we don't have currently

```go
    // Find the session with the latest timestamp in a single pass.
    latestSession := sessions[0]
    for _, s := range sessions[1:] {
        if s.CreatedAt.After(latestSession.CreatedAt) {
            latestSession = s
        }
    }
    sessionToKeep = latestSession

    // Log the sessions that will be dropped.
    for _, s := range sessions {
        if s == sessionToKeep {
            continue
        }
        log.Warnf("Dropping duplicate session with ID %x created at %v",
            id, s.CreatedAt)
    }
```

session/sql_migration_test.go
```diff
@@ -352,6 +352,115 @@ func TestSessionsStoreMigration(t *testing.T) {
...
+
+				err = updateSessionIDAndCreatedAt(
+					store, sess5.ID, sess4.MacaroonRootKey,
+					sess2.CreatedAt.Add(time.Minute),
+				)
+				require.NoError(t, err)
+
+				err = updateSessionIDAndCreatedAt(
+					store, sess6.ID, sess4.MacaroonRootKey,
+					sess2.CreatedAt.Add(time.Minute*2),
```
bitromortac: for consistency you could use `sess4.CreatedAt` instead of `sess2`

bitromortac (approved):
LGTM :zap:

session/sql_migration.go
bitromortac: we could also traverse the sessions in reverse order (`slices.Reverse`) (assuming the kv store order is in creation time order), to only use the latest created if two timestamps are the same. I know this is totally unlikely :)

ViktorT-11: Cool! Updated to the above approach :)!

> we could also traverse the sessions in reverse order

Unfortunately that'd require that we did sort the entire `kvsessions` list, which would remove any efficiency gained from that. If we did sort it and then traversed in reverse order, we could do everything in just one pass as we wouldn't need to iterate over `sessionsByID` map since we'd know that if we hit a session with duplicate id, it wouldn't be the latest session with that ID. But since sorting the entire `kvSessions` list + 1 extra iteration would be slower than just iterating over it twice, I think that wouldn't make sense :)

session/sql_migration_test.go
ViktorT-11: Ah yeah, thanks :pray:!

ellemouton (approved):
very nice!

PR #1134, opened 2025-08-20
[sql-51] session: handle removed linked account

ViktorT-11:
This PR fixes a edge case in the sessions migration, which was found during the creation of the actions migration.

If a user creates a session linked to an account, but later deletes that account using the `litcli accounts remove` command, the KVDB session will still contain the account ID of the removed account.

During the KVDB to SQL migration, the accounts migration runs before the sessions migration. Since the deleted account no longer exists, it cannot be migrated. As a result, the migrated SQL session cannot link to the now non-existent account in the SQL store.

In such cases, we will migrate the session and not link to any account in the SQL session.

### DIFF
--- a/session/sql_migration.go
+++ b/session/sql_migration.go
@@ -13,6 +13,7 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/lightninglabs/lightning-terminal/accounts"
 	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/sqldb"
 	"github.com/pmezard/go-difflib/difflib"
 	"go.etcd.io/bbolt"
@@ -196,6 +197,7 @@ func migrateSessionsToSQLAndValidate(ctx context.Context,
 		overrideSessionTimeZone(kvSession)
 		overrideSessionTimeZone(migratedSession)
 		overrideMacaroonRecipe(kvSession, migratedSession)
+		overrideRemovedAccount(kvSession, migratedSession)
 
 		if !reflect.DeepEqual(kvSession, migratedSession) {
 			diff := difflib.UnifiedDiff{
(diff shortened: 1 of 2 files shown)


bitromortac (commented):
> In such cases, we will migrate the session and not link to any account in the SQL session.

Are there any safety concerns with removing the account id from the session? I think that any macaroon that carries an unknown account id will not be valid to make requests, rendering the session useless? If that's the case should we just not migrate the session?

ViktorTigerstrom:
> > In such cases, we will migrate the session and not link to any account in the SQL session.
> 
> Are there any safety concerns with removing the account id from the session? I think that any macaroon that carries an unknown account id will not be valid to make requests, rendering the session useless? If that's the case should we just not migrate the session?

I don't think there should be any. I tested this locally with one kvdb session which has a removed account linked, and when sending requests over that session it errors with:
`rpc error: code = Unknown desc = error getting account cd88c53a03a3e66d: account not found`

Similarly, I tested running with an SQL session, which has `null` set for the `account_id`, but everything else for the account  correctly set (i.e. `type=account` + the correct `macaroon_recipe` set). Once the account has been removed, requests sent over this session also errors with:
`rpc error: code = Unknown desc = error getting account c595104329a7cfe3: account not found`

I do see your point though, and can see the motivation for just not migrating the session as it's useless once the account has been removed. Potentially instead of dropping the session, we could also just revoke it during the migration. Would be great to get Elle's perspective on this whenever possible.

session/sql_migration.go
```diff
@@ -394,3 +406,21 @@ func overrideMacaroonRecipe(kvSession *Session, migratedSession *Session) {
...
+	if migratedSession.AccountID.IsSome() {
+		return
+	}
+
+	// There was a linked account in the kv session, but not in the
+	// migrated session. We therefore remove the account ID from the
+	// kvSession.
+	kvSession.AccountID.WhenSome(func(alias accounts.AccountID) {
+		kvSession.AccountID = fn.None[accounts.AccountID]()
+	})
```
bitromortac: an alternative (don't need to use it if you find it confusing, I find it easier to understand the cases):
```go
	kvSession.AccountID = fn.ElimOption(
		migratedSession.AccountID,

		// If the migrated session does not have a linked account, we
		// also remove it from the kv session.
		func() fn.Option[accounts.AccountID] {
			return fn.None[accounts.AccountID]()
		},

		// If the migrated session has a linked account, we keep the
		// account on the kv session.
		func(id accounts.AccountID) fn.Option[accounts.AccountID] {
			return kvSession.AccountID
		},
	)
```

bitromortac (commented):
The changes look good :fire: (have a small suggestion).

I think keeping the session is probably better since we can have other db entries that are linked to the session like actions, which we then would need to delete as well? From an auditability perspective it seems better to leave it and I like your idea of revoking the session.

session/sql_migration.go
ViktorT-11: Cool thanks, updated the PR to use your approach :)

ViktorT-11:
Rebased on master to fix the changed test function signature.

PR #1080, opened 2025-05-28
Noop HTLC itest fixes

GeorgeTsagk:
### Description

Modifies some of our existing custom channels itests to now include some manual sats keysends to provide enough liquidity on some nodes to carry out the test case.

Our previous expectation was that asset HTLCs would always pay some sats to the receiver, but now this only happens up to the chan reserve amount.

Based on: https://github.com/lightningnetwork/lnd/pull/9871 , https://github.com/lightninglabs/taproot-assets/pull/1567

### Extra note

Since we introduced NoOp HTLCs on LND/Tapd that means we are no longer compatible with older versions. That leads to some itest failures, specifically the test cases that declared a backwards compatibility version for some of the nodes in the network topology. 

[Update]
We updated tapd to guard the feature behind a cfg option, so it's off by default. This PR turns NoOp HTLCs on for a sub-set of itests, which do not define a backwards-compatibility version:
`testCustomChannelsGroupTranchesForceClose`
`testCustomChannelsBreach`
`testCustomChannelsLiquidityEdgeCasesCore` (assetID & groupKey)
`testCustomChannelsStrictForwarding`
`testCustomChannelsHtlcForceCloseMpp`
`testCustomChannelsSelfPayment`

PR #1138, opened 2025-08-28
LiT: Custom channels feature bits

GeorgeTsagk:
## Description

Bumps (lots of) dependencies to use the latest tapd feature bits, which allow us to switch on/off cutting edge features while maintaining backwards compatibility with older nodes.

The first feature we use are the noop-htlcs, a special type of HTLC that will reflect the asset balance change over the channel but not shift the satoshi balances at all. This is an example of a backwards incompatible change, as if only one party uses it, the channel will soon be force-closed as there will be a disagreement on the next channel state.

Will close https://github.com/lightninglabs/lightning-terminal/pull/1080 in favor of this PR -- now we can enable noop-htlcs on all nodes that understand it, regardless of if there's an older version in the topology. This is exactly why we added the feature bits, in order for nodes to discover other peers' supported features. So for itests that do use backwards-compatibility LiT nodes, there should be no error, as the new node knows to **not** use noop-htlcs, because the noop-htlc feature bit was not communicated.

### Go 1.24.6

By bumping everything to the latest version (master) we also indirectly needed to bump 1.24.6 for LiT itself, the first commit bumps the version of LiT accordingly. Can break this out into its own PR if needed.

### Dependencies
https://github.com/lightninglabs/lndclient/pull/239
https://github.com/lightninglabs/taproot-assets/pull/1748
https://github.com/lightninglabs/taproot-assets/pull/1779
https://github.com/lightningnetwork/lnd/pull/10182

### DIFF
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
 	"github.com/lightninglabs/taproot-assets/asset"
 	"github.com/lightninglabs/taproot-assets/itest"
 	"github.com/lightninglabs/taproot-assets/proof"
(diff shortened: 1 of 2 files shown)


make/testing_flags.mk
```diff
@@ -1,6 +1,7 @@
 include make/compile_flags.mk
 
-TEST_FLAGS =
+TEST_FLAGS = -vet=off
```
gijswijs: The changes to the test flags should be in a separate commit and not in the commit that bumps the golang version.

make/testing_flags.mk
```diff
@@ -1,6 +1,7 @@
...
+ITEST_TEST_FLAGS = $(TEST_FLAGS)
```
gijswijs: I don't think I'm ok with disabling the `vet` check.
It skips static analysis that often catches real bugs and potential vulnerabilities (e.g., incorrect Printf format strings, misuse of unsafe pointers).

go.mod
```diff
@@ -247,3 +247,13 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
 // it is a replace in the tapd repository, it doesn't get propagated here
 // automatically, so we need to add it manually.
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
+
+replace github.com/lightningnetwork/lnd => github.com/GeorgeTsagk/lnd v0.0.0-20250901145709-e24e285e3e38
```
gijswijs: nit: The bumping of the lnd `paymentsdb` isn't temporary, but these replacements are. Maybe split them out of this commit to prevent errors?

itest/network_harness.go
```diff
@@ -779,13 +775,64 @@ func (n *NetworkHarness) RestartNodeNoUnlock(node *HarnessNode,
...
 		return nil, err
 	}
 
+	cfg := &suspendCfg{}
+
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
+	// If the upgrade flag was set delete any entry from the backwards
```
gijswijs: ultranit

```suggestion
	// If the upgrade flag was set, delete any entry from the backwards
```

itest/network_harness.go
```diff
@@ -779,13 +775,64 @@ func (n *NetworkHarness) RestartNodeNoUnlock(node *HarnessNode,
...
+	// compatibility map. This will force the node to use the latest build
+	// of LiT.
+	if cfg.upgrade {
+		delete(n.backwardCompat, node.Name())
+	}
+
+	// If a downgrade version was defined update the backwards compatibility
```
gijswijs: ultranit:
```suggestion
	// If a downgrade version was defined, update the backwards compatibility
```

Watch out for the line-lengths tho... :smile:

itest/litd_custom_channels_test.go
```diff
@@ -2275,6 +2302,237 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+		"--taproot-assets.proofcourieraddr=%s://%s",
+		proof.UniverseRpcCourierType,
+		fmt.Sprintf(node.ListenerFormat, davePort),
+	))
+
+	daveFlags := append(
+		slices.Clone(lndArgs), "--nolisten", "--minbackoff=1h",
+	)
+
+	// For this simple test, we'll just have Carol -> Dave as an assets
```
gijswijs: You mention Carol here, but down the line you talk about Charlie. I'm all for self-identity, but I would like to have my nodes to keep with the same given name during the duration of an itest. :wink:

itest/litd_custom_channels_test.go
```diff
@@ -2275,6 +2302,237 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+		[]*mintrpc.MintAssetRequest{
+			{
+				Asset: itestAsset,
+			},
+		},
+	)
+	cents := mintedAssets[0]
+	assetID := cents.AssetGenesis.AssetId
+
+	t.Logf("Minted %d lightning cents, syncing universes...", cents.Amount)
```
gijswijs: nit

```suggestion
	t.Logf("Minted %d itest asset cents, syncing universes...", cents.Amount)
```

`itest-asset-cents` is the name of the `itestAsset`

itest/litd_custom_channels_test.go
```diff
@@ -2275,6 +2302,237 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+			AssetAmount:        fundingAmount,
+			AssetId:            assetID,
+			PeerPubkey:         dave.PubKey[:],
+			FeeRateSatPerVbyte: 5,
+		},
+	)
+	require.NoError(t.t, err)
+	t.Logf("Funded channel between Charlie and Dave: %v", assetFundResp)
+
+	// With the channel open, mine a block to confirm it.
```
gijswijs:
```suggestion
	// With the channel open, mine 6 blocks to confirm it.
```

itest/litd_custom_channels_test.go
```diff
@@ -2275,6 +2302,237 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+	require.NoError(t.t, t.lndHarness.AssertNodeKnown(charlie, dave))
+	require.NoError(t.t, t.lndHarness.AssertNodeKnown(dave, charlie))
+
+	logBalance(t.t, nodes, assetID, "start")
+
+	// Let's dispatch 5 asset & 5 keysend payments from Charlie to Dave. At
+	// this point Charlie is running the old version of LiT.
+	for range 5 {
+		sendAssetKeySendPayment(
+			t.t, charlie, dave, 50, assetID, fn.None[int64](),
```
gijswijs, line 2435: :+1:

Unrelated to this PR, but that `fn.Option[int64]` parameter is actual a great way of passing optional parameters to a go func, other than functional options or variadic arguments.

itest/litd_custom_channels_test.go
```diff
@@ -2275,6 +2302,237 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+		)
+		sendKeySendPayment(t.t, charlie, dave, 1_000)
+	}
+
+	logBalance(t.t, nodes, assetID, "before upgrade")
+
+	// Let's assert that Charlie & Dave actually run different versions of
```
gijswijs, line 2442: Maybe clarify in a comment that Charlie is on an older version because of the `backwardCompat` settings for this test in `litd_test_list_on_test.go`

gijswijs: Also, checking that Charlie & Dave are on different versions doesn't necessarily confirm that one supports a feature and the other one does not.

itest/litd_custom_channels_test.go
```diff
@@ -2275,6 +2302,237 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+
+	// Let's send another 5 asset and keysend payments from Charlie to Dave.
+	// Charlie is now on the latest version of LiT and the channel upgraded.
+	for range 5 {
+		sendAssetKeySendPayment(
+			t.t, charlie, dave, 50, assetID, fn.None[int64](),
+		)
+		sendKeySendPayment(t.t, charlie, dave, 1_000)
+	}
+
```
gijswijs, line 2478: We don't confirm that these payments actually use the new features. If Charlie and Dave were still using v0 proofs, or using normal htlc's instead of noop-htlc's we wouldn't find out with this itest.

itest/litd_custom_channels_test.go
```diff
@@ -2275,6 +2302,237 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+	)
+
+	t.Logf("Charlie breach transfer: %v",
+		toProtoJSON(t.t, charlieBreachTransfer))
+	t.Logf("Dave breach transfer: %v",
+		toProtoJSON(t.t, daveBreachTransfer))
+
+	// With the breach transaction mined, Dave should now have a transaction
+	// in the mempool sweeping *both* commitment outputs.
+	daveJusticeTxid, err := waitForNTxsInMempool(
```
gijswijs, line 2560: So this is a non-stxo justice tx, right? (Because it's from a breach that used a commitment created before the upgrade). Shouldn't we also test the scenario where we sweep an stxo-tx?

gijswijs (changes requested):
I did a first pass and commented on a bunch of things, mostly nits.

There are two things that are of real importance:

1. Disabling the `vet` check is something we should discuss thoroughly, because of the security implications.
2. The itest should make a few more assertions to confirm that what we think is happening actually *is* happening.

Looking forward to your thoughts on this.

make/testing_flags.mk
GeorgeTsagk: When https://github.com/lightninglabs/lightning-terminal/pull/1144 is merged I'll drop this commit, after rebasing.

make/testing_flags.mk
GeorgeTsagk: See https://github.com/lightninglabs/lightning-terminal/pull/1138#discussion_r2358319661.

Also in that PR we work around disabling govet, so all rules are still active and we patch the call sites directly.

go.mod
GeorgeTsagk: Wdym? Eventually we'd also drop the replace directive and point to the appropriate LND state:
https://github.com/lightningnetwork/lnd/pull/10182

itest/litd_custom_channels_test.go
GeorgeTsagk: Oops, leftover. Changed :+1:

itest/litd_custom_channels_test.go
GeorgeTsagk: That means we'd have to introduce some kind of top-level RPC endpoint, or modify an existing one, to expose the internal feature-bit related state. Is that what you're suggesting?

itest/litd_custom_channels_test.go
GeorgeTsagk: Yeap!

We test the normal breach scenario with both peers supporting v1 in an existing itest: `testCustomChannelsBreach`

go.mod
```diff
@@ -247,3 +247,11 @@ replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-d
...
 // automatically, so we need to add it manually.
 replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate/v4 v4.18.2-9023d66a-fork-pr-2
+
+replace github.com/lightninglabs/loop => github.com/GeorgeTsagk/loop v0.0.0-20250925130456-37683918ef3e
+
+replace github.com/lightninglabs/loop/swapserverrpc => github.com/GeorgeTsagk/loop/swapserverrpc v0.0.0-20250925130456-37683918ef3e
+
+replace github.com/lightninglabs/loop/looprpc => github.com/GeorgeTsagk/loop/looprpc v0.0.0-20250925130456-37683918ef3e
+
+replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.19.0-beta.rc5.0.20250925062351-f293566849f2
```
Roasbeef: I think this can be removed now that the lnd PR is merged in. 

Can the others be removed now too?

Roasbeef: I think this can be removed now that the lnd PR is merged in. 

Can the others be removed now too?

itest/litd_node.go
```diff
@@ -652,7 +652,10 @@ func renameFile(fromFileName, toFileName string) {
 func (hn *HarnessNode) overrideFlagsAndBinary(backwardCompat map[string]string,
 	binary string, args []LitArgOption) (string, []LitArgOption) {
 
-	if backwardCompat == nil {
+	noopArg := WithLitArg("taproot-assets.channel.noop-htlcs", "")
```
Roasbeef, line 655: Seeing this here now, seems that this is actually opt in. Perhaps we should make it opt out?

itest/litd_custom_channels_test.go
```diff
@@ -1538,8 +1559,9 @@ func testCustomChannelsGroupTranchesHtlcForceClose(ctx context.Context,
 	// First, we'll send over some funds from Charlie to Dave, as we want
 	// Dave to be able to extend HTLCs in the other direction.
 	const (
-		numPayments   = 10
-		keySendAmount = 2_500
+		numPayments      = 10
+		keySendAmount    = 2_500
+		keySendSatAmount = 5000
```
Roasbeef: We should have this and the above be dynamically computed. So just `numPayments * `keySendSatAmount, that way future devs have one less thing to change.

itest/litd_custom_channels_test.go
Roasbeef: Correct that we don't directly introspect into the added HTLCs. They'll still be present, but have the extra TLV records. The indirect assertion here is that after the keysend, the balances don't change, which is the very thing that the noop HTLCs are meant to implement.

itest/litd_custom_channels_test.go
```diff
@@ -2302,6 +2303,318 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+	require.NoError(t.t, err)
+
+	proofFile, err = proof.DecodeFile(proofResp.RawProofFile)
+	require.NoError(t.t, err)
+	require.Equal(t.t, 4, proofFile.NumProofs())
+	latestProof, err = proofFile.LastProof()
+	require.NoError(t.t, err)
+
+	// This proof should contain the STXO exclusion proofs
+	stxoProofs = latestProof.InclusionProof.CommitmentProof.STXOProofs
```
Roasbeef, line 2614: 👍

itest/litd_custom_channels_test.go
Roasbeef: Ah looks like this was a dated comment, I now see that the relevant stxo proofs are examined.

go.mod
GeorgeTsagk: The only outstanding ref is the loop one ([see PR](https://github.com/lightninglabs/loop/pull/1021)).

itest/litd_node.go
GeorgeTsagk: Are you saying we should instead explicitly enable no-op on **some** itests? 

I think the above line captures the desired environment, where all upgraded node **use** no-op, while old ones ignore it / don't understand it. This way we maximize coverage of both no-op HTLC usage and also feature-bit based backwards compatibility.

itest/litd_custom_channels_test.go
GeorgeTsagk: The num of payments and the amounts below are not really related. We actually don't need to keep track of the total amount sent (`numPayments * keySendAmount`) as it's not directly required for any assertions in this specific itest.

PR #1143, opened 2025-09-11
[sql-52] firewalldb: add migration code for actions from kvdb to SQL

ViktorT-11:
Based on #1146 

This PR introduces the migration logic for transitioning the actions from kvdb to SQL.

I will cleanup the code a little bit and add some more details to the commit messages before requesting reviews.

### DIFF
--- a/firewalldb/sql_migration.go
+++ b/firewalldb/sql_migration.go
@@ -4,12 +4,20 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"reflect"
+	"sort"
+	"time"
 
+	"github.com/davecgh/go-spew/spew"
+	"github.com/lightninglabs/lightning-terminal/accounts"
 	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightninglabs/lightning-terminal/session"
 	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/sqldb"
+	"github.com/pmezard/go-difflib/difflib"
 	"go.etcd.io/bbolt"
 )
 
@@ -80,7 +88,8 @@ type privacyPairs = map[int64]map[string]string
 // NOTE: As sessions may contain linked sessions and accounts, the sessions and
 // accounts sql migration MUST be run prior to this migration.
 func MigrateFirewallDBToSQL(ctx context.Context, kvStore *bbolt.DB,
-	sqlTx SQLQueries) error {
+	sqlTx SQLQueries, sessionDB session.SQLQueries,
+	accountDB accounts.SQLQueries, macRootKeyIDs [][]byte) error {
 
 	log.Infof("Starting migration of the rules DB to SQL")
 
@@ -94,9 +103,14 @@ func MigrateFirewallDBToSQL(ctx context.Context, kvStore *bbolt.DB,
 		return err
 	}
 
-	log.Infof("The rules DB has been migrated from KV to SQL.")
+	err = migrateActionsToSQL(
+		ctx, kvStore, sqlTx, sessionDB, accountDB, macRootKeyIDs,
+	)
+	if err != nil {
+		return err
+	}
 
-	// TODO(viktor): Add migration for the action stores.
+	log.Infof("The rules DB has been migrated from KV to SQL.")
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


ViktorT-11:
Rebased on master and adjusted the PR to take into account the new changes that #1146 introduced.

This PR should now be ready for review :rocket:!

firewalldb/sql_migration_test.go
```diff
@@ -35,6 +36,50 @@ var (
...
+	// store.
+	rootKeys [][]byte
+}
+
+// addRootKeyFromIDSuffix adds a new root key to the store, using the passed
+// 4 byte suffix. The function generates a root key that ends with the 4 byte
+// suffix, prefixed by 4 random bytes.
+func (r *rootKeyMockStore) addRootKeyFromIDSuffix(suffix [4]byte) uint64 {
+	// As a real root key is 8 bytes, we need to generate a random 4 byte
+	// prefix to append to the passed 4 byte suffix.
```
bitromortac: nit: `prepend`

firewalldb/sql_migration.go
```diff
@@ -774,3 +787,769 @@ func validateGroupPairsMigration(ctx context.Context, sqlTx SQLQueries,
...
+		var sessAlias [4]byte
+		copy(sessAlias[:], sess.Alias[:4])
+
+		if _, ok := sessMap[sessAlias]; ok {
+			// NOTE: This should be unreachable, as we shouldn't
+			// have multiple sessions with the same Alias, as the
+			// sessions store has already been migrated to SQL here,
+			// and the session's table has a UNIQUE constraint on
+			// the Alias column.
+			return nil, fmt.Errorf("should'nt have multiple "+
```
bitromortac: nit: `shouldn't`

firewalldb/sql_migration.go
```diff
@@ -774,3 +787,769 @@ func validateGroupPairsMigration(ctx context.Context, sqlTx SQLQueries,
...
+// look up any potential linked session for each action based on the macaroon
+// identifier (which is the last 4 bytes of the root key ID).
+// The function returns a map where the key is the 4 byte Alias, and the
+// value is the corresponding session.
+func mapSessions(sessions []sqlc.Session) (map[[4]byte]sqlc.Session, error) {
+	sessMap := make(map[[4]byte]sqlc.Session)
+
+	for _, sess := range sessions {
+		if len(sess.Alias) != 4 {
+			return nil, fmt.Errorf("session alias must be 4 "+""+
```
bitromortac: nit: empty string addition

firewalldb/sql_migration.go
```diff
@@ -4,13 +4,20 @@ import (
...
+	"github.com/lightninglabs/lightning-terminal/accounts"
 	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightninglabs/lightning-terminal/session"
 	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/sqldb"
+	"github.com/pmezard/go-difflib/difflib"
 	"go.etcd.io/bbolt"
+	"reflect"
+	"sort"
+	"time"
```
bitromortac: nit: imports not correctly grouped/sorted

also, it would be good to go through the comments and format them as sentences, as there are a few that are not

firewalldb/sql_migration.go
```diff
@@ -774,3 +787,769 @@ func validateGroupPairsMigration(ctx context.Context, sqlTx SQLQueries,
...
+		return fn.None[sqlc.Account](), nil
+	}
+
+	attempted := action.AttemptedAt
+
+	// 1) Do some initial filtering of the accounts
+	filtered := make([]sqlc.Account, 0, len(actAccts))
+	for _, a := range actAccts {
+		// Exclude if expired before the attempt time
+		if a.Expiration.Before(attempted) && !a.Expiration.IsZero() {
```
bitromortac: nit: do the `!a.Expiration.IsZero()` check first

firewalldb/sql_migration_test.go
```diff
@@ -317,6 +339,38 @@ func TestFirewallDBMigration(t *testing.T) {
...
+
+		// Then assert that the actions in the SQLDB match the
+		// expected actions.
+		for i, migratedAction := range dbActions {
+			expAction := expectedActions[i]
+
+			// As the expected actions are set to the kvdb actions,
+			// we need to override the macaroon root key ID of the
+			// migrated SQL action before we can compare them.
+			overrideMacRootKeyID(migratedAction)
```
bitromortac: do we need this here?

firewalldb/sql_migration_test.go
```diff
@@ -963,11 +1066,687 @@ func randomPrivacyPairs(t *testing.T, ctx context.Context,
...
+	// occur with for the root key ID with an already existing session.
+	rootKey1 := rStore.addRootKeyFromIDSuffix(sess1.ID)
+
+	actionReq1 := testActionReq
+	actionReq1.MacaroonRootKeyID = fn.Some(rootKey1)
+	// However, as the session wasn't active at the time of the action
+	// creation, we don't link the session as the action wasn't linked to
+	// the session when it was created.
+	actionReq1.SessionID = fn.None[session.ID]()
+	actionReq1.AccountID = fn.None[accounts.AccountID]()
```
bitromortac, line 1152: done already in the next call, same below

firewalldb/sql_migration_test.go
```diff
@@ -963,11 +1066,687 @@ func randomPrivacyPairs(t *testing.T, ctx context.Context,
...
+	boltDB *BoltDB, sessStore session.Store, acctStore accounts.Store,
+	rStore *rootKeyMockStore) *expectedResult {
+
+	// Add a session with a linked account.
+	sess, acct, _ := testSessionWithAccount(
+		t, ctx, sessStore, acctStore,
+	)
+
+	rootKey := rStore.addRootKeyFromIDSuffix(sess.ID)
+	_ = rStore.addRootKeyFromAcctID(acct.ID)
```
bitromortac, line 1193: I'm not really sure why we add those to the root key store, shouldn't those be only identifiers that come from lnd when a custom root key (id) was created?

firewalldb/sql_migration_test.go
```diff
@@ -963,11 +1066,687 @@ func randomPrivacyPairs(t *testing.T, ctx context.Context,
...
+
+	actionReq := testActionReq
+	actionReq.MacaroonRootKeyID = fn.Some(rootKey)
+	// As the session the action is linked to does have a linked account,
+	// we also link the action to the account.
+	actionReq.SessionID = fn.Some(sess.ID)
+	actionReq.AccountID = fn.Some(acct.ID)
+
+	action := addAction(t, ctx, boltDB, &actionReq)
+	action.SessionID = fn.Some(sess.ID)
```
bitromortac: it seems like `action.SessionID = fn.Some(sess.ID)` isn't needed here, looking into `DeserializeAction`

firewalldb/sql_migration.go
ViktorT-11: Thanks :pray:! I also went through the comments and formated them as sentences. Let me know if you spot any that I missed though :pray:!

firewalldb/sql_migration_test.go
ViktorT-11: So I overhauled and cleaned up this a bit, as it actually was/is necessary to set these both before and after the action is created with `addAction`. This is since the `addAction` previously returned the pure deserialized kvdb action, which would therefore always have:
* The `MacaroonRootKeyID` set to only have values set for the last 4 bytes.
* `SessionID` set to short 4 byte `MacaroonID`, no matter if the action is actually linked to a session or not.
* Never any set `AccountID`, regardless of if there should be one linked or not.

As the returned `actions` in the `expectedResult` which all tests return, are expected to contain the "expected action", i.e. the SQL action and not the kvdb version, these needed to be set once more for the action returned by the `addAction` function, as that function only returned the kvdb version.

However, with this latest push, I realized that it's much more clean that the `addAction` function returns the expected SQL action instead of the kvdb action, and I therefore changed that the function itself sets the values for  `SessionID`, `AccountID` & `MacaroonRootKeyID` the returned action correctly.

That removes the need for the extra setting of the fields after the `addAction` function has been called.
Hope you agree with my reasoning here :pray: :)

firewalldb/sql_migration_test.go
ViktorT-11: This is related to https://github.com/lightninglabs/lightning-terminal/pull/1143#discussion_r2418142651: 
This was needed previously, as I never set `MacaroonRootKeyID` for the returned action from the `addAction` function, and the expected result would therefore contain the kvdb version which only has the last 4 bytes set to any values.

However, I realized that it's a lot more correct that the unit test ensures that the migrated action has actually have had the correct full 8 bytes set for the `MacaroonRootKeyID` field, and I therefore updated the `addAction` to also set the `MacaroonRootKeyID` of the returned action to the full 8 bytes, to represent the expected SQL version.

firewalldb/sql_migration_test.go
ViktorT-11: Removed the need for this line with:
https://github.com/lightninglabs/lightning-terminal/pull/1143#discussion_r2418142651

firewalldb/sql_migration_test.go
ViktorT-11: I do this to _simulate_ that `lnd` has indeed actually created a root key identifier as your referring to above. But since I can't use the real `lnd` macaroon store in the unit tests, this is a mock that represents that store.

I do this to ensure that the fetching of the full 8 byte root key ID is done correctly during the migration of the actions, as the kvdb actions only have the 4 byte version.

If it's helpful in order to understand how the root key ids in the `lnd` macaroon store relates to accounts and sessions I'm attaching a few examples below:

---- 
NOTE!! This is how the kvdb stores the actions.

Example 1:

Action triggered by linked session only:

action:
```
{
    "actor_name":  "Autopilot",
    "feature_name":  "AutoFees",
    "trigger":  "check if fee change was successful",
    "intent":  "",
    "structured_json_data":  "",
    "rpc_method":  "/lnrpc.Lightning/FeeReport",
    "rpc_params_json":  "{}",
    "timestamp":  "1754936907",
    "state":  "STATE_DONE",
    "error_reason":  "",
    "session_id":  "511216ee",
    "macaroon_identifier":  "511216ee"
},
```

Linked session:
```
"session":  {
    "id":  "511216ee",
    ....
}
```

`lnd` macaroons:
```
"root_key_ids": [
    ...,
    "18441921393731966702"
]
```

Which when converted to hex:
`0xFFEEDDCC511216EE`

I.e. last for bytes of the hex is the session ID

----- 

Example 2:

Action triggered by account only:

action:
```
{
    "actor_name":  "",
    "feature_name":  "",
    "trigger":  "",
    "intent":  "",
    "structured_json_data":  "",
    "rpc_method":  "/lnrpc.Lightning/GetNodeInfo",
    "rpc_params_json":  "{\n    \"pub_key\": \"0329e7c00fd7f66f1f81df7fc4d8cc411b50af81b8b3fb656b6f4d2e67598961bb\",\n    \"include_channels\": false\n}",
    "timestamp":  "1754937488",
    "state":  "STATE_DONE",
    "error_reason":  "",
    "session_id":  "82bc3f70",
    "macaroon_identifier":  "82bc3f70"
},
```

Linked account:
```
"account":  {
    "id":  "82bc3f70286b9616",
    "initial_balance":  "1000000000",
    "current_balance":  "1000000000",
    "last_update":  "1754937434",
    "expiration_date":  "0",
    "invoices":  [],
    "payments":  [],
    "label":  ""
},
```

`lnd` macaroons:
```
"root_key_ids": [
    ...,
    "18441921394565201776"
]
```

Converted to hex:
`0xFFEEDDCC82BC3F70`

I.e. last 4 bytes of the macaroon hex is used as the account ID (and also incorrectly set in the kvdb version to the "session_id" of the action).

----- 

Example 3:

Action triggered by session which by itself is linked to an account:

action:
```
{
    "actor_name":  "",
    "feature_name":  "",
    "trigger":  "",
    "intent":  "",
    "structured_json_data":  "",
    "rpc_method":  "/lnrpc.Lightning/WalletBalance",
    "rpc_params_json":  "{\n    \"account\": \"\",\n    \"min_confs\": 0\n}",
    "timestamp":  "1755176895",
    "state":  "STATE_DONE",
    "error_reason":  "",
    "session_id":  "4061d87a",
    "macaroon_identifier":  "4061d87a"
}
```

Linked account:
```
"account":  {
    "id":  "576638a1bf66832c",
    "initial_balance":  "1000000000",
    "current_balance":  "1000000000",
    "last_update":  "1755176609",
    "expiration_date":  "0",
    "invoices":  [],
    "payments":  [],
    "label":  ""
},
```

Linked session:
```
"session":  {
    "id":  "4061d87a",
    "label":  "test",
    "session_state":  "STATE_CREATED",
    "session_type":  "TYPE_MACAROON_ACCOUNT",
    "expiry_timestamp_seconds":  "1762952817",
    "mailbox_server_addr":  "mailbox.terminal.lightning.today:443",
    "dev_server":  false,
    "pairing_secret":  "d2197c51fd03c0936f24198901e8",
    "pairing_secret_mnemonic":  "...",
    "local_public_key":  "034061d87a563a5ad1d1db33c71358a97fbdab30fec58974c9cb63c6038e4b1815",
    "remote_public_key":  "",
    "created_at":  "1755176817",
    "macaroon_recipe":  {
        "permissions":  [],
        "caveats":  [
            "lnd-custom account 576638a1bf66832c"
        ]
    },
    "account_id":  "576638a1bf66832c",
    "autopilot_feature_info":  {},
    "revoked_at":  "0",
    "group_id":  "4061d87a",
    "feature_configs":  {},
    "privacy_flags":  "0"
}
```

`lnd` macaroons:
```
"root_key_ids": [
    ...,
    "18441921393451980922",
    "18441921393838143649"
]
```

Where `18441921393451980922` is the session id:
`0xFFEEDDCC4061D87A`

And `18441921393838143649` is the account id:
`0xFFEEDDCC576638A1`

bitromortac: Thank you, that's very helpful! I had the wrong impression that the macaroon root keys for accounts and sessions aren't added to lnd, which they are via `BakeSuperMacaroon`, because lnd needs to first check if the macaroon is allowed to be handed to the interceptor. So it makes sense that even if we only test for example account alias collision in `actionWithMultipleAccounts`, we add the identifiers to lnd's root key id store :+1:

firewalldb/sql_migration_test.go
```diff
@@ -889,24 +1057,890 @@ func randomPrivacyPairs(t *testing.T, ctx context.Context,
...
+	// Mock the root keys for both accounts.
+	_ = rStore.addRootKeyFromAcctID(acct1.ID)
+	rootKey := rStore.addRootKeyFromAcctID(newAcct2ID)
+
+	actionReq := testActionReq
+	actionReq.MacaroonRootKeyID = fn.Some(rootKey)
+	actionReq.SessionID = fn.None[session.ID]()
+	// When two colliding accounts exist, the account with the latest expiry
+	// should be linked to the action. In our case, that's acct2.
+	actionReq.AccountID = fn.Some(newAcct2ID)
```
bitromortac, line 1374: nit: maybe we should reverse this heuristic, as it is more likely the older account was used? same could go for the macaroon root key ids, right now we use the newer ones, but this is all very hypothetical

firewalldb/sql_migration_test.go
bitromortac: Yes, that's cleaner indeed, great changes, thank you for the explanation as well!

firewalldb/sql_migration_test.go
ViktorT-11: I updated so that the older account has priority as it it's more likely that this existed during the creation of the action, but choose to keep the logic with the macaroon root key ids, as it's more likely IMO that newer macaroons tend to be used. Can change that too if you prefer though :)

PR #1144, opened 2025-09-11
build: bump go to v1.24.6

GeorgeTsagk:
## Description

Bumps the version of golang to v1.24.6

### Notes

With this bump we also include a new `govet` rule that disallows printing non-constant strings (containing `%s` for example), giving many errors like the following:
```
printf: non-constant format string in call to ...
```

For this we add a rule to `.golangci.yml` to exclude the `printf` check from `govet` when linting, and also we disable `govet` when running tests -- could not find a workaround that directly targets the `printf` rule.

The latter is important as we use `make unit` for the "check commits" CI step.

### Rationale

The motivation behind this bump is the following PR: https://github.com/lightninglabs/lightning-terminal/pull/1138

which bumps a lot of dependencies that indirectly bump go to v1.24.6 (the above PR currently includes a commit that bumps golang for LiT, but decided to split it off into its own PR)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ env:
   # If you change this value, please change it in the following files as well:
   # /Dockerfile
   # /dev.Dockerfile
-  GO_VERSION: 1.23.12
+  GO_VERSION: 1.24.6
 
 jobs:
   ########################
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN apk add --no-cache --update alpine-sdk \
 
 # The first stage is already done and all static assets should now be generated
 # in the app/build sub directory.
-FROM golang:1.23.12-alpine3.22@sha256:383395b794dffa5b53012a212365d40c8e37109a626ca30d6151c8348d380b5f as golangbuilder
+FROM golang:1.24.6-alpine3.22@sha256:c8c5f95d64aa79b6547f3b626eb84b16a7ce18a139e3e9ca19a8c078b85ba80d as golangbuilder
 
 # Instead of checking out from git again, we just copy the whole working
 # directory of the previous stage that includes the generated static assets.
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ PUBLIC_URL :=
 # GO_VERSION is the Go version used for the release build, docker files, and
 # GitHub Actions. This is the reference version for the project. All other Go
 # versions are checked against this version.
-GO_VERSION = 1.23.12
+GO_VERSION = 1.24.6
 
 # LITD_COMPAT_VERSIONS is a space-separated list of litd versions that are
 # installed before running the integration tests which include backward
--- a/autopilotserverrpc/go.mod
+++ b/autopilotserverrpc/go.mod
@@ -1,6 +1,6 @@
 module github.com/lightninglabs/lightning-terminal/autopilotserverrpc
 
-go 1.23.12
+go 1.24.6
 
 require (
 	google.golang.org/grpc v1.56.3
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -20,7 +20,7 @@ RUN cd /go/src/github.com/lightninglabs/lightning-terminal/app \
 
 # The first stage is already done and all static assets should now be generated
 # in the app/build sub directory.
-FROM golang:1.23.12-alpine3.22@sha256:383395b794dffa5b53012a212365d40c8e37109a626ca30d6151c8348d380b5f as golangbuilder
+FROM golang:1.24.6-alpine3.22@sha256:c8c5f95d64aa79b6547f3b626eb84b16a7ce18a139e3e9ca19a8c078b85ba80d as golangbuilder
 
 # Instead of checking out from git again, we just copy the whole working
 # directory of the previous stage that includes the generated static assets.
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/lightninglabs/lightning-terminal
 
-go 1.23.12
+go 1.24.6
 
 require (
 	github.com/btcsuite/btcd v0.24.3-0.20250318170759-4f4ea81776d6
--- a/litrpc/Dockerfile
+++ b/litrpc/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.23.12-bookworm@sha256:d22457a90736648a605cf54c33a7a310bb290e1ac78af2bcbc1d87821c52c4b7
+FROM golang:1.24.6-bookworm@sha256:ab1d1823abb55a9504d2e3e003b75b36dbeb1cbcc4c92593d85a84ee46becc6c
 
 RUN apt-get update && apt-get install -y \
   git \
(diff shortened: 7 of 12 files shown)


make/testing_flags.mk
```diff
@@ -1,6 +1,6 @@
 include make/compile_flags.mk
 
-TEST_FLAGS =
+TEST_FLAGS = -vet=off
```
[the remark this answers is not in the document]
GeorgeTsagk: This was my main concern with the current way of disabling `govet` for `make unit`, interested to see what people think too

.golangci.yml
```diff
@@ -17,6 +17,8 @@ linters-settings:
   govet:
     # Don't report about shadowed variables
     check-shadowing: false
+    disable:
+      - printf
```
ViktorT-11: Instead of having to add this exception, I'd suggest that it'd probably be better to see if you can solve the underlaying issue the linter is erroring on. After testing, it seems to me like the linter is only erroring if we call a function like:
```
func test(errStr string, params ...interface{}){
      // Something that does:
      fmt.Sprintf(errStr, params...)
}
```

Where the call site doesn't actually set any value for `params`. Call sites that do set a value for `params` seems to not error. So I suggest updating such cases in `litd`, instead of updating the linter exception here.

For example, when I tried doing so for the `status` `Manager`, this update seems to work:
```
// SetErrored can be used to set the status of a sub-server as not Running
// and also to set an error message for the sub-server.
//
// NOTE: This will silently fail if the referenced sub-server has not yet been
// registered.
func (s *Manager) SetErrored(name string, errStr string) {
	s.mu.Lock()
	defer s.mu.Unlock()

	log.Debugf("Setting the %s sub-server as errored: %s", name, errStr)

	ss, ok := s.subServers[name]
	if !ok {
		return
	}

	log.Errorf("could not start the %s sub-server: %s", name, errStr)

	ss.running = false
	ss.err = errStr
	ss.customStatus = ""
}

func (s *Manager) SetErroredf(name string, errStr string,
	params ...interface{}) {

	errStr = fmt.Sprintf(errStr, params...)

	s.SetErrored(name, errStr)
}
```

NOTE: only example code above, and haven't considered if there's better ways to do so than the above :).

GeorgeTsagk: Another alternative that seems to work is to simply pass `nil` as `params ...interface{}` when no extra params are present, so on the callsites that would be
```
			s.statusServer.SetErrored(ss.Name(), err.Error(), nil)
```

I do like it as it's a bit more lean (-- we don't create an extra helper func)

I'll push that version and see how CI behaves -- will also strip away all the custom govet related rule exceptions

rpcmiddleware/proto.go
```diff
@@ -54,7 +54,7 @@ func RPCErr(req *lnrpc.RPCMiddlewareRequest,
 	err error) (*lnrpc.RPCMiddlewareResponse, error) {
 
 	if err != nil {
-		return RPCErrString(req, err.Error())
+		return RPCErrString(req, err.Error(), nil)
```
ViktorT-11: I think you should add these changes which are intended to solve the linter issue in a separate commit prior to the `golang` bump commit, and add a description to the commit to explain why they are needed.

rpc_proxy.go
```diff
@@ -365,7 +365,7 @@ func (p *rpcProxy) makeDirector(allowLitRPC bool) func(ctx context.Context,
 		handled, conn, err := p.subServerMgr.GetRemoteConn(requestURI)
 		if err != nil {
 			return outCtx, nil, status.Errorf(
-				codes.Unavailable, err.Error(),
+				codes.Unavailable, err.Error(), nil,
```
ViktorT-11: nit: There's already a `status.Error` function available that won't require the passing of `nil` here, so I would suggest using that instead.

ViktorTigerstrom (commented):
Awesome that you managed to get rid of the linter rule exception :rocket::tada:!

Regarding https://github.com/lightninglabs/lightning-terminal/pull/1144#discussion_r2351813035, I personally prefer creating helper functions as suggested in https://github.com/lightninglabs/lightning-terminal/pull/1144#discussion_r2348588689 over having to pass `nil` to the functions.
My motivation for why that's preferred is that it'll otherwise be very hard for callers to understand that they need to pass `nil` to that `params ...X` arg, or they'll end up getting linter errors. If we stick to the `nil` passing alternative though, I think you should at least update the docs for those functions and add a comment which explains that you need to set `nil` for the `params ...X` arg if no params exists, and explain why that's needed.

I don't have a too strong preference for my suggestion though, so I'd be ok with your approach if others agree that it's better :)! So I'd be interested to hear what @ellemouton thinks about this.

rpcmiddleware/proto.go
GeorgeTsagk: They're now a separate commit :+1:

rpc_proxy.go
GeorgeTsagk: Yeap followed this approach instead of passing in `nil` to all call sites.

I added the helper you recommended above.

There's only 1 call site where I pass in `nil` to `RPCErrString`. This was only a single instance, so figured it wasn't really worth splitting `RPCErrString` into two helpers.

ellemouton (approved):
lgtm! 
thanks for the update, and thanks to @ViktorTigerstrom for the helper suggestion 🙏

status/manager.go
```diff
@@ -258,24 +258,34 @@ func (s *Manager) SetStopped(name string) {
...
 	ss.customStatus = ""
 }
+
+// SetErroredf can be used to set the status of a sub-server as not Running and
+// also set an error message for the sub-server. This function can also accept
+// non-constant strings which may contain extra arguments.
+//
+// NOTE: This will silently fail if the referenced sub-server has not yet been
+// registered.
+func (s *Manager) SetErroredf(name string, errStr string,
```
ViktorT-11: Hmmm I realized that one alternative approach that you may like better than the `helper` function alterantive would be to add an `if` case in the `SetErrored` function like this:

```
func (s *Manager) SetErrored(name string, errStr string,
	params ...interface{}) {

	s.mu.Lock()
	defer s.mu.Unlock()

	log.Debugf("Setting the %s sub-server as errored: %s", name, errStr)

	ss, ok := s.subServers[name]
	if !ok {
		return
	}

	log.Errorf("could not start the %s sub-server: %s", name, errStr)

	if len(params) > 0 {
		errStr = fmt.Sprintf(errStr, params...)
	}

	ss.running = false
	ss.err = errStr
	ss.customStatus = ""
}
```

And something similar in the `RPCErrString` like this:

```
func RPCErrString(req *lnrpc.RPCMiddlewareRequest, format string,
	args ...interface{}) (*lnrpc.RPCMiddlewareResponse, error) {

	feedback := &lnrpc.InterceptFeedback{}
	resp := &lnrpc.RPCMiddlewareResponse{
		RefMsgId: req.MsgId,
		MiddlewareMessage: &lnrpc.RPCMiddlewareResponse_Feedback{
			Feedback: feedback,
		},
	}

	if len(args) > 0 {
		format = fmt.Sprintf(format, args...)
	}

	if format != "" {
		feedback.Error = format
	}

	return resp, nil
}
```

I'm totally ok with either version though, so feel free to keep the PR and merge it as if you'd like 😃 :tada:!

GeorgeTsagk: Ah, much better approach! Will actually apply it. Thanks again :100:

PR #1145, opened 2025-09-11
README.md fix LND version

ZZiigguurraatt:
The LiT to LND table is updated to match what is stated in

- https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.15.2-alpha
- https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.15.1-alpha
- https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.15.0-alpha

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,9 +78,9 @@ in remote mode (meaning that `lnd-mode=remote` is set). It shows the
 
 | LiT               | LND          |
 |-------------------|--------------|
-| **v0.15.2-alpha** | v0.19.0-beta |
-| **v0.15.1-alpha** | v0.19.0-beta |
-| **v0.15.0-alpha** | v0.19.0-beta |
+| **v0.15.2-alpha** | v0.19.3-beta |
+| **v0.15.1-alpha** | v0.19.2-beta |
+| **v0.15.0-alpha** | v0.19.1-beta |
 | **v0.14.1-alpha** | v0.18.5-beta |
 | **v0.14.0-alpha** | v0.18.4-beta |
 | **v0.13.6-alpha** | v0.17.1-beta |


README.md
```diff
@@ -78,9 +78,9 @@ in remote mode (meaning that `lnd-mode=remote` is set). It shows the
 
 | LiT               | LND          |
 |-------------------|--------------|
-| **v0.15.2-alpha** | v0.19.0-beta |
-| **v0.15.1-alpha** | v0.19.0-beta |
-| **v0.15.0-alpha** | v0.19.0-beta |
+| **v0.15.2-alpha** | v0.19.3-beta |
+| **v0.15.1-alpha** | v0.19.2-beta |
+| **v0.15.0-alpha** | v0.19.1-beta |
```
ViktorT-11, line 83: @ZZiigguurraatt, this can be a bit confusing when reading the README, but actually `v0.19.0-beta` is correct here as this details the `minimum` `lnd` version that's compatible with these `litd` versions, when running `lnd` in remote mode. This is described in the text above, but I agree that this can be a bit confusing :). This change is therefore not needed.

Perhaps this should be described in a more easy to understand way, to make this less confusing? If you think that's the case, feel free to change the PR to update how this information is displayed. Else, if you find the text above to be clear enough, I suggest we close the PR.

ZZiigguurraatt: OK, basically you are right I just failed to read.

However, for people who are just quickly skimming around, would it make more sense to put

https://github.com/lightninglabs/lightning-terminal/blob/bce4c049a95a8484b04eddbf856c9356c8a64833/README.md?plain=1#L171-L231

up first towards the top (before https://github.com/lightninglabs/lightning-terminal/blob/bce4c049a95a8484b04eddbf856c9356c8a64833/README.md?plain=1#L67) and rename the title from "Daemon Versions packaged with LiT" to "Daemon Versions packaged with LiT in **_Integrated Mode_**" 

and also change

https://github.com/lightninglabs/lightning-terminal/blob/bce4c049a95a8484b04eddbf856c9356c8a64833/README.md?plain=1#L67

to

"LND Version Compatibility When Used In **_Remote Mode_**"?

ViktorT-11: Sure! That sounds like a great suggestion. Feel free to update the PR to update that instead :)

PR #1146, opened 2025-09-15
[sql-53] firewalldb: actions migration prep 2 - persist full macaroon root key ID in the SQL actions store

ViktorT-11:

This PR addresses an issue I noticed we need to address prior to https://github.com/lightninglabs/lightning-terminal/pull/1143:

When migrating the actions store from kvdb to sql, we will need to update the existing actions to include the full mac root key, instead of just the last 4 bytes (currently called `MacaroonIdentifier`). In order to do so, we need to change the sql implementation of the `actions` store to persist the full mac root key, instead of just the last 4 bytes. As no production data in the sql actions store exists for users yet, it's fine for us to change this without having to address old sql actions which only stored the last 4 bytes.

This PR adds this change.

Note though that  this PR doesn't update the kvdb implementation, and the full macaroon root key will be ignored by the kvdb store even if set. Therefore, the rest of `litd` will still have to just expect the last 4 bytes of the mac root key when accessing an `Action`'s MacaroonIdentifier. Therefore, we we currently never expose the rest of the mac root key outside of the sql actions store.

Once the kvdb store has been fully deprecated and removed, we can then update the rest of `litd` to also use the full mac root key, and change the `Action` struct's field to reflect this.

### DIFF
--- a/firewall/request_logger.go
+++ b/firewall/request_logger.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/lightninglabs/lightning-terminal/firewalldb"
+	litmac "github.com/lightninglabs/lightning-terminal/macaroons"
 	mid "github.com/lightninglabs/lightning-terminal/rpcmiddleware"
 	"github.com/lightninglabs/lightning-terminal/session"
 	"github.com/lightningnetwork/lnd/fn"
@@ -182,14 +183,25 @@ func (r *RequestLogger) Intercept(ctx context.Context,
 func (r *RequestLogger) addNewAction(ctx context.Context, ri *RequestInfo,
 	withPayloadData bool) error {
 
-	var macaroonID fn.Option[[4]byte]
+	var (
+		rootKeyID  fn.Option[uint64]
+		macaroonID fn.Option[[4]byte]
+	)
+
 	if ri.Macaroon != nil {
 		var err error
-		macID, err := session.IDFromMacaroon(ri.Macaroon)
+
+		fullRootKeyID, err := litmac.RootKeyIDFromMacaroon(
+			ri.Macaroon,
+		)
 		if err != nil {
-			return fmt.Errorf("could not extract ID from macaroon")
+			return fmt.Errorf("could not extract root key ID from "+
+				"macaroon: %w", err)
 		}
 
+		macID := session.IDFromMacRootKeyID(fullRootKeyID)
+
+		rootKeyID = fn.Some(fullRootKeyID)
 		macaroonID = fn.Some([4]byte(macID))
 	}
 
@@ -197,6 +209,7 @@ func (r *RequestLogger) addNewAction(ctx context.Context, ri *RequestInfo,
 		SessionID:          ri.SessionID,
 		AccountID:          ri.AccountID,
 		MacaroonIdentifier: macaroonID,
+		MacaroonRootKeyID:  rootKeyID,
 		RPCMethod:          ri.URI,
 	}
 
--- a/firewall/request_logger.go
+++ b/firewall/request_logger.go
@@ -9,7 +9,6 @@ import (
 	"github.com/lightninglabs/lightning-terminal/firewalldb"
 	litmac "github.com/lightninglabs/lightning-terminal/macaroons"
 	mid "github.com/lightninglabs/lightning-terminal/rpcmiddleware"
-	"github.com/lightninglabs/lightning-terminal/session"
 	"github.com/lightningnetwork/lnd/fn"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/macaroons"
@@ -184,8 +183,7 @@ func (r *RequestLogger) addNewAction(ctx context.Context, ri *RequestInfo,
 	withPayloadData bool) error {
 
 	var (
-		rootKeyID  fn.Option[uint64]
-		macaroonID fn.Option[[4]byte]
+		rootKeyID fn.Option[uint64]
 	)
 
 	if ri.Macaroon != nil {
@@ -198,19 +196,14 @@ func (r *RequestLogger) addNewAction(ctx context.Context, ri *RequestInfo,
 			return fmt.Errorf("could not extract root key ID from "+
 				"macaroon: %w", err)
 		}
-
-		macID := session.IDFromMacRootKeyID(fullRootKeyID)
-
 		rootKeyID = fn.Some(fullRootKeyID)
-		macaroonID = fn.Some([4]byte(macID))
 	}
 
 	actionReq := &firewalldb.AddActionReq{
-		SessionID:          ri.SessionID,
-		AccountID:          ri.AccountID,
-		MacaroonIdentifier: macaroonID,
-		MacaroonRootKeyID:  rootKeyID,
-		RPCMethod:          ri.URI,
+		SessionID:         ri.SessionID,
+		AccountID:         ri.AccountID,
+		MacaroonRootKeyID: rootKeyID,
+		RPCMethod:         ri.URI,
 	}
 
 	if withPayloadData {
--- a/firewalldb/actions.go
+++ b/firewalldb/actions.go
@@ -39,6 +39,15 @@ type AddActionReq struct {
 	// If no macaroon was used for the action, then this will not be set.
 	MacaroonIdentifier fn.Option[[4]byte]
 
+	// MacaroonRootKeyID is the uint64 / full 8 bytes of the root key ID of
+	// the macaroon used to perform the action.
+	// If no macaroon was used for the action, then this will not be set.
+	//
+	// NOTE: for our BoltDB impl, only the lower 32 bits / last 4 bytes of
+	// this uint64 are stored. When read back, the upper 32 bits / first 4
+	// bytes are zeroed.
+	MacaroonRootKeyID fn.Option[uint64]
+
 	// SessionID holds the optional session ID of the session that this
 	// action was performed with.
 	//
--- a/firewalldb/actions.go
+++ b/firewalldb/actions.go
@@ -34,11 +34,6 @@ const (
 // It contains all the information that is needed to create a new Action in the
 // ActionStateInit State.
 type AddActionReq struct {
-	// MacaroonIdentifier is a 4 byte identifier created from the last 4
-	// bytes of the root key ID of the macaroon used to perform the action.
-	// If no macaroon was used for the action, then this will not be set.
-	MacaroonIdentifier fn.Option[[4]byte]
-
 	// MacaroonRootKeyID is the uint64 / full 8 bytes of the root key ID of
 	// the macaroon used to perform the action.
 	// If no macaroon was used for the action, then this will not be set.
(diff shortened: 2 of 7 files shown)


ellemouton:
@ViktorTigerstrom - what's the main reasoning around storing the full thing? 
iiuc, we just want to full thing at migration time to determine possible linked sessions/actions. But do we actually need the full thing to be persisted? 

i guess no harm in doing so, but just checking understanding

firewalldb/test_kvdb.go
```diff
@@ -59,8 +57,9 @@ func newDBFromPathWithSessions(t *testing.T, dbPath string,
 
 func assertEqualActions(t *testing.T, expected, got *Action) {
 	// Accounts are not explicitly linked in our bbolt DB implementation.
+	actualAccountID := got.AccountID
 	got.AccountID = expected.AccountID
 	require.Equal(t, expected, got)
 
-	got.AccountID = fn.None[accounts.AccountID]()
+	got.AccountID = actualAccountID
```
ellemouton, line 81: 👍

firewall/request_logger.go
```diff
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	"github.com/lightninglabs/lightning-terminal/firewalldb"
+	litd_macaroons "github.com/lightninglabs/lightning-terminal/macaroons"
```
ellemouton: suggest `litmac`?

firewall/request_logger.go
```diff
@@ -182,21 +183,33 @@ func (r *RequestLogger) Intercept(ctx context.Context,
...
+
+		rootKeyID = fn.Some(fullRootKeyID)
 		macaroonID = fn.Some([4]byte(macID))
 	}
 
 	actionReq := &firewalldb.AddActionReq{
 		SessionID:          ri.SessionID,
 		AccountID:          ri.AccountID,
 		MacaroonIdentifier: macaroonID,
+		MacaroonRootKeyID:  rootKeyID,
```
ellemouton: ok and we need both cause kvdb still stores old identifier but now sql stores full rootkey id?

firewalldb/actions.go
```diff
@@ -39,6 +39,11 @@ type AddActionReq struct {
 	// If no macaroon was used for the action, then this will not be set.
 	MacaroonIdentifier fn.Option[[4]byte]
 
+	// MacaroonRootKeyID is the uint64 / full 8 bytes of the root key ID of
+	// the macaroon used to perform the action.
+	// If no macaroon was used for the action, then this will not be set.
+	MacaroonRootKeyID fn.Option[uint64]
+
 	// SessionID holds the optional session ID of the session that this
```
ellemouton: related to above question, if one of these is stored in kvdb only and the other in sql only, lets add this warning in the comments

ellemouton: or actually, to keep the API intuitive, cant se instead _only_ have the MacRootKeyID field here & remove the MacID field. Then extract the relevant format at the DB layer. ie, in the kvdb layer, take the root key ID and compute the mac ID.

firewalldb/actions_sql.go
```diff
@@ -393,9 +397,17 @@ func unmarshalAction(ctx context.Context, db SQLActionQueries,
...
+	//
+	// TODO(viktor): Remove this when we no longer need to be compatible
+	// with the kvdb version.
 	var macID fn.Option[[4]byte]
-	if len(dbAction.MacaroonIdentifier) > 0 {
-		macID = fn.Some([4]byte(dbAction.MacaroonIdentifier))
+	if len(dbAction.MacaroonIdentifier) >= 4 {
+		dbMacID := dbAction.MacaroonIdentifier
+		macID = fn.Some([4]byte(dbMacID[len(dbMacID)-4:]))
 	}
```
ellemouton, line 408: not populating the new field below? (see comment about only needing 1 field potentially anyways)

ViktorTigerstrom:
Thanks for the review @ellemouton.

I'm answering this clarification first, as I believe that's the deciding factor regarding wether we want to go ahead of with this PR or not:

> iiuc, we just want to full thing at migration time to determine possible linked sessions/actions. But do we actually need the full thing to be persisted?

So after working through the actions migration, it has become clear that the only real reason we'd need to persist the full macaroon root key id (8 bytes), compared to the short version (just the last 4 bytes), is to make it possible to link which exact macaroon was used to when the action was generated.
Otherwise, there can be the scenario that the user has generated 2 different macaroons where there's a collision for the last 4 bytes. In that scenario, the info is lost when looking at the action which macaroon was actually used to generate it. When we discussed this offline previously, we determined it to be worth it to persist the full macaroon root key id for this reason. 

However, I'm not super firm that this protection is totally needed, and IMO it's worth doing for the the sake of it being more clean than opening up for the collision risk. If we do not care about that collision protection however, I don't think we actually need to determine the full root key ID during the migration time as it's not really needed to link possible linked sessions/actions. If so, we can drop the whole short mac key ID -> full root key mapping in the actions migration.

Let me know if you think we should proceed with this, or just not try to solve the collision issue :)

ViktorTigerstrom:
I've realized one scenario where persisting the full root key ID miiiiight be useful is if the user deletes their `lnd` macaroon DB. Right now, we only store 4 bytes of the 8-byte root key ID for the `action`, which would mean the first 4 bytes of the root key id used for the action would be lost permanently for the user. If we persisted all 8 bytes, the user could still recover all 8 bytes. With the full ID, they’d be able to pass that full `root_key_id` into `lncli bakemacaroon` and generate a new macaroon.

Not sure if there are many scenarios where this would be useful though :)

ellemouton:
Thanks for the info @ViktorTigerstrom  - i think yes it makes sense to store the full thing 🙏

firewall/request_logger.go
```diff
@@ -182,21 +183,33 @@ func (r *RequestLogger) Intercept(ctx context.Context,
...
+	var (
+		rootKeyID  fn.Option[uint64]
+		macaroonID fn.Option[[4]byte]
+	)
+
 	if ri.Macaroon != nil {
 		var err error
-		macID, err := session.IDFromMacaroon(ri.Macaroon)
+
+		fullRootKeyID, err := litd_macaroons.RootKeyIDFromMacaroon(
```
bitromortac: Still a bit unsure why it's needed, looking at `buildSession`, we derive the full root key id from the last four bytes of the local pubkey and a common prefix, see `NewSuperMacaroonRootKeyID`. So it seems like we can always reconstruct the full macaroon root key id from the four bytes that are stored?

firewalldb/actions_test.go
```diff
@@ -60,10 +61,13 @@ func TestActionStorage(t *testing.T) {
 	acct1, err := accountsDB.NewAccount(ctx, 0, time.Time{}, "foo")
 	require.NoError(t, err)
 
+	sess1RootKeyID := macIDToRootKeyID(sess1.ID)
```
bitromortac: so here we could use `NewSuperMacaroonRootKeyID` instead?

ViktorT-11: Thanks, updated to use your suggestion!

firewall/request_logger.go
ViktorT-11: Thanks a lot for the deep investigation! Unfortunately though, s also communicated offline we an have actions listed in litd which have a different prefix than the `SuperMacaroonRootKeyPrefix`.
If you run litd with the `firewall.request-logger.level=all` config option set, any calls using the pure lnd `admin.macaroon`, will log actions that start with bytes of just zeros.
Additionally, if users have created a macaroon with `lncli bakemacaroon` command and set a value for the `root_key_id` param that uses another prefix than the `SuperMacaroonRootKeyPrefix`, calls using that macaroon will also use another prefix than the `SuperMacaroonRootKeyPrefix`.
I do however think that no macaroons that were used for litd sessions nor accounts should ever start with any other prefix than the `SuperMacaroonRootKeyPrefix`.

We therefore decided to go ahead with this PR.

firewall/request_logger.go
ViktorT-11: Correct, the `kvdb` store will only store the 4 byte MacaroonID, and not the full 8 bytes. Therefore all callsites that uses the persisted `AddActionReq` will still only use the `MacaroonIdentifier` still.

firewalldb/actions.go
ViktorT-11: > or actually, to keep the API intuitive, cant se instead _only_ have the MacRootKeyID field here & remove the MacID field. Then extract the relevant format at the DB layer. ie, in the kvdb layer, take the root key ID and compute the mac ID.

Hmmm I'm not sure if I misunderstand you here, but as this is also used as the return value for an `action`, we would need to set the `MacRootKeyID` when returning an `kvdb` action if we were to remove the `MacaroonIdentifier` field. However, since we can't regenerated the full `MacRootKeyID` from just the `MacaroonIdentifier` that wouldn't be possible. OR am I misunderstanding your suggestion? 

> related to above question, if one of these is stored in kvdb only and the other in sql only, lets add this warning in the comments

Good idea. I added a warning to the `MacaroonRootKeyID` that it will not be set for `kvdb` database backends.

ViktorT-11: Potentially, we could make the struct returned from a persisted action, different from the struct used when persisting a new action. I.e. not use the `AddActionReq` struct for both.

firewalldb/actions_sql.go
ViktorT-11: Idid not set the field previously to currently only have one field set when reading a persisted action, to make it as little confusing as possible at callsites, and ensure no callsites used the `MacaroonRootKeyID` field yet. But I agree that this potentially even makes it more confusing. So therefore changed so that we set the `MacaroonRootKeyID` field when reading persisted actions on a an sql databasebackend :)

ViktorT-11:
Thanks a lot for the reviews @ellemouton & @bitromortac! I've addressed your feedback in the best way I could. I do agree that it's a bit confusing that there's both a `MacaroonIdentifier` & a `MacaroonRootKeyID` in the `AddActionReq` struct. I did not find a way to remove one of them though, so if you have feedback on how we'd do so, that'd be much appreciated :)

firewalldb/actions.go
bitromortac: Perhaps you could move `MacaroonIdentifier` from `AddActionReq` to `Action` and keep the full root key unpopulated on retrieval?

ellemouton: > we would need to set the MacRootKeyID when returning an kvdb action if we were to remove the MacaroonIdentifier field

but we only ever use 4 of the bytes right? and i think we only ever use it in ListActions for the RPC? 
so could potentially just let the bbolt impl append 4 zero bytes to whatever it has stored in order to construct the full rootkeyId?. 

bitromortac's idea is also good - we can also just completely remove the Req struct from being embedded in the action. We could rather just have an `ActionsInfo` struct which both the ActionsReq and Actions structs can embed. then it makes things clearer and doesnt require that we store everything that we needed to construct the ActionsReq. 

Anyways, gonna leave this up to you - but i do think it makes things easier to read. So leaving as non-blocking comment

ViktorT-11: > so could potentially just let the bbolt impl append 4 zero bytes to whatever it has stored in order to construct the full rootkeyId?.

Ah good idea! I actually liked that version most, so I went a head and implemented that. I therefore also added a last commit that removes the `MacaroonIdentifier` field. If we want to further change the structure of the `Action` & `AddActionReq` struct, I'd suggest we take that in a future follow-up once we've decided how we'd like it structured :)

bitromortac: Those changes look good! Just a small suggestion to add a method to cut down on code repetition: [patch.txt](https://github.com/user-attachments/files/22642788/patch.txt).

bitromortac (approved):
LGTM :tada: (just have a minor suggestion to cut down code repetition a bit)

firewalldb/actions.go
ViktorT-11: Cool thanks :pray:! Applied the patch as an extra helper function commit :)

PR #1153, opened 2025-10-10
[sql-54] firewalldb: handle deleted sessions during the kv stores + privacy mapper migration

ViktorT-11:
If the user has deleted their `session.db` file, but kept their `rules.db` file, there can exist kv entries + privacy mapper pairs  that point to a now deleted session ID. Such kv entries and privacy mapper pairs should be ignored during the migration, as they are cannot be used anymore.

This PR addresses that edge case.

### DIFF
--- a/firewalldb/sql_migration_test.go
+++ b/firewalldb/sql_migration_test.go
@@ -459,6 +459,14 @@ func TestFirewallDBMigration(t *testing.T) {
 			populateDB: multipleSessionsAndPrivacyPairs,
 		},
 		{
+			name:       "deleted session with privacy pair",
+			populateDB: deletedSessionWithPrivPair,
+		},
+		{
+			name:       "deleted and existing sessions with privacy pairs",
+			populateDB: deletedAndExistingSessionsWithPrivPairs,
+		},
+		{
 			name:       "random privacy pairs",
 			populateDB: randomPrivacyPairs,
 		},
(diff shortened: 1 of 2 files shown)


firewalldb/sql_migration_test.go
```diff
@@ -585,6 +593,34 @@ func sessionSpecificEntries(t *testing.T, ctx context.Context, boltDB *BoltDB,
...
+func sessionSpecificEntriesDeletedSession(t *testing.T, ctx context.Context,
+	boltDB *BoltDB, sessionStore session.Store, _ accounts.Store,
+	_ *rootKeyMockStore) *expectedResult {
+
+	groupAlias := getNewSessionAlias(t, ctx, sessionStore)
+
+	_ = insertTempAndPermEntry(
+		t, ctx, boltDB, testRuleName, groupAlias, fn.None[string](),
+		testEntryKey, testEntryValue,
+	)
```
bitromortac, line 626: nit: you could perhaps add one kv entry (not deleting its session), to ensure that there is actually a kv entry created using this pattern in the happy case in addition to this code, same in the other test

ViktorT-11: Thanks :pray:! Added two extra tests for this, as I do think it adds value to have a test that clearly shows that no kv entries are migrated at all if all linked sessions are deleted :)

PR #1097, opened 2025-06-23
Multi rfq send itest

GeorgeTsagk:
## Description

Enhances the multi-rfq itest to cover multi-rfq send functionality. In the following topology
```
	  /---[sats]--> Erin --[assets]--\
	 /                                \
	/                                  \

 Charlie  -----[sats]-->  Dave  --[assets]---->Fabia

	\                                  /
	 \                                /
	  \---[sats]--> Yara --[assets]--/
```

we now also get Fabia to pay Charlie back, with amounts that exceed the capacity of each individual channel. We also use a hold invoice to validate that multiple HTLCs were added in different channels.

Related PRs (in order of merge sequence):

- [x] https://github.com/lightningnetwork/lnd/pull/9980
- [ ] https://github.com/lightninglabs/taproot-assets/pull/1613
- [ ] https://github.com/lightninglabs/loop/pull/968

PR #1155, opened 2025-10-15
Merge tapd side-branch to `master`

GeorgeTsagk:
### Description

For the last couple of months we have been isolating tapd specific PRs on the `tapd-main-branch` side branch. This has been continuously rebasing on top of `master` and any conflicts have been gradually resolved.

The PRs that contributed to `tapd-main-branch` are:

1. https://github.com/lightninglabs/lightning-terminal/pull/1119
2. https://github.com/lightninglabs/lightning-terminal/pull/1097
3. https://github.com/lightninglabs/lightning-terminal/pull/1106
4. https://github.com/lightninglabs/lightning-terminal/pull/1138

Some single-commits that are not contained in any PR were direct pushes to the `tapd-main-branch`, which was not protected. These commits do routine/maintenance stuff like bumping versions or changing config values, so I hope they're easy to go through.

#### Purpose of this PR

This is definitely not a call to go through all the diff and verify the content itself, but more like an ACK around version bumps or any changes in the test harness itself. The content of the itests has been thoroughly reviewed and context should be available in the PRs linked above.

### DIFF
--- a/docs/release-notes/release-notes-0.16.0.md
+++ b/docs/release-notes/release-notes-0.16.0.md
@@ -1,0 +1,45 @@
+# Release Notes
+
+- [Lightning Terminal](#lightning-terminal)
+    - [Bug Fixes](#bug-fixes)
+    - [Functional Changes/Additions](#functional-changesadditions)
+    - [Technical and Architectural Updates](#technical-and-architectural-updates)
+- [Integrated Binary Updates](#integrated-binary-updates)
+    - [LND](#lnd)
+    - [Loop](#loop)
+    - [Pool](#pool)
+    - [Faraday](#faraday)
+    - [Taproot Assets](#taproot-assets)
+- [Contributors](#contributors-alphabetical-order)
+## Lightning Terminal
+
+### Bug Fixes
+
+### Functional Changes/Additions
+
+### Technical and Architectural Updates
+
+## RPC Updates
+
+## Integrated Binary Updates
+
+- [Bumped dependencies](https://github.com/lightninglabs/lightning-terminal/pull/1155/commits/537ed05776f64a3bbbdbd9177c8be329bf847890)
+for LND, taproot-assets, loop, lndclient.
+- Enhanced tap-channel integration tests to cover more edge cases and features,
+and also added coverage to verify support for channel versioning:
+    - https://github.com/lightninglabs/lightning-terminal/pull/1106
+    - https://github.com/lightninglabs/lightning-terminal/pull/1097
+    - https://github.com/lightninglabs/lightning-terminal/pull/1138
+ 
+
+### LND
+
+### Loop
+
+### Pool
+
+### Faraday
+
+### Taproot Assets
+
+# Contributors (Alphabetical Order)


proto/lnd.proto
```diff
@@ -341,6 +341,13 @@ service Lightning {
     */
     rpc SubscribeInvoices (InvoiceSubscription) returns (stream Invoice);
 
+    /* lncli: `deletecanceledinvoice`
+    DeleteCanceledInvoice removes a canceled invoice from the database. If the
+    invoice is not in the canceled state, an error will be returned.
+    */
+    rpc DeleteCanceledInvoice (DelCanceledInvoiceReq)
```
gijswijs, line 348: It took me a while to understand this commit: 4e52894d9cb040e0fae21d30052c23caea9e9f0d

The confusion stems from the description: regenerate protos

I didn't understand where the new endpoints where coming from if you are just regenerating protos. Obviously they come from the current state of the protos in the `lnd` and `loop` projects. Maybe explain that in the commit message?

`Copying in latest version of proto-files from lnd and loop and regenerating protos.`

itest/litd_custom_channels_test.go
```diff
@@ -4512,7 +4996,7 @@ func testCustomChannelsDecodeAssetInvoice(ctx context.Context,
 	}
 
 	const decimalDisplay = 6
-	itestAsset = &mintrpc.MintAsset{
+	tcAsset := &mintrpc.MintAsset{
```
gijswijs, line 4999: That's a great catch from @jtobin. Shouldn't we also fix this for `testCustomChannelsOraclePricing`? Because that's also a test that mutates the global `itestAsset` variable. I would also add a godoc comment to `itestAsset` to explain that we should consider it immutable.

```
// itestAsset is a predefined asset that is used in the tests below. It
// should be considered immutable.
```

itest/litd_node.go
```diff
@@ -652,7 +652,10 @@ func renameFile(fromFileName, toFileName string) {
 func (hn *HarnessNode) overrideFlagsAndBinary(backwardCompat map[string]string,
 	binary string, args []LitArgOption) (string, []LitArgOption) {
 
-	if backwardCompat == nil {
+	noopArg := WithLitArg("taproot-assets.channel.noop-htlcs", "")
```
gijswijs, line 655: Referring back to @Roasbeef comment here: https://github.com/lightninglabs/lightning-terminal/pull/1138/commits/9b7c17252e89554c5d7d69830cd070525b2f4b73#r2422820720

iiuc he suggests to turn `noop-htlcs` on by default. Not specifically in the context of itests, but just for the `tapd` client in general. The flag would then be used to turn `noop-htlcs` off. (Rename the flag to `no-noop-htlcs` or something)

For the itest it would mean you wouldn't have to pass the argument here, but you *would* have to pass it for the cases you want to have it specifically turned off.

itest/network_harness.go
```diff
@@ -777,13 +777,64 @@ func (n *NetworkHarness) RestartNodeNoUnlock(node *HarnessNode,
...
+	cfg := &suspendCfg{}
+
+	for _, opt := range opts {
+		opt(cfg)
+	}
+
+	// If the upgrade flag was set, delete any entry from the backwards
+	// compatibility map. This will force the node to use the latest build
+	// of LiT.
+	if cfg.upgrade {
```
gijswijs, line 827: :+1: :+1: I just love how this integrates so nicely with Oli's `backwardCompat` stuff. Super elegant.

itest/litd_custom_channels_test.go
```diff
@@ -2265,6 +2303,318 @@ func testCustomChannelsBreach(ctx context.Context, net *NetworkHarness,
...
+	require.NoError(t.t, err)
+
+	proofFile, err = proof.DecodeFile(proofResp.RawProofFile)
+	require.NoError(t.t, err)
+	require.Equal(t.t, 4, proofFile.NumProofs())
+	latestProof, err = proofFile.LastProof()
+	require.NoError(t.t, err)
+
+	// This proof should contain the STXO exclusion proofs
+	stxoProofs = latestProof.InclusionProof.CommitmentProof.STXOProofs
```
gijswijs, line 2614: :+1:

gijswijs (changes requested):
Did a first pass and made some comments. I think it's almost there.

itest/litd_node.go
GeorgeTsagk: I think we can make the switch to being by-default on, without changing the semantics of the config itself. We can just let it be "true" if not defined in our default config in tapd. This way the above line leads to no harm (won't need to update it too).

proto/lnd.proto
ViktorT-11: I think potentially you can also include these changes in the previous commit, and mention the protos bump in the commit description. This is to ensure that every commit will `build` correctly.

go.mod
```diff
@@ -25,22 +25,22 @@ require (
 	github.com/lightninglabs/lightning-terminal/autopilotserverrpc v0.0.3
 	github.com/lightninglabs/lightning-terminal/litrpc v1.0.2
 	github.com/lightninglabs/lightning-terminal/perms v1.0.1
-	github.com/lightninglabs/lndclient v0.19.0-12
-	github.com/lightninglabs/loop v0.31.2-beta
-	github.com/lightninglabs/loop/looprpc v1.0.8
-	github.com/lightninglabs/loop/swapserverrpc v1.0.15
+	github.com/lightninglabs/lndclient v0.20.0-1
+	github.com/lightninglabs/loop v0.31.3-beta.0.20251014011913-03de0a8ed734
```
ViktorT-11, line 29: This bump is required in order to be compatible with `lnd` & `tapd` right? Because with this bump, we'll need to wait for the next `loop` release (which should be imminent) before we can do a `litd` release.

go.mod
```diff
@@ -25,22 +25,22 @@ require (
...
+	github.com/lightninglabs/loop/looprpc v1.0.10-0.20251009095951-719355211f63
+	github.com/lightninglabs/loop/swapserverrpc v1.0.17-0.20251009095951-719355211f63
 	github.com/lightninglabs/pool v0.6.6-beta
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.3
 	github.com/lightninglabs/pool/poolrpc v1.0.1
-	github.com/lightninglabs/taproot-assets v0.6.1
-	github.com/lightninglabs/taproot-assets/taprpc v1.0.8-0.20250716163904-2ef55ba74036
-	github.com/lightningnetwork/lnd v0.19.3-beta
+	github.com/lightninglabs/taproot-assets v0.7.0-rc1.0.20251014100421-92189161aa7a
```
ViktorT-11, line 35: Also, with this, we'll need to wait for `tapd` `v0.7.0-rc2` before we can do a release as well.

go.mod
GeorgeTsagk: Yes, `loop` had to be bumped in order to reflect some of our latest changes in the `tapd` codebase (which it depends on).

go.mod
GeorgeTsagk: Yes, when all deps are properly tagged we'll have to make one final version bumping PR.

Issue #1158, opened 2025-10-21
upgrade go version to 1.24.8


ZZiigguurraatt:
lnd is currently using go 1.24.8 (https://github.com/lightningnetwork/lnd/blob/ab0177c1a10f0ef8227dc56eb44f25dc55716eed/go.mod#L219) and litd is currently using go 1.24.6 (https://github.com/lightninglabs/lightning-terminal/blob/d6b75578be9b034b612f8d67def9f830bc99c116/go.mod#L3) .

I can't compile litd master with lnd master because of this mismatch.

PR #1159, opened 2025-10-21
build: bump go to v1.24.8

ZZiigguurraatt:
Fixes https://github.com/lightninglabs/lightning-terminal/issues/1158

I'm not sure if there are any other code changes that should be made to accommodate golang version 1.24.8 and why litd has not yet upgraded but it looks like what I've done is similar to https://github.com/lightninglabs/lightning-terminal/pull/1144/commits/69ca6587b72a6324cf9b9660ebec6804b56936c2 .


ZZiigguurraatt:
Gemini's comments don't make any sense to me.

PR #1162, opened 2025-10-27
[sql-56] session: fix `FeatureConfig`s with an empty config set in kvdb to sql migration

ViktorT-11:
In the bbolt store, an empty config for a feature is represented as an empty array, while in for the SQL store, the same config is represented as nil. Therefore, in the scenario where a specific feature has an empty config, we override the SQL FeatureConfig for that feature to also be set to an empty array. This is needed to ensure that the deep equals check in the migration validation does not fail in this scenario.

### DIFF
--- a/session/sql_migration.go
+++ b/session/sql_migration.go
@@ -198,6 +198,7 @@ func migrateSessionsToSQLAndValidate(ctx context.Context,
 		overrideSessionTimeZone(migratedSession)
 		overrideMacaroonRecipe(kvSession, migratedSession)
 		overrideRemovedAccount(kvSession, migratedSession)
+		overrideFeatureConfig(kvSession, migratedSession)
 
 		if !reflect.DeepEqual(kvSession, migratedSession) {
 			diff := difflib.UnifiedDiff{
@@ -529,4 +530,30 @@ func overrideRemovedAccount(kvSession *Session, migratedSession *Session) {
 			return kvSession.AccountID
 		},
 	)
+}
+
+// overrideFeatureConfig overrides a specific feature's config for the SQL
+// session in a certain scenario:
+//
+// In the bbolt store, an empty config for a feature is represented as an empty
+// array, while in for the SQL store, the same config is represented as nil.
+// Therefore, in the scenario where a specific feature has an empty config, we
+// override the SQL FeatureConfig for that feature to also be set to an empty
+// array. This is needed to ensure that the deep equals check in the migration
+// validation does not fail in this scenario.
+func overrideFeatureConfig(kvSession *Session, mSession *Session) {
+	// If FeatureConfig is not set for both sessions, we return early.
+	if kvSession.FeatureConfig == nil || mSession.FeatureConfig == nil {
+		return
+	}
+
+	migratedConf := *mSession.FeatureConfig
+	for featureName, config := range *kvSession.FeatureConfig {
+		// If the config is empty for the bbolt feature, and nil for the
+		// SQL version, we override the SQL version to match the bbolt
+		// version.
+		if len(config) == 0 && migratedConf[featureName] == nil {
+			migratedConf[featureName] = make([]byte, 0)
+		}
+	}
 }
--- a/session/sql_migration_test.go
+++ b/session/sql_migration_test.go
@@ -81,6 +81,10 @@ func TestSessionsStoreMigration(t *testing.T) {
 			// session doesn't.
 			overrideRemovedAccount(kvSession, sqlSession)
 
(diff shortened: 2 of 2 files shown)


session/sql_migration.go
```diff
@@ -198,6 +198,7 @@ func migrateSessionsToSQLAndValidate(ctx context.Context,
 		overrideSessionTimeZone(migratedSession)
 		overrideMacaroonRecipe(kvSession, migratedSession)
 		overrideRemovedAccount(kvSession, migratedSession)
+		overrideFeatureConfig(kvSession, migratedSession)
```
bitromortac, line 201: Another approach would be to convert `nil` configs to empty ones in `unmarshalFeatureConfigs`. Wouldn't that be better because that way we ensure that both session (kvdb/sql) retrievals are consistent?

ViktorT-11: We discussed this offline, and we decided to continue with the override approach in favour of changing the  `unmarshal` function. While that approach will likely work, we'll keep the same approach we've used to tackle similar issues previously, and consider to in the future instead change to the `unmarshal` approach. But when doing so, we'll change for all current overrides, and not just this case.

session/sql_migration.go
```diff
@@ -530,3 +531,30 @@ func overrideRemovedAccount(kvSession *Session, migratedSession *Session) {
...
+	if kvSession.FeatureConfig != nil && mSession.FeatureConfig != nil {
+		for featureName, config := range *kvSession.FeatureConfig {
+			// If the config is empty for the feature, we override
+			// the SQL version.
+			if len(config) == 0 {
+				migratedConf[featureName] = make([]byte, 0)
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
ViktorT-11: Applied the suggested cleanup.

session/sql_migration.go
```diff
@@ -530,3 +531,28 @@ func overrideRemovedAccount(kvSession *Session, migratedSession *Session) {
...
+//
+// In the bbolt store, an empty config for a feature is represented as an empty
+// array, while in for the SQL store, the same config is represented as nil.
+// Therefore, in the scenario where a specific feature has an empty config, we
+// override the SQL FeatureConfig for that feature to also be set to an empty
+// array. This is needed to ensure that the deep equals check in the migration
+// validation does not fail in this scenario.
+func overrideFeatureConfig(kvSession *Session, mSession *Session) {
+	// If FeatureConfig is not set for both sessions, we return early.
+	if kvSession.FeatureConfig == nil || mSession.FeatureConfig == nil {
```
bitromortac, line 546: Should this be an `&&`? Otherwise it could be that `mSession.FeatureConfig` is nil, which would panic below.

session/sql_migration.go
```diff
@@ -530,3 +531,28 @@ func overrideRemovedAccount(kvSession *Session, migratedSession *Session) {
...
+		return
+	}
+
+	migratedConf := *mSession.FeatureConfig
+	for featureName, config := range *kvSession.FeatureConfig {
+		// If the config is empty for the feature, we override
+		// the SQL version.
+		if len(config) == 0 {
+			migratedConf[featureName] = make([]byte, 0)
```
bitromortac, line 556: What if we erroneously migrated an empty feature config to a non-empty one, then we'd silently override the migrated one, right? So maybe we could just convert the migrated config if it is nil to an empty one?

session/sql_migration.go
bitromortac: From a call-site perspective I think it's fine to keep the `nil` here https://github.com/lightninglabs/lightning-terminal/blob/5e22106fb57bb0af71bc25be72a4c6abde723e39/session_rpcserver.go#L1550-L1555

This just leads to an empty string.

session/sql_migration.go
ViktorT-11: Added an extra `nil` just to ensure this can't be an issue :)

session/sql_migration.go
ViktorT-11: Hmmm no, it's the opposite right? If it was an `&&`, then we could panic if `kvSession.FeatureConfig` is `!= nil` and `mSession.FeatureConfig` is `== nil`?

session/sql_migration.go
ViktorT-11: Thanks for checking :pray:!

ellemouton (approved):
🙏

session/sql_migration.go
bitromortac: Ah totally, sorry got that mixed up

PR #1170, opened 2025-11-10
update lndclient and enable waiting until chain notifier is ready

starius:
Bump lndclient to include https://github.com/lightninglabs/lndclient/pull/255

Use new option BlockUntilChainNotifier in LiT itself and in Loop.


ViktorT-11 (approved):
LGTM :pray:!

PR #1173, opened 2025-11-13
go.mod: bump lnd, loop, taproot-assets

ffranr:
#### Target Versions

* `lnd@v0.20.0`
* `taproot-assets@v0.7.0`
* `taproot-assets/taprpc@v1.0.12`
* `loop@v0.31.6-beta`
* `loop/looprpc@v1.0.12`
* `loop/swapserverrpc@v1.0.19`

### DIFF
--- a/docs/release-notes/release-notes-0.16.0.md
+++ b/docs/release-notes/release-notes-0.16.0.md
@@ -39,13 +39,13 @@ and also added coverage to verify support for channel versioning:
 
 ### LND
 
-* Updated [`lnd` to
-  `v0.20.0-beta.rc2`](https://github.com/lightninglabs/lightning-terminal/pull/1163).
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1173): Bump:
+  `lnd@v0.20.0-beta`.
 
 ### Loop
 
-* Updated [`loop` to
-  `v0.31.5-beta`](https://github.com/lightninglabs/lightning-terminal/pull/1163).
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1173): Bump:
+  `loop@v0.31.6-beta`, `loop/looprpc@v1.0.12`, `loop/swapserverrpc v1.0.19`.
 
 ### Pool
 
@@ -53,7 +53,7 @@ and also added coverage to verify support for channel versioning:
 
 ### Taproot Assets
 
-* Updated [`taproot-assets` to
-  `v0.7.0-rc2`](https://github.com/lightninglabs/lightning-terminal/pull/1163).
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1173): Bump:
+  `taproot-assets@v0.7.0`, `taproot-assets/taprpc@v1.0.11`.
 
 # Contributors (Alphabetical Order)


ffranr:
I'll update this PR once we have new tags for tap.

docs/release-notes/release-notes-0.16.0.md
```diff
@@ -41,11 +41,15 @@ and also added coverage to verify support for channel versioning:
 
 * Updated [`lnd` to
```
darioAnongba: doesn't make sense to keep this update in the changelog right? Since `0.16.0` updates to `0.20.0-beta` and not `0.20.0-beta.rc2`

docs/release-notes/release-notes-0.16.0.md
```diff
@@ -41,11 +41,15 @@ and also added coverage to verify support for channel versioning:
...
   `v0.20.0-beta.rc2`](https://github.com/lightninglabs/lightning-terminal/pull/1163).
+* Updated [`lnd` to
+  `v0.20.0-beta`](https://github.com/lightninglabs/lightning-terminal/pull/1173).
 
 ### Loop
 
 * Updated [`loop` to
```
darioAnongba: Same remark here

docs/release-notes/release-notes-0.16.0.md
```diff
@@ -55,5 +59,7 @@ and also added coverage to verify support for channel versioning:
 
 * Updated [`taproot-assets` to
```
darioAnongba: and here

docs/release-notes/release-notes-0.16.0.md
ffranr: I've been thinking about that also. I think it would be unnecessarily costly to, as standard, reconcile the release doc across all changes such that it reflected the final release state only. Personally, i think that it's sufficient that it's a sort of append only list of changes.

ffranr: I'll reconcile here. Removed the old entry.

.github/workflows/main.yml
```diff
@@ -21,7 +21,7 @@ env:
   # If you change this value, please change it in the following files as well:
   # /Dockerfile
   # /dev.Dockerfile
-  GO_VERSION: 1.24.9
+  GO_VERSION: 1.24.10
```
ViktorT-11: We should try to avoid going past the `lnd` Golang version which is currently at `1.24.9` in `v0.20.0`. Taking this offline to see if we can resolve this :)

GeorgeTsagk (approved):
LGTM, tACK, bACK (build ACK? new term I guess)

pending protos recompilation

docs/release-notes/release-notes-0.16.0.md
```diff
@@ -39,21 +39,21 @@ and also added coverage to verify support for channel versioning:
...
 ### Pool
 
 ### Faraday
 
 ### Taproot Assets
 
-* Updated [`taproot-assets` to
-  `v0.7.0-rc2`](https://github.com/lightninglabs/lightning-terminal/pull/1163).
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1173): Bump:
+  `taproot-assets@v0.7.0`, `taproot-assets/taprpc@v1.0.11`.
```
ViktorT-11, line 57: Nit: Just a note that this differs how we normally write these bumps in the release notes + we don't really mention the actual RPC dependency updates here. But let's not wait for a full CI job just to address that. Instead, I'll update this in my upcoming release PR.

PR #1174, opened 2025-11-14
sessions: delete session after registration failure

bitromortac:
Due to recent changes we'll always create a session first with status `reserved`. After the session is created and if the autopilot rejects session registration we are left with a session that can't be used. This is problematic for session linking, as the reserved session would need to be revoked first in order to link another session, but we don't return a reference to the new session for revocation. The session isn't usable anyhow so I suggest to delete it after a failure. I only did that for the registration failure, but we could extend the case to other errors later in the session creation.

An alternative to this would be to move the session into a terminal `failed` state, which we don't have so far and we'd end up storing non-useful sessions, but it could be useful for debugging.

### DIFF
--- a/session_rpcserver.go
+++ b/session_rpcserver.go
@@ -1286,6 +1286,23 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
 		privacyFlags.Serialize(),
 	)
 	if err != nil {
+		// If we tried to link to a previous session, we delete the
+		// newly created session in the case of errors to avoid having
+		// non-revoked sessions lying around.
+		if len(req.LinkedGroupId) != 0 {
+			log.Infof("Session registration with autopilot " +
+				"server failed, deleting the newly created " +
+				"session")
+
+			deleteErr := s.cfg.db.DeleteReservedSession(
+				ctx, sess.ID,
+			)
+			if deleteErr != nil {
+				log.Errorf("Error deleting session after "+
+					"failed linking attempt: %v", deleteErr)
+			}
+		}
+
 		return nil, fmt.Errorf("error registering session with "+
 			"autopilot server: %v", err)
 	}


session_rpcserver.go
```diff
@@ -1213,6 +1213,18 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
 		return nil, fmt.Errorf("error creating new session: %v", err)
 	}
 
+	// If we tried to link to a previous session, we delete the newly
+	// created session in the case of errors to avoid having non-revoked
+	// sessions lying around.
+	fail := func(err error) error {
```
ViktorT-11: Nit: Not sure we need a separate lambda function for this when it's only being called once?

session_rpcserver.go
```diff
@@ -1286,8 +1298,8 @@ func (s *sessionRpcServer) AddAutopilotSession(ctx context.Context,
 		privacyFlags.Serialize(),
 	)
 	if err != nil {
-		return nil, fmt.Errorf("error registering session with "+
-			"autopilot server: %v", err)
+		return nil, fail(fmt.Errorf("error registering session with "+
```
ViktorT-11: Given that we're not 100% sure here that this errors due to the session not having been stored on the autopilot server (as this could simplify fail for loss of the connection mid execution for example), I can see the motivation still persisting this on the `litd` node, to aim to not have sessions stored on the autopilot server, which do not exist on the corresponding `litd` node.
But since that can already occur for various other reasons as well, I think it's fine to just delete it.

session_rpcserver.go
bitromortac: right, I wasn't sure if we also want to delete the session on an error later on, so I left it for now

db/sqlc/queries/sessions.sql
```diff
@@ -18,6 +18,10 @@ WHERE id = $2;
 DELETE FROM sessions
 WHERE state = $1;
 
+-- name: DeleteSession :exec
```
ellemouton, line 21: could maybe in future make things more defensive (both here and above by restricting this to only allowing it for "where status=reserved" - but i guess it's ok for now since your store interface does this protection

session/kvdb_store.go
```diff
@@ -458,69 +461,120 @@ func (db *BoltStore) DeleteReservedSessions(_ context.Context) error {
...
+		if err != nil {
+			return err
+		}
+
+		if session.State != StateReserved {
+			return fmt.Errorf("session not in reserved state, is "+
+				"%v", session.State)
+		}
+
+		return deleteSession(sessionBucket, session)
```
ellemouton, line 577: as mentioned above, could just make the db query enforce this constraint

session_rpcserver.go
bitromortac: I removed that and added it directly

db/sqlc/queries/sessions.sql
bitromortac: right, I see your point enforcing that on the db layer would be safer, one downside of this is that we'd need to hard code an integer here for the type (i.e. `4`), so we'd lose some type safety, right? for this reason I kept the current approach, but I'll definitely consider that approach in future work :pray:

session_rpcserver.go
ViktorT-11: Not sure if you did 😄? But this is non-blocking. So I'm ok to Merge as is.

PR #1183, opened 2025-12-02
subservers: fail LiT startup when integrated sub-server boot fails

### LINKED ISSUE #1181: Fail `litd` startup on `tapd` startup error when `taproot-assets-mode=enable`
Since `lnd` may call into `tapd` when it is enabled, `tapd` must be started successfully before those calls are made.

Previously, we've allowed `litd` to start successfully even if one or more sub-servers (except `lnd`) failed during startup. The reasoning behind this decision was that bugs in individual sub-servers should not prevent users from running the rest of their `litd` node until those bugs are patched.

However, because `tapd` is tightly coupled to `lnd` when it is in use, we need to modify this behavior specifically for the `tapd` sub-server. If `tapd` is enabled and errors during the startup, `litd` should also fail to start.

ffranr:
Closes https://github.com/lightninglabs/lightning-terminal/issues/1181

---

* Treat integrated sub-servers as fatal to startup and return an error if any fail to start.
* Propagate integrated sub-server startup errors to LiT so it stops launching and records the failure status.
* Update docs to reflect integrated sub-servers are now critical to startup.

### DIFF
--- a/itest/litd_mode_integrated_test.go
+++ b/itest/litd_mode_integrated_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/lightninglabs/faraday/frdrpc"
 	"github.com/lightninglabs/lightning-node-connect/mailbox"
 	"github.com/lightninglabs/lightning-terminal/litrpc"
+	"github.com/lightninglabs/lightning-terminal/subservers"
 	"github.com/lightninglabs/loop/looprpc"
 	"github.com/lightninglabs/pool/poolrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc"
@@ -29,6 +30,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/walletrpc"
 	"github.com/lightningnetwork/lnd/macaroons"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/net/http2"
 	"google.golang.org/grpc"
@@ -498,6 +500,135 @@ func testCriticalTapStartupFailure(ctx context.Context, net *NetworkHarness,
 	case <-time.After(5 * time.Second):
 		t.Fatalf("litd did not exit after tapd startup failure")
 	}
+}
+
+// testNonCriticalLoopStartupFailure ensures LiT continues running when a
+// non-critical integrated sub-server (loopd) fails to start during boot.
+func testNonCriticalLoopStartupFailure(ctx context.Context,
+	net *NetworkHarness, t *harnessTest) {
+
+	// Force loopd into an invalid config combination to guarantee a
+	// startup failure in integrated mode.
+	node, err := net.NewNode(
+		t.t, "NonCriticalLoop", nil, false, false,
+		"--loop.maxl402cost=1",
+		"--loop.maxlsatcost=1",
+	)
+	require.NoError(t.t, err)
+
+	stopNodeGracefully := func() error {
+		conn, err := node.ConnectRPC(true)
+		if err != nil {
+			return err
+		}
+		defer conn.Close()
+
+		stopCtx, cancel := context.WithTimeout(ctx, defaultTimeout)
+		defer cancel()
+
+		lndConn := lnrpc.NewLightningClient(conn)
+		_, err = lndConn.StopDaemon(stopCtx, &lnrpc.StopRequest{})
+		return err
+	}
+
+	defer func() {
+		t.Logf("Shutting down node: %s", node.Cfg.LitDir)
+		if err := stopNodeGracefully(); err != nil {
+			t.Logf("Graceful shutdown failed, killing node: %v",
+				err)
+			_ = net.KillNode(node)
+		}
+		select {
+		case <-node.processExit:
+		case <-time.After(5 * time.Second):
+			_ = net.KillNode(node)
+		}
+		_ = node.cleanup()
+	}()
+
+	// Wait for the TLS cert so we can connect to the status service.
+	t.Logf("Waiting for lit TLS cert to be created: %s",
+		node.Cfg.LitTLSCertPath)
+	require.Eventually(t.t, func() bool {
+		_, err := os.Stat(node.Cfg.LitTLSCertPath)
+		return err == nil
+	}, 10*time.Second, 1*time.Second,
+		"expected lit TLS cert to be created",
+	)
+
+	t.Logf("Connecting to litd service")
+	rawConn, err := connectLitRPC(
+		ctx, node.Cfg.LitAddr(), node.Cfg.LitTLSCertPath, "",
+	)
+	require.NoError(t.t, err)
+	defer rawConn.Close()
+
+	statusConn := litrpc.NewStatusClient(rawConn)
+
+	fetchStatus := func() (*litrpc.SubServerStatusResp, error) {
+		callCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
+		defer cancel()
+
+		return statusConn.SubServerStatus(
+			callCtx, &litrpc.SubServerStatusReq{},
+		)
+	}
+
+	waitForStatus := func(desc string,
+		predicate func(*litrpc.SubServerStatusResp) bool) {
+
+		t.t.Helper()
+
+		var lastErr error
+		var lastResp *litrpc.SubServerStatusResp
+
+		ok := assert.Eventually(t.t, func() bool {
+			lastResp, lastErr = fetchStatus()
+			if lastErr != nil || lastResp == nil {
+				return false
+			}
+
+			return predicate(lastResp)
+		}, 20*time.Second, 200*time.Millisecond)
+		if ok {
+			return
+		}
+
+		if lastErr != nil {
+			t.t.Fatalf("%s: last error: %v", desc, lastErr)
+		}
+
+		t.t.Fatalf("%s: last status: %#v", desc, lastResp.SubServers)
+	}
+
+	t.Logf("Waiting for loop to report non-critical loop startup failure")
+	waitForStatus("expected loop startup error to be reported",
+		func(resp *litrpc.SubServerStatusResp) bool {
+			status, ok := resp.SubServers[subservers.LOOP]
+			if !ok || status.Running || status.Error == "" {
+				return false
+			}
+
+			res := strings.Contains(status.Error, "maxl402cost")
+			return res
+		},
+	)
+
+	t.Logf("Waiting for lnd to report running state")
+	waitForStatus("expected lnd to be running",
+		func(resp *litrpc.SubServerStatusResp) bool {
+			status, ok := resp.SubServers[subservers.LND]
+			return ok && status.Running
+		},
+	)
+
+	t.Logf("Waiting for litd to report running state")
+	waitForStatus("expected litd to be running",
+		func(resp *litrpc.SubServerStatusResp) bool {
+			status, ok := resp.SubServers[subservers.LIT]
+			return ok && status.Running
+		},
+	)
 }
 
 // integratedTestSuite makes sure that in integrated mode all daemons work
--- a/itest/litd_mode_integrated_test.go
+++ b/itest/litd_mode_integrated_test.go
@@ -465,6 +465,41 @@ func testModeIntegrated(ctx context.Context, net *NetworkHarness,
 	)
 }
 
+// testCriticalTapStartupFailure ensures LiT exits quickly when a critical
+// integrated sub-server (tapd) fails to start during boot.
+func testCriticalTapStartupFailure(ctx context.Context, net *NetworkHarness,
+	t *harnessTest) {
+
+	// Force tapd to use a postgres backend with an invalid host to
+	// guarantee a startup failure in integrated mode. This config will
+	// error during tapd's startup (after wallet unlock) rather than during
+	// litd's config validation phase.
+	node, err := net.NewNode(
+		t.t, "FailFastTap", nil, false, false,
+		"--taproot-assets.databasebackend=postgres",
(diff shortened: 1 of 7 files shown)


subservers/manager.go
```diff
@@ -104,13 +105,16 @@ func (s *Manager) GetServer(name string) (SubServer, bool) {
 }
 
 // StartIntegratedServers starts all the manager's sub-servers that should be
-// started in integrated mode.
+// started in integrated mode. An error is returned if any integrated sub-server
+// fails to start.
```
ViktorT-11: We only want errors in the `tapd` sub-server to cause the startup process to error, and not **any** sub-server. The motivation for why this should specifically target `tapd` right now, is that we've realized that `tapd` is critically coupled to `lnd` in the scenario that the user has taproot-asset channels, in the sense that `lnd` can't function if `tapd` registers as the aux-unit, but then shuts down.
 
Therefore, I suggest that you add a `map` of "critical" sub-server names as a param to the `StartIntegratedServers` function. If a sub-server then errors during the startup, you'll check if that sub-server name exists in the passed `map`, and if so return an error.

We actually originally designed this function specifically to not error the whole startup process if any of the integrated sub-servers fails to start. The motivation behind that is that we do not want bugs in one sub-server, to disable users from using the other sub-servers until a fix for the erroring sub-server has been released. I.e. a bug that causes `pool` to error on startup, shouldn't disable people from performing loops until it's been fixed. Of course users can alawys `disable` `pool` in that scenario, but for UX purposes we do not want require users to understand that it's `pool` that's actually failing the startup process, and then additionally understand how they `disable` pool, as this adds friction.

subservers/manager.go
```diff
@@ -126,11 +130,19 @@ func (s *Manager) StartIntegratedServers(lndClient lnrpc.LightningClient,
 		)
 		if err != nil {
 			s.statusServer.SetErrored(ss.Name(), err.Error())
+			errs = append(errs, fmt.Sprintf("%s: %v", ss.Name(), err))
```
ViktorT-11: I think we want to return an error imminently here, and not proceed to start all sub-servers before we return the error. The motivation for that is that we want `lnd` to shutdown asap if a "critical" sub-server errors during startup, and that could be delayed if we proceed to start the rest of the sub-servers before returning the error, and hence proceed to shutdown `lnd`.

terminal.go
```diff
@@ -768,9 +768,12 @@ func (g *LightningTerminal) start(ctx context.Context) error {
 	// Both connection types are ready now, let's start our sub-servers if
 	// they should be started locally as an integrated service.
 	createDefaultMacaroons := !g.cfg.statelessInitMode
-	g.subServerMgr.StartIntegratedServers(
+	err = g.subServerMgr.StartIntegratedServers(
 		g.basicClient, g.lndClient, createDefaultMacaroons,
 	)
+	if err != nil {
+		return fmt.Errorf("could not start integrated sub-servers: %w", err)
+	}
```
ViktorT-11, line 802: Just returning an error here will currently actually not shutdown `lnd`. Instead this will cause the `Run` function to wait until the `shutdownInterceptor.ShutdownChannel()` is triggered, and keep `lnd` up and running until that happens. This will lead to the same errors that we've seen previously, as `lnd` will still proceed to try to call into `tapd`.

Instead we want to make sure `lnd` is shutdown asap here. One way of achieving that would be to call `shutdownInterceptor.RequestShutdown()` here. Alternatively, you could call `g.basicClient.StopDaemon()` here.
I suggest the second alternative, as the single `Interceptor` pattern is a bit of a code smell that we eventually actually want to get rid of :).

ViktorT-11 (commented):
Thanks for this @ffranr :pray:!

In addition to the feedback I've commented below, this new behaviour definitely needs `itest` coverage. 

I think we actually started working on this at the same time, and I have local branch with draft code implementing this + `itest` coverage. If you want to, i can clean that up and push it so that you can cherry-pick that to make it more simple for you. Let me know if that'd be helpful :).

ellemouton (commented):
havent actually looked at the diff here yet, just want to make a note in case: we should just make sure that the daemon still runs & ie, that the status server still gets served

subservers/manager.go
```diff
@@ -126,11 +133,18 @@ func (s *Manager) StartIntegratedServers(lndClient lnrpc.LightningClient,
 		)
 		if err != nil {
 			s.statusServer.SetErrored(ss.Name(), err.Error())
+
+			if criticalIntegratedSubServers.Contains(ss.Name()) {
+				return fmt.Errorf("%s: %v", ss.Name(), err)
+			}
+
```
ellemouton, line 187: but i think this then will result in complete shutdown which i dont think we want? we want the status server to remain running.

itest/litd_mode_integrated_test.go
```diff
@@ -465,6 +465,38 @@ func testModeIntegrated(ctx context.Context, net *NetworkHarness,
 	)
 }
 
+// testCriticalTapStartupFailure ensures LiT exits quickly when a critical
+// integrated sub-server (tapd) fails to start during boot.
+func testCriticalTapStartupFailure(ctx context.Context, net *NetworkHarness,
```
ViktorT-11, line 472: Thanks for this :pray:! Similar to the unit tests, we should also have an `itest` which covers a startup when a non-critical sub-server errors during startup.

subservers/manager.go
```diff
@@ -126,11 +156,18 @@ func (s *Manager) StartIntegratedServers(lndClient lnrpc.LightningClient,
 		)
 		if err != nil {
 			s.statusServer.SetErrored(ss.Name(), err.Error())
+
+			if criticalIntegratedSubServers.Contains(ss.Name()) {
```
ViktorT-11, line 181: Thanks :pray:!

Note that the commit message is outdated and does not cover the "criticalIntegratedSubServers" part.

terminal.go
```diff
@@ -405,6 +405,31 @@ func (g *LightningTerminal) Run(ctx context.Context) error {
...
+			defer cancel()
+
+			_, err := client.StopDaemon(
+				stopCtx, &lnrpc.StopRequest{},
+			)
+			if err != nil {
+				log.Warnf("Error stopping lnd after failed "+
+					"start: %v", err)
+			}
+		}
```
ViktorT-11: This will now attempt to shutdown `lnd` for all types of startup errors, not just the "critical sub-server" error.

I'm not sure that this is exactly what we want, for 2 main reasons:
1. This will apply in `remote` mode as well, and attempt to shutdown the remote `lnd` node for any `litd` related startup error.
2. Even in integrated mode, it's not certain that we've gotten to executing the code that sets `basicLNDClient` when `g.start` errors, despite having started `lnd`. Meaning that `lnd` will not be shutdown in that scenario. That leads to a quite unpredictable behaviour, where `lnd` will sometimes be shutdown when `g.start` errors, and sometimes not.

I therefore think we should only execute this if we specifically error during the startup of a critical `litd` sub-server, to keep the behaviour predictable, and not effect remote `lnd` nodes. You may want to guard the shutdown request and require that `litd` is not running with a "remote" `lnd` node, even though the current critical sub-server startup error logic cannot occur when `lnd` is in remote mode.

terminal.go
```diff
@@ -405,6 +405,31 @@ func (g *LightningTerminal) Run(ctx context.Context) error {
...
+
+		if err := g.shutdownSubServers(); err != nil {
+			log.Errorf("Error shutting down after failed start: %v",
+				err)
+		}
+
+		return startErr
```
ViktorT-11: Similar to what @ellemouton has already commented, this will now shutdown `litd` before the `shutdownInterceptor.ShutdownChannel()` has been triggered, which we would like to avoid.

terminal.go
```diff
@@ -405,6 +405,31 @@ func (g *LightningTerminal) Run(ctx context.Context) error {
...
+			if err != nil {
+				log.Warnf("Error stopping lnd after failed "+
+					"start: %v", err)
+			}
+		}
+
+		if err := g.shutdownSubServers(); err != nil {
+			log.Errorf("Error shutting down after failed start: %v",
+				err)
+		}
```
ViktorT-11: We should only do this after the `shutdownInterceptor.ShutdownChannel()` has been triggered, similar to the current logic below :)

ViktorT-11 (commented):
Thanks for the updates :rocket:! Leaving some additional feedback in addition to @ellemouton's feedback.

terminal.go
```diff
@@ -405,6 +406,42 @@ func (g *LightningTerminal) Run(ctx context.Context) error {
...
+		// follow the same shutdown ordering as normal exits.
+		shutdownInterceptor.RequestShutdown()
+		<-shutdownInterceptor.ShutdownChannel()
+
+		if err := g.shutdownSubServers(); err != nil {
+			log.Errorf("Error shutting down after failed start: %v",
+				err)
+		}
+
+		return startErr
```
ViktorT-11: This comment still applies due to these lines:

https://github.com/lightninglabs/lightning-terminal/pull/1183#discussion_r2631337670

This now also applies to any type of `startErr` and not just a critical startup error.
We don't want to trigger a shutdown of `litd`, and instead want the user initiate it, so that `litd`'s status server remains up and running, until the user decides to shutdown `litd`.

I think your best choice is just to remove these lines I have marked :). Then let the logic outside of the `if` case scope handle the waiting for the shutdown like we already do prior to this PR.

ViktorT-11 (commented):
Thanks for the updates :pray:! Looking better, but commenting of a few lines I think you should just remove to address some previously mentioned feedback.

itest/litd_mode_integrated_test.go
```diff
@@ -465,6 +467,167 @@ func testModeIntegrated(ctx context.Context, net *NetworkHarness,
...
+	t *harnessTest) {
+
+	// Force tapd to bind to an invalid port to guarantee a startup failure
+	// in integrated mode.
+	node, err := net.NewNode(
+		t.t, "FailFastTap", nil, false, false,
+		"--taproot-assets.rpclisten=127.0.0.1:65536",
```
ViktorT-11: I believe this config flag will make the node error on the `config` validation logic in `litd`, and not when `tapd` is attempted to be started after the wallet is unlocked.

I suggest using these config options instead, as those will actually error during `tapd`'s startup:
```
taproot-assets.databasebackend=postgres
taproot-assets.postgres.host=tapd-postgres.invalid
```

itest/litd_mode_integrated_test.go
```diff
@@ -465,6 +467,167 @@ func testModeIntegrated(ctx context.Context, net *NetworkHarness,
...
+		_ = net.ShutdownNode(node)
+	}()
+
+	select {
+	case procErr := <-net.ProcessErrors():
+		require.ErrorContains(t.t, procErr, "invalid port")
+	case <-time.After(15 * time.Second):
+		t.Fatalf("expected tapd startup failure to be reported")
+	}
+
```
ViktorT-11, line 496: Depending on @ellemouton's response to my broader question in the main review comment, we likely want to expand the test code here, to ensure that the `status` endpoint & `lnd` node is at the state we expect it to be.

ViktorT-11 (commented):
After testing this PR locally, I’ve realized that this comment isn’t really addressable with how `litd` is currently designed:
https://github.com/lightninglabs/lightning-terminal/pull/1183#discussion_r2631337670

The reason is that shutting down `lnd` will always trigger the `shutdownInterceptor.ShutdownChannel()` to be sent over, which in turn also shuts down `litd`. This behavior stems from the single-interceptor pattern we currently use in `litd`, which we eventually want to move away from.

As a result, I don’t think it’s possible to keep `litd`’s status endpoint up and running while still shutting down `lnd` in response to critical sub-server startup errors.

Given that, I believe the behavior introduced by this PR is effectively the best we can do until the single-interceptor pattern is addressed.

@ellemouton, do you think it’s critical that `litd`, and therefore the status endpoint, remains up when critical sub-servers fail on startup (which also causes `lnd` to shut down)? If so, we’ll need to address the single-interceptor issue first. Otherwise, can we ship this as-is for now and tackle the interceptor issue at a later stage?

itest/litd_mode_integrated_test.go
ffranr: Issue to track interceptor decoupling: https://github.com/lightninglabs/lightning-terminal/issues/1201

ellemouton:
> @ellemouton, do you think it’s critical that litd, and therefore the status endpoint, remains up when critical sub-servers fail on startup (which also causes lnd to shut down)? 

I think we need to bring in @jamaljsr  & @levmi  in on this discussion since the main reason for the status endpoint is to provide context to TW re why things are not starting up. The alternative is: just dont fully start up meaning that LNC sessions wont work & users will just see that they cannot connect to their node

litbot-9000:
@ellemouton: review reminder
@ziggie1984: review reminder
@ffranr, remember to re-request review from reviewers when ready

litbot-9000:
@ellemouton: review reminder
@ziggie1984: review reminder
@ffranr, remember to re-request review from reviewers when ready

litbot-9000:
@ellemouton: review reminder
@ziggie1984: review reminder
@ffranr, remember to re-request review from reviewers when ready

litbot-9000:
@ellemouton: review reminder
@ziggie1984: review reminder
@ffranr, remember to re-request review from reviewers when ready

litbot-9000:
@ellemouton: review reminder
@ziggie1984: review reminder
@ffranr, remember to re-request review from reviewers when ready

litbot-9000:
@ziggie1984: review reminder
@ellemouton: review reminder
@ffranr, remember to re-request review from reviewers when ready

litbot-9000:
@ziggie1984: review reminder
@ellemouton: review reminder
@ffranr, remember to re-request review from reviewers when ready

litbot-9000:
@ziggie1984: review reminder
@ellemouton: review reminder
@ffranr, remember to re-request review from reviewers when ready

PR #1186, opened 2025-12-06
Add tranche-based parallel itests and enable in CI

ffranr:
* Add tranche splitting/shuffling support to itest harness and wire per-tranche logging/exit reporting.
* Introduce `itest_parallel` runner/Make targets (with and without backward-compat binaries) and per-tranche log directories.
* Update GitHub Actions itest matrix to use parallel tranches (4).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -258,9 +258,9 @@ run-itest-only:
 
 itest-only: build-itest install-backward-compat-versions run-itest-only
 
-itest: app-build build-itest itest-only
+itest: app-build itest-only
 
-itest-no-backward-compat: app-build build-itest build-itest run-itest-only
+itest-no-backward-compat: app-build build-itest run-itest-only
 
 itest-parallel: app-build build-itest install-backward-compat-versions build-itest-binary
 	@$(call print, "Running integration tests in parallel.")


ffranr:
I think branch protection rules need to be updated for CI to pass.

itest/litd_test.go
```diff
@@ -12,6 +15,40 @@ import (
...
+	// explicit tranche is selected.
+	defaultRunTranche uint = 0
+)
+
+var (
+	// testCasesSplitTranches controls how many tranches the full itest list
+	// is divided into for parallel execution.
+	testCasesSplitTranches = flag.Uint(
+		"splittranches", defaultSplitTranches,
+		"split the test cases in this many tranches and run the tranche "+
```
ViktorT-11: Nit: Line length exceeds 80 chars in my IDE here.

itest/litd_test.go
```diff
@@ -12,6 +15,40 @@ import (
...
+	// balance workload across tranches.
+	shuffleSeedFlag = flag.Uint64(
+		"shuffleseed", 0, "if set, shuffles the test cases using this "+
+			"as the source of randomness",
+	)
+
+	// testCasesRunTranche selects which tranche (0-based) to execute.
+	testCasesRunTranche = flag.Uint(
+		"runtranche", defaultRunTranche,
+		"run the tranche of the split test cases with the given (0-based) "+
```
ViktorT-11: Same as above.

itest/litd_test.go
```diff
@@ -12,6 +15,40 @@ import (
...
+
+	// defaultRunTranche is the default tranche index to execute when no
+	// explicit tranche is selected.
+	defaultRunTranche uint = 0
+)
+
+var (
+	// testCasesSplitTranches controls how many tranches the full itest list
+	// is divided into for parallel execution.
+	testCasesSplitTranches = flag.Uint(
```
ViktorT-11: Nit: could potentially include some of the docs present in `lnd` here:
```
	// testCasesSplitParts is the number of tranches the test cases should
	// be split into. By default this is set to 1, so no splitting happens.
	// If this value is increased, then the -runtranche flag must be
	// specified as well to indicate which part should be run in the current
	// invocation.
```

itest/litd_test.go
ffranr: These lines don't look too long in my IDE. I've also checked with this commit which passes:
https://github.com/lightninglabs/lightning-terminal/pull/1189

ellemouton: @ffranr  - viktor is correct. These are more than 80 chars.

ellemouton (commented):
great addition! looks good. Viktor is correct about the line length though, so LGTM after that. 

I've updated the branch rules - so CI should work now

itest/litd_test.go
ffranr: My IDE was misconfigured. I added an `.editorconfig` to this PR so future developers won’t run into the same issue.

PR #1189, opened 2025-12-09
multi: enforce 80-char line length and add `nolint:ll` comments

ffranr:
- Updated `.golangci.yml` to enforce a maximum line length of 80 characters.
- Reformatted code comments and function parameters to adhere to the new line-length restriction.
- Added `nolint: lll` comments where line wrapping was not applicable.


ellemouton (commented):
i think maybe we should take this opportunity to use the same custom `ll` linter that LND uses? that way things work nicely for structured logging

ellemouton (approved):
thank you!! much appreciated 🚀

firewall/privacy_mapper.go
```diff
@@ -511,9 +513,10 @@ func handleFeeReportResponse(db firewalldb.PrivacyMapDB,
 
 				chanPoint := c.ChannelPoint
 				if !flags.Contains(session.ClearChanIDs) {
-					chanPoint, err = firewalldb.HideChanPointStr(
-						ctx, tx, chanPoint,
-					)
+					chanPoint, err =
+						firewalldb.HideChanPointStr(
+							ctx, tx, chanPoint,
+						)
```
ViktorT-11, line 519: I don't think we've used this pattern previously in `litd`, where you'd start the values on the right side of a `=` char on a new line. I'm ok with introducing it though, but I'm also in slight favour of just using `nolint:ll` for these types of lines. There's quite of few of these examples throughout the changed files.

I'll let you decide and I'm definitely ok with merging this as is if you think we should use this pattern :)!

ffranr: I'm not a big fan of the newline after `=` style. But I do think that in some cases its better than the alternative. It is used in tap and loop:
https://github.com/lightninglabs/loop/blob/f76c15ca16442b19b296b332ff77fb684729ecfa/staticaddr/withdraw/manager.go#L660-L664

So I think it should be ok for now. Thanks for the review!

PR #1198, opened 2026-01-13
itest: add channel capacity assertions

jtobin:
Adds assertions that check a responder's view of channel capacity after funding. This should catch if passive assets have been included amongst the funding proofs (see lightninglabs/taproot-assets#1932).

### DIFF
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -3,6 +3,7 @@ package itest
 import (
 	"context"
 	"crypto/rand"
+	"encoding/hex"
 	"fmt"
 	"math"
 	"math/big"
@@ -35,6 +36,7 @@ import (
 	"github.com/lightningnetwork/lnd/lntest"
 	"github.com/lightningnetwork/lnd/lntest/node"
 	"github.com/lightningnetwork/lnd/lntest/port"
+	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/lightningnetwork/lnd/lntypes"
 	"github.com/lightningnetwork/lnd/lnwallet/chainfee"
 	"github.com/lightningnetwork/lnd/lnwire"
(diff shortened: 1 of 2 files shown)


jtobin:
Note that this itest uses a group key to create a passive asset, whereas the user in lightninglabs/taproot-assets#1932 did not. It doesn't actually matter here, because the effect is the same -- it was just easier to construct an itest this way.

itest/litd_custom_channels_test.go
```diff
@@ -5480,3 +5482,146 @@ func testCustomChannelsMultiChannelPathfinding(ctx context.Context,
...
+	require.Equal(t.t, outpointA, outpointB,
+		"assets must be in same UTXO for passive asset test")
+
+	syncUniverses(t.t, charlieTap, dave)
+	mineBlocks(t, net, 1, 0)
+
+	// Fund a channel using the group key. We request only fundingAmount
+	// worth of assets, which should come from Asset A. Asset B should
+	// become passive and not appear in Dave's view of the channel.
+	const fundingAmount = 50_000
```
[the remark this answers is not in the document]
jtobin: I think assetA might be chosen deterministically anyway, but addressed in 9e641ba30cfd4a2b2fce35e02181833c25c514e7.

jtobin:
(FYI this passes for taproot-assets's current 'main'.)

itest/litd_custom_channels_test.go
```diff
@@ -5480,3 +5482,147 @@ func testCustomChannelsMultiChannelPathfinding(ctx context.Context,
 		t.t, alice, bob, invoiceResp.PaymentRequest, assetIDPences,
 	)
 }
+
+// testCustomChannelsPassiveAssets tests that passive assets (assets in the same
+// input commitment but not used for channel funding) are not incorrectly
+// included in the funding proofs sent to the responder
```
ffranr: `not incorrectly included in` -> `absent from`

itest/litd_custom_channels_test.go
```diff
@@ -5480,3 +5482,147 @@ func testCustomChannelsMultiChannelPathfinding(ctx context.Context,
...
+			return fmt.Errorf("capacity should be %d, not "+
+				"inflated by passive assets, got %d",
+				fundingAmount, chanData.Capacity)
+		}
+
+		return nil
+	}, defaultTimeout)
+	require.NoError(t.t, err)
+
+	// Clean up.
```
ffranr: drop trivial comment?

itest/litd_custom_channels_test.go
ffranr: looking at the test body, I dont think we're actually inspecting proofs but just that the passive assets are not included as funding assets. So i think this doc comment is too strong and our check is actually at a higher level. If that's correct I think we just need to adjust the itest doc comment.

ffranr (approved):

itest/litd_custom_channels_test.go
jtobin: Fixed in a5c54f98c5d52a597d62ae6fdbfe38b7f8c9bc3f.

itest/litd_custom_channels_test.go
jtobin: Fixed in a5c54f98c5d52a597d62ae6fdbfe38b7f8c9bc3f.

PR #1200, opened 2026-01-14
itest: fix nil error deref

ellemouton:
Here we fix a nil pointer panic that is caused by an 'err' var being deferenced before first checking that it is not nil. This was caused by an oversite in a refactor commit that was attempting to improve line length.

Panic introduced in: https://github.com/lightninglabs/lightning-terminal/pull/1189/files#diff-62c19b273dec93f7183dca48128ca4e3022a0c9736ffa5cfa47ad9bdfdd59508


ffranr (approved):

jtobin (approved):
LGTM. :+1:

PR #1207, opened 2026-01-21
[sql-57] Add accounts kvdb->SQL migration itest

ViktorT-11:
Based on #1114

Implements part of step 6. of "Phase 3" in https://github.com/lightninglabs/lightning-terminal/issues/917.

This PR adds the foundational components for testing the kvdb -> SQL migration within the itest framework, and implements itest coverage for the migration of the accounts store.

I'll add follow up PRs which will also add implement itest coverage for the sessions store, and the full firewalldb store.   

The benefit of adding migration coverage through itests compared to just unit tests, is that this will test the full migration flow in litd, as it would be executed in production for a user who switches from a bbolt database backend to an SQL database backend.
Additionally, as itests have access to the full litcli, we can also assert parts of the migration through litcli commands, as the end user would experience it.

### DIFF
--- a/itest/litd_migration_test.go
+++ b/itest/litd_migration_test.go
@@ -4,11 +4,22 @@ package itest
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
+	"path/filepath"
+	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
 	terminal "github.com/lightninglabs/lightning-terminal"
+	"github.com/lightninglabs/lightning-terminal/accounts"
+	"github.com/lightninglabs/lightning-terminal/db/sqlcmig6"
+	"github.com/lightninglabs/lightning-terminal/litrpc"
+	"github.com/lightningnetwork/lnd/clock"
+	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lntest"
+	"github.com/lightningnetwork/lnd/lntypes"
+	"github.com/lightningnetwork/lnd/lnwire"
+	"github.com/lightningnetwork/lnd/sqldb/v2"
 	"github.com/stretchr/testify/require"
 )
 
@@ -32,14 +43,16 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
 		t.t.Skipf("Skipping kvdb migration test for bbolt backend")
 	}
 
+	ctxt, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+
 	// Step 1: Start a node with a bbolt backend.
 	// We want to start from an explicit bbolt backend regardless of the
 	// command line flag used to run the itests.
 	migNode, err := net.NewNode(
 		t.t, "Migrator", nil, false, true,
 		fmt.Sprintf(
-			"--databasebackend=%s",
-			terminal.DatabaseBackendBbolt,
+			"--databasebackend=%s", terminal.DatabaseBackendBbolt,
 		),
 	)
 	require.NoError(t.t, err)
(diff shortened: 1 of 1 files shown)


itest/litd_migration_test.go
```diff
@@ -60,12 +76,41 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
 	)
 	defer closeChannelAndAssert(t, net, migNode, channelOp, false)
 
+	// Setup a raw gRPC connection used to set up RPC clients.
+	rawConn, err := connectRPC(
+		ctxt, migNode.Cfg.LitAddr(), migNode.Cfg.LitTLSCertPath,
+	)
+	require.NoError(t.t, err)
+	defer rawConn.Close()
```
bitromortac: I think we don't need that as it's closed later explicitly

itest/litd_migration_test.go
```diff
@@ -93,41 +153,448 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
...
+		{
+			hash:   lntypes.Hash{4, 4, 4, 4},
+			amount: 200,
+			status: lnrpc.Payment_FAILED,
+		},
+	}
+
+	for _, payment := range payments {
+		_, err = store.UpsertAccountPayment(
+			context.Background(), *parsedID, payment.hash,
```
bitromortac: Could you use the test context here?

itest/litd_migration_test.go
```diff
@@ -93,41 +153,448 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
...
+// assertAccountMigrationDataSQL connects to the SQL DB and queries account
+// data to assert migration results.
+func assertAccountMigrationDataSQL(ctx context.Context, t *harnessTest,
+	node *HarnessNode, data accountMigrationData) {
+
+	dbPath := filepath.Join(
+		node.Cfg.LitDir, node.Cfg.NetParams.Name, "litd.db",
+	)
+
+	sqlStore, err := sqldb.NewSqliteStore(
```
bitromortac: Is testing with Sqlite enough?

itest/litd_migration_test.go
```diff
@@ -93,41 +153,448 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
...
+	expectations["migration-invoices"] = invoiceExpectation
+
+	// Mimics accounts/sql_migration_test.go "account with payments".
+	// NOTE: As there's no way to insert payments into the accounts store
+	// directly by just using `litcli`, we insert the payments outside
+	// of this function, by directly connecting to the bbolt db.
+	createAccount(50000, 0, "migration-payments")
+	paymentExpectation := expectations["migration-payments"]
+	paymentExpectation.payments = 4
+	expectations["migration-payments"] = paymentExpectation
```
bitromortac, line 270: Can't we take use the RouterClient and add the account to the context to make some payments? `payNode` seems to be ready for that

itest/litd_migration_test.go
```diff
@@ -93,41 +153,448 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
...
+	}
+
+	// Note that other migration tests for sessions and firewalldb may add
+	// accounts, and therefore we only check that the accounts store
+	// contains at least the expected amount, and not exactly the amount.
+	if len(accountsList) < len(data.expectations) {
+		return fmt.Errorf(
+			"expected at least %d accounts, got %d",
+			len(data.expectations), len(accountsList),
+		)
```
bitromortac: You could pass in the test to do require assertions

itest/litd_migration_test.go
```diff
@@ -93,41 +153,448 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
 	require.NoError(t.t, err)
 
 	// Step 4: Assert data via litcli where possible.
-	assertMigrationDataViaLitCLI()
+	assertMigrationDataViaLitCLI(ctxt, t, migNode, accountsData)
```
bitromortac: Can't we just use the RPC to query the contents? I'm not sure it adds a lot using litcli, as it's only a wrapper

itest/litd_migration_test.go
```diff
@@ -93,41 +153,448 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
...
 	)
 	require.NoError(t.t, err)
 
 	// Step 6: Assert data in SQL via direct database access.
-	assertMigrationDataSQL()
+	assertMigrationDataSQL(ctxt, t, migNode, accountsData)
 
 	// Step 7: Assert data via litcli where possible.
-	assertMigrationDataViaLitCLI()
+	assertMigrationDataViaLitCLI(ctxt, t, migNode, accountsData)
```
bitromortac: It seems like here it would also be enough to query the RPC?

itest/litd_migration_test.go
ViktorT-11: I added coverage for postgres also here.

itest/litd_migration_test.go
ViktorT-11: I changed this to use the RPCs directly instead. The reason I used `litcli` previously is that I intended to mimic the what most users experience when they query LiT, but I agree that it probably don't make most sense.

itest/litd_migration_test.go
ViktorT-11: Do I think this is a bigger questions for not just this PR, but the rest of the migration `itest` PRs as well.

We can either try to create all the data we can through the different clients, but with the drawback of not being able to mimic the `unit` tests as they cover more data variations than what's possible to generate through the clients. One example here is to manage to generate an account payment with the status `unkonwn`, which some users have in their DB due to having generated it with old releases, but which is hard to generate through the clients today.

Therefore I think we have to make the choice here if we want to try to cover as much data variation as possible in the `itests`, which requires a direct connection to the `db`, or if we're ok with just generating the data we can through the clients.

I can see the motivation for both alternatives:
- If we generate through the clients, the data becomes much more realistic and may show errors with `nil` handling in the migrations for example. But we won't be able to mimic all unit tests.
- If we do it through a direct connection to the db, we'll be able to cover more data variation.

Which variant do you think most sense? Would be interesting to hear @ellemouton opinion here as well after she has reviewed the PR.

bitromortac: I'd argue that we want to test different flows here that couple different systems. If we replicate all unit tests, itests become heavier and then one could (provocatively) ask why do we have the unit tests? I can see that sometimes redundancy is good, but it also comes at a cost, so I would lean to replicate as much as possible only using RPCs, but not a very strong opinion

ViktorT-11: Yeah I guess my point of view and what I've tried to simulate is that the `itest` tests and ensures the full migration flow works for a user which has a bbolt db populated with every type of data variation possible. Therefore, the means of how the user actually generated that data is less relevant.

But I can also see the motivation for not attempting to have such full coverage, and instead slim the `itest` down to only have:
* One account
* One session
* One kv stores entry
* One privacy mapper value
* One action

That then just tests that the full migration flow itself works and ignores the data variation, as the unit tests only covers data variation and doesn't really cover the full migration flow.

ViktorT-11: So IMO:

> I would lean to replicate as much as possible only using RPCs, but not a very strong opinion

I would disagree with this, as IMO it makes most sense to either decide to:
* Replicate all data variations in the `itest`, and how we generate that data is then less relevant.

OR 
* Ignore the data variations as the unit tests should cover that, and go with the minimal set of data needed to trigger all paths of the migration flow to ensure that the full migration flow itself works in the `itest`.

Having coverage for some data variation but not all in the `itest` seems like the least clear alternative as to what we're actually trying to test.

Or do you alternative viewpoint on this :)?

PR #1215, opened 2026-01-29
rules: allow non-present channels in restriction

bitromortac:
The `channel-restriction` rule previously failed if any channel in its deny list wasn't currently open. A routine channel closure could block a request with `invalid channel ID`. This change removes that strict check to ensure the firewall remains resilient to channel changes. It is fine to remove the check, because we can't operate on closed channels anyhow.

### DIFF
--- a/docs/release-notes/release-notes-0.16.1.md
+++ b/docs/release-notes/release-notes-0.16.1.md
@@ -16,6 +16,11 @@
 
 ### Bug Fixes
 
+* [Improve channel restriction
+  resilience](https://github.com/lightninglabs/lightning-terminal/pull/1215):
+  The channel restriction rule no longer fails to initialize when restricted
+  channels are closed.
+
 ### Functional Changes/Additions
 
 ### Technical and Architectural Updates
@@ -38,3 +43,5 @@
 ### Taproot Assets
 
 # Contributors (Alphabetical Order)
+
+* bitromortac


rules/channel_restrictions.go
```diff
@@ -139,28 +139,13 @@ func (c *ChannelRestrictMgr) maybeUpdateChannelMaps(ctx context.Context,
...
 		c.chanIDToPoint[id] = point
-
-		if id == chanID {
-			found = true
-		}
-	}
-
-	if !found {
-		return fmt.Errorf("invalid channel ID")
 	}
```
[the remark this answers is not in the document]
bitromortac: This is a valid point, but it looks like this solution has the same problem as with only using `ListChannels` will not populate the `chanIDToPoint` with closed channels or invalid ids. I came up with a different solution that uses a negative cache with cache refresh on querying for an unknown channel point.

rules/channel_restrictions.go
```diff
@@ -118,19 +129,28 @@ func (c *ChannelRestrictMgr) EmptyValue() Values {
...
-	// maps.
-	_, ok := c.chanIDToPoint[chanID]
-	if ok {
+	// If we already know about all these channels or have checked them,
+	// then we don't need to do anything.
+	if !needsSync {
+		c.mu.Unlock()
 		return nil
 	}
+	c.mu.Unlock()
```
ViktorT-11: Doesn't releasing the lock here create concurrency risks?
Say you have 2 concurrent calls to `maybeUpdateChannelMaps`:

1.
Call 1:
- Takes lock, resolves in `needsSync = true`
- Releases lock here

2.
Call 2: 
- Takes lock, resolves in `needsSync = false`
- Releases lock here and returns

3.
Call 1: 
- Syncs and returns

The issue with the above, is that the sync by "Call 1", could lead to "Call 2" requiring a `needsSync = true`?

bitromortac: Yes, I think there can be concurrency issues, good call (I believe the worst case would be a retry error). I opted for the simpler locking during the whole function call to simplify the logic.

bitromortac:
@claude review

rules/channel_restrictions.go
```diff
@@ -174,6 +184,67 @@ func (c *ChannelRestrictMgr) getChannelID(point string) (uint64, bool) {
...
+	}
+
+	// Channel not found, but all deny list entries are mapped, so this
+	// unknown channel is definitely not in our restriction list.
+	return 0, false, false
+}
+
+// hasCheckedIDs returns true if any of the given channel IDs are present in the
+// manager's checkedIDs map.
+func (c *ChannelRestrictMgr) hasCheckedIDs(chanIDs []uint64) bool {
```
ViktorT-11: Like lint it alluding to, this & `clearCheckedIDs` is now unused.

bitromortac: Done :+1:

docs/release-notes/release-notes-0.16.1.md
```diff
@@ -37,4 +42,7 @@
 
 ### Taproot Assets
 
-# Contributors (Alphabetical Order)
\ No newline at end of file
+# Contributors (Alphabetical Order)
+
+
+
```
bitromortac: will fix this in with the next push

PR #1216, opened 2026-01-29
firewall: add request logger disable flag

ellemouton:
This introduces a new firewall.request-logger.disable config option to completely disable request logging. When disabled, the request logger interceptor is not instantiated, avoiding all logging overhead including database writes and request processing.

This is implemented as a separate disable flag rather than adding a new log level because the goal is to bypass the logging system entirely for performance reasons, not just filter events. A log level would still process and filter each request through the interceptor.

The change also adds validation to ensure autopilot remains enabled only when request logging is active, since autopilot relies on action logs for rule enforcement and auditing.

### DIFF
--- a/docs/release-notes/release-notes-0.16.1.md
+++ b/docs/release-notes/release-notes-0.16.1.md
@@ -23,6 +23,10 @@
 
 ### Functional Changes/Additions
 
+* Add a `firewall.request-logger.disable` [config 
+  flag](https://github.com/lightninglabs/lightning-terminal/pull/1216) and 
+  require action logging when autopilot is enabled.
+
 ### Technical and Architectural Updates
 
 ## RPC Updates
@@ -34,7 +38,7 @@
 ### Loop
 
 * [PR](https://github.com/lightninglabs/lightning-terminal/pull/1191): Bump:
-    `loop@v0.31.7-beta`.
+  `loop@v0.31.7-beta`.
 
 ### Pool
 
@@ -45,3 +49,4 @@
 # Contributors (Alphabetical Order)
 
 * bitromortac
+* Elle Mouton


docs/release-notes/release-notes-0.16.1.md
```diff
@@ -18,6 +18,10 @@
 
 ### Functional Changes/Additions
 
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1216): Add a
+  `firewall.request-logger.disable` config flag and require action logging when
+  autopilot is enabled.
```
ViktorT-11: You're adding and editing these in 2 different commits. So I would recommend only doing it in the last.

firewall/config.go
```diff
@@ -11,13 +11,21 @@ type Config struct {
 //
 //nolint:ll
 type RequestLoggerConfig struct {
+	// Disable completely disables request logging. This option exists as a
+	// separate flag rather than a log level because there are scenarios
+	// where logging should be entirely skipped (no interceptor, no
+	// database writes) rather than just filtered. Disabling improves
+	// performance by avoiding all logging overhead, whereas a log level
+	// would still process and filter each request.
+	Disable            bool               `long:"disable" description:"Disable request logging completely."`
```
ViktorT-11: I think we should probably also add a note here that the autopilot must be turned off if this option is set to  true.

terminal.go
```diff
@@ -1166,7 +1159,41 @@ func (g *LightningTerminal) startInternalSubServers(ctx context.Context,
...
 		privacyMapper,
 		g.accountService,
-		requestLogger,
+	}
+
+	// Autopilot relies on action logging for rule enforcement/auditing.
+	if !g.cfg.Autopilot.Disable && g.cfg.Firewall.RequestLogger.Disable {
+		return fmt.Errorf("request logger must be enabled when " +
+			"autopilot is active")
+	}
```
ViktorT-11: IMO this should error during the config validation process, and not error here. Erroring here will keep the status server up and running, and will make it harder for users to understand why the startup failed.

ellemouton:
looks good 🙏 nice itest!
 for future - good practice to keep og commiter as author of the original commits they added (even if you amend to them) and then add new things as new commiter :)

ViktorT-11:
Thanks :tada:!

> for future - good practice to keep og commiter as author of the original commits they added (even if you amend to them) and then add new things as new commiter :)

Yes sorry about that 😞. As I needed to remove stuff from the first commit to split out the release docs from the commit, I noticed now that splitting that commit removed you as the author. I then fixed up the actual release docs commit, with the "second" split commit from the first commit, which resulted in me becoming the author for the second commit as well. I can try to fix that if you'd like?

PR #1219, opened 2026-01-30
itest: bump to latest LND

darioAnongba:
Bumps LND to the latest version that includes needed fixes. Some fixes and other dependencies needed to be bumped as well to align with the latest LND and taproot-assets.
- `lnd`
- `sqldb`
- `migrate`
- `go` to `1.24.11`

**FIxes needed**

1. Fixes a race condition in the `custom_channels` itests that was causing intermittent failures after recent LND changes to channel close confirmation handling. The test was checking `ClosedChannels` RPC immediately after the close transaction was confirmed, but the channel arbitrator processes close events asynchronously. The `ClosedChannels` RPC filters out channels where `IsPending == true`, so if the query happens before `MarkChanFullyClosed` is called, the channel won't be found.
2. Increase minimum timelock delta to at least 24 as that is the new minimum accepted by LND. See https://github.com/lightningnetwork/lnd/pull/10331

### DIFF
--- a/itest/litd_custom_channels_test.go
+++ b/itest/litd_custom_channels_test.go
@@ -3266,7 +3266,7 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
 		},
 		BaseFeeMsat:   31337,
 		FeeRatePpm:    443322,
-		TimeLockDelta: 19,
+		TimeLockDelta: 25,
 	})
 	require.NoError(t.t, err)
 	require.Empty(t.t, resp.FailedUpdates)
@@ -3277,7 +3277,7 @@ func testCustomChannelsLiquidityEdgeCasesCore(ctx context.Context,
 		},
 		BaseFeeMsat:   42069,
 		FeeRatePpm:    223344,
-		TimeLockDelta: 18,
+		TimeLockDelta: 24,
 	})
 	require.NoError(t.t, err)
 	require.Empty(t.t, resp.FailedUpdates)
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -32,15 +32,15 @@ require (
 	github.com/lightninglabs/pool v0.6.6-beta
 	github.com/lightninglabs/pool/auctioneerrpc v1.1.3
 	github.com/lightninglabs/pool/poolrpc v1.0.1
-	github.com/lightninglabs/taproot-assets v0.7.0
-	github.com/lightninglabs/taproot-assets/taprpc v1.0.11
-	github.com/lightningnetwork/lnd v0.20.0-beta
+	github.com/lightninglabs/taproot-assets v0.7.1-0.20260127171608-666ec2acc675
+	github.com/lightninglabs/taproot-assets/taprpc v1.0.12-0.20260127171608-666ec2acc675
+	github.com/lightningnetwork/lnd v0.20.0-beta.rc4.0.20260129204118-7ccab028f0a8
```
ZZiigguurraatt, line 37: What version is this you are targeting? I feel like go often times uses the wrong tag names when it does this automatically. We have https://github.com/lightningnetwork/lnd/releases/tag/v0.20.1-beta.rc1 and are nearing `v0.20.1-beta.rc2`, not `v0.20.0-beta.rc4`. Or, are you targeting lnd `master`?

ZZiigguurraatt:
> * `go` to `1.24.11`

Does it make sense to move this to a separate PR so we can build litd with https://github.com/lightningnetwork/lnd/releases/tag/v0.20.1-beta.rc1 and not lnd master?

go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/lightninglabs/lightning-terminal
```
ellemouton, line 1: think this should be 2 separate commits. please also flesh out the commit message explaining the commit

ellemouton (commented):
please use commit messages to explain the purpose of each commit 🙏

PR #1225, opened 2026-02-06
Bump LND to latest

darioAnongba:
Bump of LND to latest version requires:

- Upgrading LND to 0.21.0 introduces a race condition in the `assets_test`. This race is introduced by changes in the chancloser. The channel may not be market as fully closed immediately after the close transaction is confirmed.
- LND 0.21.0 introduces a new limit on the timelock delta of `24`. Previous limit was 18 so a change is required to bump this limit in all tests to avoid errors.
- Bump `sqldb`
- Bump `taproot-assets` and `taprpc`

### DIFF
--- a/docs/release-notes/release-notes-0.16.1.md
+++ b/docs/release-notes/release-notes-0.16.1.md
@@ -26,6 +26,9 @@
 
 ### LND
 
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1225): Bump:
+    `lnd@v0.21.0-beta`, `lnd/sqldb@v1.0.13`.
+
 ### Loop
 
 * [PR](https://github.com/lightninglabs/lightning-terminal/pull/1191): Bump:
@@ -36,5 +39,8 @@
 ### Faraday
 
 ### Taproot Assets
+
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1225): Bump:
+    `taproot-assets@master`, `taproot-assets/sqltaprpcdb@master`.
 
 # Contributors (Alphabetical Order)


go.mod
```diff
@@ -34,13 +34,13 @@ require (
 	github.com/lightninglabs/pool/poolrpc v1.0.1
 	github.com/lightninglabs/taproot-assets v0.7.0
 	github.com/lightninglabs/taproot-assets/taprpc v1.0.11
-	github.com/lightningnetwork/lnd v0.20.0-beta
+	github.com/lightningnetwork/lnd v0.20.1-beta.rc1
```
ziggie1984: why only 20.1 tho, I would prefer referencing the current master commit ?

darioAnongba: me as well but that is forbidden by LiT policy apparently. It makes sense as referencing random commits that are not stable releases (or at least RCs) is not a great idea, but if tests pass then it should be okay.

cc: @ViktorT-11 

On a side note, LiT doesn't yet support `0.20.1`, that is not even released yet I guess the first step is to align LiT to `0.20.1` and then to `0.21.0` ? or is `0.20.1` just a patch?

ziggie1984: 20.1 is just a patch so we go first 20.1 and then master commit LND, but then you also need to update your comments you made during this PR, because you were referencing 21

ViktorT-11: Yes, we should preferebly not bump to an untagged version, so I'm in favour of using `v0.20.1-beta.rc1` here.
The reason being is that if we need make hotfix release, which for example https://github.com/lightninglabs/lightning-terminal/pull/1215 could motivate (as that solves issues for the `AutoFees` algo), we wouldn't be blocked by having an unreleased version of `lnd` merged to `master`.

By referencing `v0.20.1-beta.rc1`, we could instead make a `litd` RC release which ships with `lnd`  `v0.20.1-beta.rc1`.

darioAnongba: this PR does target `0.21`, I can have an other one for `0.20.1 `but this one can stay open in draft until it becomes relevant (when we have our first 0.21 RC tag?). It is used for testing before LND officially releases.

See this PR as example: https://github.com/lightninglabs/taproot-assets/pull/1958

ZZiigguurraatt: > On a side note, LiT doesn't yet support `0.20.1`

litd master now compiles with https://github.com/lightningnetwork/lnd/pull/10556 if I manually replace it

darioAnongba: Sorry I meant doesn't support as it's not the targeted version in `go.mod` on `master`. As per @ziggie1984  comment, because it's only a patch, we can skip that version and go straight to `0.21` with this PR ( that will stay open waiting on the RC1 tag).

darioAnongba:
ok the PR is now green targeting latest LND and TAPD. No need to approve but would be nice to get your thoughts on it. We can reference this branch anywhere we need latest LND (in taproot-assets or in tests, etc).

docs/release-notes/release-notes-0.16.1.md
```diff
@@ -26,6 +26,9 @@
 
 ### LND
 
+* [PR](https://github.com/lightninglabs/lightning-terminal/pull/1225): Bump:
+    `lnd@v0.21.0-beta`, `lnd/sqldb@v1.0.13`.
```
ViktorT-11, line 30: Since the `taproot-assets` entries below explicitly mentions that it updates the dependencies to `master`, maybe we should do the same here.

itest/litd_mode_integrated_test.go
```diff
@@ -314,15 +314,14 @@ var (
...
-		macaroonFn:        emptyMacaroonFn,
+		macaroonFn:        tapMacaroonFn,
 		requestFn:         tapUniverseRequestFn,
 		successPattern:    "\"runtime_id\":",
 		disabledPattern:   "unknown request",
 		allowedThroughLNC: true,
 		grpcWebURI:        "/universerpc.Universe/Info",
 		restWebURI:        "/v1/taproot-assets/universe/info",
 		canDisable:        true,
-		noAuth:            true,
```
ViktorT-11, line 325: Using a macaroon here & removing the `noAuth: true` contradicts what this is trying to test as it's no longer a whitelisted call, and makes the test redundant due to the `taprpc` test above.

Instead I recommend adding the extra `taproot-assets.allow-public-uni-proof-courier` config we added in the `sql-migration-base` branch:
https://github.com/lightninglabs/lightning-terminal/commit/e5a9bdad7e46eff7f35b412f01f33b415ff59f6f

ViktorT-11 (commented):
Generally looks good! We still need to discuss within the terminal team if we'd want to actually merge this to `master` or a side branch, so I'll let you know once we've made a decision.

docs/release-notes/release-notes-0.16.1.md
darioAnongba: Yes, I was waiting either on a tag or a decision on your end to update this to the correct tag. PR being blocked in the meantime.

PR #1233, opened 2026-02-19
[sql-63] session: add database benchmarking for session package

ViktorT-11:
Reopening #1231 as that PR was closed due to the base branch being deleted.

Based on https://github.com/lightninglabs/lightning-terminal/pull/1227

Implements part of step 7. of "Phase 3" in https://github.com/lightninglabs/lightning-terminal/issues/917.

This PR implements benchmarking tests for different database backends in the session package.

### DIFF
--- a/docs/benchmark_db_backends.md
+++ b/docs/benchmark_db_backends.md
@@ -8,6 +8,7 @@ This document describes how to run LiT benchmarks to compare the kvdb
 The benchmark suite currently covers these stores:
 
 - `accounts`: account CRUD and payment/invoice updates.
+- `session`: session creation, lookup, and state updates.
 
 ## Running Benchmarks
 
@@ -19,6 +20,7 @@ all supported backends in that package (kvdb-bbolt and native-sqlite).
 
 ```bash
 make unit-bench pkg=accounts
+make unit-bench pkg=session
 ```
 
 ### postgres (optional)


session/benchmark_test.go
```diff
@@ -0,0 +1,526 @@
...
+
+	for i := 0; i < b.N; i++ {
+		priv, err := btcec.NewPrivateKey()
+		require.NoError(b, err)
+
+		err = store.UpdateSessionRemotePubKey(
+			ctx, sess.ID, priv.PubKey(),
+		)
+		require.NoError(b, err)
+	}
```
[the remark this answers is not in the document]
ViktorT-11: Addressed.

session/benchmark_test.go
```diff
@@ -0,0 +1,528 @@
...
+		err = store.ShiftState(ctx, sess.ID, StateCreated)
+		require.NoError(b, err)
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		_, err := store.ListSessionsByState(
+			ctx, StateReserved,
```
bitromortac: does it matter if we use `StateReserved` or `StateCreated` here?

session/benchmark_test.go
```diff
@@ -0,0 +1,528 @@
...
+	require.NoError(b, err)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		_, err := store.GetSession(ctx, sess.ID)
+		require.NoError(b, err)
+	}
+}
```
bitromortac, line 235: nit: These type of tests could be misleading, as we only generate data once and then will get cached results (also different behavior of sqlite/postgres). Not sure if we should only test non-cached situations?

session/benchmark_test.go
ViktorT-11: It shouldn't really matter as the test creates 50 sessions of each state, but just to be sure I added a sub-test for querying each state :)

session/benchmark_test.go
ViktorT-11: I did it that way as generating a unique session for each step of the `for` loop dramatically increases the overall runtime for this test (not the benchmark time which is paused).

But I do agree that this is valid feedback, so I changed that. Alternatively, we could do something like this to limit the amount of sessions created (even though this will mean querying the same session multiple times):

```
// benchmarkSessionGet measures session lookup performance.
func benchmarkSessionGet(b *testing.B, store Store) {
	ctx := context.Background()

	// Seed a bounded pool of unique sessions outside the timed section.
	// Creating a fresh session on every loop iteration makes this benchmark
	// very slow in wall-clock time because each iteration still performs an
	// untimed write. A bounded pool keeps runtime reasonable while avoiding
	// repeatedly looking up a single hot ID.
	const maxSeedSessions = 2000
	seedCount := b.N
	if seedCount > maxSeedSessions {
		seedCount = maxSeedSessions
	}

	sessionIDs := make([]ID, 0, seedCount)
	expiry := time.Now().Add(time.Hour)
	for i := 0; i < seedCount; i++ {
		sess, err := store.NewSession(
			ctx, "bench", TypeMacaroonAdmin, expiry,
			"mailbox.test",
		)
		require.NoError(b, err)

		sessionIDs = append(sessionIDs, sess.ID)
	}

	b.ReportAllocs()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		_, err := store.GetSession(ctx, sessionIDs[i%seedCount])
		require.NoError(b, err)
	}
}
```

ellemouton (approved):
lgtm!

PR #1238, opened 2026-02-27
[sql-65] multi: improve `ListAccounts` performance for SQL database backends

ViktorT-11:
As noted in PR #1227, https://github.com/lightninglabs/lightning-terminal/pull/1227#pullrequestreview-3855419957, the `ListAccounts` performance in SQL was significantly slower than the `kvdb` version.

This PR improves that performance from roughly 14-20x slower, to around 2-2.5x slower based on my testing.

### DIFF
--- a/accounts/store_test.go
+++ b/accounts/store_test.go
@@ -141,6 +141,76 @@ func TestAccountStore(t *testing.T) {
 	require.ErrorIs(t, err, ErrAccNotFound)
 }
 
+// TestAccountStoreAccountsGrouping verifies that Accounts groups invoices and
+// payments by account ID and does not leak linked data across accounts.
+func TestAccountStoreAccountsGrouping(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	store := NewTestDB(t, clock.NewTestClock(time.Now()))
+
+	accountA, err := store.NewAccount(ctx, 1_000, time.Time{}, "group-a")
+	require.NoError(t, err)
+
+	accountB, err := store.NewAccount(ctx, 2_000, time.Time{}, "group-b")
+	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


db/sqlc/queries/accounts.sql
```diff
@@ -70,11 +70,21 @@ SELECT *
 FROM account_payments
 WHERE account_id = $1;
 
+-- name: ListAllAccountPayments :many
+SELECT *
+FROM account_payments
+ORDER BY account_id;
```
bitromortac: What's the motivation to sort by account id (here and later), is that needed?

accounts/store_sql.go
```diff
@@ -518,9 +528,36 @@ func (s *SQLStore) Accounts(ctx context.Context) ([]*OffChainBalanceAccount,
...
+		if err != nil {
+			return err
+		}
+
+		dbPayments, err := db.ListAllAccountPayments(ctx)
+		if err != nil {
+			return err
+		}
+
+		accountInvoices := make(map[int64][]sqlc.AccountInvoice)
```
bitromortac, line 541: nit: we could add a multi-account test to test the grouping logic

db/sqlc/queries/accounts.sql
ViktorT-11: The motivation was mainly that `ListAllAccounts` did so as well, and that it could be good to ensure that the result order from the queries are the same regardless of `sqlite` or `postgres` is used. But I agree that given how this is used, that's not needed, so I removed it :)

accounts/store_sql.go
ViktorT-11: Added a `TestAccountStoreAccountsGrouping` for this.

PR #1226, opened 2026-02-09
[sql-61] Add actions kvdb->SQL migration itest

ViktorT-11:
Based on https://github.com/lightninglabs/lightning-terminal/pull/1218

Implements the last part of step 6. of "Phase 3" in https://github.com/lightninglabs/lightning-terminal/issues/917.

This PR implements itest coverage for the kvdb->SQL migration of actions.

PR #1243, opened 2026-03-02
itest: add kvdb -> SQL migration test

ViktorT-11:
Implements step 6. of "Phase 3" in https://github.com/lightninglabs/lightning-terminal/issues/917.

Replaces #1207, #1211, #1217, #1218, #1226.

This PR implements a simplified kvdb -> SQL migration itest for `litd`.
The itest is only intended to cover the full migration flow of the kvdb -> SQL migration in `litd`, and does so by adding one object per database file to litd's bbolt database (one account, one session & one action). `litd` is thereafter restarted with an SQL database backend, which triggers the migration.

If reviewers agree that this is the responsibility of the kvdb -> SQL `itest` should have, I'll proceed to close the replaced PRs (#1207, #1211, #1217, #1218, #1226) once this PR has been merged.

### DIFF
--- a/itest/litd_migration_test.go
+++ b/itest/litd_migration_test.go
@@ -1,0 +1,397 @@
+//go:build itest
+
+package itest
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	terminal "github.com/lightninglabs/lightning-terminal"
+	"github.com/lightninglabs/lightning-terminal/accounts"
+	"github.com/lightninglabs/lightning-terminal/db/sqlc"
+	"github.com/lightninglabs/lightning-terminal/db/sqlcmig6"
+	"github.com/lightninglabs/lightning-terminal/firewalldb"
+	"github.com/lightninglabs/lightning-terminal/litrpc"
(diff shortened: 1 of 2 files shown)


itest/litd_migration_test.go
```diff
@@ -0,0 +1,397 @@
...
+	ctxt, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+
+	// Step 1: Start a node with a bbolt backend.
+	// We want to start from an explicit bbolt backend regardless of the
+	// command line flag used to run the itests.
+	migNode, err := net.NewNode(
+		t.t, "Migrator", nil, false, true,
+		fmt.Sprintf(
+			"--databasebackend=%s", terminal.DatabaseBackendBbolt,
```
ellemouton, line 62: iirc, we will remove the ability to startup with bbolt right?

ViktorT-11: Yes, correct. We'll also remove the bbolt implementation from the codebase, so at that state this test & the other migration tests will need to be removed.

bitromortac (approved):
LGTM :tada:

PR #1248, opened 2026-03-06
mod: bump to latest lnd pre v0.21 with channel updates

bitromortac:
This updates lnd to a master commit that supports channel updates in the channel events RPC (see https://github.com/lightningnetwork/lnd/pull/10543). It also bumps lndclient to a version that supports it. This is merged into a side branch because of the replaces, which I tried to get rid of, but couldn't (also future faraday related changes are planned to go into this side branch).

Todo:
- [ ] figure out if we can get rid of replaces


go.mod
```diff
@@ -252,3 +252,10 @@ replace github.com/golang-migrate/migrate/v4 => github.com/lightninglabs/migrate
 
 // The upstream nhooyr.io/websocket repository has been moved to github.com/coder/websocket.
 replace nhooyr.io/websocket => github.com/coder/websocket v1.8.7
+
+// The taprpc submodule in taproot-assets uses a replace directive that doesn't
+// propagate to consumers. We need to manually specify the version that
+// includes the portfoliopilotrpc package and aligns with the chancloser fix.
+replace github.com/lightninglabs/taproot-assets/taprpc => github.com/lightninglabs/taproot-assets/taprpc v1.0.12-0.20260223100136-c6e45b478517
+
+replace github.com/lightningnetwork/lnd => github.com/lightningnetwork/lnd v0.20.0-beta.rc4.0.20260223110936-dd65ba2b0106
```
ViktorT-11, line 261: As this is merged to a side branch, I think potentially that it makes just set these as the actual dependency above, instead of using `replace`. That way when we merge the side branch into the `master` in the future, there's no risk that we'll miss removing these replaces.

bitromortac: For taproot assets `c6e45b478517` it's currently not possible as itself has a replace which is inherited. For lnd I tried that, but `go mod tidy` doesn't like it (some dependency strangeness). I'd suggest to leave it that way until we have the releases, but would also be grateful if you'd know a solution :pray:

itest/litd_firewall_test.go
```diff
@@ -987,7 +987,7 @@ func testSessionLinking(net *NetworkHarness, t *harnessTest) {
 			Scope: &lnrpc.PolicyUpdateRequest_ChanPoint{
 				ChanPoint: chanPoint,
 			},
-			TimeLockDelta: 20,
+			TimeLockDelta: 40,
```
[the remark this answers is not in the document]
bitromortac: it's documented in the commit message, lnd requires a larger minimum delta

PR #1262, opened 2026-03-24
config: add options to modify obfuscation settings

bitromortac:
Based on https://github.com/lightninglabs/lightning-terminal/pull/1251

These changes are meant for easing the dev process. The no-privacy flag is hidden and the time variation can only be altered in dev mode.

### DIFF
--- a/cmd/litcli/autopilot.go
+++ b/cmd/litcli/autopilot.go
@@ -152,6 +152,16 @@ var addAutopilotSessionCmd = cli.Command{
 				"flags of the previous session. Example: " +
 				"\"ClearPubkeys|ClearAmounts\"",
 		},
+		cli.BoolFlag{
+			Name: "noprivacy",
+			Usage: `Switch off privacy mapping for this session.
+			This is not recommended except for debugging purposes,
+			as it can leak sensitive information to autopilot
+			servers. If this flag is set, all privacy flags will be
+			ignored and privacy mapping will be turned off for this
+			session.`,
+			Hidden: true,
+		},
 	},
 }
 
@@ -449,6 +459,7 @@ func initAutopilotSession(cli *cli.Context) error {
 			LinkedGroupId:          groupID,
 			PrivacyFlags:           privacyFlags,
 			PrivacyFlagsSet:        privacyFlagsSet,
+			NoPrivacyMapper:        cli.Bool("noprivacy"),
 		},
 	)
 	if err != nil {


firewall/privacy_mapper.go
```diff
@@ -60,21 +63,25 @@ var _ mid.RequestInterceptor = (*PrivacyMapper)(nil)
...
 
-// NewPrivacyMapper returns a new instance of PrivacyMapper. The randIntn
-// function is used to draw randomness for request field obfuscation.
-func NewPrivacyMapper(newDB firewalldb.PrivacyMapper,
-	randIntn func(int) (int, error),
+// NewPrivacyMapper returns a new instance of PrivacyMapper. The time variation
+// parameter specifies the variation space for timestamp obfuscation. If zero,
+// DefaultTimeVariation is used. The randIntn function is used to draw
+// randomness for request field obfuscation.
+func NewPrivacyMapper(timeVariation time.Duration,
```
ViktorT-11: nit: The `timeVariation time.Duration` arg shouldn't be the first arg of the constructor here.

firewall/privacy_mapper_test.go
```diff
@@ -1014,7 +1018,7 @@ func TestPrivacyMapper(t *testing.T) {
 		// seconds as there can be numerical inaccuracies with the
 		// nanosecond one.
 		timestamp := msg.ForwardingEvents[0].TimestampNs / 1e9
-		timestampInterval := uint64(timeVariation) / 1e9
+		timestampInterval := uint64(DefaultTimeVariation) / 1e9
```
ViktorT-11: nit: Use the value from the PrivacyMapper (`p`) to have test coverage that it's actually set?

config_prod.go
```diff
@@ -5,21 +5,29 @@ package terminal
...
-// tag. This struct is embedded in the main Config struct, and it adds no new
-// functionality in a production build.
-type DevConfig struct{}
+// DevConfig is a shell struct that allows us to build without the dev
+// tag. This struct is embedded in the main Config struct. In production
+// builds, it only holds default values for privacy configuration.
+type DevConfig struct {
+	// PrivacyTimestampVariation holds the default timestamp variation.
+	// This field is not configurable in production builds.
+	PrivacyTimestampVariation time.Duration
```
ViktorT-11: Now that we're setting actual `prod` values to this struct, I think this isn't really a `DevConfig` any longer.

I think we should consider having a new `Config` struct (or another suitable name), which has this `PrivacyTimestampVariation` field, and also embeds the `DevConfig`. Then the functions which exports the `DevConfig` today should export the outer `Config` struct. That could be done in a separate refactoring commit.

bitromortac: I found another solution keeping the dev config empty for prod builds, without having to add another struct. Instead we add it to the prod config without making it configurable and keep the dev config we had. We then override any prod setting with the validate function (a pattern that was used before, without that it doesn't pick up the changed setting).

PR #1266, opened 2026-03-26
[sql-66] multi: Use final `sqldb/v2` dependency

ViktorT-11:
This PR updates the `sql-migration-base` to use the final released version of the `sqldb/v2` dependency.

### DIFF
--- a/config_dev.go
+++ b/config_dev.go
@@ -9,7 +9,7 @@ import (
 
 	"github.com/lightninglabs/lightning-terminal/accounts"
 	"github.com/lightninglabs/lightning-terminal/db"
-	"github.com/lightninglabs/lightning-terminal/db/migstreams"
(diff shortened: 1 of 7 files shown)


ViktorT-11:
Ah sorry I should probably have mentioned in the PR description that the dependency is referencing this branch in lnd:
https://github.com/lightningnetwork/lnd/tree/sqldb-v2

Maybe we could ask them to tag that in lnd, so it becomes clearer?

bitromortac:
> Ah sorry I should probably have mentioned in the PR description that the dependency is referencing this branch in lnd: https://github.com/lightningnetwork/lnd/tree/sqldb-v2
> 
> Maybe we could ask them to tag that in lnd, so it becomes clearer?

I was aware, but not sure if there will be another merge commit when that branch is merged into master. But a tag sounds good as well.

PR #1269, opened 2026-03-28
subservers: fail litd startup on critical sub-server error

### LINKED ISSUE #1181: Fail `litd` startup on `tapd` startup error when `taproot-assets-mode=enable`

Antisys:
## Summary

Closes #1181.

- Mark `tapd` as a critical sub-server when `taproot-assets-mode` is not `disable`, so litd fails to start if tapd fails to start (preventing a broken state where lnd calls into a non-running tapd)
- Use a two-pass approach in `StartIntegratedServers`/`ConnectRemoteSubServers`: attempt all sub-servers first, then return error if any critical one failed — avoids non-determinism from Go's random map iteration order
- Fix pre-existing build failure in `perms/mock.go` and `perms/mock_dev.go` where Config struct fields were referenced without the required RPC build tags

## Changes

| File | What |
|------|------|
| `subservers/manager.go` | Add `critical` set + `SetCritical()`, return errors from `StartIntegratedServers`/`ConnectRemoteSubServers` for critical servers |
| `terminal.go` | Mark TAP as critical when enabled, handle new error returns |
| `subservers/manager_test.go` | 5 unit tests (critical failure, non-critical failure, no critical, disabled critical, status verification) |
| `perms/mock.go`, `perms/mock_dev.go` | Remove field initializers for empty (untagged) Config structs |

## Test plan

- [x] `go build $(go list ./... | grep -v /itest)` — clean
- [x] `go test $(go list ./... | grep -v /itest)` — all pass
- [x] `go test -race ./subservers/...` — no races
- [ ] Integration test: start litd with `taproot-assets-mode=integrated` and a tapd config that will fail → litd should exit with error
- [ ] Integration test: start litd with `taproot-assets-mode=disable` and a broken tapd config → litd should start normally

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/perms/mock.go
+++ b/perms/mock.go
@@ -5,8 +5,6 @@ package perms
 import (
 	"net"
 
-	"github.com/lightningnetwork/lnd/autopilot"
-	"github.com/lightningnetwork/lnd/chainreg"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/autopilotrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/chainrpc"
@@ -19,9 +17,7 @@ import (
 	"github.com/lightningnetwork/lnd/lnrpc/walletrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/watchtowerrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/wtclientrpc"
-	"github.com/lightningnetwork/lnd/lntest/mock"
 	"github.com/lightningnetwork/lnd/routing"
-	"github.com/lightningnetwork/lnd/sweep"
 )
 
 // mockConfig implements lnrpc.SubServerConfigDispatcher. It provides the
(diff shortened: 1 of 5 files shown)


ViktorT-11:
Hi @Antisys,

Thank you for your PR contribution, unfortunately though this issue already has a linked PR which is actively being 
worked on: https://github.com/lightninglabs/lightning-terminal/pull/1183. This PR is therefore unfortunately redundant and I must sadly therefore recommend you to close this PR, and contribute in the already opened PR instead. Ways to contribute to that PR can either be adding a code review, or expressing your need for this functionality. Let me know if there's anything else I can clarify further!

Antisys:
Thanks for the heads up @ViktorT-11 — makes sense to consolidate. I'll follow #1183 and contribute there if I have anything useful to add. Closing this.

ViktorT-11:
Thanks a lot @Antisys!

PR #1271, opened 2026-03-31
firewall: add interceptor for non-LND calls

bitromortac:
Adds a gRPC unary interceptor to LNC sessions that applies privacy mapping to non-LND sub-daemon calls (e.g. Faraday). LND URIs are skipped since they're already handled by LND's middleware chain.

I added a diagram to see how the two flows work for lnd/subdaemon calls:
https://github.com/bitromortac/lightning-terminal/blob/eee3d033fff0f808ec5aa3119721fb5d3fd95c9c/docs/privacy-mapping-flows.md

If the design is ok, I can also use a streaming interceptor.

### DIFF
--- a/firewall/privacy_mapper.go
+++ b/firewall/privacy_mapper.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	// privacyMapperName is the name of the RequestLogger interceptor.
+	// privacyMapperName is the name of the PrivacyMapper interceptor.
 	privacyMapperName = "lit-privacy-mapper"
 
 	// amountVariation and DefaultTimeVariation are used to set the
@@ -120,7 +120,7 @@ func (p *PrivacyMapper) Intercept(ctx context.Context,
 		return nil, err
 	}
 
-	log.Tracef("PrivacyMapper: Intercepting %v", ri)
+	log.TraceS(ctx, "PrivacyMapper: Intercepting", "info", ri)
 
 	switch r := req.InterceptType.(type) {
 	case *lnrpc.RPCMiddlewareRequest_StreamAuth:
--- a/firewall/privacy_mapper.go
+++ b/firewall/privacy_mapper.go
@@ -415,6 +415,45 @@ func (p *PrivacyMapper) UnaryInterceptor() grpc.UnaryServerInterceptor {
 	}
 }
 
+// StreamServerInterceptor returns a gRPC stream server interceptor that
+// enforces fail-close privacy for non-LND services. LND streaming RPCs are
+// passed through (they use LND's own middleware chain). Non-LND streams are
+// passed through for sessions without privacy mapping, and blocked for
+// privacy-enabled sessions because the privacy mapper does not yet support
+// stream-level request/response rewriting.
+func (p *PrivacyMapper) StreamServerInterceptor() grpc.StreamServerInterceptor {
+	return func(srv any, ss grpc.ServerStream, info *grpc.StreamServerInfo,
+		handler grpc.StreamHandler) error {
+
+		ctx := ss.Context()
+
+		log.TraceS(ctx, "LNC privacy stream interceptor called",
+			"method", info.FullMethod)
+
+		sess, passThrough, err := p.resolveLNCSession(
+			ctx, info.FullMethod,
+		)
+		if err != nil {
+			return err
+		}
+		if passThrough {
+			return handler(srv, ss)
+		}
+
+		// Fail-close: non-LND streaming RPCs are not yet supported
+		// by the privacy mapper. Block them to prevent cleartext
+		// leaks to LNC sessions with privacy enabled.
+		log.WarnS(ctx, "Blocking non-LND stream: privacy mapping "+
+			"not supported for streams", nil,
+			"method", info.FullMethod,
+			"session_id", sess.ID)
+
+		return status.Errorf(codes.PermissionDenied,
+			"streaming RPCs for non-LND services are not "+
+				"supported with privacy mapping")
+	}
+}
+
 // isLndURI returns true if the URI belongs to LND (uses permissions manager
 // to identify all LND services: lnrpc, routerrpc, signrpc, walletrpc, etc).
 func (p *PrivacyMapper) isLndURI(uri string) (bool, error) {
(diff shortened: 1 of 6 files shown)


firewall/privacy_mapper.go
```diff
@@ -193,6 +206,172 @@ func (p *PrivacyMapper) Intercept(ctx context.Context,
...
+
+		log.Tracef("LNC privacy interceptor called for: %s",
+			info.FullMethod)
+
+		// Fail-close: block if permissions manager isn't ready.
+		isLnd, err := p.isLndURI(info.FullMethod)
+		if err != nil {
+			log.Errorf("Permissions manager not initialized, "+
+				"blocking %s", info.FullMethod)
+			return nil, status.Errorf(codes.Unavailable,
```
ViktorT-11, line 232: Nit: There's a lot of cases here when there's no new line before the return statement. I think you should add those.

ViktorT-11 (commented):
Generally this looks good, thanks :fire:! Worth looking into some of the review-bot comments though :).

Also: In the future, we'll likely want to make the `accounts` system be able to act on requests to other sub-systems than just `lnd` (specifically `taproot-assets` seems likely). I think a potential solution of how we would achieve that would be to create some a generic account interceptor layer that works on normal gRPC requests/responses, and not just `lnrpc.RPCMiddlewareRequest`s. I think it's likely out of scope for this PR, but when we do so, it impact what's being added by this PR.
Just thought I'd add a comment regarding that, in case that in some way changes how you'd like to design this PR :). 

Additionally, this PR should add a release note IMO.

bitromortac:
The unit-race test failure is independent of this PR, fixed in https://github.com/lightninglabs/lightning-terminal/pull/1281.

firewall/privacy_mapper.go
```diff
@@ -16,7 +16,12 @@ import (
 	"github.com/lightninglabs/lightning-terminal/firewalldb"
```
ellemouton, line 16: pls add more descriptive commit messages 🙏

firewall/privacy_mapper.go
```diff
@@ -67,21 +72,29 @@ type PrivacyMapper struct {
 	randIntn      func(int) (int, error)
 	sessionDB     firewalldb.SessionDB
 	timeVariation time.Duration
+	permsMgr      PermissionsManager
+}
+
+// PermissionsManager defines the interface for checking if a URI belongs to a
+// specific sub-server.
+type PermissionsManager interface {
+	IsSubServerURI(subServer string, uri string) bool
```
ellemouton, line 81: godoc

terminal.go
```diff
@@ -1071,6 +1073,34 @@ func (g *LightningTerminal) startInternalSubServers(ctx context.Context,
 
 	log.Infof("Starting LiT session server")
 
+	// Privacy mapper for LNC sessions. Written once during startup
+	// via Store; read by gRPC handler goroutines via Load.
+	var lncPrivacyMapper atomic.Pointer[firewall.PrivacyMapper]
```
ellemouton, line 1078: perhaps explain the circular dependency that makes this needed

terminal.go
```diff
@@ -1071,6 +1073,34 @@ func (g *LightningTerminal) startInternalSubServers(ctx context.Context,
...
+
+			pm := lncPrivacyMapper.Load()
+			if pm == nil {
+				return nil, grpcstatus.Errorf(
+					codes.Unavailable,
+					"privacy mapper not ready",
+				)
+			}
+
+			return pm.UnaryInterceptor()(
```
ellemouton, line 1097: nit: fits on one line

firewall/privacy_mapper.go
```diff
@@ -199,6 +212,182 @@ func (p *PrivacyMapper) Intercept(ctx context.Context,
...
+			log.Errorf("Permissions manager not initialized, "+
+				"blocking %s", info.FullMethod)
+
+			return nil, status.Errorf(codes.Unavailable,
+				"permissions manager not ready, please retry")
+		}
+
+		// LND URIs are exempted (they use LND's middleware chain).
+		if isLnd {
+			log.Tracef("Skipping LNC privacy for LND URI: %s",
```
ellemouton, line 238: there is a `ctx` available, so rather use structured logging pls :)

ellemouton: it makes it easier to do trace logging

ellemouton (approved):
concept ACK

litbot-9000:
@bitromortac, remember to re-request review from reviewers when ready

firewall/privacy_mapper.go
bitromortac: done, sorry about that!

firewall/privacy_mapper.go
bitromortac: updated to structured logging in that file :tada:

firewall/privacy_mapper.go
```diff
@@ -193,22 +238,259 @@ func (p *PrivacyMapper) Intercept(ctx context.Context,
...
+		// leaks to LNC sessions with privacy enabled.
+		log.WarnS(ctx, "Blocking non-LND stream: privacy mapping "+
+			"not supported for streams", nil,
+			"method", info.FullMethod,
+			"session_id", sess.ID)
+
+		return status.Errorf(codes.PermissionDenied,
+			"streaming RPCs for non-LND services are not "+
+				"supported with privacy mapping")
+	}
```
ViktorT-11, line 454: Just wanted to check that you've made sure that this can't cause any issues with `tapd` (and potentially `loop`) if we merge this as is?

firewall/privacy_mapper.go
```diff
@@ -193,22 +238,259 @@ func (p *PrivacyMapper) Intercept(ctx context.Context,
...
+}
+
+// isLndURI returns true if the URI belongs to LND (uses permissions manager
+// to identify all LND services: lnrpc, routerrpc, signrpc, walletrpc, etc).
+func (p *PrivacyMapper) isLndURI(uri string) (bool, error) {
+	if p.permsMgr == nil {
+		return false, fmt.Errorf("permissions manager not initialized")
+	}
+	return p.permsMgr.IsSubServerURI(subservers.LND, uri), nil
```
ViktorT-11, line 463: nit: new line before return.

ViktorT-11 (approved):
Nice, uTACK LGTM :fire:!

firewall/privacy_mapper.go
bitromortac: I tested the following:
- loop out via TW session
- local unary frcli call
- local streaming loop call (monitor)
- payment using accounts via a session

So I'd consider this well tested and will merge.

firewall/privacy_mapper.go
bitromortac: will skip due to CI run

PR #1272, opened 2026-04-01
[sql-67] multi: tombstone bbolt databases on successful SQL migration

ViktorT-11:
Based on #1266

This PR tombstones/deprecates the different bbolt db files once a `litd` node has successfully migrated from kvdb -> SQL.

The files are intentionally kept with the same name, but are updated by adding an extra bucket in the different files, so that startup of `litd` will fail if the user decides to change back their `databasebackend` to `bbolt` after the migration.

The motivation for this is to ensure that the user only has one source of data, and so that the user can't continue to operate with a `bbolt` database and add new data to it after the migration, as that data cannot be migrated over to the users new SQL database if we did allow the user to continue to run with the `bbolt` backend.

The tombstone is intentionally designed so that the `accounts.db` will cause a reading failure if the user decides to downgrade their `litd` node to a version prior to the tombstone logic existing. If the user runs a `litd` version which includes the tombstone logic, the startup will fail when reading the `session.db` & `rules.db` files as well, in the scenario that for example the user deletes their `accounts.db` file, but keeps their tombstoned `session.db` & `rules.db` files.

Additionally, the logic is designed so that in case the user deletes their SQL database file after migration, they can still re-run a new kvdb -> SQL migration with their old tombstoned `bbolt` database files.

### DIFF
--- a/accounts/kvdb_deprecation.go
+++ b/accounts/kvdb_deprecation.go
@@ -1,0 +1,138 @@
+package accounts
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/lightningnetwork/lnd/kvdb"
+	"go.etcd.io/bbolt"
+)
+
+const (
+	// dbFilePermission is the default permission the account database file
+	// is created with.
+	dbFilePermission = 0600
+)
+
+var (
+	// ErrKVDBDeprecated signals that the legacy kvdb database was already
+	// migrated to SQL and should not be opened again.
+	ErrKVDBDeprecated = errors.New("kvdb database has been migrated to " +
+		"SQL and can no longer be used")
+
+	// deprecatedBucketKey marks a kvdb database as deprecated in a way that
+	// older kvdb account readers will fail once they iterate the account
+	// bucket.
+	deprecatedBucketKey = []byte("kvdb-sql-migrated")
+
+	deprecatedReasonKey = []byte("reason")
+)
+
+// DeprecateKVDB marks the accounts kvdb file as deprecated after a successful
+// SQL migration. It does so by creating a dedicated tombstone bucket named
+// `kvdb-sql-migrated` inside the top-level accounts bucket and storing the
+// deprecation reason in that bucket. New code checks for that tombstone before
+// opening the accounts kvdb file, while older account readers will fail once
+// they iterate the accounts bucket and encounter the unexpected tombstone
+// bucket.
+func DeprecateKVDB(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return nil
+	}
+
+	db, err := bbolt.Open(path, dbFilePermission, &bbolt.Options{
+		Timeout: DefaultAccountDBTimeout,
+	})
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	return db.Update(func(tx *bbolt.Tx) error {
+		accountBucket, err := tx.CreateBucketIfNotExists(
+			accountBucketName,
+		)
+		if err != nil {
+			return err
+		}
+
+		deprecatedBucket, err := accountBucket.CreateBucketIfNotExists(
+			deprecatedBucketKey,
+		)
+		if err != nil {
+			return err
+		}
+
+		return deprecatedBucket.Put(
+			deprecatedReasonKey, []byte(ErrKVDBDeprecated.Error()),
+		)
+	})
+}
+
+// CheckKVDBDeprecated returns a clear error if the accounts kvdb file was
+// marked as deprecated.
+func CheckKVDBDeprecated(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return nil
+	}
+
+	db, err := bbolt.Open(path, dbFilePermission, &bbolt.Options{
+		Timeout: DefaultAccountDBTimeout,
+	})
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	var deprecated bool
+	err = db.View(func(tx *bbolt.Tx) error {
+		accountBucket := tx.Bucket(accountBucketName)
+		if accountBucket == nil {
+			return nil
+		}
+
+		deprecated = accountBucket.Bucket(deprecatedBucketKey) != nil
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	if deprecated {
+		return fmt.Errorf(
+			"%w: %s", ErrKVDBDeprecated, filepath.Base(path),
+		)
+	}
+
+	return nil
+}
+
+// checkKVDBDeprecated returns a clear error if the opened accounts kvdb was
+// marked as deprecated.
+func checkKVDBDeprecated(db kvdb.Backend, path string) error {
+	var deprecated bool
+	err := kvdb.View(db, func(tx kvdb.RTx) error {
+		accountBucket := tx.ReadBucket(accountBucketName)
+		if accountBucket == nil {
+			return nil
+		}
+
+		deprecated = accountBucket.NestedReadBucket(
+			deprecatedBucketKey,
+		) != nil
+		return nil
+	}, func() {})
+	if err != nil {
+		return err
+	}
+
+	if deprecated {
+		return fmt.Errorf(
+			"%w: %s", ErrKVDBDeprecated, filepath.Base(path),
+		)
+	}
+
+	return nil
+}
--- a/accounts/kvdb_deprecation_test.go
+++ b/accounts/kvdb_deprecation_test.go
@@ -1,0 +1,52 @@
+package accounts
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/lightningnetwork/lnd/clock"
+	"github.com/stretchr/testify/require"
+)
+
+// TestKVDBDeprecation verifies that a deprecated accounts kvdb file refuses to
+// reopen.
+func TestKVDBDeprecation(t *testing.T) {
+	t.Parallel()
+
+	dbDir := t.TempDir()
+	clk := clock.NewDefaultClock()
+
+	store, err := NewBoltStore(dbDir, DBFilename, clk)
+	require.NoError(t, err)
+	require.NoError(t, store.Close())
+
+	dbPath := filepath.Join(dbDir, DBFilename)
+
+	err = DeprecateKVDB(dbPath)
+	require.NoError(t, err)
+
+	_, err = NewBoltStore(dbDir, DBFilename, clk)
+	require.Error(t, err)
+	require.True(t, errors.Is(err, ErrKVDBDeprecated))
+	require.Contains(t, err.Error(), ErrKVDBDeprecated.Error())
+
+	store, err = NewBoltStoreForMigration(dbDir, DBFilename, clk)
+	require.NoError(t, err)
+	require.NoError(t, store.Close())
+}
+
+// TestDeprecateKVDBMissingFile verifies that deprecating a missing accounts
+// kvdb file is a no-op.
+func TestDeprecateKVDBMissingFile(t *testing.T) {
+	t.Parallel()
+
+	dbPath := filepath.Join(t.TempDir(), DBFilename)
+
+	err := DeprecateKVDB(dbPath)
+	require.NoError(t, err)
+
+	_, err = os.Stat(dbPath)
+	require.ErrorIs(t, err, os.ErrNotExist)
+}
--- a/accounts/store_kvdb.go
+++ b/accounts/store_kvdb.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/btcsuite/btcwallet/walletdb"
@@ -67,6 +68,21 @@ type BoltStore struct {
 // NewBoltStore creates a BoltStore instance and the corresponding bucket in the
 // bolt DB if it does not exist yet.
 func NewBoltStore(dir, fileName string, clock clock.Clock) (*BoltStore, error) {
+	return newBoltStore(dir, fileName, clock, false)
+}
+
+// NewBoltStoreForMigration opens the accounts kvdb store even if it was
+// already marked as deprecated. This is only intended for rerunning the kvdb
+// to SQL migration after the SQL database was removed or downgraded.
+func NewBoltStoreForMigration(dir, fileName string,
+	clock clock.Clock) (*BoltStore, error) {
+
+	return newBoltStore(dir, fileName, clock, true)
+}
+
+func newBoltStore(dir, fileName string, clock clock.Clock,
+	allowDeprecated bool) (*BoltStore, error) {
+
 	// Ensure that the path to the directory exists.
 	if _, err := os.Stat(dir); os.IsNotExist(err) {
 		if err := os.MkdirAll(dir, dbPathPermission); err != nil {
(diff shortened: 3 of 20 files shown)


litbot-9000:
@ellemouton: review reminder
@bitromortac: review reminder
@viktort-11, remember to re-request review from reviewers when ready

accounts/kvdb_deprecation.go
```diff
@@ -0,0 +1,138 @@
...
+
+		return deprecatedBucket.Put(
+			deprecatedReasonKey, []byte(ErrKVDBDeprecated.Error()),
+		)
+	})
+}
+
+// CheckKVDBDeprecated returns a clear error if the accounts kvdb file was
+// marked as deprecated.
+func CheckKVDBDeprecated(path string) error {
```
bitromortac: I think this would need to be removed as well.

db/migsets/post_migration_callbacks_dev.go
```diff
@@ -54,6 +54,29 @@ func MakePostStepCallbacksMig6(ctx context.Context,
...
+		// committed successfully. Returning an error here would only
+		// cause the programmatic migration to rerun on the next startup
+		// and reprocess data that is already present in SQL.
+		//
+		// We still want the failure to be highly visible because the
+		// legacy bbolt files were not tombstoned and may therefore
+		// still be opened unexpectedly.
+		err = deprecateKVDBStores(filepath.Dir(macPath))
+		if err != nil {
+			log.Criticalf("CRITICAL: kvdb -> SQL migration "+
```
bitromortac: this will lead to a shutdown using a critical log call

accounts/kvdb_deprecation_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	dbPath := filepath.Join(dbDir, DBFilename)
+
+	err = DeprecateKVDB(dbPath)
+	require.NoError(t, err)
+
+	_, err = NewBoltStore(dbDir, DBFilename, clk)
+	require.Error(t, err)
+	require.True(t, errors.Is(err, ErrKVDBDeprecated))
+	require.Contains(t, err.Error(), ErrKVDBDeprecated.Error())
```
bitromortac: can use `require.ErrorIs`, same in other spots

itest/litd_migration_test.go
```diff
@@ -395,3 +585,148 @@ func openMigrationSQLStore(t *harnessTest,
...
+// expected error string for the LiT sub-server.
+func assertLitStatusError(t *harnessTest, node *HarnessNode,
+	expectedErr string) {
+
+	t.t.Helper()
+
+	deadline := time.Now().Add(defaultTimeout)
+	var lastErr error
+
+	for time.Now().Before(deadline) {
```
bitromortac: we could replace this with `wait.NoError`

bitromortac (commented):
Looks great conceptually :pray:, only had some minor comments.

itest/litd_migration_test.go
```diff
@@ -107,29 +132,113 @@ func testKvdbSQLMigration(ctx context.Context, net *NetworkHarness,
...
+	)
+
+	// Step 9: Delete the SQL database and verify that starting with the
+	// selected SQL backend reruns the kvdb -> SQL migration successfully.
+	rerunSQLMigrationAndAssert(
+		t, net, migNode, newStepCtx, newAdminCtx, beforeMigration,
+		migrationRefs,
+	)
+
+	// Step 10: If the backward compatibility binary exists, verify that an
```
bitromortac, line 193: I think this is still based on an older branch containing backward compatibility tests, right? If so we could add a compile target in the make file to add an older binary.

accounts/kvdb_deprecation.go
ViktorT-11: Thanks, addressed!

db/migsets/post_migration_callbacks_dev.go
ViktorT-11: Ah good catch!

itest/litd_migration_test.go
ViktorT-11: After discussing this offline, we agreed on including this in this branch to show that the backwards compatibility works, but when this is merged to `master` we'll remove this part of the test to not have to reintroduce backwards compatibility test support to `master`.

db/migsets/post_migration_callbacks_dev.go
```diff
@@ -193,3 +216,32 @@ func kvdbToSqlMigrationCallback(ctx context.Context,
 
 	return nil
 }
+
+// deprecateKVDBStores marks the old kvdb stores as deprecated after the SQL
+// migration committed successfully. We do this after the SQL transaction is
+// committed so a failed SQL migration cannot strand the user with an unusable
+// kvdb backend.
+func deprecateKVDBStores(dbDir string) error {
+	err := accounts.DeprecateKVDB(
```
bitromortac: We could use `errors.Join` in the end to try all deprecations?

itest/litd_migration_test.go
```diff
@@ -395,3 +586,144 @@ func openMigrationSQLStore(t *harnessTest,
...
+	err := wait.NoError(func() error {
+		ctxt, cancel := context.WithTimeout(
+			context.Background(), 2*time.Second,
+		)
+		defer cancel()
+
+		rawConn, err := connectLitRPC(
+			ctxt, node.Cfg.LitAddr(), node.Cfg.LitTLSCertPath, "",
+		)
+		defer rawConn.Close()
```
bitromortac: nit: better call this after the error check, to avoid nil dereferences

db/migsets/post_migration_callbacks_dev.go
```diff
@@ -54,6 +54,29 @@ func MakePostStepCallbacksMig6(ctx context.Context,
...
+		//
+		// At this point the kvdb -> SQL data migration is already
+		// committed successfully. Returning an error here would only
+		// cause the programmatic migration to rerun on the next startup
+		// and reprocess data that is already present in SQL.
+		//
+		// We still want the failure to be highly visible because the
+		// legacy bbolt files were not tombstoned and may therefore
+		// still be opened unexpectedly.
+		err = deprecateKVDBStores(filepath.Dir(macPath))
```
bitromortac, line 73: I think we should add a changelog entry to document that this adds a downgrade protection for the sql migration.

bitromortac (approved):
LGTM :tada: (also saw that the backward compatibility test ran).

ViktorT-11:
Thanks for the review @bitromortac! I'll address your latest feedback after the next review round (from @ellemouton) :tada:

itest/litd_node.go
```diff
@@ -792,15 +792,21 @@ func (hn *HarnessNode) Start(litdBinary string,
...
-		if err != nil {
-			litdError <- fmt.Errorf("%v\n%v\n", err, errb.String())
-		}
-
 		// Signal any onlookers that this process has exited.
 		close(hn.processExit)
 
 		// Make sure log file is closed and renamed if necessary.
 		finalizeLogfile()
+
```
ellemouton, line 807: the code should document why the error check is so much later on else it looks like a bug & also at a glance just looks like an unchecked error

session/kvdb_store.go
```diff
@@ -111,11 +127,24 @@ func NewDB(dir, fileName string, clock clock.Clock,
 	if err != nil {
 		return nil, err
 	}
+	err = checkKVDBDeprecated(db, path)
```
ellemouton: new line above

db/migsets/post_migration_callbacks_dev.go
ellemouton: agreed. atm it could land in a state where first one succeeds, second fails and then third isnt attempted.

accounts/kvdb_deprecation.go
```diff
@@ -0,0 +1,100 @@
+package accounts
```
ellemouton: i dont think it is clear why the tomb-stoning is done differently for accounts vs session/firewall packages. i think this could be made clearer (if there is a reason). 

Ideally there is one `Deprecation` helper/set of helpers and all three packages just call that (can just pass in a top-level bucket?). or is the idea that each file's keys are kept in that package? if so , can let the helper take in a "tombstone key prefix" perhaps.  

See how LND does tombstoneing (invoices package for example).

accounts/kvdb_deprecation.go
```diff
@@ -0,0 +1,100 @@
...
+	// migrated to SQL and should not be opened again.
+	ErrKVDBDeprecated = errors.New("kvdb database has been migrated to " +
+		"SQL and can no longer be used")
+
+	// deprecatedBucketKey marks a kvdb database as deprecated in a way that
+	// older kvdb account readers will fail once they iterate the account
+	// bucket.
+	deprecatedBucketKey = []byte("kvdb-sql-migrated")
+
+	deprecatedReasonKey = []byte("reason")
```
ellemouton: Deprecation reason is always set to "deprecated" error? dont think we need this. isnt read anywhere anyways

accounts/store_kvdb.go
```diff
@@ -67,13 +68,29 @@ type BoltStore struct {
...
+	allowDeprecated bool) (*BoltStore, error) {
+
 	// Ensure that the path to the directory exists.
 	if _, err := os.Stat(dir); os.IsNotExist(err) {
 		if err := os.MkdirAll(dir, dbPathPermission); err != nil {
 			return nil, err
 		}
 	}
 
+	dbPath := filepath.Join(dir, fileName)
```
ellemouton: move to where it is used?

db/migsets/post_migration_callbacks_dev.go
ellemouton: i think we shoudl too but surely only when it is in the prod code?

session/kvdb_deprecation.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// DeprecateKVDB marks the session kvdb file as deprecated after a successful
+// SQL migration. It does so by writing a sentinel database version value of
+// `^uint32(0)` into the metadata bucket. Session startup reads that metadata
+// version before opening the store for normal use, so the sentinel version
+// cleanly blocks the kvdb backend while still leaving the file available for
+// explicit migration reruns.
+func DeprecateKVDB(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
ellemouton: while this is a neat solution, im not sure we should do this since: 1) this is not the same way it is done for the accounts db and 2) this doesnt have the same "backwards" behaviour that the other method has. For the other method (adding a new tombstone key), a user can decide to spin up an older binary which would then not know about that key & so would succeed. whereas with this method, an older node _would_ see the change and wont startup due to the "revert". This might actually be nicer behaviour but it doesnt match our usual pattern. So just want to be sure that this was intentional and not just an agent deciding to use `syncVersions` cause it saw it lying around. It does override the version field which isnt exactly the same as just marking it as deprecated. 

thoughts?

ellemouton: ie, wondering if we cant have one sharee "tomestone marker/checker" impl used by all three?

session/kvdb_store.go
ellemouton: I thikn the order used in `accounts` is better: 1) first check if deprecated _and then_ 2) init the db. initing a db first and then checking deprecation seems odd

itest/litd_migration_test.go
```diff
@@ -395,3 +586,144 @@ func openMigrationSQLStore(t *harnessTest,
...
+
+		if err != nil {
```
ellemouton, line 643: you cant check the error after the defer. gotta first check the error else you can nil-pointer panic on the defer if there is an error

itest/litd_node.go
```diff
@@ -792,15 +792,21 @@ func (hn *HarnessNode) Start(litdBinary string,
...
+		if err != nil {
+			select {
+			case litdError <- fmt.Errorf(
+				"%v\n%v\n", err, errb.String(),
+			):
+
```
ellemouton: worried about this dropping errors for the broader itests... cant we use a buffered channel instead?

db/migsets/post_migration_callbacks_dev.go
```diff
@@ -54,6 +54,29 @@ func MakePostStepCallbacksMig6(ctx context.Context,
...
+		if err != nil {
+			log.Errorf("CRITICAL: kvdb -> SQL migration "+
```
ellemouton, line 75: is there a way for us to retry later on? how does lnd do it?

ellemouton (commented):
exciting!

left some comments

db/migsets/post_migration_callbacks_dev.go
ViktorT-11: Thanks for the suggestion, addressed!

itest/litd_node.go
ViktorT-11: Updated with a doc, also buffered the channel.

accounts/kvdb_deprecation.go
ViktorT-11: Good idea! Updated to add a new `tombstone` package, which is used across the packages instead.  

Also updated to use the same approach `lnd` uses for tombstoneing

accounts/kvdb_deprecation.go
ViktorT-11: https://github.com/lightninglabs/lightning-terminal/pull/1272#discussion_r3201772680

accounts/store_kvdb.go
ViktorT-11: Missed adding this specific change, but will do on the next review round.

session/kvdb_deprecation.go
ViktorT-11: https://github.com/lightninglabs/lightning-terminal/pull/1272#discussion_r3201772680

session/kvdb_store.go
ViktorT-11: I updated so that all uses the order you suggested. The reason I did it the way I did before, is that this new ordering requires that we open the kvdb DB twice per startup per package, instead of just once. But agree that the code is cleaner this way.

itest/litd_node.go
ViktorT-11: Changed to use a buffered channel!

db/migsets/post_migration_callbacks_dev.go
ViktorT-11: I did not update this.

IMO, the way you'd need to implement it is that you'd need to add a flag in the `sql` db that signals if the `kvdb` has been tombstoned or not, which is initialized as `false`. We'd then retry the tombstoneing until the flag is set to `true`, which we'd do after a successful tombstoneing, or if the `kdvb` files do not exist.

That would requiring adding a separate table for this in the `sql` db, which I think might been seen as overengineering for an issue that potentially never occur for any users, and if it does, you could argue that the outcome not that bad.

If you think it makes sense to add that tracking table in `sql`, I'm open to doing so though, unless you have a better idea of how to solve it!

The current implementation in `lnd` is even less complete than the above:

```
// Set the invoice bucket tombstone to indicate
// that the migration has been completed.
//
// TODO(ziggie): The tombstone is currently
// set inside the SQL transaction callback,
// which is fragile: if the SQL transaction
// is retried (e.g. on a serialization
// error), the KV tombstone is written before
// the SQL commit is confirmed. Move this to
// run after ApplyAllMigrations returns so
// the tombstone is only set once the
// migration is durably committed.
d.logger.Debugf("Setting invoice bucket " +
"tombstone")

//nolint:ll
return dbs.ChanStateDB.SetInvoiceBucketTombstone()

```

ViktorT-11:
Thank you so much for the detailed review @ellemouton! Addressed your feedback :fire:

I'll fix the linter issue & https://github.com/lightninglabs/lightning-terminal/pull/1272#discussion_r3201786606 on the next review round.

ellemouton (approved):
lgtm!

ViktorT-11:
Thanks for the reviews :tada:!

PR #1274, opened 2026-04-04
account: add option to rename account

### LINKED ISSUE #1110: account: add option to rename account
`litcli accounts update` can change only balance and expiration, but also label change is useful for good account management. Also help text should be changed to clarify --label is not for changing but only for identifying the account.

Current:
```
NAME:
   litcli accounts update - Update an existing off-chain account.

USAGE:
   litcli accounts update [command options] [id | label] new_balance [new_expiration_date] [--save_to=]

DESCRIPTION:
   Updates an existing off-chain account and sets either a new balance or new expiration date or both.

OPTIONS:
   --id value                   The ID of the account to update.
   --label value                (optional) The unique label of the account.
   --new_balance value          The new balance of the account; -1 means do not update the balance. (default: -1)
   --new_expiration_date value  The new expiration date of the account expressed in seconds since the unix epoch; -1 means do not update the expiration date; 0 means it does not expire. (default: -1)
   
```

Cyberguru1:
This PR introduces support for renaming existing off-chain accounts in lightning-terminal. The functionality is exposed via the litcli accounts update command and backed by a new RPC implementation.

This enables users to modify account names without recreating accounts, improving usability and account management.

Closes #1110.

### DIFF
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -242,11 +242,11 @@ type Store interface {
 	// Accounts retrieves all accounts from the store and un-marshals them.
 	Accounts(ctx context.Context) ([]*OffChainBalanceAccount, error)
 
-	// UpdateAccountBalanceAndExpiry updates the balance and/or expiry of an
-	// account.
-	UpdateAccountBalanceAndExpiry(ctx context.Context, id AccountID,
+	// UpdateAccount updates the balance, expiry and/or label of an account.
+	UpdateAccount(ctx context.Context, id AccountID,
 		newBalance fn.Option[int64],
-		newExpiry fn.Option[time.Time]) error
+		newExpiry fn.Option[time.Time],
+		newLabel fn.Option[string]) error
 
 	// AddAccountInvoice adds an invoice hash to an account.
 	AddAccountInvoice(ctx context.Context, id AccountID,
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -442,3 +442,18 @@ func WithErrIfUnknown() UpsertPaymentOption {
 		o.errIfUnknown = true
 	}
 }
+
+// First, ensure that if a label is set, it can't be
+// mistaken for a hex encoded account ID.
+func checkLabel(label string) error {
+	if len(label) == hex.EncodedLen(AccountIDLen) {
+		_, err := hex.DecodeString(label)
+		if err == nil {
+			return fmt.Errorf("the label '%s'"+
+				" is not allowed as it "+
+				"can be mistaken for an account ID", label)
+		}
+	}
+
+	return nil
+}
--- a/accounts/interface.go
+++ b/accounts/interface.go
@@ -449,8 +449,8 @@ func checkLabel(label string) error {
 	if len(label) == hex.EncodedLen(AccountIDLen) {
 		_, err := hex.DecodeString(label)
 		if err == nil {
-			return fmt.Errorf("the label '%s'"+
-				" is not allowed as it "+
+			return fmt.Errorf("the label '%s' "+
+				"is not allowed as it "+
 				"can be mistaken for an account ID", label)
 		}
 	}
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -301,7 +301,8 @@ func (s *InterceptorService) NewAccount(ctx context.Context,
 // if it exists.
 func (s *InterceptorService) UpdateAccount(ctx context.Context,
 	accountID AccountID, accountBalance btcutil.Amount,
-	expirationDate int64) (*OffChainBalanceAccount, error) {
+	expirationDate int64, newLabel string) (*OffChainBalanceAccount,
+	error) {
 
 	s.Lock()
 	defer s.Unlock()
@@ -334,9 +335,17 @@ func (s *InterceptorService) UpdateAccount(ctx context.Context,
 		balance = fn.Some(int64(accountBalance) * 1000)
 	}
 
+	// If a new label was provided, wrap it in an option. An empty
+	// string is treated as "no update requested" because protobuf
+	// cannot distinguish an absent field from the zero value "".
+	var label fn.Option[string]
+	if newLabel != "" {
+		label = fn.Some(newLabel)
+	}
+
 	// Create the actual account in the macaroon account store.
-	err := s.store.UpdateAccountBalanceAndExpiry(
-		ctx, accountID, balance, expiry,
+	err := s.store.UpdateAccount(
+		ctx, accountID, balance, expiry, label,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("unable to update account: %w", err)
--- a/accounts/store_kvdb.go
+++ b/accounts/store_kvdb.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/binary"
-	"encoding/hex"
 	"fmt"
 	"math"
 	"os"
@@ -123,16 +122,12 @@ func (s *BoltStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
 
 	// If a label is set, it must be unique, as we use it to identify the
 	// account in some of the RPCs. It also can't be mistaken for a hex
-	// encoded account ID to avoid confusion and make it easier for the CLI
-	// to distinguish between the two.
-	if len(label) > 0 {
-		if _, err := hex.DecodeString(label); err == nil &&
-			len(label) == hex.EncodedLen(AccountIDLen) {
+	// encoded account ID.
+	if err := checkLabel(label); err != nil {
+		return nil, err
+	}
 
-			return nil, fmt.Errorf("the label '%s' is not allowed "+
-				"as it can be mistaken for an account ID",
-				label)
-		}
+	if len(label) > 0 {
 
 		accounts, err := s.Accounts(ctx)
 		if err != nil {
--- a/accounts/store_kvdb.go
+++ b/accounts/store_kvdb.go
@@ -128,7 +128,6 @@ func (s *BoltStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
 	}
 
 	if len(label) > 0 {
-
 		accounts, err := s.Accounts(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("error checking label "+
@@ -181,13 +180,54 @@ func (s *BoltStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
 	return account, nil
 }
 
-// UpdateAccountBalanceAndExpiry updates the balance and/or expiry of an
-// account.
+// UpdateAccount updates the balance and/or expiration date of an existing
+// off-chain account.
 //
 // NOTE: This is part of the Store interface.
-func (s *BoltStore) UpdateAccountBalanceAndExpiry(_ context.Context,
+func (s *BoltStore) UpdateAccount(ctx context.Context,
 	id AccountID, newBalance fn.Option[int64],
-	newExpiry fn.Option[time.Time]) error {
+	newExpiry fn.Option[time.Time],
+	newLabel fn.Option[string]) error {
+
+	// If a new label is set, we need to check for its uniqueness. Since the
+	// check requires a read transaction, we must do it before we start the
+	// update transaction.
+	var labelErr error
+	newLabel.WhenSome(func(label string) {
+		if err := checkLabel(label); err != nil {
+			labelErr = err
+			return
+		}
+
+		if len(label) > 0 {
+			accounts, err := s.Accounts(ctx)
+			if err != nil {
+				labelErr = fmt.Errorf("error checking label "+
+					"uniqueness: %w", err)
+				return
+			}
+
+			for _, other := range accounts {
+				if other.ID == id {
+					continue
+				}
+
+				if other.Label == label {
+					labelErr = fmt.Errorf(
+						"an account with the "+
+							"label '%s' "+
+							"already "+
+							"exists: %w",
+						label,
+						ErrLabelAlreadyExists)
+					return
+				}
+			}
+		}
+	})
+	if labelErr != nil {
+		return labelErr
+	}
 
 	update := func(account *OffChainBalanceAccount) error {
 		newBalance.WhenSome(func(balance int64) {
@@ -195,6 +235,9 @@ func (s *BoltStore) UpdateAccountBalanceAndExpiry(_ context.Context,
 		})
 		newExpiry.WhenSome(func(expiry time.Time) {
 			account.ExpirationDate = expiry
+		})
+		newLabel.WhenSome(func(label string) {
+			account.Label = label
 		})
 
 		return nil
--- a/accounts/store_sql.go
+++ b/accounts/store_sql.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/rand"
 	"database/sql"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"time"
@@ -102,17 +101,13 @@ func (s *SQLStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
 	error) {
 
 	// Ensure that if a label is set, it can't be mistaken for a hex
-	// encoded account ID to avoid confusion and make it easier for the CLI
-	// to distinguish between the two.
+	// encoded account ID.
+	if err := checkLabel(label); err != nil {
+		return nil, err
+	}
+
 	var labelVal sql.NullString
 	if len(label) > 0 {
-		if _, err := hex.DecodeString(label); err == nil &&
-			len(label) == hex.EncodedLen(AccountIDLen) {
-
-			return nil, fmt.Errorf("the label '%s' is not allowed "+
-				"as it can be mistaken for an account ID",
-				label)
-		}
 
 		labelVal = sql.NullString{
 			String: label,
--- a/accounts/store_sql.go
+++ b/accounts/store_sql.go
@@ -51,6 +51,7 @@ type SQLQueries interface {
 	// UpdateAccountAliasForTests is a query intended only for testing
 	// purposes, to change the account alias.
 	UpdateAccountAliasForTests(ctx context.Context, arg sqlc.UpdateAccountAliasForTestsParams) (int64, error)
+	UpdateAccountLabel(ctx context.Context, arg sqlc.UpdateAccountLabelParams) (int64, error)
 	UpsertAccountPayment(ctx context.Context, arg sqlc.UpsertAccountPaymentParams) error
 	GetAccountInvoice(ctx context.Context, arg sqlc.GetAccountInvoiceParams) (sqlc.AccountInvoice, error)
 }
@@ -108,7 +109,6 @@ func (s *SQLStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
 
 	var labelVal sql.NullString
 	if len(label) > 0 {
-
 		labelVal = sql.NullString{
 			String: label,
 			Valid:  true,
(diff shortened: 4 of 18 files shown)


Cyberguru1:
Hello @ViktorT-11,

Could you please help rerun the failing workflow? I’m not sure why the unit-race test is failing, as it passes locally. Also, I would appreciate it if you could review the PR.

Thank you.

ViktorT-11:
Hi @Cyberguru1, thank you very much for the contribution 🙏. I started a re-run of the failed CI job. @bitromortac will do an initial review of this PR.

Cyberguru1:
> Hi @Cyberguru1, thank you very much for the contribution 🙏. I started a re-run of the failed CI job. @bitromortac will do an initial review of this PR.

Hi, thank you for the update 🙏

I appreciate you triggering the re-run. Glad to see the checks have passed.

@bitromortac looking forward to your review.

litrpc/lit-accounts.proto
```diff
@@ -159,9 +159,12 @@ message UpdateAccountRequest {
 
     /*
     The label of the account to update. If an account has no label, then the ID
-    must be used instead.
+    must be used instead. This field is only used to identify the account.
```
bitromortac, line 162: This change should come with changes from `make rpc/protos`.

db/sqlc/queries/accounts.sql
```diff
@@ -21,6 +21,12 @@ SET last_updated = $1
 WHERE id = $2
 RETURNING id;
 
+-- name: UpdateAccountLabel :one
```
bitromortac, line 24: In this we'd expect the changes included from `make sqlc`.

.agent/rules/development-guidelines.md
```diff
@@ -0,0 +1,88 @@
+---
+trigger: always_on
```
bitromortac: Please don't check in any agentic rule sets, see also later commits.

accounts/store_sql.go
```diff
@@ -379,6 +380,51 @@ func (s *SQLStore) UpdateAccountBalanceAndExpiry(ctx context.Context,
...
+				if err == nil && isID {
+					err = fmt.Errorf(
+						"the label '%s' is not "+
+							"allowed as it can be "+
+							"mistaken for an "+
+							"account ID",
+						label)
+
+					return
+				}
```
bitromortac: Would be great if you could extract the logic in `NewAccount` that checks this and reuse it here, something like `func checkLabel(label string) err` and in the kv equivalents.

accounts/store_test.go
```diff
@@ -206,36 +207,62 @@ func TestAccountUpdateMethods(t *testing.T) {
...
+
+		// Try to rename to an existing label.
+		_, err = store.NewAccount(ctx, 0, time.Time{}, "existing")
+		require.NoError(t, err)
+
+		err = store.UpdateAccount(
+			ctx, acct.ID, fn.None[int64](), fn.None[time.Time](),
+			fn.Some("existing"),
+		)
+		require.ErrorIs(t, err, ErrLabelAlreadyExists)
```
bitromortac, line 279: Maybe test cases for `None` label, an empty `Some` label and a hex string could be interesting to add here.

proto/lit-accounts.proto
```diff
@@ -159,9 +159,12 @@ message UpdateAccountRequest {
 
     /*
     The label of the account to update. If an account has no label, then the ID
-    must be used instead.
+    must be used instead. This field is only used to identify the account.
```
bitromortac, line 162: Please add this change and the corresponding `make rpc/protos` changes in the previous proto commit.

bitromortac (commented):
Thank you @Cyberguru1 for your work! Logic-wise the changes look good, I have some feedback concerning the commit structure for PRs in this project (especially with respect to auto-generated files).

accounts/store_sql.go
Cyberguru1: Good suggestion, I’ve refactored the logic by extracting the label validation into a reusable checkLabel(label string) error function and reused it here as well as in the KV equivalents.

Cyberguru1:
@bitromortac  
Thank you for your feedback and for taking the time to review.

Regarding the commit structure, I’ve updated it so that each source change is grouped together with its generated output, ensuring that no commit leaves the repository in an inconsistent state.

accounts/store_sql.go
```diff
@@ -381,24 +376,14 @@ func (s *SQLStore) UpdateAccount(ctx context.Context,
 		}
 
 		newLabel.WhenSome(func(label string) {
+			if err := checkLabel(label); err != nil {
+				return
+			}
```
bitromortac, line 383: Here you silently ignore the error, but you'd need to set the `err` from the outer scope to catch it in the later error check. So `if err = checkLabel(label); err != nil {` would fix that (no colon).

Also, with this commit you are rewriting changes that you introduced in `accounts: implement account rename support in stores`. In general we try to avoid this to not needing to review the code twice (reviewers go through the review commit by commit). The best pattern would be to do a pure refactor commit for the existing code where you extract `checkLabel` first. Then you can use the check function in `accounts: implement account rename support in stores`. We aim for the "ideal" commit structure, where it's ok to rewrite commits. I know it's a bit technical, but wanted to make you aware of the patterns used here.

bitromortac (commented):
> Regarding the commit structure, I’ve updated it so that each source change is grouped together with its generated output, ensuring that no commit leaves the repository in an inconsistent state.

Thank you for your changes and the added tests. We'd just need to slightly change the order of commits a bit (see comment), but otherwise looks quite close.

accounts/store_sql.go
Cyberguru1: Thank you for the feedback and the guidance, I appreciate it. Nice catch on the error handling.

I’ve addressed it by assigning the result of `checkLabel` to the outer `err` variable so it is properly captured. I also restructured the commits by first introducing a pure refactor commit to extract `checkLabel`, and then applying it in the account rename implementation.

accounts/store_sql.go
```diff
@@ -379,6 +375,41 @@ func (s *SQLStore) UpdateAccountBalanceAndExpiry(ctx context.Context,
 			return err
 		}
 
+		newLabel.WhenSome(func(label string) {
+			if err = checkLabel(label); err != nil {
+				return
+			}
+
+			// First, ensure that if a label is set, it can't be
+			// mistaken for a hex encoded account ID.
```
bitromortac: That comment should be above `checkLabel`

Cyberguru1:
Hello @bitromortac,

When you have some time, could you please review this PR? I look forward to your feedback.

Thank you.

accounts/interface.go
```diff
@@ -442,3 +442,19 @@ func WithErrIfUnknown() UpsertPaymentOption {
 		o.errIfUnknown = true
 	}
 }
+
+// checkLabel ensures that if a label is set, it can't be mistaken for a hex
+// encoded account ID to avoid confusion and make it easier for the CLI to
+// distinguish between the two.
+func checkLabel(label string) error {
```
bitromortac, line 448: nit: usually when doing such a refactor we also swap out the existing code with the new code in the same commit as this is a unit of operation (a clean refactor), so that the new function is already used somewhere, which gives more context for a reviewer (I marked this as a `nit` that you can address in another round, but nits usually are non-blocking).

accounts/store_kvdb.go
```diff
@@ -202,6 +197,46 @@ func (s *BoltStore) UpdateAccountBalanceAndExpiry(_ context.Context,
...
+		var labelErr error
+		newLabel.WhenSome(func(label string) {
+			if err := checkLabel(label); err != nil {
+				labelErr = err
+				return
+			}
+
+			// If a label is set, it must be unique.
+			if len(label) > 0 {
+				accounts, err := s.Accounts(ctx)
```
bitromortac: This opens a view inside an update transaction, which is considered unsafe for bbolt (could lead to deadlocks), so we'd need to do the uniqueness check as an initial step, even though that creates a tiny concurrency gap (this is acceptable since we'll remove kvdb soon, I think).

```go
func (s *BoltStore) UpdateAccount(ctx context.Context,
	id AccountID, newBalance fn.Option[int64],
	newExpiry fn.Option[time.Time],
	newLabel fn.Option[string]) error {

	// If a new label is set, we need to check for its uniqueness. Since the
	// check requires a read transaction, we must do it before we start the
	// update transaction.
	var labelErr error
	newLabel.WhenSome(func(label string) {
		if err := checkLabel(label); err != nil {
			labelErr = err
			return
		}

		if len(label) > 0 {
			accounts, err := s.Accounts(ctx)
			if err != nil {
				labelErr = fmt.Errorf("error checking label "+
					"uniqueness: %w", err)
				return
			}

			for _, other := range accounts {
				if other.ID == id {
					continue
				}

				if other.Label == label {
					labelErr = fmt.Errorf("an account with the "+
						"label '%s' already exists: %w",
						label, ErrLabelAlreadyExists)
					return
				}
			}
		}
	})
	if labelErr != nil {
		return labelErr
	}

	update := func(account *OffChainBalanceAccount) error {
		newBalance.WhenSome(func(balance int64) {
			account.CurrentBalance = balance
		})
		newExpiry.WhenSome(func(expiry time.Time) {
			account.ExpirationDate = expiry
		})
		newLabel.WhenSome(func(label string) {
			account.Label = label
		})

		return nil
	}

	return s.updateAccount(id, update)
}
```

Cyberguru1:
Hello @bitromortac,

Regarding the failing test case, it looks like there is an issue with the test causing it to fail intermittently. I ran the tests locally for this PR and they pass, so it might just need a re-run to pass the CI checks.

I would be happy to take a closer look into it in another PR.

accounts/store_test.go
```diff
@@ -206,36 +213,118 @@ func TestAccountUpdateMethods(t *testing.T) {
...
+
```
bitromortac, line 280: we need another test case here:
```go
		// Test updating an account with its existing label doesn't fail
		err = store.UpdateAccount(
			ctx, acct.ID, fn.None[int64](), fn.None[time.Time](),
			fn.Some(newLabel),
		)
		require.NoError(t, err)
```

This will show you that 
`go test -v -run TestAccountUpdateMethods ./accounts` and `go test -v -tags="test_db_sqlite" -run TestAccountUpdateMethods ./accounts` will behave differently. We should make the behavior consistent (act as a no-op).

bitromortac (commented):
I focussed a bit more on the logic and still found some issues.

accounts/store_sql.go
Cyberguru1: I’ve moved the comment above checkLabel to better reflect its purpose.

accounts/interface.go
Cyberguru1: Thanks for pointing that out. I’ve updated the refactor so that the new checkLabel function is introduced together with its usage in the same commit.

accounts/store_kvdb.go
Cyberguru1: Good catch on the deadlock state, and thank you for the code reference. I’ve moved the label validation and uniqueness check outside the update transaction so it runs as an initial step.

accounts/store_test.go
Cyberguru1: Thanks for pointing that out. I’ve added the additional test case to ensure that updating an account with its existing label behaves as a no-op and does not fail.

After running the two test cases, I observed the difference in behavior with one failing, which I have addressed.

bitromortac: I think the test got lost, or am I missing it somewhere?

bitromortac (commented):
I think it's just missing to add the test, otherwise LGTM.

accounts/store_test.go
Cyberguru1: I must have missed it while amending the commit, I’ve now added it.

accounts/service.go
```diff
@@ -334,9 +335,17 @@ func (s *InterceptorService) UpdateAccount(ctx context.Context,
 		balance = fn.Some(int64(accountBalance) * 1000)
 	}
 
+	// If a new label was provided, wrap it in an option. An empty
+	// string is treated as "no update requested" because protobuf
+	// cannot distinguish an absent field from the zero value "".
+	var label fn.Option[string]
+	if newLabel != "" {
+		label = fn.Some(newLabel)
+	}
```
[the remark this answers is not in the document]
bitromortac: I was aware of that, it would either need an additional bool to signal unsetting the label, or the optional approach. We can add that later if it's needed.

accounts/store_kvdb.go
```diff
@@ -186,13 +180,54 @@ func (s *BoltStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
...
-func (s *BoltStore) UpdateAccountBalanceAndExpiry(_ context.Context,
+func (s *BoltStore) UpdateAccount(ctx context.Context,
 	id AccountID, newBalance fn.Option[int64],
-	newExpiry fn.Option[time.Time]) error {
+	newExpiry fn.Option[time.Time],
+	newLabel fn.Option[string]) error {
+
+	// If a new label is set, we need to check for its uniqueness. Since the
+	// check requires a read transaction, we must do it before we start the
+	// update transaction.
```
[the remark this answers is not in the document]
bitromortac: The small toctou race was accepted because the kvdb implementation will be phased out soon and to avoid having to make more changes to the account querying. Would be open to change that

accounts/store_kvdb.go
```diff
@@ -186,13 +180,54 @@ func (s *BoltStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
...
+					labelErr = fmt.Errorf(
+						"an account with the "+
+							"label '%s' "+
+							"already "+
+							"exists: %w",
+						label,
+						ErrLabelAlreadyExists)
+					return
+				}
+			}
```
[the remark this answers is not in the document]
bitromortac: See above.

bitromortac (approved):

accounts/service.go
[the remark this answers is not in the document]
Cyberguru1: Thanks for the clarification. That makes sense. I would be happy to address this in a follow up PR by introducing either an optional field or an additional flag if we decide to support explicit label clearing.

Cyberguru1:
Hello @ViktorT-11

Since @bitromortac has given a review and approved the PR, I would appreciate it if you could take another look when you have some time.

Looking forward to your review, thank you.

accounts/interface.go
```diff
@@ -442,3 +442,18 @@ func WithErrIfUnknown() UpsertPaymentOption {
...
+
+// First, ensure that if a label is set, it can't be
+// mistaken for a hex encoded account ID.
+func checkLabel(label string) error {
+	if len(label) == hex.EncodedLen(AccountIDLen) {
+		_, err := hex.DecodeString(label)
+		if err == nil {
+			return fmt.Errorf("the label '%s'"+
+				" is not allowed as it "+
+				"can be mistaken for an account ID", label)
```
ViktorT-11: nit: The space should be moved to the line before, to ensure that the new line doesn't start with a space char. 
```suggestion
			return fmt.Errorf("the label '%s' "+
				"is not allowed as it "+
				"can be mistaken for an account ID", label)
```

accounts/store_kvdb.go
```diff
@@ -123,17 +122,12 @@ func (s *BoltStore) NewAccount(ctx context.Context, balance lnwire.MilliSatoshi,
...
-			return nil, fmt.Errorf("the label '%s' is not allowed "+
-				"as it can be mistaken for an account ID",
-				label)
-		}
+	// encoded account ID.
+	if err := checkLabel(label); err != nil {
+		return nil, err
+	}
 
+	if len(label) > 0 {
```
ViktorT-11, line 130: Nit: This isn’t just a refactor, since it the logic is changed slightly with the introduction of the `checkLabel` function.

For example, if the passed label has a length of `3` but `hex.DecodeString(label)` returns an error, the added `checkLabel` function will return `nil` instead of an error, whereas the previous code would have left the error in the `err` variable.

That said, the previous code also handled this incorrectly, since it left the error unchecked and likely should have returned an error in that case instead of treating the `label` value as valid.

Because the previous behavior was already incorrect, I’m fine merging this as-is for now if you’d prefer not to address it here. We can open an issue to track it and fix it in a follow-up PR.

Let me know if you prefer to fix this in this PR or in a follow-up.

ViktorT-11 (approved):
Thanks a lot for the contribution @Cyberguru1! 

Apologies for the delay in the review! I'm leaving two non-blocking suggestions below.
Let me know if you prefer to fix them now, or if you want to fix especially the second suggestion in a follow-up PR as I'm ok with merging this as is.

accounts/interface.go
Cyberguru1: Noted on the formatting, I’ll address this in a new PR.

accounts/store_kvdb.go
Cyberguru1: Thanks for the detailed explanation, that makes sense. I’ll handle this in a new PR.

Cyberguru1:
> Thanks a lot for the contribution @Cyberguru1!
> 
> Apologies for the delay in the review! I'm leaving two non-blocking suggestions below. Let me know if you prefer to fix them now, or if you want to fix especially the second suggestion in a follow-up PR as I'm ok with merging this as is.

Thanks @ViktorT-11, i appreciate the review.

I’d prefer to address both suggestions in a follow-up PR so they can be handled properly there. I’ll open a new PR to cover the updates.

cmd/litcli/accounts.go
```diff
@@ -146,18 +146,25 @@ var updateAccountCommand = cli.Command{
 	Name:      "update",
 	ShortName: "u",
 	Usage:     "Update an existing off-chain account.",
-	ArgsUsage: "[id | label] new_balance [new_expiration_date] " +
-		"[--save_to=]",
+	ArgsUsage: "[id | label] [new_balance] [new_expiration_date] " +
+		"[--new_label=]",
```
ViktorT-11: This argument usage description needs to be updated, as placing the new_label option as described will not actually update the label. I.e. the description currently suggests the ordering of the arguments should be as follows:
```
litcli accounts update d2c41adea29b3eae 10 10 --new_label="test"
```

In practise that won't have update the label though, and the argument order should instead be as below to update the label:
```
litcli accounts update --new_label="test" d2c41adea29b3eae 10 10
```

Therefore, please update the text to the below:

```suggestion
	ArgsUsage: "[id | label] new_balance new_expiration_date",
```

That will produce a USAGE text as follows:
```
litcli accounts update command [command options] [id | label] new_balance new_expiration_date
```

That's in line how we'd represent similar examples in `lncli` (we don't any have similar examples in `litcli` yet.

cmd/litcli/accounts.go
```diff
@@ -146,18 +146,25 @@ var updateAccountCommand = cli.Command{
...
 	Description: "Updates an existing off-chain account and sets " +
-		"either a new balance or new expiration date or both.",
+		"a new balance, new expiration date or a new label.",
```
ViktorT-11:
```suggestion
		"a new balance, new expiration date and optionally a new label.",
```

cmd/litcli/accounts.go
ViktorT-11, line 188: Alternatively, instead of passing the `--new_label` flag to the main `litcli accounts update` command to update the label, you could potentially create a new sub-command that specifically only updates the label.

There's a main benefit to doing so: 
Updating the account balance with `litcli accounts update` is not safe if the accounts system is used at the same time as described in:
https://github.com/lightninglabs/lightning-terminal/issues/648

Currently, the main `litcli accounts update` requires that you also set a new account balance and expiration when executing it. I.e. just executing the command like this:
```
litcli accounts update --new_label="test" d2c41adea29b3eae
```
will indeed update the account's label, but it will also set the account balance & expiration to `0`.

Therefore, it could be motivated to add a separate sub-command to allow you to just update the label.

Worth noting, we are going to deprecate updating the balance through the `litcli accounts update` command: https://github.com/lightninglabs/lightning-terminal/issues/1276

I think we should potentially also make it optional to update the expiry with the command as well.

Due to the above, I think it is fine to keep the `--new_label` flag in the `litcli accounts update` command if you prefer it that way. 
Alternatively, if you like the sub-command idea but would prefer to do that in a follow-up PR, and just want to update the command argument description for now, that'd be ok for me as well! 

Therefore, please update the PR with either the command argument description update, or with the sub-command idea.

ViktorT-11 (changes requested):
Sorry @Cyberguru1, when testing this more before merging the PR, I noticed that you need to update the argument description for the command before we can merge this.

Please see the below feedback, once that's been addressed, I'm ready to merge this.

cmd/litcli/accounts.go
Cyberguru1: Thanks for the detailed feedback @ViktorT-11, this makes sense.

I went ahead with the `--new_label` approach for now to keep the change minimal within this PR. I agree with the concern around litcli accounts update not being safe in its current form, especially with how it implicitly affects balance and expiry.

My plan is to pick up  #1276 once this is merged, and address the balance update deprecation there.

ViktorT-11 (approved):
Thanks for addressing the feedback @Cyberguru1! tACK LGTM :fire:! Going ahead with merging this now.

> My plan is to pick up https://github.com/lightninglabs/lightning-terminal/issues/1276 once this is merged, and address the balance update deprecation there.

That sounds great, thanks a lot!

PR #1295, opened 2026-05-01
docs: note that FreeBSD users need gmake to build

### LINKED ISSUE #859: [docs] Add gmake references to compilations instructions to support FreeBSD
This is just a short note for FreeBSD users they want to compile this project. I compiled the project successfully [on FreeBSD v13.3] according to the instruction [Compile from source code](https://github.com/lightninglabs/lightning-terminal/blob/master/docs/compile.md#compile-from-source-code). The only difference is to use `gmake install` instead of `make install`.

0xfandom:
Closes #859.

The `Makefile` uses GNU Make features, which on FreeBSD requires the `gmake` port rather than the base system `make`. Add a short note to `docs/compile.md` so FreeBSD users know to install `gmake` and substitute it for `make` in the build commands.

Docs-only change — can add a release-notes entry or apply `no-changelog` if preferred.

### DIFF
--- a/docs/compile.md
+++ b/docs/compile.md
@@ -26,6 +26,11 @@ need to download those binaries from the
 [pool](https://github.com/lightninglabs/pool/releases), and
 [faraday](https://github.com/lightninglabs/faraday/releases) repos manually.
 
+**Note for FreeBSD users:** the `Makefile` requires GNU Make. Install it with
+`pkg install gmake` and substitute `gmake` for `make` in the commands above
+(for example, `gmake install`). The same applies to the other `make` targets
+mentioned later in this document.
+
 ## Building a docker image
 
 There are two flavors of Dockerfiles available:
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ our [troubleshooting guide](./docs/troubleshooting.md).
 If you’d prefer to compile from source code please
 follow [these instructions](./docs/compile.md).
 
+**Note for FreeBSD users:** the `Makefile` requires GNU Make. Install it with
+`pkg install gmake` and use `gmake` in place of `make` (for example,
+`gmake install`).
+
 ## Component Compatibility
 
 This section describes version and mode compatibility across Lightning Terminal


docs/release-notes/release-notes-0.17.0.md
```diff
@@ -31,6 +31,11 @@
 
 ### Technical and Architectural Updates
 
+* [Document `gmake` requirement for FreeBSD
```
ViktorT-11: I would remove this release note & commit, as this isn't an actual change to any of `litd`'s functionality.

docs/compile.md
```diff
@@ -26,6 +26,11 @@ need to download those binaries from the
 [pool](https://github.com/lightninglabs/pool/releases), and
 [faraday](https://github.com/lightninglabs/faraday/releases) repos manually.
 
+**Note for FreeBSD users:** the `Makefile` requires GNU Make. Install it with
```
ViktorT-11, line 29: You could potentially add a similar note in the the following section in the `README.md` file, as many user will likely read that and go through that document before reading this file:

## LND

Note that LiT requires `lnd` to be built with **all of its subservers** and
requires running at least v0.11.0. Download the
latest [official release binary](https://github.com/lightningnetwork/lnd/releases/latest)
or build `lnd` from source by following
the [installation instructions](https://github.com/lightningnetwork/lnd/blob/master/docs/INSTALL.md).
If you choose to build `lnd` from source, use the following command to enable
all the relevant subservers:

```shell
⛰  make install tags="signrpc walletrpc chainrpc invoicesrpc"
```

ViktorT-11 (commented):
Thanks @0xfandom! This generally looks good, just a few small suggestions.

docs/release-notes/release-notes-0.17.0.md
0xfandom: Good call — dropped the release-notes commit, the docs change alone is enough.

docs/compile.md
0xfandom: Done — added an equivalent FreeBSD note under the README's *Build from source* section.

ViktorT-11 (approved):
Thanks!

bitromortac (approved):

PR #1305, opened 2026-05-12
[sql-70] Enable SQL backends in production & trigger the kvdb to sql migration

ViktorT-11:
Replaces steps 8, 9 & 11 of "Phase 3" in https://github.com/lightninglabs/lightning-terminal/issues/917.

This PR enables SQL database backends for non `dev` builds, and also adds to kvdb to SQL migration to the `prod` environment.  

This PR also sets the default database backend to Sqlite, and hence will automatically trigger the kvdb to sql migration for users which have not explicitly set their database backend to `bbolt`. Note that once the migration has been completed, the user's `bbolt` database will be tombstoned.


config_dev.go
```diff
@@ -2,232 +2,20 @@
...
-				"BoltDB: %v", err)
-		}
-
-		stores.firewall = firewalldb.NewDB(firewallBoltDB)
-		stores.closeFns["bbolt-firewalldb"] = firewallBoltDB.Close
-	}
-
-	return stores, nil
+// Validate checks that all the values set in our DevConfig are valid.
+func (c *DevConfig) Validate() error {
```
ViktorT-11, line 19: I think there's still value in keeping the structure for the `dev` config, which is why I didn't remove this & the `config_prod.go` file. Let me know if you disagree.

ViktorT-11:
Rebased on `master` without any changes. 

I intentionally cancelled the CI workflow as there's been no changes since the last run when CI succeeded.
I'll ensure to run the CI run after the next review round.

ViktorT-11:
Will address the review bots comment after the next review-round!

db/migsets/sql_migrations_dev.go
```diff
@@ -64,17 +64,14 @@ func MakeMigrationSets(ctx context.Context,
...
-			// after the dev migration number for the uint key in
-			// the map has been applied. If no entry exists for a
-			// given uint, then no callback will be executed for
-			// that migration number. This is useful for adding a
-			// code migration step as a callback to be run
-			// after a specific migration of a given number has been
-			// applied.
+			// Any programmatic migrations added to this map will be
+			// executed when the dev migration number for the uint
+			// key ia applied. If no entry exists for a
```
bitromortac: nit: `ia` -> `is`

db/sqlc/migrations_dev/000001_removed_dev_migration.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Comment to ensure the file created and picked up in the migration stream.
```
bitromortac: Not 100% sure I understand why this is kept.

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -24,6 +24,31 @@
 
 ### Functional Changes/Additions
 
+* [Support for SQL database
+  backends](https://github.com/lightninglabs/lightning-terminal/pull/1305):
+  Litd now supports SQLite and postgres database backends. The old BBolt
```
bitromortac, line 29: nice :tada::tada:

bitromortac (commented):
Nice work :pray:, first pass only

db/sqlc/migrations/000006_code_migration_kvdb_to_sql.down.sql
ViktorT-11, line 1: Note to self, this file should be removed.

db/sqlc/migrations_dev/000001_removed_dev_migration.up.sql
ViktorT-11: As discussed offline, this was mainly kept to ensure no old test DB would get errors. But we decided that that's less of an issue than the confusion of keeping this file.

db/migsets/programmatic_migrations_dev.go
```diff
@@ -22,11 +22,9 @@ import (
 	"github.com/lightningnetwork/lnd/sqldb/v2"
 )
 
-// MakePostStepCallbacksMig6 turns the post migration checks into a map of post
-// step callbacks that can be used with the migrate package. The keys of the map
-// are the migration versions, and the values are the callbacks that will be
-// executed after the migration with the corresponding version is applied.
-func MakePostStepCallbacksMig6(ctx context.Context,
+// Mig6ProgrammaticMigration generates and returns the ProgrammaticMigrEntry
```
bitromortac: nit: could write out programmatic migration entry

db/migrations.go
```diff
@@ -12,7 +14,15 @@ const (
...
-	LatestMigrationVersion = 5
+	//
+	// TODO: On the next update, when the migration version is updated
+	// be beyond version 6, make sure that the default
+	// `SkipMigrationDbBackup` value for Sqlite is set to false.
+	LatestMigrationVersion = 6
+
+	// KVDBtoSQLMigVersion is the version of the migration that migrates the
+	// kvdb to the sql database.
+	KVDBtoSQLMigVersion = 6
```
bitromortac, line 28: We should probably state that this is a constant and should not be changed?

config.go
```diff
@@ -299,7 +299,10 @@ func (c *Config) NewStores(ctx context.Context,
 	case DatabaseBackendSqlite:
 		// Before we initialize the SQLite store, we'll make sure that
 		// the directory where we will store the database file exists.
-		err := makeDirectories(networkDir)
+		// Note that the c.Sqlite.DatabaseFileName may already have
+		// overwritten in loadAndValidateConfig, so that the correct
```
bitromortac, line 308: I don't fully understand this, could you clarify?

db/migsets/programmatic_migrations.go
```diff
@@ -6,7 +6,6 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"path/filepath"
```
bitromortac, line 11: So wouldn't we ideally like to have all dbs in the same folder?

bitromortac (commented):
Code-wise lgtm! I will just need to do some manual tests :pray:

config.go
ViktorT-11: So what I'm referring to here is that `loadAndValidateConfig` has overwrites the `c.Sqlite.DatabaseFileName` if is not set (i.e. when the default value is assigned), and adds the correct network folder to the path.
If that wasn't done, we wouldn't be able to set the sqlite database folder path into the correct `network` folder, as we can't vary the default value. 
 
See the code below from `loadAndValidateConfig`:
```
	if cfg.Sqlite != nil &&
		cfg.Sqlite.DatabaseFileName == defaultSqliteDatabasePath {

		cfg.Sqlite.DatabaseFileName = filepath.Join(
			litDir, cfg.Network, defaultSqliteDatabaseFileName,
		)
	}
```

Let me know if you think this needs to be clarified somehow in the docs.

db/migsets/programmatic_migrations.go
ViktorT-11: Yes we definitely should... But this is a bug that we have for current users which means that their `accounts` db may be in another location than the rest of their dbs. I think the reason we haven't had any bug reports of this yet is that the `accounts` subsystem has been sparsely used so far,  and usage of that that in combination of specifying a custom macaroon path is probably quite rare.

When we have migrated over to `sql` though, this issue will not be an issue as there'll only be one database file.

ViktorT-11:
Thanks a lot for the review @bitromortac :tada:! I've address your feedback and answered your questions above.

db/migsets/programmatic_migrations.go
bitromortac: Ah, right, thank you!

bitromortac (approved):
LGTM :tada:

ellemouton (approved):
Let's goooooooooo 🚀

config.go
```diff
@@ -244,6 +283,166 @@ type Config struct {
...
+			return stores, fmt.Errorf("error creating firewall "+
+				"BoltDB: %v", err)
+		}
+
+		stores.firewall = firewalldb.NewDB(firewallBoltDB)
+		stores.closeFns["bbolt-firewalldb"] = firewallBoltDB.Close
+	}
+
+	return stores, nil
+}
```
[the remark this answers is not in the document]
ViktorT-11: Aa this is instead handled through our startup process in `terminal.go`, I'm ignoring this feedback.

PR #1313, opened 2026-05-20
go.mod: bump pool to v0.7.0-beta

Roasbeef:
In this PR, we bump the embedded `pool` dependency up to [`v0.7.0-beta`](https://github.com/lightninglabs/pool/releases/tag/v0.7.0-beta). The main motivation is to pull in two reliability fixes around the auctioneer subscription path that were causing the order subscribe stream to silently die after extended runtime.

The first fix is the EOF handling in the subscribe stream (lightninglabs/pool#518). Before, an EOF from the auctioneer would leave the client in a state where the gRPC connection looked healthy, but no further order updates would actually arrive. We now treat EOF the same as any other transient stream error and trigger a reconnect rather than swallowing it. The second piece (also lightninglabs/pool#518) adds jittered backoff on the reconnect path, which avoids the thundering-herd scenario where every connected trader retries at the same instant after the auctioneer comes back up from a brief restart.

The release also pulls in the stream handling fix in `consumePendingOpenChannels` (lightninglabs/pool#507) so we don't get stuck on pending channel finalization, plus the usual round of transitive dependency bumps (`go-viper/mapstructure/v2`, `opencontainers/runc`, `golang-jwt/jwt/v4`, `golang.org/x/net`).

The diff itself is minimal: just `go.mod`/`go.sum` and a one-bullet entry under `### Pool` in the upcoming `0.17.0` release notes.

## Test plan

- [x] `go build ./...` passes on the main lit module.
- [ ] CI passes.

### DIFF
--- a/docs/release-notes/release-notes-0.17.0.md
+++ b/docs/release-notes/release-notes-0.17.0.md
@@ -45,6 +45,12 @@
 
 ### Pool
 
+* Updated [`pool` to
+  `v0.7.0-beta`](https://github.com/lightninglabs/lightning-terminal/pull/1313),
+  which includes a fix for the auctioneer subscription stream silently
+  dying on EOF, jittered reconnect backoff, and a fix for stream
+  handling in the pending-open-channel consumer.
+
 ### Faraday
 
 ### Taproot Assets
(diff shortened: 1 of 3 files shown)


bitromortac (approved):
LGTM :zap:

ViktorT-11 (approved):
Thanks :tada:!

PR #1314, opened 2026-05-22
go.mod: bump pool to v0.7.1-beta

Roasbeef:
In this PR, we bump the embedded `pool` dependency up to [`v0.7.1-beta`](https://github.com/lightninglabs/pool/releases/tag/v0.7.1-beta). The motivation is a single auctioneer reliability fix that follows the same severity class as the EOF reconnect work we pulled in for `v0.7.0-beta`.

The fix is for `HandleServerShutdown` (lightninglabs/pool#521). Before, when the trader's subscription stream went down and the client tried to re-subscribe every account against the new stream, the loop bailed on the first per-account failure. Because the accounts were drained out of the in-memory `subscribedAccts` map *before* the loop ran, any account after the failing one in iteration order was silently lost. The trader looked healthy locally, but the auctioneer saw it as offline and filtered its orders at matching time until the process restarted. The symptom was also intermittent in practice, since Go's randomized map iteration meant the set of "survivors" changed between runs.

The new behavior is best-effort: per-account errors are logged with the trader key for operator visibility, every account still gets a re-subscribe attempt against the new stream, and the aggregated error is returned to the caller via `errors.Join` so the rpcserver still sees a non-nil signal when anything went wrong.

The diff itself is minimal: just `go.mod`/`go.sum` and a one-bullet entry under `### Pool` in the upcoming `0.17.0` release notes.

## Test plan

- [x] `go build .` on the main lit module is unaffected by the bump (pre-existing `perms/mock.go` lnd-API breakage is in a separate sub-module).
- [ ] CI passes.

### DIFF
--- a/docs/release-notes/release-notes-0.17.0.md
+++ b/docs/release-notes/release-notes-0.17.0.md
@@ -62,6 +62,12 @@
   dying on EOF, jittered reconnect backoff, and a fix for stream
   handling in the pending-open-channel consumer.
 
+* Updated [`pool` to
+  `v0.7.1-beta`](https://github.com/lightninglabs/lightning-terminal/pull/1314),
+  which includes a fix for `HandleServerShutdown` silently dropping
+  every account after the first per-account re-subscribe failure on a
+  reconnect.
+
 ### Faraday
 
 ### Taproot Assets
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,7 @@ require (
 	github.com/lightninglabs/loop v0.31.8-beta
 	github.com/lightninglabs/loop/looprpc v1.0.13
 	github.com/lightninglabs/loop/swapserverrpc v1.0.20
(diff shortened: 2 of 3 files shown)


docs/release-notes/release-notes-0.17.0.md
```diff
@@ -62,6 +62,12 @@
   dying on EOF, jittered reconnect backoff, and a fix for stream
   handling in the pending-open-channel consumer.
 
+* Updated [`pool` to
```
ViktorT-11, line 65: Nit: We usually remove the old upgrade entry in the release-notes for the sub-servers, if there's a newer one available. I.e. this should replace the above `pool` entry, and not be added separately.

I can edit this in my upcoming release PRs though, so ok to leave it as is if you want to.

ViktorT-11 (approved):
Thanks! LGTM :fire:

bitromortac (approved):

PR #1315, opened 2026-05-22
[sql-71] Prompt user before starting KVDB -> SQL migration

ViktorT-11:
Based on #1305

This PR adds a prompt to `litd`, if the user's `litd` node is configured to use an SQL db and the KVDB -> SQL migration hasn't been triggered yet.

The prompt will explicitly warn the user the their bbolt database is about to be migrated to SQL, and that when doing so their `bbolt` database will be tombstoned and won't be useable with `litd` any longer.

Additionally, this PR also adds a `auto-migrate-to-sql` config flag. When that flag is set, the prompt will not be shown and the migration will be started automatically. This is intended for users which do not have access to enter input into `stdin`.

### DIFF
--- a/db/tombstone/tombstone.go
+++ b/db/tombstone/tombstone.go
@@ -85,6 +85,31 @@ func CheckKVDBDeprecated(path string, bucketKey []byte,
 	return nil
 }
 
+// HasActiveKVDB reports whether the legacy bbolt database at the given path
+// exists and has not yet been tombstoned by a SQL migration in the specified
+// top-level bucket. A missing file or a tombstoned database is reported as
+// inactive, so callers can use this to detect kvdb state that is still pending
+// migration to SQL.
+func HasActiveKVDB(path string, bucketKey []byte,
+	timeout time.Duration) (bool, error) {
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return false, nil
+	}
+
+	err := CheckKVDBDeprecated(path, bucketKey, timeout)
+	switch {
+	case errors.Is(err, ErrKVDBDeprecated):
+		return false, nil
+
+	case err != nil:
+		return false, err
+
+	default:
+		return true, nil
+	}
+}
+
 // IsMigrationTombstoneKey returns true if the given key is the kvdb migration
 // tombstone marker.
 func IsMigrationTombstoneKey(key []byte) bool {
--- a/migration_prompt.go
+++ b/migration_prompt.go
@@ -1,0 +1,180 @@
+package terminal
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/lightninglabs/lightning-terminal/accounts"
+	"github.com/lightninglabs/lightning-terminal/firewalldb"
+	"github.com/lightninglabs/lightning-terminal/session"
+)
+
+var kvdbToSQLMigrationPromptLines = []string{
+	"",
+	"CAUTION: litd is about to migrate your existing data to a new SQL " +
+		"database.",
+	"After this, litd will use the SQL database for your existing data " +
+		"and any new data added after that point.",
+	"However, after the migration you will not be able to switch back " +
+		"to your old database, as it will be incompatible with litd " +
+		"after the migration.",
+	"NOTE: This also means that you will not be able to downgrade litd " +
+		"to a version prior to when SQL database support was added " +
+		"(v0.17.0-alpha).",
+	"",
+	"SQL databases are more performant, quicker to start, and much less " +
+		"prone to database corruption.",
+	"It is therefore strongly recommended that you proceed with this " +
+		"database migration.",
+	"",
+	"If you want to abort the migration and keep using the old database " +
+		"type instead, stop now and restart litd with the following " +
+		"config option set: `databasebackend=bbolt`.",
+	"Please note though that your old database type (bbolt) is " +
+		"deprecated, and support for it will be removed in a future " +
+		"release. Migration to SQL will at that point be mandatory.",
+	"",
+}
+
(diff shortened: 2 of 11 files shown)


bitromortac (commented):
Concept ACK. I think you can squash the changes as it adds more review confusion and dev friction than necessary with the rewrites. I think the code is well reviewable in its final form.

ViktorT-11:
> Concept ACK. I think you can squash the changes as it adds more review confusion and dev friction than necessary with the rewrites. I think the code is well reviewable in its final form.

Thanks for the review :pray:! Squashed some of the commits :)

migration_prompt.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return err
+	}
+
+	reader := bufio.NewReader(input)
+	answer, err := reader.ReadString('\n')
+	if err != nil {
+		return fmt.Errorf("manual confirmation required before kvdb "+
+			"migration can continue: %w", err)
+	}
```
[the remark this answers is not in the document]
bitromortac: I think we don't need it as the common flow would be to use the config flag?

db/tombstone/tombstone.go
```diff
@@ -85,6 +85,28 @@ func CheckKVDBDeprecated(path string, bucketKey []byte,
 	return nil
 }
 
+// HasMigrationTombstone reports whether the given legacy bbolt database
+// contains the SQL migration tombstone marker in any top-level bucket.
+// Missing files are treated as not tombstoned so callers can distinguish
+// between a fresh legacy store and one that was already migrated.
+func HasMigrationTombstone(path string, timeout time.Duration) (bool, error) {
```
bitromortac: Maybe it would be cleaner to add a
```golang
// HasActiveKVDB reports whether the legacy bbolt database at the given path
// exists and has not yet been tombstoned by a SQL migration in the specified
// top-level bucket. A missing file or a tombstoned database is reported as
// inactive, so callers can use this to detect kvdb state that is still pending
// migration to SQL.
func HasActiveKVDB(path string, bucketKey []byte,
	timeout time.Duration) (bool, error) {

	if _, err := os.Stat(path); os.IsNotExist(err) {
		return false, nil
	}

	err := CheckKVDBDeprecated(path, bucketKey, timeout)
	switch {
	case errors.Is(err, ErrKVDBDeprecated):
		return false, nil

	case err != nil:
		return false, err

	default:
		return true, nil
	}
}
```

we could then add
```golang
// HasActiveKVDB reports whether the rules kvdb file in the given db directory
// still holds data that is pending migration to SQL.
func HasActiveKVDB(dbDir string) (bool, error) {
	return tombstone.HasActiveKVDB(
		filepath.Join(dbDir, DBFilename), rulesBucketKey,
		DefaultRulesDBTimeout,
	)
}
```
to each db and then change `hasActiveLegacyKVDB` to do the checks to have a single source of truth.

terminal.go
```diff
@@ -444,6 +444,13 @@ func (g *LightningTerminal) start(ctx context.Context) error {
 		return fmt.Errorf("could not create network directory: %v", err)
 	}
 
+	if g.cfg.DatabaseBackend != DatabaseBackendBbolt {
```
bitromortac: I think we also need to check the `--sqlite.skipmigrations=true` flags here? Otherwise there could be a discrepancy.

migration_prompt.go
```diff
@@ -57,6 +65,21 @@ func (c *Config) confirmPendingKVDBToSQLMigration() error {
 func (c *Config) confirmPendingKVDBToSQLMigrationWithInput(
 	input io.Reader, output io.Writer) error {
 
+	// If the user has enabled auto migration to SQL without the prompt,
+	// we can return early.
+	if c.AutoMigrateKVDB {
+		return nil
+	}
+
+	autoMigrateKVDB, err := autoMigrateKVDBFromEnv()
```
bitromortac: It may be better to parse the env at the config layer, similar to `readUIPassword`. We also should clarify the precedence of the flag and the env. For example setting the flag to false does not override the env, which I'd see as the expected behavior, since the flag is more explicit.

migration_prompt_test.go
```diff
@@ -41,7 +41,53 @@ func TestConfirmPendingKVDBToSQLMigration(t *testing.T) {
 		)
 		require.ErrorContains(t, err, "manual confirmation declined")
 	})
-	
+
+	t.Run("skips prompt on auto migration config", func(t *testing.T) {
```
bitromortac: It could make sense to convert this to a table-driven test.

migration_prompt.go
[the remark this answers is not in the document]
ViktorT-11: Still added this as it feels like there's no con in doing so.

db/tombstone/tombstone.go
ViktorT-11: Addressed :)

terminal.go
ViktorT-11: Added!

migration_prompt.go
ViktorT-11: Addressed

migration_prompt_test.go
ViktorT-11: Updated

ellemouton (approved):
nice!

db/tombstone/tombstone.go
bitromortac: Thanks! `HasMigrationTombstone` and `hasBoltBucketTombstone` can be deleted then though. Maybe we should switch the "unused" linter on

migration_prompt.go
bitromortac: nice, thanks!

bitromortac (approved):
LGTM :tada: except for some code removal

db/tombstone/tombstone.go
ViktorT-11: Thanks! I'll remove those before merging this :)

PR #1316, opened 2026-05-26
accounts: add account payments subcommand

### LINKED ISSUE #935: accounts: Enable showing the history of payments through the `accounts` cli
It could be useful from a superuser's perspective of the `accounts` subsystem to be able to show the history of accounts payments, instead of the very sparsely detailed info that `litcli accounts list` offers today, compared to `lncli listpayments`.

The superuser is therefore required to query `lnd` in order to get that detailed information, which isn't ideal from a user experience perspective.

Cyberguru1:
closes #935

This PR implements the off-chain account payments history subcommand (`litcli accounts payments`) and its backend support in the Accounts gRPC subserver.

It adds database store methods for BoltStore and SQLStore to fetch account payment hashes in insertion order, exposes the `AccountPayments` RPC endpoint on the RPC server with support for paginated limits, offsets, reversing order, and concurrently fetching payment details from LND, and updates the release documentation.


accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+	// Fetch all payment hashes from the store.
+	hashes, err := s.service.store.ListAccountPayments(ctx, accountID)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"unable to list account payments: %w", err,
+		)
+	}
+
+	// Fetch the account to get the cached payment statuses.
+	account, err := s.service.store.Account(ctx, accountID)
```
ViktorT-11: I think this shows an issue with your added `ListAccountPayments` API, as the payment's fetched from the db already contains the status, but since only the hash is returned, you need to add this extra check db call + check.

Instead you should change the `ListAccountPayments` API to return the full payment, and not just the hash.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch account: %w", err)
+	}
+
+	// Filter out payments if we don't want incomplete ones.
+	var targetHashes []lntypes.Hash
+	for _, h := range hashes {
+		if !req.IncludeIncomplete {
```
ViktorT-11: I'm a bit opposed to this `IncludeIncomplete` flag, as with in combination with the `offset` pagination calls, there's a chance that the user will miss payments which have been completed between the first `RPC` call and the last call, and therefore the full returned paginated result will not contain all completed payments. 

I'd be in favour of removing this flag for now, despite that `lncli` has it. We can add it at a later stage if needed.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+	maxPayments := req.MaxPayments
+	if maxPayments == 0 {
+		maxPayments = 20
+	} else if maxPayments > 50 {
+		maxPayments = 50
+	}
+
+	if uint64(len(targetHashes)) > maxPayments {
+		targetHashes = targetHashes[:maxPayments]
+	}
```
ViktorT-11: I think this offset handling & `maxPayments` should be handled in the `db` layer instead of here, i.e. in the SQL implementation (directly in the query) and the `kvdb` implementation.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+			payment, err := stream.Recv()
+			if err != nil {
+				log.Warnf("Failed to receive payment "+
+					"update for %x: %v", h[:], err)
+				return nil
+			}
+
+			trackCancel()
+			payments[i] = payment
+			return nil
```
ViktorT-11: nit: please add a new line before these `return` statement rows.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+		targetHashes = targetHashes[:maxPayments]
+	}
+
+	// Fetch the detailed payments concurrently from LND.
+	rawCtx, _, client := s.service.routerClient.RawClientWithMacAuth(ctx)
+	g, gCtx := errgroup.WithContext(rawCtx)
+	payments := make([]*lnrpc.Payment, len(targetHashes))
+
+	for i, h := range targetHashes {
+		i, h := i, h
```
ViktorT-11: nit: Also add a new line after this row please :)

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+			}
+
+			payment, err := stream.Recv()
+			if err != nil {
+				log.Warnf("Failed to receive payment "+
+					"update for %x: %v", h[:], err)
+				return nil
+			}
+
+			trackCancel()
```
ViktorT-11: I don't think the extra `trackCancel()` is necessary here, as you already have deferred it above, and the only logic that's done before the return here is the line:
`payments[i] = payment`

proto/lit-accounts.proto
```diff
@@ -244,4 +252,40 @@ message AccountIdentifier {
...
+    */
+    bool include_incomplete = 5;
+}
+
+message AccountPaymentsResponse {
+    /*
+    The detailed payments associated with the account.
+    */
+    repeated lnrpc.Payment payments = 1;
 }
```
ViktorT-11, line 306: We'd also want to add some info in the response that can be used for the pagination to request the next payments. 
This is what `listpayments` return in `lncli`:
```
type ListPaymentsResponse struct {
	state protoimpl.MessageState `protogen:"open.v1"`
	// The list of payments
	Payments []*Payment `protobuf:"bytes,1,rep,name=payments,proto3" json:"payments,omitempty"`
	// The index of the first item in the set of returned payments. This can be
	// used as the index_offset to continue seeking backwards in the next request.
	FirstIndexOffset uint64 `protobuf:"varint,2,opt,name=first_index_offset,json=firstIndexOffset,proto3" json:"first_index_offset,omitempty"`
	// The index of the last item in the set of returned payments. This can be used
	// as the index_offset to continue seeking forwards in the next request.
	LastIndexOffset uint64 `protobuf:"varint,3,opt,name=last_index_offset,json=lastIndexOffset,proto3" json:"last_index_offset,omitempty"`
	// Will only be set if count_total_payments in the request was set. Represents
	// the total number of payments (complete and incomplete, independent of the
	// number of payments requested in the query) currently present in the payments
	// database.
	TotalNumPayments uint64 `protobuf:"varint,4,opt,name=total_num_payments,json=totalNumPayments,proto3" json:"total_num_payments,omitempty"`
	unknownFields    protoimpl.UnknownFields
	sizeCache        protoimpl.SizeCache
}
 ```

I think we'd want to copy that, and while you're at it, you could also support the `TotalNumPayments` field as well, which is only returned if a `--count_total_payments` flag is set in the request.

File: cmd/litcli/accounts.go [hunk 8]
```diff
@@ -541,3 +542,77 @@ func parseIDOrLabel(ctx *cli.Context) (string, string, cli.Args, error) {
...
+			Name: "max_payments",
+			Usage: "The maximum number of payments to " +
+				"return. The default value is 20 and " +
+				"the maximum is 50.",
+			Value: 20,
+		},
+		cli.Uint64Flag{
+			Name: "index_offset",
+			Usage: "The index offset of the first " +
+				"payment to return.",
```
ViktorT-11: You should include more info of how this argument is used. In `lncli` it states:
````
The index of a payment that will be used as either the start (in forwards mode) or end (in reverse mode) of a query to determine which payments should be returned in the response, where the index_offset is excluded. If index_offset is set to zero in reversed mode, the query will end with the last payment made. 
```

I think you can effectively copy that. You'd also want to use a similar description for the protos field.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+		maxPayments = 50
+	}
+
+	if uint64(len(targetHashes)) > maxPayments {
+		targetHashes = targetHashes[:maxPayments]
+	}
+
+	// Fetch the detailed payments concurrently from LND.
+	rawCtx, _, client := s.service.routerClient.RawClientWithMacAuth(ctx)
+	g, gCtx := errgroup.WithContext(rawCtx)
```
ViktorT-11: So we haven't really used `errgroup` in any of our other repos, so I suggest not introducing that pattern here.
Instead I suggest using a standard `sync.WaitGroup`, and add to that for every started goroutine.

Alternatively, you could look into the using the `github.com/lightningnetwork/lnd/fn` `GoroutineManager` if you'd want something similar to the `errgroup` implementation.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+					NoInflightUpdates: true,
+				},
+			)
+			if err != nil {
+				log.Warnf("Failed to track payment %x: %v",
+					h[:], err)
+				return nil
+			}
+
+			payment, err := stream.Recv()
```
ViktorT-11, line 536: I think we should timeout here if we don't receive anything over `stream.Recv()` for like 5-10 seconds.

accounts/store_test.go
```diff
@@ -848,3 +850,64 @@ func TestLastInvoiceIndexes(t *testing.T) {
...
+
+	_, err = store.UpsertAccountPayment(
+		ctx, acct.ID, hash2, 200, lnrpc.Payment_SUCCEEDED,
+	)
+	require.NoError(t, err)
+
+	_, err = store.UpsertAccountPayment(
+		ctx, acct.ID, hash3, 300, lnrpc.Payment_FAILED,
+	)
+	require.NoError(t, err)
```
ViktorT-11, line 958: We'd definitely would like to test payments with these statuses in the `itest` you've added, and not just in the `store` test. The main reason for this is that we'd want to ensure that we correctly cancel the `TrackPaymentV2` for real payments which are in-flight for example, and that those streaming RPC calls doesn't remain pending until the payment has succeeded.

ViktorT-11 (commented):
Thanks a lot again for the contribution @Cyberguru1 :fire::pray:! I've added a few comments below as an initial review.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+	// Sort target hashes lexicographically to ensure pagination and
+	// ordering are deterministic across all store backends.
+	sort.Slice(targetHashes, func(i, j int) bool {
+		return bytes.Compare(targetHashes[i][:], targetHashes[j][:]) < 0
+	})
+
+	// Apply reversed ordering.
+	if req.Reversed {
+		slices.Reverse(targetHashes)
+	}
```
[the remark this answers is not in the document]
Cyberguru1: I updated the documentation and CLI help to clarify that ordering is lexicographical rather than chronological.

Since the `account_payments` table doesn't store timestamps, achieving true chronological ordering would require a `TrackPaymentV2` streaming RPC call per payment to fetch the necessary details, which doesn't scale well for large accounts when not paginated.

accounts/rpcserver.go
```diff
@@ -371,3 +377,137 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+	} else if maxPayments > 50 {
+		maxPayments = 50
+	}
+
+	if uint64(len(targetHashes)) > maxPayments {
+		targetHashes = targetHashes[:maxPayments]
+	}
+
+	// Fetch the detailed payments concurrently from LND.
+	rawCtx, _, client := s.service.routerClient.RawClientWithMacAuth(ctx)
```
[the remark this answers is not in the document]
Cyberguru1: `RawClientWithMacAuth` doesn't return a cleanup function, the second return value is a `time.Duration` representing the macaroon timeout. [Reference](https://github.com/lightninglabs/lndclient/blob/master/lnd_services.go#L85)

accounts/rpcserver.go
Cyberguru1: Yes, that makes sense. Returning the full payment details would be better than returning only the hashes and then fetching the payment details again.

accounts/rpcserver.go
Cyberguru1: That is a high possibility, nice catch. I would remove the flag implementation for now and we can add it later if it becomes necessary.

accounts/rpcserver.go
Cyberguru1: I have updated this to use a custom query in the `db` layer. This approach is better since it queries only what is needed instead of fetching everything first and then applying pagination.

accounts/rpcserver.go
Cyberguru1: Agreed, I’ve removed the extra `trackCancel()` call.

proto/lit-accounts.proto
Cyberguru1: I’ve added the extra fields in the response to provide more context during pagination, and also included support for `TotalNumPayments`.

cmd/litcli/accounts.go
Cyberguru1: Updated the description to include how `index_offset` is used, based on the `lncli` docs, and aligned the proto field as well.

accounts/rpcserver.go
Cyberguru1: I used the standard library `sync.WaitGroup` and also collected any errors returned from each goroutine.

accounts/rpcserver.go
Cyberguru1: Yes, that makes sense. I’ve added an 8-second timeout.

accounts/store_test.go
Cyberguru1: I’ve updated `litd_accounts_test.go` to include tests for the three statuses.

Cyberguru1:
> Thanks a lot again for the contribution @Cyberguru1 🔥🙏! I've added a few comments below as an initial review.
Hello @ViktorT-11,

Thanks again for taking the time to review the PR. I’ve addressed the comments you pointed out and pushed the updates.

Regarding the failing CI, I investigated the issue and tried reproducing it locally multiple times, but the tests pass on my end. From the logs, it seems the failure happens while starting the Bob node, where the `lit.macaroon` file is not yet available when the RPC connection is attempted. It might need a re-run.

Looking forward to a follow-up review.

db/sqlc/accounts_custom.go
```diff
@@ -0,0 +1,132 @@
...
+	AccountID int64
+	Status    sql.NullInt16
+	Reversed  bool
+	Limit     int32
+	Offset    int32
+}
+
+// ListAccountPaymentsPaginated retrieves a paginated, sorted, and optionally
+// filtered list of account payments.
+func (q *Queries) ListAccountPaymentsPaginated(ctx context.Context,
```
ViktorT-11: In order to not complicate the sql db implementation, I would suggest that we don't add a separate "custom SQL query" for this, so that querier can be auto implemented with `make sqlc`.

I therefore suggest you add this query & the `CountAccountPayments` to `db/sqlc/queries/accounts.sql` instead.

The only real benefit having this as a custom query, is that you can optimize the `index` usage for the `db`. However, this is only applicable as you have filtering by `status` added as an option in the `db` layer.

As none of the production code actually uses that filtering though, I suggest you remove that filtering by `status` option, and instead add queries like these to the `db/sqlc/queries/accounts.sql` file:

```
-- name AccountPaymentsAsc :many
SELECT *
FROM account_payments
WHERE account_id = @account_id
ORDER BY hash ASC
LIMIT @num_limit OFFSET @num_offset;

-- name: AccountPaymentsDesc :many
SELECT *
FROM account_payments
WHERE account_id = @account_id
ORDER BY hash DESC
LIMIT @num_limit OFFSET @num_offset;

-- name: CountAccountPayments :one
SELECT COUNT(*)
FROM account_payments
WHERE account_id = @account_id;
```

Feel free to rename and update those queries in whichever way you find suitable.

cmd/litcli/accounts.go
```diff
@@ -541,3 +542,83 @@ func parseIDOrLabel(ctx *cli.Context) (string, string, cli.Args, error) {
...
+		},
+		cli.StringFlag{
+			Name:  labelName,
+			Usage: "(optional) The unique label of the account.",
+		},
+		cli.Uint64Flag{
+			Name: "max_payments",
+			Usage: "The maximum number of payments to " +
+				"return. The default value is 20 and " +
+				"the maximum is 50.",
```
ViktorT-11: As the description specifies that `50` is the maximum value, I suggest that this command should error whenever a value over 50 is specified. That should be both done in this file & in the `accounts/rpcserver.go` file.

accounts/rpcserver.go
```diff
@@ -371,3 +374,155 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+	if req.CountTotalPayments {
+		total, err := s.service.store.CountAccountPayments(
+			ctx, accountID, lnrpc.Payment_UNKNOWN,
+		)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"unable to count account payments: %w", err,
+			)
+		}
+		totalNumPayments = total
```
ViktorT-11, line 464: nit: New line above this line.

accounts/rpcserver.go
```diff
@@ -371,3 +374,155 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+
+	// Determine limits.
+	limit := req.MaxPayments
+	if limit == 0 {
+		limit = 20
+	} else if limit > 50 {
+		limit = 50
+	}
+
+	offset := req.IndexOffset
```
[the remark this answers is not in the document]
ViktorT-11: Sure, you could add this check.

cmd/litcli/accounts.go
```diff
@@ -541,3 +542,83 @@ func parseIDOrLabel(ctx *cli.Context) (string, string, cli.Args, error) {
...
+				"query to determine which payments should " +
+				"be returned in the response, where the " +
+				"index_offset is excluded. If index_offset " +
+				"is set to zero in reversed mode, the " +
+				"query will end with the last payment made.",
+		},
+		cli.BoolFlag{
+			Name: "reversed",
+			Usage: "If true, payments are returned in " +
+				"reverse lexicographical order.",
```
ViktorT-11: You could maybe add here that it's the "lexicographical order" for the payment's hash.

itest/litd_accounts_test.go
```diff
@@ -164,12 +168,189 @@ func runAccountSystemTest(t *harnessTest, node *HarnessNode, hostPort,
...
+			},
+			MaxPayments: 2,
+		},
+	)
+	require.NoError(t.t, err)
+	require.Len(t.t, paymentsResp.Payments, 2)
+	require.EqualValues(t.t, 0, paymentsResp.FirstIndexOffset)
+	require.EqualValues(t.t, 2, paymentsResp.LastIndexOffset)
+
+	// Query with pagination offset out of bounds.
```
ViktorT-11, line 329: Nice, I like this new `itest` coverage a lot!

One more small thing you could add there is that you could also add a call above here where `IndexOffset=2` and also with `MaxPayments=2` set. That'll add test coverage when the number of payments returned are fewer than the `MaxPayments`.

ViktorT-11 (commented):
Thanks a lot for the updates @Cyberguru1! This is looking much better :tada:. I'm leaving a few more comments, but I think this PR is almost there now :fire:.

accounts/rpcserver.go
```diff
@@ -371,3 +374,155 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+		}()
+	}
+
+	wg.Wait()
+
+	if len(errs) > 0 {
+		return nil, fmt.Errorf(
+			"failed to fetch payment details: %w", errs[0],
+		)
+	}
```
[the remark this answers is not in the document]
ViktorT-11: I think this is a valid point, but not sure that the account should still be valid if payments have been removed from `lnd`'s database. But you could implement this, and then instead of logging "Warn", log with "Error" instead. Though if there's no payments returned from `lnd`, and this `if len(errs) > 0` check is `true`, I still think the entire RPC should error.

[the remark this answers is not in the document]
Cyberguru1: Good point. I’ve updated it to continue returning the available payments while logging individual tracking failures as errors. The RPC will still return an error if no payments can be retrieved and tracking errors are present.

db/sqlc/accounts_custom.go
Cyberguru1: That was my initial thought as well. I anticipated that filtering by status might be useful in the future, which is why I opted for the custom query approach.

I agree that it adds unnecessary complexity for the current use case and also bypasses the benefits of having the queries generated by sqlc. I've updated it to use generated sqlc queries and removed the status filtering.

cmd/litcli/accounts.go
Cyberguru1: Updated

cmd/litcli/accounts.go
Cyberguru1: Looks more informative. I’ve updated it to return an error whenever a value greater than 50 is specified, and added the validation in both locations.

itest/litd_accounts_test.go
Cyberguru1: Nice, thanks for the suggestion. I’ve added another test case to cover the case where fewer payments are returned than the requested limit.

cmd/litcli/accounts.go
```diff
@@ -541,3 +542,89 @@ func parseIDOrLabel(ctx *cli.Context) (string, string, cli.Args, error) {
...
+				"index_offset is excluded. If index_offset " +
+				"is set to zero in reversed mode, the " +
+				"query will end with the last payment made.",
+		},
+		cli.BoolFlag{
+			Name: "reversed",
+			Usage: "If true, payments are returned in " +
+				"reverse lexicographical order of " +
+				"the payment's hash.",
+		},
```
ViktorT-11: Hmmm I've been thinking more about this reverse flag, and sorry for realising it this late in the PR, but I don't think it adds any value in `litd`, as there is no way we can sort the payments by insertion order with the data we have in `litd` for the payments.. Compare that to `lnd` where it's also present, where it's very simple to sort the payments by creation time. Instead the only sorting option we have in `litd` is the payment's `hash`.

We could add an ID field to the `account_payments` sql table which would be able to track insertion order, but the issue is that the the `kvdb` store contains no way to track the order of already inserted payments.

Therefore, I've come to the conclusion that I unfortunately think it makes more sense from a UX point of view to remove this `reversed` option, as it brings no real value in `litd`. Sorry @Cyberguru1  for realising this earlier.

cmd/litcli/accounts.go
```diff
@@ -541,3 +542,89 @@ func parseIDOrLabel(ctx *cli.Context) (string, string, cli.Args, error) {
...
+		cli.Uint64Flag{
+			Name: "index_offset",
+			Usage: "The index of a payment that will be " +
+				"used as either the start (in forwards " +
+				"mode) or end (in reverse mode) of a " +
+				"query to determine which payments should " +
+				"be returned in the response, where the " +
+				"index_offset is excluded. If index_offset " +
+				"is set to zero in reversed mode, the " +
+				"query will end with the last payment made.",
```
ViktorT-11: This comment mentioning "end with the last payment made" is incorrect. I suggest you mention that the returned result will sort be ordered by the lexicographical order of the payment hash instead.

db/sqlc/queries/accounts.sql
```diff
@@ -80,6 +80,25 @@ WHERE account_id = $1;
 SELECT *
 FROM account_payments;
 
+-- name: AccountPaymentsAsc :many
```
ViktorT-11: If you remove the `reverse` flag, I suggest you rename this to something like `AccountPaymentsPaginated` or something similar, and then remove the other `AccountPaymentsDesc` query.

ViktorT-11 (commented):
Generally this looks good now! I realized though that the `reverse` flag doesn't really add any value in `litd`, and I've left a comment regarding that below. Really sorry for not catching this earlier! 

If you agree with my reasoning, I'll also request a review from @bitromortac once that's been addressed, as this PR is generally in a great state now!

cmd/litcli/accounts.go
Cyberguru1: Yes, I agree. Users would be most interested in time-sorted payment entries. Because of the KVDB store limitation for already inserted payments, this is not possible right now. 

Once the database migration to SQLite/Postgres is fully complete, we could store the payment timestamps returned from LND in the `account_payments` table to easily sort by time. We can plan this as a future improvement.

db/sqlc/queries/accounts.sql
Cyberguru1: Updated :heavy_check_mark:

cmd/litcli/accounts.go
Cyberguru1: Included the sorting order in the description of the command

litrpc/lit-accounts.proto
```diff
@@ -244,4 +252,57 @@ message AccountIdentifier {
...
+message AccountPaymentsResponse {
+    /*
+    The detailed payments associated with the account, sorted in ascending
+    lexicographical order of their payment hash.
+    */
+    repeated lnrpc.Payment payments = 1;
+
+    /*
+    The index of the first payment returned. This can be used as the
+    index_offset in a subsequent query to paginate backwards.
```
ViktorT-11: Very minor nit: Don't think the last part is true any longer after the removal of the `reverse` flag, i.e. "This can be used as the index_offset in a subsequent query to paginate backwards.".

Don't fix this unless `bitmortac` finds other things to address though, as it so minor.

ViktorT-11 (approved):
Nice looks great, thanks a lot for the fixes :fire::tada:!

LGTM! I'll test this thoroughly tomorrow, but since the `itest` is quite thorough, I don't think I'll have any additional feedback after the testing round. Meanwhile, I'm requesting a review from @bitromortac.

litrpc/lit-accounts.proto
Cyberguru1: I must have missed updating this description, thanks for pointing it out. Sure, I would address it after the follow-up review from @bitromortac.

litbot-9000:
@bitromortac: review reminder
@cyberguru1, remember to re-request review from reviewers when ready

accounts/rpcserver.go
```diff
@@ -371,3 +374,158 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+
+	wg.Wait()
+
+	// If we failed to track any payments successfully and there were
+	// errors, return the error.
+	if len(payments) == 0 && len(errs) > 0 {
+		return nil, fmt.Errorf(
+			"failed to fetch payment details: %w", errs[0],
+		)
+	}
```
bitromortac: If some payments fail to track but others succeed, we quietly drop the failed ones from the response while still
reporting the full offset window and total count. That means a caller can't tell anything went missing. Could we fail the whole request when any lookup errors so the page stays honest?

accounts/rpcserver.go
```diff
@@ -371,3 +374,158 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+	if limit == 0 {
+		limit = 20
+	} else if limit > 50 {
+		return nil, fmt.Errorf("max_payments cannot exceed 50")
+	}
+
+	if req.IndexOffset > 0x7fffffff {
+		return nil, fmt.Errorf("index_offset out of range")
+	}
+	offset := req.IndexOffset
```
bitromortac, line 440: Could we lift these into named constants? The default of 20, the cap of 50, the offset guard, and the timeout are magic numbers and the 50 cap is repeated over in litcli too, so it'd be nice to have the server own that as the single source of truth.

accounts/rpcserver.go
```diff
@@ -371,3 +374,158 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+		errs     []error
+	)
+
+	rawCtx, _, client := s.service.routerClient.RawClientWithMacAuth(ctx)
+
+	for _, entry := range paymentsFromStore {
+		entry := entry
+
+		wg.Add(1)
+		go func() {
```
bitromortac, line 482: We fire off one TrackPaymentV2 stream per payment with no bound here. It's fine today since the page is capped at 50, but a small semaphore would make the concurrency explicit and keep us safe if that cap ever grows.

proto/lit-accounts.proto
```diff
@@ -244,4 +252,57 @@ message AccountIdentifier {
...
+    */
+    uint64 max_payments = 2 [jstype = JS_STRING];
+
+    /*
+    The index of a payment that will be used as the start of a query to
+    determine which payments should be returned in the response, where the
+    index_offset is excluded. The payments are returned in ascending
+    lexicographical order of their payment hash.
+    */
+    uint64 index_offset = 3 [jstype = JS_STRING];
```
bitromortac, line 274: nit: this reads like lnd's excluded cursor index, but it's really just a plain row offset. Could you reword it (and the matching litcli flag) so it's clear it's an offset into the sorted list?

proto/lit-accounts.proto
```diff
@@ -244,4 +252,57 @@ message AccountIdentifier {
...
+
+    /*
+    If set, the total number of payments matching the query will be returned
+    in the response.
+    */
+    bool count_total_payments = 6;
```
bitromortac: The field numbers jump from 3 to 6 here. Since this message is new, it'd be cleaner to make them contiguous.

itest/litd_accounts_test.go
```diff
@@ -164,12 +168,235 @@ func runAccountSystemTest(t *harnessTest, node *HarnessNode, hostPort,
 		ctxa, t, rawConn, charlie, acctBalance,
 	)
 
+	// Initiate a HODL payment (which remains IN_FLIGHT) and a FAILED
+	// payment to verify that the AccountPayments RPC correctly retrieves
+	// and reports payments across all potential lifecycles (SUCCEEDED,
+	// IN_FLIGHT, FAILED).
```
bitromortac, line 175: This comment lists SUCCEEDED among the lifecycles, but the succeeded payment isn't created here, it's the 4444-sat payment made earlier in testAccountRestrictions. Could we note that in the comment so it's clear
where the third payment in the assertions comes from?

bitromortac (changes requested):
Nice addition, the account payment is a useful endpoint. I left a few inline notes: one about how we handle a failed lookup and the rest are small polish and doc things.

One structural thought on the commits: it'd be nice to introduce the unit tests in the same commit as the code they exercise, so each commit stays self-contained.

Thanks for putting this together!

accounts/rpcserver.go
Cyberguru1: Added a semaphore bound via MaxConcurrentTrackPayments.

proto/lit-accounts.proto
Cyberguru1: updated the descriptions

proto/lit-accounts.proto
Cyberguru1: True, updated it

itest/litd_accounts_test.go
Cyberguru1: Updated the [comment](https://github.com/lightninglabs/lightning-terminal/pull/1316/changes/128f31d88655c6c21eb7c43771ec48af757c3ceb#diff-288e0b6402591efec3fcc5b96f361713d392a139c3d9f4b757642e2a3ce49b2dR263) to clarify the source of the SUCCEEDED payment.

accounts/rpcserver.go
Cyberguru1: Moved values to a named constant and used them in litcli.

accounts/rpcserver.go
Cyberguru1: Updated this flow to fail the whole request if any lookup fails.

Cyberguru1:
> Nice addition, the account payment is a useful endpoint. I left a few inline notes: one about how we handle a failed lookup and the rest are small polish and doc things.
> 
> One structural thought on the commits: it'd be nice to introduce the unit tests in the same commit as the code they exercise, so each commit stays self-contained.
> 
> Thanks for putting this together!

@bitromortac  Thanks for the review! :pray: Addressed the inline notes and folded the test commit into the code commit it exercises, so history's self-contained now.

Looking forward to the next review round! :rocket:

litrpc/lit-accounts.proto
Cyberguru1: Updated the description

accounts/rpcserver.go
bitromortac: I discussed this out of band with @ViktorT-11, his argument was that there can be situations where the payment hashes in the lit database become de-synced with the lnd node's, perhaps because a user deleted payments or started from a new lnd db, reusing old lnd data. I'm not sure how to handle that, but I prefer an explicit approach.

bitromortac: Maybe another solution could be to allow gaps if we encounter "doesn't exist errors". We'd need to see what error lnd would return for a payment keyed by an absent payment hash.

bitromortac: So it seems like `TrackPaymentV2` on a hash lnd doesn't know returns a gRPC `codes.NotFound` error (message "payment isn't initiated").

Path: `TrackPaymentV2` -> `subscribePayment` -> `Tower.SubscribePayment` -> `db.FetchPayment`, which returns `ErrPaymentNotInitiated`; `subscribePayment` maps that to `status.Error(codes.NotFound, ...)`.

@Cyberguru1 can you check if we can use this to skip over gaps where lit and lnd are de-synced concerning absent/deleted payments and that we do the right pagination?

accounts/rpcserver.go
```diff
@@ -371,3 +394,170 @@ func marshalAccount(acct *OffChainBalanceAccount) *litrpc.Account {
...
+			if err != nil {
+				log.Errorf("Failed to receive payment "+
+					"update for %x: %v", entry.Hash[:], err)
+
+				mu.Lock()
+				errs = append(errs, err)
+				mu.Unlock()
+
+				return
+			}
```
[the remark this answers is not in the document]
bitromortac: Would be done with the suggestion above.

accounts/store_kvdb.go
```diff
@@ -445,6 +446,60 @@ func (s *BoltStore) DeleteAccountPayment(_ context.Context, id AccountID,
...
+	})
+
+	// Apply pagination limits and offsets.
+	total := int32(len(matchedPayments))
+	if offset >= total {
+		return nil, nil
+	}
+
+	end := offset + limit
+	if limit <= 0 || end > total {
```
bitromortac: tiny nit: to make the behavior between kvdb and sql consistent we'd need to have `limit < 0` I think here, to pass back zero entries

bitromortac (commented):
Thank you for the changes. Could you look into the thread https://github.com/lightninglabs/lightning-terminal/pull/1316#discussion_r3534248760 about skipping pagination if we fail for a single query?

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -91,6 +91,12 @@
   `lncli bakemacaroon`, for sessions of type `custom`. Supports repeated flags,
   comma-separated lists, and mixed input.
 
+* [Add accounts payments history subcommand](https://github.com/lightninglabs/lightning-terminal/pull/1316):
```

accounts/rpcserver.go
Cyberguru1: @bitromortac apologies for the late reply, yes we can use it, i've updated the flow to check for the `codes.NotFound` error. I used the count of actually returned/tracked payments (`finalPayments`) instead of the database count of fetched entries (`paymentsFromStore`) to calculate the `lastIndexOffset` for the returned pagination.

accounts/store_kvdb.go
Cyberguru1: Updated this.

docs/release-notes/release-notes-0.17.0.md
Cyberguru1: Thanks, I've updated this.

accounts/rpcserver.go
Cyberguru1: Follow-up thought on the pagination, I was thinking that instead of skipping the desynced entries, we could still include them in the response but explicitly mark them with a "not found" status or return a placeholder entry that indicates that. That way, the user can clearly see in the response which payments are missing due to a desync. What do you think?

ViktorT-11: > Follow-up thought on the pagination, I was thinking that instead of skipping the desynced entries, we could still include them in the response but explicitly mark them with a "not found" status or return a placeholder entry that indicates that. That way, the user can clearly see in the response which payments are missing due to a desync. What do you think?

Yes, that sounds like a good alternative! I think that especially makes sense as it would become clearer for the user why certain payment_hashes are shown in the `litcli accounts list`, but not in the command which actually  lists the payments.

Cyberguru1: Thanks! I updated it to follow that approach. Desynced entries are now included in the response with the `lnrpc.Payment_UNKNOWN` status instead of being skipped.

bitromortac: Looks good, thanks!

accounts/store_kvdb.go
```diff
@@ -445,6 +446,60 @@ func (s *BoltStore) DeleteAccountPayment(_ context.Context, id AccountID,
...
+		hashI := matchedPayments[i].Hash[:]
+		hashJ := matchedPayments[j].Hash[:]
+		return bytes.Compare(hashI, hashJ) < 0
+	})
+
+	// Apply pagination limits and offsets.
+	total := int32(len(matchedPayments))
+	if offset >= total {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
bitromortac: it's positive via the `req`

bitromortac (approved):
LGTM, thanks for the updates and additional test coverage :tada:. Due to a merge of another PR we need another rebase unfortunately :pray:.

Cyberguru1:
> LGTM, thanks for the updates and additional test coverage 🎉. Due to a merge of another PR we need another rebase unfortunately 🙏.

@bitromortac  Thanks! 🙏

I've rebased the branch and resolved the merge conflicts.

ViktorT-11:
Thanks again, merging :fire::rocket:!

PR #1319, opened 2026-06-04
[sql-72] Preserve SQL ordering and uniqueness parity with kvdb

ViktorT-11:
## Summary

This PR tightens SQL schema and migration behavior so it matches the legacy
kvdb model more closely in a few places:

- preserve session macaroon caveat and permission order explicitly in SQL
- tighten kvstore namespace uniqueness to match kvdb semantics
- preserve firewall action ordering during kvdb -> SQL migration
- stabilize SQL action query ordering with as a tie-breaker

## Why

These changes fix cases where the SQL schema or migration behavior was weaker
than kvdb:

- session macaroon caveat order can affect the baked macaroon, so SQL needs to
  preserve recipe order explicitly
- kvstores previously relied on a UNIQUE constraint across nullable columns,
  which could allow duplicate global or group-scoped rows that kvdb would never
  permit
- firewall actions in kvdb have a real global order via , but
  the SQL migration previously walked grouped buckets instead
- action queries ordered only by , which was not stable when
  multiple rows shared the same timestamp

## Notes

This PR modifies previous SQL migrations directly. That is OK because SQL
databases are currently only active in `dev` builds.

This PR should be merged before #1305, since that change enables SQL databases
in production.

### DIFF
--- a/db/sqlc/migrations/000003_kvstores.up.sql
+++ b/db/sqlc/migrations/000003_kvstores.up.sql
@@ -50,8 +50,29 @@ CREATE TABLE IF NOT EXISTS kvstores (
     entry_key TEXT NOT NULL,
 
     -- The value of the entry.
-    value BLOB NOT NULL
+    value BLOB NOT NULL,
+
+    -- Feature-scoped kv stores must always belong to a session group.
+    CHECK (feature_id IS NULL OR group_id IS NOT NULL)
 );
 
-CREATE UNIQUE INDEX IF NOT EXISTS kvstores_lookup_idx
-    ON kvstores (entry_key, rule_id, perm, group_id, feature_id);
+-- Mirror the legacy KVDB namespace semantics precisely. A kv store record is
+-- uniquely identified by one of three namespace shapes:
+--   1. Global: entry_key + rule_id + perm
+--   2. Group scoped: entry_key + rule_id + perm + group_id
+--   3. Feature scoped: entry_key + rule_id + perm + group_id + feature_id
+--
+-- A single UNIQUE index across nullable columns is not sufficient here, as
+-- SQL NULL handling can allow duplicates for the global and group-scoped
+-- cases that the KVDB bucket layout would never permit.
+CREATE UNIQUE INDEX IF NOT EXISTS kvstores_global_lookup_idx
+    ON kvstores (entry_key, rule_id, perm)
+    WHERE group_id IS NULL AND feature_id IS NULL;
+
+CREATE UNIQUE INDEX IF NOT EXISTS kvstores_group_lookup_idx
+    ON kvstores (entry_key, rule_id, perm, group_id)
+    WHERE group_id IS NOT NULL AND feature_id IS NULL;
+
+CREATE UNIQUE INDEX IF NOT EXISTS kvstores_feature_lookup_idx
+    ON kvstores (entry_key, rule_id, perm, group_id, feature_id)
+    WHERE feature_id IS NOT NULL;
--- a/db/sqlc/migrations/000005_actions.down.sql
+++ b/db/sqlc/migrations/000005_actions.down.sql
@@ -1,5 +1,5 @@
 DROP INDEX IF NOT EXISTS actions_state_idx;
 DROP INDEX IF NOT EXISTS actions_session_id_idx;
 DROP INDEX IF NOT EXISTS actions_feature_name_idx;
-DROP INDEX IF NOT EXISTS actions_created_at_idx;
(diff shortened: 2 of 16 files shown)


session/sql_migration.go
```diff
@@ -650,10 +651,6 @@ func overrideSessionTimeZone(session *Session) {
 // as nil in the bbolt store. Therefore, we also override the permissions
 // or caveats to nil for the migrated session in that scenario, so that the
 // deep equals check does not fail in this scenario either.
-//
-// Additionally, we sort the caveats & permissions of both the kv and sql
-// sessions by their ID, so that they are always comparable in a deterministic
-// way with deep equals.
 func overrideMacaroonRecipe(kvSession *Session, migratedSession *Session) {
 	if kvSession.MacaroonRecipe != nil {
 		kvPerms := kvSession.MacaroonRecipe.Permissions
```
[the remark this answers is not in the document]
ViktorT-11: Updated it to set it to `migratedSession.MacaroonRecipe = &MacaroonRecipe{}` instead of returning if that happens (even though I think this case should be unreachable).

firewalldb/sql_migration.go
```diff
@@ -970,65 +980,43 @@ func migrateActionsToSQL(ctx context.Context, kvStore *bbolt.DB,
...
+			}
 
-				sessionID, err := session.IDFromBytes(macID)
-				if err != nil {
-					// This should be unreachable, as the
-					// macID should always be 4 bytes long.
-					return fmt.Errorf("invalid session ID "+
-						"format %x: %v", macID, err)
-				}
+			log.Tracef("Migrated Action: SeqNo: %x, Macaroon ID: +"+
```
ellemouton: stray "+"

db/sqlc/migrations/000003_kvstores.up.sql
```diff
@@ -50,8 +50,29 @@ CREATE TABLE IF NOT EXISTS kvstores (
...
     -- The value of the entry.
-    value BLOB NOT NULL
+    value BLOB NOT NULL,
+
+    -- Feature-scoped kv stores must always belong to a session group.
+    CHECK (feature_id IS NULL OR group_id IS NOT NULL)
 );
 
-CREATE UNIQUE INDEX IF NOT EXISTS kvstores_lookup_idx
-    ON kvstores (entry_key, rule_id, perm, group_id, feature_id);
```
ellemouton, line 57: think we have to update the drop files too (both to remove old one and add new ones?)

ellemouton (approved):
lgtm!

checking: do we test the "(created_at, id)" tie-breaker?

db/sqlc/migrations/000003_kvstores.up.sql
ViktorT-11: Ah yeah, good catch :pray:! Totally forgot, my bad

ViktorT-11:
Thanks for the review @ellemouton :pray:! I address your comments and also added some test coverage for actions with the same timestamp in the `firewalldb` migrations test.

I also addressed gemini's comments above as I do believe they make sense.

db/sqlc/migrations/000005_actions.down.sql
```diff
@@ -1,5 +1,5 @@
 DROP INDEX IF NOT EXISTS actions_state_idx;
 DROP INDEX IF NOT EXISTS actions_session_id_idx;
 DROP INDEX IF NOT EXISTS actions_feature_name_idx;
-DROP INDEX IF NOT EXISTS actions_created_at_idx;
+DROP INDEX IF NOT EXISTS actions_created_at_id_idx;
```
bitromortac: preexisting: `IF NOT EXISTS` -> `IF EXISTS`

session/sql_migration.go
```diff
@@ -650,15 +651,18 @@ func overrideSessionTimeZone(session *Session) {
...
 		kvCaveats := kvSession.MacaroonRecipe.Caveats
 
+		// If the migratedSession.MacaroonRecipe is nil, we set it to
+		// an empty MacaroonRecipe, as that can be correct if kvPerms &
+		// kvCaveats.
```
bitromortac: `... correct when both kvPerms and kvCaveats are nil.`?

bitromortac (approved):
LGTM, great catches :pray:

db/sqlc/migrations/000005_actions.down.sql
ViktorT-11: Ah yeah, good catch!

session/sql_migration.go
ViktorT-11: Yes, my bad 😃

PR #1321, opened 2026-06-09
firewall: enable and obfuscate `faraday.ForwardingAbility`

bitromortac:
Enables and obfuscates the faraday `ForwardingAbility` endpoint from https://github.com/lightninglabs/faraday/pull/243.


firewall/privacy_mapper.go
```diff
@@ -609,6 +615,68 @@ func (p *PrivacyMapper) checkers(db firewalldb.PrivacyMapDB,
...
+				// pseudonym across RPCs, then decode the pseudo
+				// pubkey back to raw bytes for the wire.
+				pseudo, err := firewalldb.HideString(
+					ctx, tx, hex.EncodeToString(p),
+				)
+				if err != nil {
+					return err
+				}
+
+				pseudoBytes, err := hex.DecodeString(pseudo)
```
Cyberguru1: This would be nice if we could use the existing [HideBytes](https://github.com/lightninglabs/lightning-terminal/blob/1903010be566cbcf50defe6548e4526569989bd1/firewalldb/privacy_mapper.go#L208) as it does the same logic.

ViktorT-11 (approved):
Nice, LGTM :fire:

firewall/privacy_mapper_test.go
```diff
@@ -151,6 +151,27 @@ func TestPrivacyMapper(t *testing.T) {
 		msg                 proto.Message
 		expectedReplacement proto.Message
 	}{
+		{
+			name:    "ForwardingAbility Response",
+			uri:     "/frdrpc.FaradayServer/ForwardingAbility",
+			msgType: rpcperms.TypeResponse,
+			msg: &frdrpc.ForwardingAbilityResponse{
+				Peers: [][]byte{{0x01, 0x02, 0x03, 0x04}},
```
Cyberguru1, line 160: Would be nice to also have an additional test case covering multiple peers, something like `[PeerA, PeerB, PeerA]`, and verify the obfuscated indexed peer mapping.

Also, if you agree, it would be good to add another test case where `privacyFlags` contains `session.ClearPubkeys`, which should assert that no obfuscation is applied in that scenario.

firewall/privacy_mapper.go
```diff
@@ -609,6 +615,68 @@ func (p *PrivacyMapper) checkers(db firewalldb.PrivacyMapDB,
...
+				if err != nil {
+					return err
+				}
+
+				peers[i] = pseudoBytes
+			}
+
+			return nil
+		})
```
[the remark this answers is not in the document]
Cyberguru1: I agree with point 1. We can return early when pubkey hiding is disabled or when there is nothing to process,

Cyberguru1 (commented):
@bitromortac nice addition to the firewall privacy :fire:. I’ve left a few comments and suggestions for improving test coverage.

firewall/privacy_mapper.go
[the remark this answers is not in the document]
bitromortac: Done.

firewall/privacy_mapper.go
bitromortac: Right, thanks!

firewall/privacy_mapper_test.go
bitromortac: Good call, have added those.

Cyberguru1 (approved):

bitromortac:
The itests fail because we can't connect to the mailbox (ALPN issue), a fix is in master. I had a successful itest run rebasing on master locally. The fix will land in a rebase of the feature branch on master in the feature branch merge PR.

PR #1322, opened 2026-06-09
accounts: don't mask payment errors when request values are absent

### LINKED ISSUE #963: LND Accounts: Improve error message for insufficient funds
**Expected behavior**

I am using `litd` with LND Accounts. I have a balance and am trying to pay an invoice for an amount larger than that.

I expect the `lncli payinvoice` request to fail with an error along the lines of:

`Insufficient amount. Invoice of Y sats is larger than balance of X sats`

**Actual behavior**

Instead, I get the following error message

`[lncli] rpc error: code = Unknown desc = no request values found for request: 29618`

**System information**

`litd version 0.14.0-alpha commit=v0.14.0-alpha-dirty`

Vandit1604:
Fixes #963.

### Problem

When an account payment fails, users can see a confusing error:

```
[lncli] rpc error: code = Unknown desc = no request values found for request: 29618
```

instead of lnd's actual failure reason (e.g. an insufficient balance).

### Cause

`SendPaymentV2`/`SendToRouteV2` are streaming endpoints, and lnd intercepts the request and every streamed response separately under one middleware request ID:

1. The request is checked and the request values are registered (`checkers.go` `RegisterValues`).
2. A terminal `Payment` response is handled and the request values are deleted (`checkSendResponse` -> `DeleteValues`).
3. lnd then sends a terminal error response for the same request. `erroredPaymentHandler` looks up the (already deleted) request values, finds none, and **aborts with `no request values found for request: <id>`**, replacing lnd's real error.

The two sibling response handlers that run in the same streaming context (`checkSendResponse` and `sendToRouteHTLCResponseHandler`) already treat a missing request-values entry as a no-op. `erroredPaymentHandler` was the only one missing that guard.

### Fix

When no request values are found there is no payment state to clean up, so the error handler now passes the original error through unchanged (`return nil, nil`) instead of masking it.

### Testing

Added a case to `TestSendPaymentV2` that replays the streaming sequence (register -> delete on failed response -> error response). It fails on `master` with the exact `no request values found for request: 6` error and passes with this change.

### DIFF
--- a/accounts/checkers.go
+++ b/accounts/checkers.go
@@ -668,10 +668,18 @@ func erroredPaymentHandler(service Service) mid.ErrorHandler {
 			return nil, err
 		}
 
+		// On streaming sends a call can return a response and then a
+		// terminal error, so the same request's error may be processed
+		// more than once. If the request values are already gone there
+		// is nothing left to clean up and we let the error pass
+		// through.
 		reqVals, ok := service.GetValues(reqID)
 		if !ok {
-			return nil, fmt.Errorf("no request values found for "+
-				"request: %d", reqID)
+			log.Tracef("No request values found for request: %d, "+
+				"passing the response error through unchanged",
+				reqID)
+
+			return nil, nil
 		}
 
 		log.Tracef("Handling payment request error for payment with "+
--- a/accounts/checkers_test.go
+++ b/accounts/checkers_test.go
@@ -677,6 +677,42 @@ func TestSendPaymentV2(t *testing.T) {
 	_, err = service.checkers.handleErrorResponse(ctx, uri, nil)
 	require.NoError(t, err)
 	assertBalance(acct.ID, 2000)
+
+	// Finally, replicate the streaming scenario. Since SendPaymentV2 is a
+	// streaming endpoint, the request values are deleted as soon as a
+	// terminal Payment response is handled. If lnd then sends a terminal
+	// error response for the same request, the error handler finds no
+	// request values. It must pass the original error through unchanged
+	// instead of masking it with a confusing "no request values found"
+	// error.
+	reqID5 := nextRequestID()
+	ctx = AddRequestIDToContext(ctxWithAcct, reqID5)
+
+	// Send a valid request so the request values are registered.
+	err = service.checkers.checkIncomingRequest(
+		ctx, uri, &routerrpc.SendPaymentRequest{
+			AmtMsat:     1000,
+			PaymentHash: testHash2[:],
+		},
+	)
+	require.NoError(t, err)
+
+	// A failed Payment response deletes the stored request values.
+	_, err = service.checkers.replaceOutgoingResponse(
+		ctx, uri, &lnrpc.Payment{
+			PaymentHash: hex.EncodeToString(testHash2[:]),
+			Status:      lnrpc.Payment_FAILED,
+		},
+	)
+	require.NoError(t, err)
+
+	// A subsequent error response for the same request must not be masked.
+	originalErr := fmt.Errorf("account balance insufficient")
+	returnedErr, err := service.checkers.handleErrorResponse(
+		ctx, uri, originalErr,
+	)
+	require.NoError(t, err)
+	require.ErrorContains(t, returnedErr, "account balance insufficient")
 }
 
 // TestSendToRouteV2 performs test coverage on the SendToRouteV2 checker.
--- a/itest/litd_accounts_test.go
+++ b/itest/litd_accounts_test.go
@@ -170,6 +170,12 @@ func runAccountSystemTest(t *harnessTest, node *HarnessNode, hostPort,
 		ctxm, t, rawConn, newAcctBalance, acctResp.Account.Id,
 	)
 
+	// Make sure a payment that the account checker rejects surfaces the
+	// real error instead of the masked "no request values found" error.
+	testAccountPaymentErrorPassthrough(
+		ctxa, t, rawConn, charlie, newAcctBalance,
+	)
+
 	// Clean up our channel and payments, so we can start the next test
 	// iteration with a clean slate.
 	closeChannelAndAssert(t, net, node, channelOp, false)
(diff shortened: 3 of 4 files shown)


accounts/checkers.go
```diff
@@ -668,10 +668,19 @@ func erroredPaymentHandler(service Service) mid.ErrorHandler {
 			return nil, err
 		}
 
+		// If no request values were registered for this request ID, then
+		// the request was rejected before it was ever forwarded to lnd
+		// (for example due to an insufficient account balance). In that
```
ViktorT-11: I'm not sure that the linked issue is referencing that the `account` has insufficient balance to pay the invoice, but rather that the "admin" `lnd` wallet itself has insufficient balance, while the account has more balance than what exists in the actual wallet.
I do think that if the account balance insufficient, the `checkSend` function will error and we'll actually never reach this `erroredPaymentHandler`.

I might be incorrect regarding the above though, so please see my next comment regarding adding an `itest` as that should make that clear.

I do think the fix you've included do solve the linked issue though, but it's just that the linked issue is referencing another issue than "insufficient account balance".

accounts/checkers_test.go
```diff
@@ -677,6 +677,44 @@ func TestSendPaymentV2(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+
+	// A subsequent error response for the same request must not be masked.
+	originalErr := fmt.Errorf("account balance insufficient")
+	returnedErr, err := service.checkers.handleErrorResponse(
+		ctx, uri, originalErr,
+	)
+	require.NoError(t, err)
+	require.ErrorContains(t, returnedErr, "account balance insufficient")
```
ViktorT-11, line 715: Instead of adding test coverage for the `handleErrorResponse` just through unit tests, the test coverage here would be much more sufficient if you instead added an `itest` for this in the `itest/litd_accounts_test.go` file.

In that test, you can test the behaviour if the `account` balance is insufficient and also a test for when the overall `lnd` wallet balance is insufficient but not the actual account balance. 

That will actually trigger and show how the `handleErrorResponse` handles this in a real world scenario :)

ViktorT-11 (commented):
Thank you very much for the contribution @Vandit1604 :pray::fire:!

In general, I think the fix you've included here does solve the linked issue, but the linked issue is referencing another type of insufficient balance. See my suggestions below, as I think those would add a lot of value!

Vandit1604:
Thanks @ViktorT-11, that makes sense.

You're right. If the account balance is too low, `checkSend` errors before the request is ever sent to lnd, so we never reach `erroredPaymentHandler` on that path. I've fixed the comment and the release note so they describe what actually happens: a streaming send where lnd returns a terminal error after the request values were already cleaned up.

I also added an itest (`testAccountPaymentErrorPassthrough`) for the account insufficient case. It checks that the caller gets the real "account balance insufficient" error and not the masked one, and that the balance stays the same.

For the "lnd wallet is low but the account is fine" case, I looked into it and ran into something. The masking only happens when lnd sends a terminal error after the payment values are already deleted. A normal failed or no-route payment comes back as a FAILED payment status, not a gRPC error, so it never hits that branch. A test like that would pass on master too, so it wouldn't really guard the fix.

Do you know a good way to force that terminal-error-after-cleanup case in the itest harness? If not, I think it's cleanest to keep the unit test as the deterministic guard for that branch and use the new itest for the account insufficient case.

accounts/checkers.go
```diff
@@ -668,10 +668,21 @@ func erroredPaymentHandler(service Service) mid.ErrorHandler {
...
 		}
 
+		// If no request values are found for this request ID, then there
+		// is no payment state left for us to clean up. This happens for
+		// the streaming send endpoints (SendPaymentV2/SendToRouteV2):
+		// the request values are deleted as soon as a terminal Payment
+		// response is handled, and lnd may then send a terminal error
+		// for the same request. In that case we pass the original error
+		// through unchanged instead of masking it with a confusing "no
+		// request values found" error.
```
ViktorT-11:
```suggestion
		// If no request values are found for this request ID, then there
		// is no payment state left for us to clean up. This happens for
		// the streaming send endpoints (SendPaymentV2/SendToRouteV2)
		// when lnd delivers a terminal stream error for a request that
		// no longer has any associated request values. That can happen
		// either because we already handled a terminal Payment response
		// and deleted the request values, or because the request was
		// rejected before any request values were registered. In either
		// case we pass the original error through unchanged instead of
		// masking it with a confusing "no request values found" error.
```

itest/litd_accounts_test.go
```diff
@@ -313,6 +319,58 @@ func testAccountRestrictions(ctxa context.Context, t *harnessTest,
...
+// testAccountPaymentErrorPassthrough verifies that a payment which the account
+// checker rejects because the account balance is insufficient surfaces a clear
+// account-balance error to the caller, and never the masked "no request values
+// found for request: <id>" error from issue #963.
+//
+// When the account balance is insufficient, checkSend errors before the request
+// is ever forwarded to lnd, so no request values are registered and the
+// erroredPaymentHandler is never reached on this path. This confirms that the
+// masking only ever occurred for streaming sends where lnd returns a terminal
+// error after the request values had already been cleaned up.
```
ViktorT-11:
```suggestion
// When the account balance is insufficient, checkSend errors before any request
// values are registered. But because SendPaymentV2 is a streaming RPC, lnd
// still reports that request-side rejection back through the stream's terminal
// error path, which then reaches erroredPaymentHandler without any stored
// request values. This confirms that the masking occurs on streaming sends
// whenever lnd emits a terminal error for a request that has no remaining
// request values, whether they were already cleaned up or were never
// registered.
```

itest/litd_accounts_test.go
```diff
@@ -313,6 +319,58 @@ func testAccountRestrictions(ctxa context.Context, t *harnessTest,
 	return initialAccountBalance + inboundPaymentAmt - outboundPaymentAmt
 }
 
+// testAccountPaymentErrorPassthrough verifies that a payment which the account
+// checker rejects because the account balance is insufficient surfaces a clear
+// account-balance error to the caller, and never the masked "no request values
+// found for request: <id>" error from issue #963.
```
ViktorT-11: We generally don't mention specific issue numbers in the code base (which you've done here and in a few other places), as those issues are github specific and not part of the codebase.

ViktorT-11 (commented):
Thanks for the changes @Vandit1604. I've actually now tested and double checked exactly how this works, and after testing I've noticed that your fix and referenced issue is actually solving the "account balance insufficient" for an account case, and not insufficient balance for the general wallet. Sorry for the confusion! 

The cause for the confusion and the reason why this is solving that, is that SendPaymentV2 is a streaming RPC. Therefore, when checkSend rejects the request, lnd still surfaces that rejection through the stream’s terminal error path, which then reaches erroredPaymentHandler. If this were not a streaming RPC, that request-side rejection would not go through erroredPaymentHandler in the same way.

I've left a few suggestions of how the docs can be improved below. Other than that, I think this PR should be good to go :rocket:! Thanks again :pray:

ViktorT-11:
Additionally, you can squash your commits into one commit, as this is a minor change :)

Vandit1604:
Thanks for the review! @ViktorT-11 
Updated the PR with squashed commit and your doc-polish suggestions.

ViktorT-11 (approved):
Thanks for the updates and for the contribution, tACK LGTM :fire:

Vandit1604:
@bitromortac - Gentle Review Ping

litbot-9000:
@bitromortac: review reminder

accounts/checkers.go
```diff
@@ -668,10 +668,24 @@ func erroredPaymentHandler(service Service) mid.ErrorHandler {
...
+		// there is no payment state left for us to clean up. This
+		// happens for the streaming send endpoints (SendPaymentV2/
+		// SendToRouteV2) when lnd delivers a terminal stream error
+		// for a request that no longer has any associated request
+		// values. That can happen either because we already handled
+		// a terminal Payment response and deleted the request
+		// values, or because the request was rejected before any
+		// request values were registered. In either case we pass
+		// the original error through unchanged instead of masking
+		// it with a confusing "no request values found" error.
```
bitromortac: imo this adds too much "legacy" explanation and is a burden when it comes to refactors. I think it would suffice if you'd describe that streaming errors can send back a reply and a terminal error such that we may process an "error" twice, such that we can hit this scenario, if I understand correctly.

Maybe something like:
```
		// On streaming sends a call can return a response and then a
		// terminal error, so the same request's error may be processed
		// more than once. If the request values are already gone there
		// is nothing left to clean up and we let the error pass
		// through.
```

But please check that the understanding is right. The itest docstring is very similar and should be updated as well.

bitromortac (commented):
Thank you for the fix :pray:. I'd only have a comment about some docstrings.

accounts/checkers.go
Vandit1604: Thanks for the review! Updated both docstrings with your suggested wording. Squashed and pushed.

bitromortac (approved):
LGTM :tada:. The test flakes are unrelated.

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -16,6 +16,14 @@
 
 ### Bug Fixes
 
+* [Don't mask account payment errors when request values are
```

bitromortac: And rebase on master to let CI pass.

Vandit1604:
Moved the release note to v0.17.1 and rebased on master. Thanks!

Vandit1604:
The macOS build failure is a flake. It hung in the `setup go 1.25.11` step for 46 minutes until the runner killed it, so it never reached the build. Linux and Windows built the same code fine, and the other 21 checks are green.

Could someone re-run that job? I can't.

PR #1324, opened 2026-06-11
terminal: auto-bake super macaroon on startup

### LINKED ISSUE #910: BakeSuperMacaroon on startup if it does not exist
Would like an easier way to bake a super macaroon.

Add a config option to litd that will allow a super macaroon to be baked automatically on startup if it does not already exist.

Use the gRPC or CLI (https://lightning.engineering/api-docs/api/lit/proxy/bake-super-macaroon/) to do it manually after first creating the node. This is annoying because the user needs to coordinate a one time task to determine whether the macaroon exists or not and deal with yet another macaroon to control litd to then create the super macaroon.

**Additional context**
When using taproot assets, you often need to use both lnd and tapd RPC's and having to manage multiple connections to the same litd instance with multiple macaroons is a nuisance.

When creating docker containers for litd, normally the one time tasks of doing container setup are done in the Dockerfile, but litd does not normally start and create the node data until after the docker container is started and appropriate volumes are mounted. With this paradigm in mind, all other macaroons are automatically created on startup if they do not exist, and the super macaroon is an exception that I don't think we should have to deal with.

Cyberguru1:
closes #910

This PR implements the ability to automatically bake a super macaroon on startup if configured and it doesn't already exist on disk, and validates/regenerates it dynamically if the active permissions change.


Cyberguru1:
I’ll fix the lint error after a follow-up review.

config.go
```diff
@@ -94,6 +94,14 @@ const (
 	// autogenerated lit macaroon.
 	DefaultMacaroonFilename = "lit.macaroon"
 
+	// DefaultSuperMacaroonFilename is the default file name for the
+	// autogenerated super macaroon.
+	DefaultSuperMacaroonFilename = "super.macaroon"
+
+	// defaultBakeSuperMacaroon is the default value for the
+	// bake-super-macaroon configuration option.
+	defaultBakeSuperMacaroon = "none"
```
ViktorT-11: Instead of naming the "none" value to "default", I would rather do it like this:

```
noneChoice = "none"

// defaultBakeSuperMacaroon is the default value for the
// bake-super-macaroon configuration option.
defaultBakeSuperMacaroon = noneChoice
```

That way we can add checking of the `none` value in the rest of the codebase without it breaking if we change the "default" value.

config.go
```diff
@@ -727,6 +746,39 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
...
+	switch cfg.BakeSuperMacaroon {
+	case defaultBakeSuperMacaroon, "read-only", "read-write":
+		// valid, do nothing
+	default:
+		return nil, fmt.Errorf("invalid value for "+
+			"--bake-super-macaroon: %q, must be one of: "+
+			"none, read-only, read-write",
+			cfg.BakeSuperMacaroon,
+		)
+	}
```
ViktorT-11: I think this check is redundant, as it's handled automatically by setting the `bake-super-macaroon` to a "choice" config option.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+			}
+		}
+	}
+
+	return true, nil
+}
+
+// setupSuperMacaroon bakes a super macaroon and writes it to disk if needed.
+func (g *LightningTerminal) setupSuperMacaroon(ctx context.Context) error {
+	if g.cfg.BakeSuperMacaroon == defaultBakeSuperMacaroon {
```
ViktorT-11: Switch the `defaultBakeSuperMacaroon` to the `noneChoice` instead here, as we may change the `defaultBakeSuperMacaroon` value in the future, which would break the logic of this function.

ViktorT-11: It'd also add value to add a comment that here in this early return that we won't bake a macaroon when the choice is `none`.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+		return nil
+	}
+
+	if g.cfg.statelessInitMode {
+		log.Infof("Stateless init mode is active, skipping writing " +
+			"super macaroon to disk")
+		return nil
+	}
```
ViktorT-11, line 2217: I actually think it make sense to check this in the config validation instead of here, and fail the validation if `g.cfg.BakeSuperMacaroon != noneChoice & g.cfg.statelessInitMode`.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+
+	path := g.cfg.SuperMacaroonPath
+
+	readOnly := g.cfg.BakeSuperMacaroon == "read-only"
```
ViktorT-11: Also make a `const readOnlyChoice = "read-only"` for this string value the `config.go` file, just like the `noneChoice`.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+
+	if superMacaroonExists(path) {
+		matches, err := macaroonMatchesPermissions(
+			path, activePerms,
+		)
+
+		if err == nil && matches {
+			log.Infof("Super macaroon already exists at "+
+				"%v and matches configuration, "+
+				"skipping bake", path)
```
ViktorT-11: I think it make sense to change this to a `log.Debugf`.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+			return nil
```
ViktorT-11, line 2234: nit: new line before the return statement.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+		}
+		if err != nil {
+			log.Warnf("Unable to verify super "+
+				"macaroon permissions: %v", err)
```
ViktorT-11: I think we should return an error instead of just logging?
It'd maybe also be valuable to inform the user to delete the macaroon at the specified path if the issue persists on next startup.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+	superMacBytes, err := hex.DecodeString(superMacHex)
+	if err != nil {
+		return fmt.Errorf("unable to decode baked super macaroon: %w", err)
+	}
+
+	// Write the macaroon to disk.
+	if err := os.WriteFile(path, superMacBytes, 0600); err != nil {
+		return fmt.Errorf("unable to write super macaroon to %v: %w",
+			path, err)
+	}
```
ViktorT-11: I think it makes sense to move this part of the code into the `macaroons/super_mac.go` file, along with the `macaroonMatchesPermissions` & `superMacaroonExists` functions.

terminal.go
```diff
@@ -2159,3 +2167,151 @@ func randId(n int) string {
...
+	if _, err := os.Stat(path); err != nil {
+		return false
+	}
+
+	return true
+}
+
+// macaroonMatchesPermissions checks if the macaroon at the given path contains
+// exactly the expected permissions (no more and no less).
+func macaroonMatchesPermissions(path string,
```
ViktorT-11: If the `lndclient` version of this function would satisfy your needs, I think you could open a PR in the `lndclient` repo that exports this function, and then update `litd` to use that function instead once it's been merged. Fine to keep this as is for now though :)!

itest/litd_mode_integrated_test.go
```diff
@@ -1481,3 +1482,157 @@ func bakeSuperMacaroon(t *testing.T, cfg *LitNodeConfig,
...
+			WithLitArg("loop-mode", "disable"),
+			WithLitArg("pool-mode", "disable"),
+			WithLitArg("faraday-mode", "disable"),
+		},
+	)
+	require.NoError(t.t, err)
+	net.ConnectNodes(t.t, net.Alice, net.Bob)
+
+	require.FileExists(t.t, superMacPath)
+
```
ViktorT-11, line 1705: This test should verify that the disabling of the different sub-servers means that the permissions for the different sub-servers were also removed from the super macaroon, and then verify once more that they're then present once they've been re-enabled in the next startup below.

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -85,6 +85,10 @@
   `litcli accounts update debit` and `litcli accounts update credit` commands
   to modify an account's balance.
 
+* [Auto-bake super macaroon on startup](https://github.com/lightninglabs/lightning-terminal/pull/1324):
+  Added config options `--bake-super-macaroon` (choice: `none`, `read-only`, `read-write`) and `--super-macaroon-path` to
+  automatically bake a super macaroon on startup and keep its permissions in sync.
```
ViktorT-11: nit: line length.
I also think you should briefly explain what the functionality does once `read-only` & `read-write` is set.

ViktorT-11 (commented):
Awesome, thanks for this @Cyberguru1 :tada:! Generally this looks great, but leaving some suggestions below for some additional improvements :fire:

itest/litd_mode_integrated_test.go
Cyberguru1: Good call, added them

Cyberguru1:
Thank you for taking the time to review @ViktorT-11 :pray: , I have addressed your feedbacks.

terminal.go
Cyberguru1: Sounds good, I’ll consider that as a follow-up improvement and open a PR in `lndclient`.

terminal.go
Cyberguru1: I think the check for `statelessInitMode` wouldn’t be possible during config validation since it isn’t a static config value. It gets determined [dynamically](https://github.com/lightninglabs/lightning-terminal/blob/1903010be566cbcf50defe6548e4526569989bd1/terminal.go#L917) at runtime during startup, after connecting to LND.

config.go
```diff
@@ -201,6 +223,9 @@ type Config struct {
 
 	MacaroonPath string `long:"macaroonpath" description:"Path to write the macaroon for litd's RPC and REST services if it doesn't exist."`
 
+	BakeSuperMacaroon string `long:"bake-super-macaroon" description:"Bake a super macaroon on startup if it doesn't exist." choice:"none" choice:"read-only" choice:"read-write"`
+	SuperMacaroonPath string `long:"super-macaroon-path" description:"Path to write the auto-baked super macaroon."`
```
ViktorT-11: I think we should clarify in the description for the `SuperMacaroonPath`, that the value for this config option should be the directory to store macaroon in as well as the name of the macaroon file that will be generated. We should also clarify that the macaroon file should be suffixed with `.macaroon`. 

We should also validate in the config validation that the value of this field ends with the `.macaroon` suffix if set.

terminal.go
```diff
@@ -789,6 +789,21 @@ func (g *LightningTerminal) start(ctx context.Context) error {
 		return fmt.Errorf("could not start litd sub-servers: %v", err)
 	}
 
+	// If the super macaroon baking option is enabled, we cannot run in
+	// stateless initialization mode as it won't write any macaroons to the
+	// filesystem.
+	if g.cfg.BakeSuperMacaroon != noneChoice && g.cfg.statelessInitMode {
+		return fmt.Errorf("cannot use bake-super-macaroon " +
+			"with stateless-init mode")
+	}
```
ViktorT-11: I think it makes sense to move this into the `setupSuperMacaroon` function.

ViktorT-11 (commented):
Thanks for the updates @Cyberguru1 :pray:!

This looks great and I've tested locally that this works as expected :fire:! I think this looks good to go from my end once you've fixed the merge conflicts & addressed some minor comments I've added below.

config.go
Cyberguru1: Agreed, updated the description and added validation for it.

terminal.go
Cyberguru1: I've moved the check.

terminal.go
```diff
@@ -2172,3 +2179,63 @@ func randId(n int) string {
...
+				"skipping bake", path)
+
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf(
+				"unable to verify super macaroon permissions: "+
+					"%w; please delete the macaroon at %v "+
+					"if the issue persists", err, path,
+			)
```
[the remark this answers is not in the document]
ViktorT-11: This is valid feedback to address.

macaroons/super_mac.go
```diff
@@ -118,3 +122,123 @@ func BakeSuperMacaroon(ctx context.Context, lnd lnrpc.LightningClient,
...
+			path, err)
+	}
+
+	return nil
+}
+
+// ValidateSuperMacaroonPath checks that the super macaroon path ends with the
+// expected suffix.
+func ValidateSuperMacaroonPath(path string) error {
+	if path != "" && !strings.HasSuffix(path, ".macaroon") {
```
ViktorT-11: Hmmm I'm not sure I understand the motivation for allowing the `path` to be an empty string (`""`)? Shouldn't that be an invalid MacaroonPath? Maybe the name of this function should also be changed to something like: "HasMacaroonSuffix".

ViktorT-11 (approved):
Nice! Thanks a lot for this @Cyberguru1, it's really appreciated that you keep contributing :pray:!

Leaving some minor nonblocking feedback below.

tACK LGTM :fire:

itest/litd_mode_integrated_test.go
```diff
@@ -1481,3 +1482,224 @@ func bakeSuperMacaroon(t *testing.T, cfg *LitNodeConfig,
...
+	// Restart Alice with read-only super macaroon baking enabled.
+	err = net.RestartNode(
+		net.Alice, nil, []LitArgOption{
+			WithLitArg("bake-super-macaroon", "read-only"),
+			WithLitArg("super-macaroon-path", superMacPath),
+		},
+	)
+	require.NoError(t.t, err)
+
+	net.ConnectNodes(t.t, net.Alice, net.Bob)
```
bitromortac: These `ConnectNodes` calls aren't needed in this itest, except for the last one, I think. This could save a few seconds.

bitromortac (approved):
LGTM, nice addition :pray::tada:

macaroons/super_mac.go
Cyberguru1: Good catch!, leaving it empty would error out, added a check for it and updated the function naming.

itest/litd_mode_integrated_test.go
Cyberguru1: Agreed. cleaned the other entries and left the last one.

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -85,6 +85,15 @@
   `litcli accounts update debit` and `litcli accounts update credit` commands
   to modify an account's balance.
 
+* [Auto-bake super macaroon on startup](https://github.com/lightninglabs/lightning-terminal/pull/1324):
```

Cyberguru1: updated this

litbot-9000:
@cyberguru1, remember to re-request review from reviewers when ready

ViktorT-11 (approved):
Thanks again for this @Cyberguru1 :pray:! I think this is ready to be merged once you've fixed the merge conflict in the release notes.

Cyberguru1:
> Thanks again for this @Cyberguru1 🙏! I think this is ready to be merged once you've fixed the merge conflict in the release notes.

Thanks again for the review and for the suggestion! 🙏

I've rebased the branch and resolved the merge conflicts.

Issue #1098, opened 2025-06-23
add `sample-litd.conf`


ZZiigguurraatt:
There is no `sample-litd.conf` in this repository. Other projects like lnd and taproot-assets have `sample-lnd.conf` and `sample-tapd.conf` respectively.

We can run `litd -h` to get some options, but the sample files are often nice to review before you even download and install the litd binary so you can see what the options are that you could set. Also, I'm not sure where someone even finds out about the `litd -h` command line option.

Running `litd -h` does give a bunch of stuff that is carryover from all the subservers, but I think `sample-litd.conf` should only include the options that are litd specific and then reference `sample-lnd.conf`, `sample-tapd.conf`, etc. for the rest of the samples.

victorabarros:
I hope this is helpful! 🤗 
https://github.com/lightninglabs/lightning-terminal/pull/1325

PR #1325, opened 2026-06-11
docs: add sample-litd.conf and reference it in README

victorabarros:
## What

Issue https://github.com/lightninglabs/lightning-terminal/issues/1098
Introduces `sample-litd.conf` — a comprehensive, well-commented configuration template for `litd` — and updates the README to point users to it.

## Why

There was no single reference file for `litd`-native configuration options. Users had to dig through `--help` output or source code to understand what settings exist, what they do, and what the defaults are. This fills that gap.

## Changes

- **`sample-litd.conf`** (new): 372-line annotated config covering application settings, TLS, Let's Encrypt, daemon modes, and more. Follows the same convention as `sample-lnd.conf` and `sample-tapd.conf` in the broader Lightning ecosystem.
- **`README.md`**: Adds a note directing users to `sample-litd.conf` as the canonical reference for litd-native config options, including descriptions and defaults.

> Note: This file intentionally covers only litd-native options. Sub-daemon configuration (lnd, taproot-assets, loop, pool, faraday) is delegated to their respective sample config files, with links provided inline.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -216,6 +216,26 @@ jobs:
         run: make sqlc-check
 
   ########################
+  # sample conf check
+  ########################
+  sample-conf-check:
+    name: Check sample-litd.conf is valid
+    runs-on: ubuntu-latest
+    steps:
+      - name: git checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: setup go ${{ env.GO_VERSION }}
+        uses: ./.github/actions/setup-go
+        with:
+          go-version: '${{ env.GO_VERSION }}'
+
+      - name: check sample-litd.conf options
+        run: make sample-conf-check
+
+  ########################
   # lint code
   ########################
   lint:
--- a/Makefile
+++ b/Makefile
@@ -281,6 +281,15 @@ fmt: $(GOIMPORTS_BIN)
 	@$(call print, "Formatting source.")
 	gofmt -l -w -s $(GOFILES_NOVENDOR)
 
+sample-conf-check:
+	@$(call print, "Checking sample-litd.conf options are valid")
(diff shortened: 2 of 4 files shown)


ViktorT-11 (commented):
Thanks for the contribution @victorabarros :pray:!

This could be a nice contribution. Since you're adding a `sample_conf` it would also be valuable if you additionally add a CI job that checks that it's been correctly updated with the new config options when such are added. `lnd` does have such a CI job, so I recommend looking into how it is implemented there and try to mimic that:
https://github.com/lightningnetwork/lnd/blob/1a6ecebcbf490de5e8932c782b0d57c5d57c5679/.github/workflows/main.yml#L67

victorabarros:
> Thanks for the contribution @victorabarros 🙏!
> 
> This could be a nice contribution. Since you're adding a `sample_conf` it would also be valuable if you additionally add a CI job that checks that it's been correctly updated with the new config options when such are added. `lnd` does have such a CI job, so I recommend looking into how it is implemented there and try to mimic that: https://github.com/lightningnetwork/lnd/blob/1a6ecebcbf490de5e8932c782b0d57c5d57c5679/.github/workflows/main.yml#L67

Sure. Right now I'm AFK, but I can do that by EOD.

victorabarros:
@ViktorT-11 
Can you review, pls? 🤗

ViktorT-11 (commented):
Thanks for the updates @victorabarros!

I've tested your changes, and the added `make sample-conf-check` currently correctly identifies if an option/flag in the `sample-litd.conf` file exists which isn't an actual config option/flag, so that's great!

However, the script currently doesn't correctly identify if `litd` has a config option/flag which doesn't have an entry in the `sample-litd.conf` file. This differs from the version in `lnd` which would identify that missing entry. You can test this easily by removing a config option/flag from the `sample-litd.conf` file , and then running the `make sample-conf-check` command.

Please update that so `litd` also has that check. I'll have another look once that's been updated :fire:

litbot-9000:
@victorabarros, remember to re-request review from reviewers when ready

PR #1326, opened 2026-06-11
[sql-75] Don't create kvdb database files in kvdb -> SQL migration if they don't exist

ViktorT-11:
This PR adds an extra check in the beginning of the kvdb -> SQL migration, which checks if any `bbolt` database files exist before we start the migration.

If they do not exists, we just skip the migration as there's nothing to migrate. Previously, the migration would create empty bbolt database files if that was the case.

NOTE: This new logic only applies if none of the `bbolt` database files exists. In case for example only 1 of the 3 files does actually exist, the logic remains the same as previously, and we'll create the remaining 2 files before executing the migration.

### DIFF
--- a/db/migsets/programmatic_migrations_test.go
+++ b/db/migsets/programmatic_migrations_test.go
@@ -1,0 +1,152 @@
+package migsets
+
+import (
+	"context"
+	"net"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/lightninglabs/lightning-terminal/accounts"
+	"github.com/lightninglabs/lightning-terminal/db"
+	"github.com/lightninglabs/lightning-terminal/db/sqlcmig6"
+	"github.com/lightninglabs/lightning-terminal/firewalldb"
+	"github.com/lightninglabs/lightning-terminal/session"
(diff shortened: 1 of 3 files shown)


db/migsets/programmatic_migrations_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	dbSessions, err := queries.ListSessions(ctx)
+	require.NoError(t, err)
+	require.Empty(t, dbSessions)
+}
+
+// TestKVDBToSQLProgrammaticMigrationRunsWith2BBoltDBFiles verifies that
+// the migration still runs when two legacy kvdb files exist and migrates both
+// stores while creating the missing rules database file.
+func TestKVDBToSQLProgrammaticMigrationRunsWithTwoBBoltDBFiles(t *testing.T) {
```
bitromortac: nit: feels like this test isn't needed

bitromortac (approved):
LGTM :zap:

ellemouton (approved):

db/migsets/programmatic_migrations_test.go
ViktorT-11: Removed the test.

PR #1327, opened 2026-06-11
[sql-76] config: block startup on mixed SQL backends

ViktorT-11:
Reject litd startup when the selected SQL backend conflicts with existing data for the other backend.

When starting with the `databasebackend` config flag set to  `postgres`, we add a check if an `sqlite` database file exists under the default path and fail the startup if it does. When starting with `sqlite`, we instead add a check if config for `postgres` database has also been set, and check if an actual database exists on those configured `postgres` values. 

This prevents accidental backend switches from one SQL backend to the other. This becomes extra important as allowing this will retrigger the kvdb->SQL migration for the new configured sql database. As the data in the would already have been migrated if that'd be the case, the data outdated and running with that outdated data can cause issues. Additionally, it would create 2 separate states for the data between the SQL databases, which we could never recover from as we wouldn't support migration from one of the SQL databases to the other.

### DIFF
--- a/config.go
+++ b/config.go
@@ -578,7 +578,9 @@ func defaultConfig() *Config {
 
 // loadAndValidateConfig loads the terminal's main configuration and validates
 // its content.
-func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
+func loadAndValidateConfig(ctx context.Context,
+	interceptor signal.Interceptor) (*Config, error) {
+
 	// Start with the default configuration.
 	preCfg := defaultConfig()
 
@@ -729,7 +731,7 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 		)
 	}
 
-	err = validateExclusiveSQLBackends(cfg, litDir)
+	err = validateExclusiveSQLBackends(ctx, cfg, litDir)
 	if err != nil {
 		return nil, err
 	}
@@ -1160,7 +1162,9 @@ func readAutoMigrateKVDB(config *Config) error {
 // inactive SQL backend still has data at its default location. This prevents
 // silently switching to a different SQL store and starting against an empty
 // database.
-func validateExclusiveSQLBackends(cfg *Config, litDir string) error {
+func validateExclusiveSQLBackends(ctx context.Context, cfg *Config,
+	litDir string) error {
+
 	switch cfg.DatabaseBackend {
 	case DatabaseBackendPostgres:
 		sqlitePath := filepath.Join(
@@ -1184,7 +1188,7 @@ func validateExclusiveSQLBackends(cfg *Config, litDir string) error {
 		}
 
 	case DatabaseBackendSqlite:
-		exists, err := postgresDatabaseExists(cfg.Postgres)
+		exists, err := postgresDatabaseExists(ctx, cfg.Postgres)
 		if err != nil {
 			return fmt.Errorf("unable to check for existing "+
 				"postgres database %q at %s:%d for user %q. "+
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -1149,6 +1156,116 @@ func readAutoMigrateKVDB(config *Config) error {
...
+	}
+
+	dbConn, err := sql.Open("postgres", cfg.DSN(false))
+	if err != nil {
+		return false, fmt.Errorf("unable to check for existing "+
+			"postgres database %q: %w", cfg.DBName, err)
+	}
+	defer dbConn.Close()
+
+	err = dbConn.Ping()
```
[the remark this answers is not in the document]
ViktorT-11: Will add this after the next review round

config.go
```diff
@@ -3,6 +3,7 @@ package terminal
 import (
 	"context"
 	"crypto/tls"
+	"database/sql"
```
bitromortac, line 6: nit: commit message needs a small pass to read clearly

config.go
```diff
@@ -1149,6 +1156,116 @@ func readAutoMigrateKVDB(config *Config) error {
...
+				"sqlite database file at %s: %w", sqlitePath,
+				err)
+		}
+		if exists {
+			return fmt.Errorf("cannot start litd with postgres "+
+				"backend: sqlite database file already exists "+
+				"at %s. If you really intend to switch to "+
+				"postgres you must delete the sqlite "+
+				"database before restarting with postgres "+
+				"again", sqlitePath)
```
bitromortac: nit: obviously the data will be gone

config_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	fixture := db.NewTestPgFixture(
+		t, db.DefaultPostgresFixtureLifetime, true,
+	)
+	t.Cleanup(func() {
+		fixture.TearDown(t)
+	})
+
+	require.NoError(t, os.MkdirAll(filepath.Dir(sqlitePath), 0700))
+	// NOTE: we don't write any file to the sqlite path here, so the sqlite
```
bitromortac, line 135: nit: new line

bitromortac (approved):
LGTM :zap:

ellemouton (approved):

config.go
ViktorT-11: Clarified the commit message a bit :)

config.go
ViktorT-11: Added a clarifying comment regarding that.

config.go
```diff
@@ -1149,6 +1156,117 @@ func readAutoMigrateKVDB(config *Config) error {
...
+	}
+
+	dbConn, err := sql.Open("postgres", cfg.DSN(false))
+	if err != nil {
+		return false, fmt.Errorf("unable to check for existing "+
+			"postgres database %q: %w", cfg.DBName, err)
+	}
+	defer dbConn.Close()
+
+	err = dbConn.Ping()
```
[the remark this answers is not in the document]
ViktorT-11: Addressed

config.go
```diff
@@ -1149,6 +1156,117 @@ func readAutoMigrateKVDB(config *Config) error {
...
+		if exists {
+			return fmt.Errorf("cannot start litd with sqlite "+
+				"backend: postgres database %q already exists "+
+				"at %s:%d for user %q; If you really intend "+
+				"to switch to sqlite you must delete the "+
+				"postgres database (effectively deleting all "+
+				"stored data) before restarting with sqlite "+
+				"again", cfg.Postgres.DBName, cfg.Postgres.Host,
+				cfg.Postgres.Port, cfg.Postgres.User)
+		}
```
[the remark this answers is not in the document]
ViktorT-11: I don't agree with this, as the action the user should take then is to remove the `postgres` database configuration. I clarified the error to make that clear to the user.

config.go
```diff
@@ -1149,6 +1158,130 @@ func readAutoMigrateKVDB(config *Config) error {
...
```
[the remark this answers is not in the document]
ViktorT-11: See https://github.com/lightninglabs/lightning-terminal/pull/1327#discussion_r3423576915

PR #1329, opened 2026-06-12
Keep litd running after lnd stops

### LINKED ISSUE #1201: Decouple litd and lnd Shutdown Interceptors
## Summary

Currently, litd and lnd share a single `signal.Interceptor` instance. This tight coupling means that any shutdown of lnd (whether intentional or due to errors) cascades to shut down litd entirely. This prevents litd from maintaining its status endpoint after lnd stops, reducing observability during failure scenarios.

This issue proposes decoupling the shutdown interceptors so that litd can continue running (with its status endpoint available) even after lnd has stopped.

### Current Architecture

litd creates a single `signal.Interceptor` and passes it to lnd:

```go
// terminal.go:269-270
shutdownInterceptor, err := signal.Intercept()

// terminal.go:579
err := lnd.Main(g.cfg.Lnd, lisCfg, implCfg, interceptor)
```

The interceptor is also stored as a package-level variable for use by the logger and RPC handlers:

```go
// log.go:33-35
var interceptor signal.Interceptor
```

### The Problem

When lnd shuts down (via `StopDaemon` RPC or any other means), it calls `interceptor.RequestShutdown()` on the **shared** interceptor. This closes the `ShutdownChannel()` that litd is waiting on:

```go
// terminal.go:435-437
<-shutdownInterceptor.ShutdownChannel()
log.Infof("Shutdown signal received")
```

**Result**: litd exits, and its status endpoint becomes unavailable.

### Why This Matters

In PR #1183, we added fail-fast behavior for critical sub-server startup failures (e.g., tapd). When tapd fails to start, we call `StopDaemon` on lnd to shut it down promptly. However, due to the shared interceptor, this also shuts down litd entirely.

Ideally, litd's status endpoint should remain available after such failures so operators can:
- Query the status endpoint to understand what failed
- Retrieve error details programmatically
- Monitor the system state via external tools

### High-Level Approach

1. Create a **separate interceptor for lnd** that doesn't cascade to litd's shutdown
2. Have litd **monitor lnd's lifecycle independently** via the existing `lndQuit` channel
3. Keep litd running after lnd stops, with the status endpoint reporting the failure state
4. Only shut down litd when:
   - An OS signal is received (SIGTERM/SIGINT)
   - The user explicitly calls litd's `StopDaemon` RPC
   - A fatal litd-specific error occurs

#### Step 1: Create a Custom Interceptor Wrapper

Create a wrapper that implements the same interface as `signal.Interceptor` but allows litd to control whether lnd's shutdown propagates:

```go
// interceptor.go (new file)
package terminal

import (
    "github.com/lightningnetwork/lnd/signal"
)

// LndInterceptor wraps a signal.Interceptor for lnd, allowing litd to
// monitor lnd shutdown without being forced to shut down itself.
type LndInterceptor struct {
    signal.Interceptor

    // lndShutdownChan is closed when lnd requests shutdown.
    lndShutdownChan chan struct{}

    // litdInterceptor is the main litd interceptor, only used for
    // OS signal propagation.
    litdInterceptor signal.Interceptor
}

// NewLndInterceptor creates a new interceptor for lnd that doesn't
// cascade shutdown to litd.
func NewLndInterceptor(litdInterceptor signal.Interceptor) *LndInterceptor {
    return &LndInterceptor{
        Interceptor:     litdInterceptor, // Inherit OS signal handling
        lndShutdownChan: make(chan struct{}),
        litdInterceptor: litdInterceptor,
    }
}

// RequestShutdown is called by lnd when it wants to shut down.
// We close our local channel but don't propagate to litd.
func (i *LndInterceptor) RequestShutdown() {
    select {
    case <-i.lndShutdownChan:
        // Already closed
    default:
        close(i.lndShutdownChan)
    }
}

// ShutdownChannel returns a channel that's closed when lnd should shut down.
// This is triggered by OS signals OR lnd's internal shutdown request.
func (i *LndInterceptor) ShutdownChannel() <-chan struct{} {
    // Merge OS signal channel with lnd-specific shutdown channel
    merged := make(chan struct{})
    go func() {
        select {
        case <-i.litdInterceptor.ShutdownChannel():
            close(merged)
        case <-i.lndShutdownChan:
            close(merged)
        }
    }()
    return merged
}

// LndShutdownChannel returns a channel that's closed only when lnd
// requests shutdown (not on OS signals).
func (i *LndInterceptor) LndShutdownChannel() <-chan struct{} {
    return i.lndShutdownChan
}
```

#### Step 2: Modify terminal.go to Use Separate Interceptors

```go
// terminal.go - Run() method

// Create litd's main interceptor for OS signals
shutdownInterceptor, err := signal.Intercept()
if err != nil {
    return fmt.Errorf("could not intercept signals: %v", err)
}

// Create a separate interceptor for lnd
lndInterceptor := NewLndInterceptor(shutdownInterceptor)

// ... later, when starting lnd ...

err := lnd.Main(g.cfg.Lnd, lisCfg, implCfg, lndInterceptor)
```

#### Step 3: Update the Main Event Loop

Modify the shutdown handling to distinguish between lnd shutdown and litd shutdown:

```go
// terminal.go - after start() returns

select {
case <-shutdownInterceptor.ShutdownChannel():
    // OS signal or litd StopDaemon - shut down everything
    log.Infof("Shutdown signal received, stopping all services")

case <-lndInterceptor.LndShutdownChannel():
    // lnd stopped but litd should keep running
    log.Infof("LND has stopped, litd status endpoint remains available")

    // Keep running until explicit shutdown
    <-shutdownInterceptor.ShutdownChannel()
    log.Infof("Shutdown signal received")
}

err = g.shutdownSubServers()
```

#### Step 4: Update the Package-Level Interceptor in log.go

The package-level `interceptor` variable in `log.go` is used for critical error shutdown. This should continue to use litd's interceptor:

```go
// log.go - no changes needed to the variable itself, but ensure
// SetupLoggers receives litd's interceptor, not lnd's
```

#### Step 5: Update rpc_proxy.go StopDaemon

The litd `StopDaemon` RPC should continue to shut down everything:

```go
// rpc_proxy.go - no changes needed
// interceptor.RequestShutdown() still uses litd's interceptor
```

#### Step 6: Handle lnd Shutdown in Status Manager

When lnd stops, update the status manager to reflect this:

```go
case <-lndQuit:
    g.statusMgr.SetErrored(
        subservers.LND, "lnd has stopped",
    )
    // Don't return error - keep litd running
```

### Files to Modify

| File | Changes |
|------|---------|
| `terminal.go` | Create separate lnd interceptor, update event loop |
| `interceptor.go` | New file - LndInterceptor wrapper |
| `log.go` | Ensure it uses litd's interceptor (may need verification) |
| `rpc_proxy.go` | Verify it uses litd's interceptor |
| `config.go` | Update `loadAndValidateConfig` to accept litd interceptor |

### Testing

1. **Unit tests**: Test the `LndInterceptor` wrapper in isolation
2. **Integration tests**:
   - Start litd in integrated mode
   - Cause lnd to shut down (e.g., via `StopDaemon`)
   - Verify litd's status endpoint is still accessible
   - Verify status endpoint reports lnd as stopped/errored
   - Verify litd shuts down cleanly on SIGTERM

#### OS Signal Handling

The `signal.Intercept()` function registers handlers for SIGINT/SIGTERM. We need to ensure:
- Only one interceptor registers OS signal handlers (litd's)
- lnd's interceptor inherits OS signal notifications from litd's

#### Graceful Degradation

When lnd stops but litd continues:
- Sub-servers that depend on lnd will fail
- Status endpoint should clearly indicate the degraded state
- Consider adding a "degraded" status in addition to "running"/"stopped"/"errored"

#### Remote Mode

This change primarily affects integrated mode. Remote mode already has lnd running separately, so the interceptor coupling doesn't apply.

#### Backward Compatibility

- External behavior change: litd will no longer exit when lnd stops in integrated mode
- Operators who rely on litd exiting when lnd stops may need to update their monitoring
- Consider a config flag to preserve the old behavior if needed

## Acceptance Criteria

- [ ] litd continues running after lnd stops in integrated mode
- [ ] Status endpoint remains accessible and reports lnd's stopped state
- [ ] OS signals (SIGTERM/SIGINT) still shut down both lnd and litd
- [ ] litd's `StopDaemon` RPC still shuts down both lnd and litd
- [ ] No regression in remote mode behavior
- [ ] Integration tests cover the new behavior
- [ ] Documentation updated to reflect the change

Vandit1604:
Closes #1201 (initial implementation).

Makes litd keep running after lnd stops in integrated mode, so the status endpoint stays up to report the failure instead of litd exiting along with lnd.

Based on #1356, which splits out the itest harness change. The bottom two commits here belong to that PR.

### Why not the wrapper from the issue

The pinned lnd's `signal.Intercept` can only be called once, its channels are unexported, and `lnd.Main` takes the interceptor by value, so a wrapper type can't be passed in. Instead litd owns its own shutdown source and stops tying its lifecycle to lnd's interceptor.

### What it does

- litd waits on its own shutdown source (its own SIGINT/SIGTERM handler plus a programmatic channel). lnd still gets the real interceptor.
- When lnd stops after startup, litd marks lnd as errored, tears down the lnd dependent sub servers (loop, pool, tapd, faraday, accounts, autopilot), and keeps running.
- OS signals, litd `StopDaemon`, and critical litd errors still stop everything. litd `StopDaemon` also stops the embedded lnd.

### Behaviour change to flag

Calling lnd's `StopDaemon` (for example `lncli stop`) no longer stops litd in integrated mode. That is the goal of the issue but it is user facing. The itest harness now stops litd through its own `StopDaemon`, and outside tooling that relied on the old cascade would need the same. I can add a config flag to keep the old behaviour if you prefer.

### Testing

- unit tests for the new shutdown source
- new itest `litd survives lnd shutdown`: stop lnd, then check the status endpoint still answers and reports lnd not running
- ran the integrated mode, stateless init mode and request logger itests locally to cover the harness change

### DIFF
--- a/itest/litd_lnd_shutdown_test.go
+++ b/itest/litd_lnd_shutdown_test.go
@@ -1,0 +1,129 @@
+package itest
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/lightninglabs/lightning-terminal/litrpc"
+	"github.com/lightninglabs/lightning-terminal/subservers"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/lntest/wait"
+	"github.com/stretchr/testify/require"
+)
+
+// testLitdSurvivesLndShutdown asserts that when lnd stops in integrated mode,
+// litd keeps running and its status endpoint stays reachable, reporting lnd as
+// stopped rather than cascading to a full litd shutdown.
+func testLitdSurvivesLndShutdown(ctx context.Context, net *NetworkHarness,
+	t *harnessTest) {
+
+	// Use a dedicated, throwaway node so that stopping its lnd cannot
+	// affect any other test.
+	node, err := net.NewNode(
+		t.t, "lnd-stop", nil, false, true, "--autopilot.disable",
+	)
+	require.NoError(t.t, err)
+
+	// The harness ShutdownNode stops litd (via its StopDaemon) and waits
+	// for the process to exit, so no manual litd shutdown is needed here.
+	defer func() {
+		_ = net.ShutdownNode(node)
+	}()
+
+	ctxt, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+
+	rawConn, err := connectLitRPC(
+		ctxt, node.Cfg.LitAddr(), node.Cfg.LitTLSCertPath,
+		node.Cfg.LitMacPath,
+	)
+	require.NoError(t.t, err)
+	defer func() { _ = rawConn.Close() }()
+
+	statusClient := litrpc.NewStatusClient(rawConn)
+
+	// lnd should be running before we stop it.
+	resp, err := statusClient.SubServerStatus(
+		ctxt, &litrpc.SubServerStatusReq{},
+	)
+	require.NoError(t.t, err)
+
+	lndStatus, ok := resp.SubServers[subservers.LND]
+	require.True(t.t, ok, "expected lnd sub-server status")
+	require.True(t.t, lndStatus.Running, "lnd should be running initially")
+
+	// Stop lnd (but not litd) via the lnrpc StopDaemon. The call may return
+	// an error as the connection is torn down, which is expected.
+	_, _ = node.LightningClient.StopDaemon(ctxt, &lnrpc.StopRequest{})
+
+	// litd's status endpoint must remain reachable and must report lnd as
+	// no longer running, with an error explaining why. We poll because lnd
+	// shutdown is asynchronous.
+	err = wait.NoError(func() error {
+		ctxs, cancels := context.WithTimeout(
+			context.Background(), defaultTimeout,
+		)
+		defer cancels()
+
+		resp, err := statusClient.SubServerStatus(
+			ctxs, &litrpc.SubServerStatusReq{},
+		)
+		if err != nil {
+			return fmt.Errorf("litd status endpoint unreachable "+
+				"after lnd stopped: %w", err)
+		}
+
+		lndStatus, ok := resp.SubServers[subservers.LND]
+		if !ok {
+			return fmt.Errorf("no lnd sub-server status returned")
+		}
+
+		if lndStatus.Running {
+			return fmt.Errorf("lnd still reported as running")
+		}
+
+		if lndStatus.Error == "" {
+			return fmt.Errorf("lnd error message not yet set")
+		}
+
+		return nil
+	}, defaultTimeout)
+	require.NoError(t.t, err)
+
+	// litd's process itself must keep running after lnd has stopped: its
+	// status endpoint stays reachable. The LIT sub-server is reported as no
+	// longer running, since litd cannot function without lnd, but the
+	// endpoint continuing to answer at all is what proves litd did not
+	// cascade down with lnd. Check this holds over time.
+	err = wait.InvariantNoError(func() error {
+		ctxs, cancels := context.WithTimeout(
+			context.Background(), defaultTimeout,
+		)
+		defer cancels()
+
+		resp, err := statusClient.SubServerStatus(
+			ctxs, &litrpc.SubServerStatusReq{},
+		)
+		if err != nil {
+			return fmt.Errorf("litd status endpoint must stay "+
+				"reachable after lnd stopped: %w", err)
+		}
+
+		litStatus, ok := resp.SubServers[subservers.LIT]
+		if !ok {
+			return fmt.Errorf("no lit sub-server status returned")
+		}
+
+		if litStatus.Running {
+			return fmt.Errorf("lit sub-server should be reported " +
+				"as stopped once lnd has stopped")
+		}
+
+		if litStatus.Error == "" {
+			return fmt.Errorf("lit sub-server error message not set")
+		}
+
+		return nil
+	}, defaultTimeout)
+	require.NoError(t.t, err)
+}
--- a/itest/litd_node.go
+++ b/itest/litd_node.go
@@ -45,7 +45,10 @@ import (
 	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/lightningnetwork/lnd/macaroons"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 	"gopkg.in/macaroon.v2"
 )
 
@@ -383,6 +386,13 @@ type HarnessNode struct {
 
 	// litConn is the underlying connection to Lit's grpc endpoint.
 	litConn *grpc.ClientConn
+
+	// litMacPath, when set, is the macaroon file used to authenticate
+	// litConn. It is set for stateless init nodes, which have no lit
+	// macaroon on disk: we bake a super macaroon from the in-memory admin
+	// macaroon and point this at it so the node can still be stopped via
+	// litd's own StopDaemon.
+	litMacPath string
 
 	// RouterClient, WalletKitClient, WatchtowerClient cannot be embedded,
 	// because a name collision would occur with LightningClient.
@@ -776,31 +786,9 @@ func (hn *HarnessNode) Start(litdBinary string, litdError chan<- error,
 		return nil
 	}
 
-	err = hn.initLightningClient(conn)
-	if err != nil {
-		return fmt.Errorf("could not init Lightning Client: %w", err)
-	}
-
-	// Also connect to Lit's RPC port for any Litd specific calls.
-	litConn, err := connectLitRPC(
-		context.Background(), hn.Cfg.LitAddr(), hn.Cfg.LitTLSCertPath,
-		hn.Cfg.LitMacPath,
-	)
-	if err != nil {
-		return fmt.Errorf("could not connect to Lit RPC: %w", err)
-	}
-	hn.litConn = litConn
-
-	ctxt, cancel := context.WithTimeout(
-		context.Background(), lntest.DefaultTimeout,
-	)
-	defer cancel()
-	return wait.NoError(func() error {
-		litConn := litrpc.NewProxyClient(hn.litConn)
-
-		_, err = litConn.GetInfo(ctxt, &litrpc.GetInfoRequest{})
-		return err
-	}, lntest.DefaultTimeout)
+	// initLightningClient also connects to Lit's RPC port and sets
+	// hn.litConn so the node can be stopped via litd's own StopDaemon.
+	return hn.initLightningClient(conn)
 }
 
 // WaitForLNDWalletReady waits until the wallet state flips from
@@ -1087,6 +1075,60 @@ func (hn *HarnessNode) initClientWhenReady(timeout time.Duration) error {
 	return hn.initLightningClient(conn)
 }
 
+// bakeLitSuperMacaroon bakes a lit super macaroon from the given admin macaroon
+// (returned in-memory during stateless init, where no lit macaroon is written
+// to disk), writes it next to the node's lit config, and records its path so
+// litConn can authenticate against it. This lets the harness stop a stateless
+// node via litd's own StopDaemon.
+func (hn *HarnessNode) bakeLitSuperMacaroon(adminMac []byte) error {
+	ctxt, cancel := context.WithTimeout(
+		context.Background(), lntest.DefaultTimeout,
+	)
+	defer cancel()
+
+	rawConn, err := connectLitRPC(
+		ctxt, hn.Cfg.LitAddr(), hn.Cfg.LitTLSCertPath, "",
+	)
+	if err != nil {
+		return err
+	}
+	defer rawConn.Close()
+
+	// Attach the admin macaroon to the request context so litd authorizes
+	// the bake call.
+	macCtx := metadata.NewOutgoingContext(ctxt, metadata.MD{
+		"macaroon": []string{hex.EncodeToString(adminMac)},
+	})
+
+	litConn := litrpc.NewProxyClient(rawConn)
+	resp, err := litConn.BakeSuperMacaroon(
+		macCtx, &litrpc.BakeSuperMacaroonRequest{
+			RootKeyIdSuffix: 0,
+			ReadOnly:        false,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// BakeSuperMacaroon hex encodes the returned macaroon.
+	superMacBytes, err := hex.DecodeString(resp.Macaroon)
+	if err != nil {
+		return err
+	}
+
+	macPath := filepath.Join(
+		filepath.Dir(hn.Cfg.LitMacPath), "lit-super.macaroon",
+	)
+	if err := os.WriteFile(macPath, superMacBytes, 0644); err != nil {
+		return err
+	}
+
+	hn.litMacPath = macPath
+
+	return nil
+}
+
 // Init initializes a harness node by passing the init request via rpc. After
 // the request is submitted, this method will block until a
 // macaroon-authenticated RPC connection can be established to the harness node.
@@ -1124,6 +1166,17 @@ func (hn *HarnessNode) Init(ctx context.Context,
 		return err == nil
 	}, lntest.DefaultTimeout); err != nil {
 		return nil, err
+	}
+
+	// In stateless init mode no lit macaroon is written to disk, so bake a
+	// super macaroon from the admin macaroon we just received. The harness
+	// uses it to stop the node via litd's own StopDaemon.
+	if initReq.StatelessInit {
+		err := hn.bakeLitSuperMacaroon(response.AdminMacaroon)
+		if err != nil {
+			return nil, fmt.Errorf("could not bake lit super "+
+				"macaroon: %w", err)
+		}
 	}
 
 	return response, hn.initLightningClient(conn)
@@ -1167,6 +1220,17 @@ func (hn *HarnessNode) InitChangePassword(ctx context.Context,
 		return err == nil
 	}, lntest.DefaultTimeout); err != nil {
 		return nil, err
+	}
+
+	// In stateless init mode no lit macaroon is written to disk, so bake a
+	// super macaroon from the admin macaroon we just received. The harness
+	// uses it to stop the node via litd's own StopDaemon.
+	if chngPwReq.StatelessInit {
+		err := hn.bakeLitSuperMacaroon(response.AdminMacaroon)
+		if err != nil {
+			return nil, fmt.Errorf("could not bake lit super "+
+				"macaroon: %w", err)
+		}
 	}
 
 	return response, hn.initLightningClient(conn)
@@ -1248,6 +1312,41 @@ func (hn *HarnessNode) initLightningClient(conn *grpc.ClientConn) error {
 	err := wait.NoError(
 		hn.FetchNodeInfo, lntest.DefaultTimeout,
 	)
+	if err != nil {
+		return err
+	}
+
+	// Also connect to Lit's RPC port for any litd specific calls. We do
+	// this here, rather than only on the no-seed path in start, so that
+	// seed based nodes (which reach this method via Init or Unlock) get a
+	// litConn too. Stop relies on it to shut the node down via litd's own
+	// StopDaemon, which is the only way to bring litd down now that its
+	// lifecycle is decoupled from lnd's. For stateless nodes litMacPath
+	// points at a baked super macaroon; otherwise the lit macaroon on disk
+	// is used.
+	macPath := hn.Cfg.LitMacPath
+	if hn.litMacPath != "" {
+		macPath = hn.litMacPath
+	}
+	litConn, err := connectLitRPC(
+		context.Background(), hn.Cfg.LitAddr(), hn.Cfg.LitTLSCertPath,
+		macPath,
+	)
+	if err != nil {
+		return fmt.Errorf("could not connect to Lit RPC: %w", err)
+	}
+	hn.litConn = litConn
+
+	ctxt, cancel := context.WithTimeout(
+		context.Background(), lntest.DefaultTimeout,
+	)
+	defer cancel()
+	err = wait.NoError(func() error {
+		litClient := litrpc.NewProxyClient(hn.litConn)
+
+		_, err := litClient.GetInfo(ctxt, &litrpc.GetInfoRequest{})
+		return err
+	}, lntest.DefaultTimeout)
 	if err != nil {
 		return err
 	}
@@ -1423,9 +1522,31 @@ func (hn *HarnessNode) Stop() error {
 		return nil
 	}
 
-	// If start() failed before creating a client, we will just wait for the
-	// child process to die.
-	if !hn.Cfg.RemoteMode && hn.LightningClient != nil {
+	// litd owns its own lifecycle, so we ask it to stop via its own
+	// StopDaemon RPC. In integrated mode this also brings down the embedded
+	// lnd.
+	switch {
+	case hn.litConn != nil:
+		ctx, cancel := context.WithTimeout(
+			context.Background(), lntest.DefaultTimeout,
+		)
+		litConn := litrpc.NewProxyClient(hn.litConn)
+
+		// litd normally answers before it shuts down, but the
+		// connection can also be torn down first as litd begins
+		// shutting down. That surfaces as an Unavailable status and
+		// is expected. Any other error is not, so surface it. We
+		// confirm the process actually exits via processExit below.
+		_, err := litConn.StopDaemon(ctx, &litrpc.StopDaemonRequest{})
+		cancel()
+		if err != nil && status.Code(err) != codes.Unavailable {
+			return err
+		}
+
+	// If start() failed before the lit connection was created, fall back to
+	// stopping the integrated lnd directly and just wait for the child
+	// process to die.
+	case !hn.Cfg.RemoteMode && hn.LightningClient != nil:
 		// Don't watch for error because sometimes the RPC connection
 		// gets closed before a response is returned.
 		req := lnrpc.StopRequest{}
@@ -1447,19 +1568,6 @@ func (hn *HarnessNode) Stop() error {
 				return nil
 			}
 		}, lntest.DefaultTimeout)
-		if err != nil {
-			return err
-		}
-	} else if hn.Cfg.RemoteMode {
-		// If lit is running in remote mode, then calling LNDs
-		// StopDaemon method will not shut down Lit, and so we need to
-		// explicitly request lit to shut down.
-		ctx, cancel := context.WithTimeout(
-			context.Background(), lntest.DefaultTimeout,
-		)
-		litConn := litrpc.NewProxyClient(hn.litConn)
-		_, err := litConn.StopDaemon(ctx, &litrpc.StopDaemonRequest{})
-		cancel()
 		if err != nil {
 			return err
 		}
--- a/rpc_proxy.go
+++ b/rpc_proxy.go
@@ -73,7 +73,8 @@ func (e *proxyErr) Unwrap() error {
 func newRpcProxy(cfg *Config, validator macaroons.MacaroonValidator,
 	superMacValidator litmac.SuperMacaroonValidator,
 	permsMgr *perms.Manager, subServerMgr *subservers.Manager,
-	statusMgr *litstatus.Manager, getLNDClient lndBasicClientFn) *rpcProxy {
+	statusMgr *litstatus.Manager, getLNDClient lndBasicClientFn,
+	litdShutdown *shutdownSource) *rpcProxy {
 
 	// The gRPC web calls are protected by HTTP basic auth which is defined
 	// by base64(username:password). Because we only have a password, we
@@ -96,6 +97,7 @@ func newRpcProxy(cfg *Config, validator macaroons.MacaroonValidator,
 		subServerMgr:      subServerMgr,
 		statusMgr:         statusMgr,
 		getBasicLNDClient: getLNDClient,
+		litdShutdown:      litdShutdown,
 	}
 	p.grpcServer = grpc.NewServer(
 		// From the grpxProxy doc: This codec is *crucial* to the
@@ -173,6 +175,10 @@ type rpcProxy struct {
 	statusMgr         *litstatus.Manager
 	getBasicLNDClient lndBasicClientFn
 
+	// litdShutdown is litd's own shutdown source. It is needed alongside
+	// the interceptor because litd's lifecycle is decoupled from lnd's.
+	litdShutdown *shutdownSource
+
 	bakeSuperMac bakeSuperMac
 
 	macValidator      macaroons.MacaroonValidator
@@ -227,6 +233,11 @@ func (p *rpcProxy) StopDaemon(_ context.Context,
 	_ *litrpc.StopDaemonRequest) (*litrpc.StopDaemonResponse, error) {
 
 	log.Infof("StopDaemon rpc request received")
+
+	// Stop litd via its own shutdown source, and stop the embedded lnd via
+	// its interceptor. Both are needed because litd's lifecycle is
+	// decoupled from lnd's interceptor.
+	p.litdShutdown.RequestShutdown()
 
 	interceptor.RequestShutdown()
 
--- a/shutdown.go
+++ b/shutdown.go
@@ -1,0 +1,85 @@
+package terminal
+
+import (
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+// shutdownSource is litd's own shutdown trigger. It is intentionally separate
+// from the signal.Interceptor that litd hands to lnd: lnd and litd are forced
+// to share a single signal.Interceptor (lnd's signal package only allows one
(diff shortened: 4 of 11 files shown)


Vandit1604:
Thanks for the review.

1. Good catch on the Windows build. `syscall.Kill` is not available there, so I moved `TestShutdownSourceSignal` into `shutdown_unix_test.go` with a `//go:build !windows` constraint. The other two tests stay cross platform.

2. The node is already started in integrated mode. The signature is `NewNode(t, name, extraArgs, remoteMode, wait, ...litArgs)`, so `remoteMode` is the 4th arg, which is `false` here. The `true` is the `wait` arg. So the test does exercise the integrated lnd shutdown, no change needed there.

ViktorT-11:
Sorry for taking some time with this one @Vandit1604, I'll review this tomorrow.

terminal.go
```diff
@@ -792,26 +827,51 @@ func (g *LightningTerminal) start(ctx context.Context) error {
...
-		if err != nil {
-			return fmt.Errorf("received critical error from "+
-				"subsystem, shutting down: %v", err)
-		}
+	// Now block until litd is told to shut down. Unlike during startup, an
+	// lnd shutdown here does not bring litd down: we mark lnd as errored,
+	// tear down the lnd-dependent sub-servers, and keep litd running so its
+	// status endpoint stays available. Only an OS signal, litd StopDaemon,
+	// a critical litd error or a fatal sub-system error stops litd.
+	for {
```
ViktorT-11, line 895: With the change you've included here, I think this observing here that remains up until `litd` is placed incorrectly. It's currently placed in the `start` function, but the right place for this should be the `Run` function. I therefore suggest you just return `nil` once you've set `LIT` as running above, and make this the observing part the responsibility of the `Run` function.

terminal.go
```diff
@@ -792,26 +827,51 @@ func (g *LightningTerminal) start(ctx context.Context) error {
...
+		select {
+		case err := <-g.errQueue.ChanOut():
```
ViktorT-11: Hmmm I don't really like here that this introduces different behaviour for what an error sent to the `g.errQueue.ChanOut` channel means depending on if it is during the startup of `litd`, or after all sub-servers have been started. I.e. with this change, an error during startup means that `litd` is shutdown, but is later changed to just mean that it's logged.

I think you need to separate that with two different channels or something similar, where sending over one of them means "fatal" errors, and the one which is used after this `Start` function has returned is `non-fatal` and logged only errors.

shutdown.go
```diff
@@ -0,0 +1,85 @@
...
+	signals chan os.Signal
+}
+
+// newShutdownSource constructs a shutdownSource and starts listening for OS
+// interrupt and termination signals.
+//
+// Note that Go delivers a signal to every channel registered via
+// signal.Notify, so listening here does not prevent lnd's own
+// signal.Interceptor from also receiving the same signal. An OS signal
+// therefore stops both lnd and litd, the same as before this change.
```
ViktorT-11: Commenting specifically regarding the "the same as before this change" part of this comment:
We usually don't mention changes from previous behaviour in the docs. This is because the previous version is irrelevant to people reviewing the current state of the code, unless knowledge about that previous version of the code is explicitly needed to understand why a specific code behaviour is implemented the way it is, which is not the case here.

rpc_proxy.go
```diff
@@ -228,6 +228,12 @@ func (p *rpcProxy) StopDaemon(_ context.Context,
 
 	log.Infof("StopDaemon rpc request received")
 
+	// Stop litd via its own shutdown source, and stop the embedded lnd via
+	// its interceptor. Both are needed now that litd no longer shuts down
+	// just because lnd's interceptor fired.
+	if litdShutdown != nil {
+		litdShutdown.RequestShutdown()
+	}
 	interceptor.RequestShutdown()
```
ViktorT-11, line 242: nit: new line before this line.

rpc_proxy.go
```diff
@@ -228,6 +228,12 @@ func (p *rpcProxy) StopDaemon(_ context.Context,
 
 	log.Infof("StopDaemon rpc request received")
 
+	// Stop litd via its own shutdown source, and stop the embedded lnd via
+	// its interceptor. Both are needed now that litd no longer shuts down
+	// just because lnd's interceptor fired.
```
ViktorT-11: See comment https://github.com/lightninglabs/lightning-terminal/pull/1329/changes#r3435314091 as it applies here as well.

itest/litd_node.go
```diff
@@ -1398,9 +1398,27 @@ func (hn *HarnessNode) Stop() error {
 		return nil
 	}
 
-	// If start() failed before creating a client, we will just wait for the
-	// child process to die.
-	if !hn.Cfg.RemoteMode && hn.LightningClient != nil {
+	// Calling lnd's StopDaemon no longer shuts down litd (litd keeps running
```
ViktorT-11: Same point regarding mentioning previous behaviour here: "no longer shuts down litd".

itest/litd_node.go
```diff
@@ -1398,9 +1398,27 @@ func (hn *HarnessNode) Stop() error {
...
+	// also brings down the embedded lnd.
+	switch {
+	case hn.litConn != nil:
+		ctx, cancel := context.WithTimeout(
+			context.Background(), lntest.DefaultTimeout,
+		)
+		litConn := litrpc.NewProxyClient(hn.litConn)
+
+		// Don't watch for an error: the connection may already be
+		// closing (e.g. litd has begun shutting down), which is fine.
```
ViktorT-11: I think you can just require that the error is that specific type if the `err != nil`, instead of not checking the error at all.

itest/litd_lnd_shutdown_test.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// testLitdSurvivesLndShutdown asserts that when lnd stops in integrated mode,
+// litd keeps running and its status endpoint stays reachable, reporting lnd as
+// stopped rather than cascading to a full litd shutdown.
+func testLitdSurvivesLndShutdown(ctx context.Context, net *NetworkHarness,
+	t *harnessTest) {
+
+	// Use a dedicated, throwaway node so that stopping its lnd cannot
+	// affect any other test. Disable autopilot to keep the node quiet.
```
ViktorT-11: Not sure what the disable autopilot to "keep the node quiet." is referring to hear. But I think you can just remove that part of the comment and keep the passed `"--autopilot.disable"` arg.

itest/litd_lnd_shutdown_test.go
```diff
@@ -0,0 +1,92 @@
...
+			return fmt.Errorf("lnd still reported as running")
+		}
+
+		if lndStatus.Error == "" {
+			return fmt.Errorf("lnd error message not yet set")
+		}
+
+		return nil
+	}, defaultTimeout)
+	require.NoError(t.t, err)
```
ViktorT-11, line 91: You should also test here that the lit sub-server remains up and running after this, i.e. that `statusClient.SubServerStatus` still responds and that the `LIT` sub-server the has "Running" set to `true`. Could be worth testing that this is the case for maybe 3 more seconds or something similar.

shutdown_unix_test.go
```diff
@@ -0,0 +1,29 @@
+//go:build !windows
```
ViktorT-11, line 1: This last commit can be squashed with the first one, instead of being a separate commit :)

ViktorT-11 (commented):
Thanks a lot for this contribution @Vandit1604! This actually works really well, and was a smaller change that I thought it would be. I've tested it locally, and it introduces the functionality that we'd like, so thanks for that :fire:.

Here are the things I've tested:
 * A `log.Critical` error in `lnd` shutsdown just `lnd` and + dependent sub-servers, and not `litd`. Something to note is that it however just leaves "lnd shutdown" in the `litcli status` error for `lnd` though, not the actual error. But I think that's good enough.
 * A runtime error in `lnd` does shutdown just `lnd` and the dependent sub-servers, but not `litd`. It does also print the real error in `litcli status` `lnd` error.
 * An error in `litd` does shut down `lnd` as well.

There is however some architectural feedback that I think needs to be resolved. I'm leaving those in my first 2 comments below. The rest of the feedback is just minor.

Great work :fire:

terminal.go
Vandit1604: Done. start now returns nil once LIT is set to running, and the steady state wait lives in Run as waitForShutdown. Run blocks there, observes lndQuit to tear down the lnd-dependent sub-servers if lnd stops, and only returns on a real litd shutdown.

terminal.go
Vandit1604: Done. The post-startup channel is now a dedicated non-fatal one (subServerErrs), read only in waitForShutdown after start has returned, and those errors are logged only. Fatal errors are no longer routed through it: startup failures are returned directly from start, and an lnd stop during startup comes in over lndQuit. The middleware is the only component that reports async errors and it only does so at runtime, so the queue now has a single, unambiguous meaning.

shutdown.go
Vandit1604: Done, dropped the comparison to the previous behaviour.

rpc_proxy.go
Vandit1604: Done, reworded without referencing the old behaviour.

rpc_proxy.go
Vandit1604: Done.

itest/litd_node.go
Vandit1604: Done, reworded.

itest/litd_node.go
Vandit1604: Done. It now returns the error unless it is an Unavailable status, which is the expected case when the connection is torn down as litd shuts down.

itest/litd_lnd_shutdown_test.go
Vandit1604: Done, removed that part of the comment and kept the --autopilot.disable arg.

itest/litd_lnd_shutdown_test.go
Vandit1604: Done. After lnd is reported stopped, the test now checks that the LIT sub-server stays reachable and Running for 3 seconds.

shutdown_unix_test.go
Vandit1604: Done, folded into the first commit.

Vandit1604:
Thanks for the thorough review. I've pushed an update addressing all the points:

- The steady state wait moved out of start and into Run as waitForShutdown. start now returns once LIT is running.
- Post-startup sub-server errors now go through a dedicated non-fatal channel (subServerErrs) that is only read after start returns and is logged only. Fatal startup errors are returned directly or signalled via lndQuit.
- Reworded the comments that referenced previous behaviour, fixed the spacing nit, tightened the harness error check, and the test now also asserts LIT stays up for a few seconds after lnd stops.
- Squashed the unix test file into the first commit.

Replied inline on each thread too.

terminal.go
```diff
@@ -520,9 +554,15 @@ func (g *LightningTerminal) start(ctx context.Context) error {
...
 		unlockChan     = make(chan struct{})
-		lndQuit        = make(chan struct{})
 		macChan        = make(chan []byte, 1)
 	)
+
+	// lndQuit is observed both here (during startup, where an lnd stop is
+	// fatal) and by Run (after startup, where it tears down the lnd-dependent
+	// sub-servers but keeps litd alive). It lives on g so Run can reach it.
+	g.lndQuit = make(chan struct{})
+	lndQuit := g.lndQuit
```
ViktorT-11: Hmmm now we have the same behaviour where wanted to avoid but for your newly added `g.lndQuit` channel instead (i.e.  that an error sent over the channel means different thing depending on wether we've started or starting)...

I think this new version is more confusing than your last version as well. I therefore suggest reverting to your previous version, and building upon that.

When looking through that version again, I noticed the following which would probably a clean way we could handle this instead:

-------------------

As the lnd shutdown doesn't trigger the `<-g.shutdown.ShutdownChannel()` to be sent over, not even during the startup phase, we can just let it be the case that errors from the `start` function just triggers the `lit` sub-server to be stopped, but the `status` server keeps being up and running. This also applies during the startup phase of the `start` function.

That way we can keep `g.errQueue` as the main error source, and not have to add these additional channels i.e. keep it the way it was during your previous version of this PR. 

However, due to the signaling change (i.e. that stopping `lnd` doesn't trigger `<-g.shutdown.ShutdownChannel()`), we need to also change one thing with the error checking of the `start` function in the `Run` function, i.e.:
```
	// Attempt to start Lit and all of its sub-servers. If an error is
	// returned, it means that either one of Lit's internal sub-servers
	// could not start or LND could not start or be connected to.
	startErr := g.start(ctx)
	if startErr != nil {
		g.statusMgr.SetErrored(
			subservers.LIT, "could not start Lit: %v", startErr,
		)
	}
```

We can't let `litd` remain up and running if the `lnd` errors during the startup, when the following hasn't been run yet in `start`:
```
	// Now start the RPC proxy that will handle all incoming gRPC, grpc-web
	// and REST requests.
	if err := g.rpcProxy.Start(g.lndConn, bakeSuperMac); err != nil {
		return fmt.Errorf("error starting lnd gRPC proxy server: %v",
			err)
	}
```

The reason for that is that users won't be able to run `litcli stop` before that if we keep the status server up and running.

Therefore, in the `startErr` check in the `Run` function, you'd want to check if the `rpcProxy` has been started or not. If it hasn't `litd` should shutdown as well.

Finally, to keep the behaviour cooherent, you should also change so that the `lit` sub-server is always stopped when `lnd` is stopped, independently of wether `lnd` stops during the startup phase, or during runtime after its been started.

Let me know if you need any clarifications to the above!

ViktorT-11 (commented):
Thanks for the updates @Vandit1604. Unfortunately though, I think these changes were a bit more confusing than the previous version and still have the same behaviour. I've therefore left a comment below of an alternative solution building upon your previous version of the PR, which would solve the issue.

Let me know if you'd want some additional clarifications!

Vandit1604:
Thanks for the detailed write up. I reworked it close to what you suggested:

- Back to a single `g.errQueue` as the only runtime error source, no extra channels.
- `start()` returns `nil` once `LIT` is running, and the keep-alive wait is back in `Run`, in a small `waitForShutdown` helper, like you noted in your earlier comment.
- The keep-alive is gated on `g.rpcProxy.hasStarted()`. If the proxy never started, litd shuts down fully, since `litcli stop` could not reach it anyway.
- `LIT` is now always marked stopped when lnd stops, both during startup and at runtime.

There is one part I could not get working: having `start()` itself block and return the error on a runtime lnd stop. That structure consistently fails the lnd shutdown itest, so I kept the wait in `Run`.

Here is exactly what fails and why, in case you see something I missed.

When the wait blocks inside `start()`, a second node that runs at the same time (Bob in the harness) fails its graph topology subscription during teardown. The harness watcher gets:

```
rpc error: code = Unknown desc = unexpected HTTP status code received from server: 200 (OK); malformed header: missing HTTP content-type
```

So the proxied `SubscribeChannelGraph` stream gets answered by litd's static file handler (HTTP 200) instead of a clean gRPC EOF. The harness treats any non EOF stream error as fatal, so the test fails. This is deterministic with the wait inside `start()`, and never happens with the wait in `Run`. I ran it many times both ways.

I ruled out:

- the `lndStopped` guard that skips the graceful lnd client `Close`
- whether `start` returns via `ctx.Done` or via the shutdown channel
- the `LND` status being marked errored (removing that did not help)

The litd shutdown log sequence is the same in both versions, so it looks like a teardown timing or scheduling difference between litd tearing down its proxy and lnd sending GOAWAY on that proxied stream. I could not pin the exact cause.

Net effect: `lndQuit` is still observed in two places, `start` during startup and `waitForShutdown` at runtime. If that is acceptable given the itest, I will keep it. If you would rather have the wait inside `start()`, any pointer on the teardown ordering would help, since I could not crack it.

terminal.go
ViktorT-11, line 577: We still want to pass the runtime errors into the `errQueue`, as we reverted to it being the "single source" of errors.

ViktorT-11, line 608: You've removed this check for the `errQueue.ChanOut()` here and in many other spots in this file. I don't think that is what you want after reverting back to it being the "single error source". We especially want it to return here if `lnd` fails during startup, so we don't proceed to start the other sub-servers (i.e. `loop`, `tapd` etc. if `lnd` startup failed)

It should be fine to return here now, as your other logic should have changed that the `litd` daemon doesn't shutdown when `lnd` has stopped, as the status server is kept up and running.

ViktorT-11 (commented):
Thanks for the update @Vandit1604. This looks better, but I still think there are some spots where you didn't revert back to using the `errQueue` as the source for errors.

You have also added the "revert" in an additional commit as the second last commit of the PR. Please squash it with the "terminal: keep litd running after lnd stops" commit, as the current commit structure makes this quite complicated to review.

Also responding to part of your comment above:
> There is one part I could not get working: having start() itself block and return the error on a runtime lnd stop.

I'm sorry if I was unclear there, that was not the approach I intended to convey, so your idea of blocking in `Run` is the correct approach.

Vandit1604:
@ViktorT-11 I pushed the review changes (errQueue back as the single error source, and squashed the rework into the "keep litd running after lnd stops" commit).

Before I push a fix for the failing CI I wanted your call on it, since it touches the itest harness.

CI fails on `terminal stateless init mode` with "process did not exit". I traced it and it is a direct consequence of the feature, not a flake.

The tests stop a node by calling lnd's `StopDaemon`. On master that also brought litd down. With this PR, stopping lnd leaves litd running on purpose, so the process never exits and the harness times out. The Bob HTTP 200 error you might see is a side effect of the same hang: while the node's teardown drags on, a co-running node's graph subscription gets answered by litd's static file handler instead of a clean gRPC EOF.

The harness already switched to stopping nodes through litd's own `StopDaemon` to handle this, but that path needs an authenticated litd connection (`litConn`), which is only set up on the no-seed path in `start`. Seed based nodes (stateless init, seed restarts) return early before it is set, so they fall back to stopping lnd and hang.

Two harness-only ways to fix it, no production change:

1. Send the litd process an OS interrupt in `Stop` instead of an RPC. litd's shutdown source already handles SIGINT/SIGTERM and brings down both litd and lnd. No macaroon needed, so it works for stateless too (where macaroons are intentionally not on disk). This is my preferred option.
2. Set up `litConn` for seed nodes too. This does not work cleanly for stateless, since that mode keeps no macaroon files on disk, which is exactly what `litConn` needs.

I am leaning towards option 1. Does that sound right, or would you prefer a different approach?

ViktorT-11:
Thanks for the changes and the writeup @Vandit1604!

Regarding the issue with stateless init mode:
> The tests stop a node by calling lnd's StopDaemon. On master that also brought litd down. With this PR, stopping lnd leaves litd running on purpose, so the process never exits and the harness times out. The Bob HTTP 200 error you might see is a side effect of the same hang: while the node's teardown drags on, a co-running node's graph subscription gets answered by litd's static file handler instead of a clean gRPC EOF.

> The harness already switched to stopping nodes through litd's own StopDaemon to handle this, but that path needs an authenticated litd connection (litConn), which is only set up on the no-seed path in start. Seed based nodes (stateless init, seed restarts) return early before it is set, so they fall back to stopping lnd and hang.

I propose one alternative solution you could try first:
For the `stateless init` setups in the `itest` setup, we could try to actually create a lit macaroon that we could use within the test harness, so that the `StopDaemon` call for `lit` can actually be executed even in stateless init mode.

This is how you'd create such a macaroon in a real stateless init setup:
1. Create a wallet in stateless init mode (be prepared to choose a password, and save your seed): 
`lncli create --stateless_init --save_to /tmp/admin.macaroon`

2. Create a lit super macaroon that contains all permissions from all different subservers
`litcli --macaroonpath /tmp/admin.macaroon bakesupermacaroon --save_to /tmp/lit.macaroon`

The outputted litd macaroon will then be saved to `/tmp/lit.macaroon`.

Obviously for the `itest` setup, this would need to be done a little bit differently, as for example the test harness has its own capabilities to create a temporary directory.

-----------------
Test if you can get that working. Else I agree with you that option 1 is best of the 2 options you suggested.
If you cannot figure out how to get that working, I can look into creating a PR to add this capability to the `itest` harness in the stateless init setup, which you could then base this PR on.

terminal.go
```diff
@@ -575,19 +608,32 @@ func (g *LightningTerminal) start(ctx context.Context) error {
...
-				errStr := fmt.Sprintf("Error running main "+
-					"lnd: %v", err)
-				log.Errorf(errStr)
-
-				g.statusMgr.SetErrored(subservers.LND, errStr)
+				lndErr := fmt.Sprintf(
+					"Error running main lnd: %v", err,
+				)
+				log.Errorf(lndErr)
+				g.statusMgr.SetErrored(subservers.LND, lndErr)
```
ViktorT-11, line 643: nit: don't remove the new line above this line.

terminal.go
```diff
@@ -844,7 +919,7 @@ func (g *LightningTerminal) checkRunning(ctx context.Context,
 
 	select {
 	case err := <-g.errQueue.ChanOut():
-		return fmt.Errorf("error from subsystem: %v", err)
+		return fmt.Errorf("error while starting: %w", err)
```
ViktorT-11: I don't think this necessarily only have to be startup issues, so I would not change this. Potentially you could keep the change to use `%w` though.

terminal.go
```diff
@@ -1614,7 +1734,7 @@ func (g *LightningTerminal) shutdownSubServers() error {
 		}
```
ViktorT-11, line 1634: Hmmm in what scenario does this change lead to that the `errQueue.ChanOut` now contains an error that hasn't been handled prior to the shutting down `litd` and that we'd therefore only want to log and not assign to `returnErr`?

ViktorT-11 (commented):
Just leaving a few minor comments after looking through the new changes. I haven't tested this again, so these are primarily quick feedback and a question for you.

Vandit1604:
Pushed the updates, thanks for the detailed review.

The three inline nits are done:
- kept the blank line above the `SetErrored` call
- reverted the message back to "error from subsystem" (kept the `%w` change)
- restored `returnErr = err` in the final drain

I also squashed the rework into the "terminal: keep litd running after lnd stops" commit, so the history is clean now.

For the stateless init issue I went with your suggestion. In `Init` and `InitChangePassword`, for stateless nodes the harness now bakes a lit super macaroon from the admin macaroon it gets back, writes it to disk, and uses it to set up `litConn`. I also moved the `litConn` setup into `initLightningClient` so every node gets one, seed or not. The harness can now stop stateless nodes through litd's own `StopDaemon`.

Locally this passes: both the stateless init test and the litd survives lnd shutdown test pass, with no regressions. One small note: the "Bob 200" line is just a non fatal log from `PrintErr`, it does not fail the test, so it can be ignored.

CI looks like it is waiting on approval to run. Once it is kicked off and passes, this should be good to merge from my side.

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -35,6 +35,14 @@
 
 ### Functional Changes/Additions
 
+* [Keep litd running after lnd
```
ViktorT-11: Sorry for the delay in the review here. I'll get this reviewed before the end of the week!

In the meantime, please update this to instead be part of the release notes for v0.17.1:

terminal.go
```diff
@@ -265,31 +279,42 @@ func (s *stores) close() error {
...
 	if err != nil {
 		return fmt.Errorf("could not intercept signals: %v", err)
 	}
 
+	// Create litd's own shutdown source and make it available to the
+	// package-level logger so that critical logs can bring litd down. It
+	// listens for OS signals independently of lnd's interceptor.
+	g.shutdown = newShutdownSource()
+	defer g.shutdown.Stop()
+	litdShutdown = g.shutdown
```
ViktorT-11: Instead of letting `litdShutdown` be a global variable in `log.go` that is assigned here, I think it'd make more sense to pass a reference to it in the `log.go SetupLoggers` & `rpc_proxy.go newRpcProxy` functions.

The reason behind that is in that in general it's better to avoid global variables if possible, and especially here it makes code readability worse as it's hard to understand that the `terminal.go Run` function needs to run prior to the executing the dependent code in `log.go` &  `rpc_rpoxy.go`.

terminal.go
```diff
@@ -802,29 +842,65 @@ func (g *LightningTerminal) start(ctx context.Context) error {
...
+			// lnd has stopped while litd was running. The
+			// lnd.Main goroutine has already set the LND
+			// sub-server status (the real error, or a generic
+			// message on a clean stop). Record that lnd is gone
+			// so the final shutdown skips the graceful lnd client
+			// teardown, mark LiT as stopped so the status endpoint
+			// reports coherently, and tear down the lnd-dependent
+			// sub-servers while keeping litd running.
+			g.lndStopped = true
+			g.statusMgr.SetStopped(subservers.LIT)
```
ViktorT-11: When we shutdown the lit sub-server due to that `lnd` has stopped, we can set it as errored with an error explaining that it was stopped as `lnd` was shut down.

itest/litd_node.go
```diff
@@ -384,6 +387,13 @@ type HarnessNode struct {
 	// litConn is the underlying connection to Lit's grpc endpoint.
 	litConn *grpc.ClientConn
 
+	// litMacPath, when set, is the macaroon file used to authenticate
```
ViktorT-11, line 390: You can break out these changes which are related to being able to shutdown the `itest` litd node in stateless init though the `litcli` `StopDaemon` into a separate PR, as it is quite separate to the rest of the changes in this PR.

We can then review that separately, and base this PR on that new PR which contains only those changes.

itest/litd_lnd_shutdown_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// litd's process itself must keep running after lnd has stopped: its
+	// status endpoint stays reachable. The LIT sub-server is reported as no
+	// longer running, since litd cannot function without lnd, but the
+	// endpoint continuing to answer at all is what proves litd did not
+	// cascade down with lnd. Check this holds for a few seconds.
+	for i := 0; i < 3; i++ {
```
ViktorT-11: Instead of doing this `for` loop with a `time.Sleep`, you can instead use the `wait.Invariant()` helper. That aligns better with how we typically test that a specific condition remains true over time (i.e. in this case that the litd status endpoint remains up).

itest/litd_lnd_shutdown_test.go
```diff
@@ -91,18 +91,22 @@ func testLitdSurvivesLndShutdown(ctx context.Context, net *NetworkHarness,
 	}, defaultTimeout)
 	require.NoError(t.t, err)
 
-	// litd itself must keep running after lnd has stopped. Check that the
-	// LIT sub-server stays reachable and reports as running for a few
-	// seconds, confirming it does not shut down shortly after lnd does.
+	// litd's process itself must keep running after lnd has stopped: its
```
ViktorT-11, line 93: This last commit can be squashed with the commit called:
"itest: add test for litd surviving lnd shutdown"

ViktorT-11 (commented):
Thanks again for this contribution @Vandit1604! Appologies for taking some time to review this again.

This looks very close from my point of view now. Just leaving a few minor comments below.

Additionally, I've also tested once more that the changes in this PR ensures the following:

 * A `log.Critical` error in `lnd` shutsdown just `lnd` and + dependent sub-servers, and not `litd`. Something to note is that it however just leaves "lnd shutdown" in the `litcli status` error for `lnd` though, not the actual error. But I think that's good enough.
 * A runtime error in `lnd` does shutdown just `lnd` and the dependent sub-servers, but not `litd`. It does also print the real error in `litcli status` `lnd` error.
 * An error in `litd` does shut down `lnd` as well.

Vandit1604:
Thanks for the review @ViktorT-11. All addressed, rebased on latest master.

`litdShutdown` is no longer a global. It goes through `loadAndValidateConfig` into `SetupLoggers` and `genSubLogger`, and is a field on `rpcProxy`. `NewGrpcLogLogger` also calls `genSubLogger`, so I gave it the param too rather than passing nil. It has no callers, but say the word if you'd rather leave its signature alone.

Lit is now set errored with "lit was stopped as lnd was shut down" instead of just stopped. The itest checks the error string too.

Also done: `wait.InvariantNoError` instead of the sleep loop, last commit squashed, release notes moved to 0.17.1.

The harness change is split out into #1356. The bottom two commits here are from it.

Locally `litd survives lnd shutdown` and `terminal stateless init mode` pass, and every commit builds on its own. `UI index page fallback` fails locally only because a locally built `litd-itest` has no UI bundle.

litbot-9000:
@viktort-11: review reminder

litbot-9000:
@viktort-11: review reminder

litbot-9000:
@viktort-11: review reminder
@ellemouton: review reminder

litbot-9000:
@viktort-11: review reminder
@ellemouton: review reminder

litbot-9000:
@viktort-11: review reminder
@ellemouton: review reminder

litbot-9000:
@viktort-11: review reminder
@ellemouton: review reminder

litbot-9000:
@viktort-11: review reminder
@ellemouton: review reminder

Vandit1604:
Hi @ViktorT-11, thanks again for all the help on this one. I rebased it onto the current #1356 commit, so both PRs now share the same stateless init fix. I also moved the release note to `release-notes-0.17.5.md`, since 0.17.2 is already out. The red postgres run is the `process did not exit` hang from #1408, in `terminal_request_logger_disable`. Whenever you have time for another look, that would be great.

PR #1330, opened 2026-06-15
[sql-77] terminal: shut down LiT when declining the SQL migration prompt

ViktorT-11:
Treat a declined kvdb-to-SQL migration prompt as a startup-abort condition that shuts litd down fully. This prevents litd from leaving the status server running after the operator refuses the migration prompt.

The reason why this is motivated, is that `litd` at this stage of the startup process will not be ready to handle an `litcli stop` RPC call.

### DIFF
--- a/migration_prompt_test.go
+++ b/migration_prompt_test.go
@@ -58,7 +58,7 @@ func TestConfirmPendingKVDBToSQLMigration(t *testing.T) {
 				createActiveAccountsKVDB(t, dbDir)
 			},
 			input:     "no\n",
-			expectErr: "manual confirmation declined",
+			expectErr: errKVDBToSQLMigrationDeclined.Error(),
 		},
 		{
 			name: "skips prompt on auto migration config",
--- a/terminal.go
+++ b/terminal.go
@@ -405,6 +405,19 @@ func (g *LightningTerminal) Run(ctx context.Context) error {
 		g.statusMgr.SetErrored(
 			subservers.LIT, "could not start Lit: %v", startErr,
 		)
+
+		// If the user has declined the migration prompt, we shut down
+		// Lit fully, and do not keep the status server up and running.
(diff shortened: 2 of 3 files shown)


migration_prompt_test.go
```diff
@@ -144,6 +144,14 @@ func TestConfirmPendingKVDBToSQLMigration(t *testing.T) {
 
 			if tc.expectErr != "" {
 				require.ErrorContains(t, err, tc.expectErr)
+				if strings.Contains(tc.expectErr,
+					"manual confirmation declined") {
+
+					require.ErrorIs(
+						t, err,
+						errKVDBToSQLMigrationDeclined,
+					)
```
ellemouton: this reads a bit strange:
1) is error equal to expected error
2) does expected error contain this specific thing
3) then again require that the error is the expected one?

terminal.go
```diff
@@ -405,6 +405,19 @@ func (g *LightningTerminal) Run(ctx context.Context) error {
...
+
+		// If the user has declined the migration prompt, we shut down
+		// Lit fully, and do not keep the status server up and running.
+		// The motivation for this is that the error occurs prior to
+		// Lit being able to accept a `litcli stop` call. Users running
+		// in an env that can't easily kill the daemon therefore need to
+		// be able to shut it down by just declining the migration.
+		// Note that no sub-servers, including `lnd` has not been
+		// started/connected to when the migration prompt is shown. We
+		// can therefore safely shut Lit down without effecting them.
```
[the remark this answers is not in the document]
ellemouton: agree

ellemouton (approved):

migration_prompt_test.go
ViktorT-11: Removed this check, and changed the `expectErr` to instead be `errKVDBToSQLMigrationDeclined.Error()`, which should provide sufficient coverage IMO.

Issue #1294, opened 2026-04-30
#1274 Follow-ups


ViktorT-11:
In #1274, the functionality to edit an account's label was introduced. 

This issue tracks follow-ups to that PR mentioned in the reviews:
* https://github.com/lightninglabs/lightning-terminal/pull/1274#discussion_r3159938962
* https://github.com/lightninglabs/lightning-terminal/pull/1274#discussion_r3159963652
* https://github.com/lightninglabs/lightning-terminal/pull/1274#discussion_r3190733934 (Addressed in https://github.com/lightninglabs/lightning-terminal/pull/1303)

PR #1334, opened 2026-06-23
accounts: add test coverage for checkLabel

0xfandom:
### Change Description

`checkLabel` (in `accounts/interface.go`) rejects account labels that could be mistaken for a hex-encoded account ID, so the CLI can reliably tell labels and IDs apart. The helper had no dedicated unit test.

This adds a table-driven `TestCheckLabel` covering both branches:

- **Accepted:** empty label, plain-text label, short hex string, and a non-hex string that happens to be the length of an encoded account ID.
- **Rejected:** lower- and upper-case hex strings of exactly the encoded account-ID length (`hex.DecodeString` accepts uppercase, so that case is covered explicitly).

This fills the test-coverage gap left by the `checkLabel` validation discussed in the #1274 follow-ups (#1294). No behaviour change.

### Steps to Test

```
go test ./accounts/ -run TestCheckLabel -v
```

### Pull Request Checklist

- [x] Test coverage added.
- [x] Commits are signed (DCO).

### DIFF
--- a/accounts/store_test.go
+++ b/accounts/store_test.go
@@ -778,3 +778,61 @@ func TestLastInvoiceIndexes(t *testing.T) {
 	require.EqualValues(t, 7, add)
 	require.EqualValues(t, 99, settle)
 }
+
+// TestCheckLabel ensures that only labels that could be mistaken for a hex
+// encoded account ID are rejected, while all other labels (including the empty
+// label) are accepted.
+func TestCheckLabel(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name      string
+		label     string
+		expectErr bool
+	}{{
+		name:  "empty label is allowed",
+		label: "",
+	}, {
+		name:  "plain text label is allowed",
+		label: "my account",
+	}, {
+		name:  "short hex label is allowed",
+		label: "00112233",
+	}, {
+		name: "non-hex label with account ID length is allowed",
+		// 16 characters long, matching an encoded account ID, but not
+		// valid hex.
+		label: "zzzzzzzzzzzzzzzz",
+	}, {
+		name: "lowercase hex label with account ID length is rejected",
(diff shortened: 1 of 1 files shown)


docs/release-notes/release-notes-0.17.0.md
```diff
@@ -79,50 +33,28 @@
...
-
-* [Add custom permissions support to sessions](https://github.com/lightninglabs/lightning-terminal/pull/1317):
-  Added the `--permission` flag to `litcli sessions add` to allow specifying
-  custom `entity:action` permissions (e.g. `info:read`), similar to
-  `lncli bakemacaroon`, for sessions of type `custom`. Supports repeated flags,
-  comma-separated lists, and mixed input.
-
 ### Technical and Architectural Updates
 
+* [Add test coverage for account label
```
ViktorT-11: As this PR adds test coverage only, and doesn't effect the user in any way, it should not have any dedicated release notes. The release notes are intended to be read by users to see what has changed that may effect their usage of the `litd`.

ViktorT-11 (approved):
Leaving a minor comment regarding removing the added release note.

0xfandom:
Good point, removed the release notes entry , this is test-only and has no user-facing impact. Thanks for the review!

0xfandom:
Heads-up: dropping the entry makes the `check release notes updated` CI job go red, since `scripts/check-release-notes.sh` greps the release-notes folder for the PR number on every PR. Happy to either re-add a minimal entry or leave it out and let you admin-merge past that check , whichever you prefer.

ViktorT-11:
Thanks!

> Heads-up: dropping the entry makes the check release notes updated CI job go red, since scripts/check-release-notes.sh greps the release-notes folder for the PR number on every PR. 

Yes, that's ok.

PR #1337, opened 2026-06-25
config: make litd -V report litd version

### LINKED ISSUE #1091: `litd -V` shows confusing version string
If I run `litd -V` I get

`litd version 0.18.0-beta commit=lightning-terminal-v0.13.993-experimental`

which leads me confused if it is litd version `0.18.0-beta` or `v0.13.993-experimental`.

Viktor edit:
The version string therefore first shows the integrated `lnd` version, followed by the `git tag` as the commit.
It would be more suitable if the string included `litd`'s version, and not `lnd`'s.

0xfandom:
### Change Description

Fixes #1091.

`litd -V` reported the integrated **lnd** version rather than litd's own version. The `-V` / `--version` flag is inherited from lnd's embedded config (`preCfg.Lnd.ShowVersion`), and its handler printed `build.Version()` / `build.Commit` from `github.com/lightningnetwork/lnd/build`:

```go
if preCfg.Lnd.ShowVersion {
    fmt.Println(appName, "version", build.Version(), "commit="+build.Commit)
    os.Exit(0)
}
```

litd's own `--version` flag (`preCfg.ShowVersion`) already printed the correct value via `RichVersion()`. This consolidates both flags so either one reports litd's version, matching the behaviour agreed on in the issue thread.

### Steps to Test

```
make build
./litd-debug -V
./litd-debug --version
```

Both now print litd's version (e.g. `litd version 0.17.0-alpha.rc1 commit=... commit_hash=...`). Before this change `-V` printed lnd's version (e.g. `0.21.0-beta`).

### Pull Request Checklist

- [x] Verified by building `litd` and running both `-V` and `--version`.
- [x] Commits are signed (DCO).

### DIFF
--- a/config.go
+++ b/config.go
@@ -365,16 +365,12 @@ func loadAndValidateConfig(interceptor signal.Interceptor) (*Config, error) {
 		return nil, fmt.Errorf("error parsing flags: %w", err)
 	}
 
-	// Show the version and exit if the version flag was specified.
+	// Show the version and exit if the version flag was specified. This
+	// applies for both the `--version` and the `-V` flag.
 	appName := filepath.Base(os.Args[0])
 	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
-	if preCfg.ShowVersion {
+	if preCfg.ShowVersion || preCfg.Lnd.ShowVersion {
 		fmt.Println(appName, "version", RichVersion())
-		os.Exit(0)
-	}
-	if preCfg.Lnd.ShowVersion {
-		fmt.Println(appName, "version", build.Version(),
-			"commit="+build.Commit)
 		os.Exit(0)
 	}
 
--- a/docs/release-notes/release-notes-0.17.0.md
+++ b/docs/release-notes/release-notes-0.17.0.md
@@ -35,6 +35,10 @@
 
 ### Technical and Architectural Updates
 
+* [Report litd's own version for `litd
+  -V`](https://github.com/lightninglabs/lightning-terminal/pull/1337): The `-V`
+  flag now prints litd's version instead of the integrated lnd version.
+
 ## RPC Updates
 
 ## Integrated Binary Updates


config.go
```diff
@@ -591,18 +365,16 @@ func loadAndValidateConfig(ctx context.Context,
 		return nil, fmt.Errorf("error parsing flags: %w", err)
 	}
 
-	// Show the version and exit if the version flag was specified.
+	// Show the version and exit if the version flag was specified. Both
+	// litd's own `--version` flag and the `-V`/`--version` flag inherited
+	// from lnd should report litd's version, not the integrated lnd version
+	// (see https://github.com/lightninglabs/lightning-terminal/issues/1091).
```
ViktorT-11: The `lint` issue is actually stemming from this line being too long even though the CI is referencing another unrelated row.

ViktorT-11: We usually don't refer to the github issue in question that we're fixing in the actual code base, unless necessary to understand why the code is structured the way it is. In this case it is not, so I would suggest updating this to something like the below: 
 
```suggestion
	// Show the version and exit if the version flag was specified. This
	// applies for both the `--version` and the `-V` flag.
```

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -79,50 +33,28 @@
...
-
-* [Add custom permissions support to sessions](https://github.com/lightninglabs/lightning-terminal/pull/1317):
-  Added the `--permission` flag to `litcli sessions add` to allow specifying
-  custom `entity:action` permissions (e.g. `info:read`), similar to
-  `lncli bakemacaroon`, for sessions of type `custom`. Supports repeated flags,
-  comma-separated lists, and mixed input.
-
 ### Technical and Architectural Updates
 
+* [Report litd's own version for `litd
```
ViktorT-11, line 38: You can squash this commit with the other commit as this is such a small change.

docs/release-notes/release-notes-0.17.0.md
```diff
@@ -79,50 +33,28 @@
...
+  -V`](https://github.com/lightninglabs/lightning-terminal/pull/1337): The
+  `-V`/`--version` flag now prints litd's version instead of the integrated lnd
```
ViktorT-11: I think you should probably only mention the `-V` flag here, as the `--version` flag already printed the `litd` version prior to this PR.

ViktorT-11 (commented):
Thanks for the contribution. I've tested this and your update works as expected! So mainly leaving some comments regarding non code specific suggestions.

0xfandom:
Thanks for testing it and for the review! Addressed all of it:
- Reworded the comment to your suggestion (also clears the line-length lint).
- Dropped the issue reference from the code comment.
- Release note now mentions only the `-V` flag.
- Squashed into a single commit.

ViktorT-11 (approved):

bitromortac (approved):

litbot-9000:
@0xfandom, remember to re-request review from reviewers when ready

PR #1353, opened 2026-07-15
terminal: gate wallet-ready status on lnd's actual RPC readiness

bitromortac:
Tries to fix https://github.com/lightninglabs/lightning-terminal/actions/runs/29286599979/job/87013613508?pr=1322 which occurred in https://github.com/lightninglabs/lightning-terminal/pull/1322.

### DIFF
--- a/itest/litd_node.go
+++ b/itest/litd_node.go
@@ -899,31 +899,56 @@ func (hn *HarnessNode) WaitUntilStarted(conn grpc.ClientConnInterface,
 			return err
 		}
 
+		// LiT itself only reports Running once it has finished baking
+		// and writing its default macaroons to disk, so waiting for
+		// this closes the race between that and callers that read
+		// LitMacPath straight off disk right after we return.
+		litStatus, ok := states.SubServers[subservers.LIT]
+		if !ok || !litStatus.Running {
+			return fmt.Errorf("LiT has not yet started")
+		}
+
 		if faradayMode != terminal.ModeDisable {
 			faraday, ok := states.SubServers[subservers.FARADAY]
-			if !ok || !faraday.Running {
-				return fmt.Errorf("faraday has not yet started")
+			if !ok {
+				return fmt.Errorf("faraday status not found")
+			}
+			if faraday.Error != "" {
+				return fmt.Errorf("faraday failed to "+
+					"start: %s", faraday.Error)
 			}
 		}
 
 		if loopMode != terminal.ModeDisable {
 			loop, ok := states.SubServers[subservers.LOOP]
-			if !ok || !loop.Running {
-				return fmt.Errorf("loop has not yet started")
+			if !ok {
+				return fmt.Errorf("loop status not found")
+			}
+			if loop.Error != "" {
+				return fmt.Errorf("loop failed to "+
+					"start: %s", loop.Error)
 			}
 		}
 
 		if poolMode != terminal.ModeDisable {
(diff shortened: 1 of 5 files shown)


bitromortac:
Added another itest failure fix.

itest/litd_node.go
```diff
@@ -899,6 +899,15 @@ func (hn *HarnessNode) WaitUntilStarted(conn grpc.ClientConnInterface,
 			return err
 		}
 
+		// LiT itself only reports Running once it has finished baking
+		// and writing its default macaroons to disk, so waiting for
+		// this closes the race between that and callers that read
+		// LitMacPath straight off disk right after we return.
+		litStatus, ok := states.SubServers[subservers.LIT]
```
ViktorT-11, line 906: As we at least currently set the LIT sub-server as running only after the rest of the sub-servers has been set to running, the the other checks below become redundant now.
Perhaps they should check if the sub-servers errored instead of if they're running.

lnd_ready_test.go
```diff
@@ -0,0 +1,116 @@
+package terminal
```
ViktorT-11, line 1: I feel that maybe we should find another file to fit this test in as a general "terminal_test" file for just this test is quite broad. Or we can extract the `waitForLndRPCReady` to a separate file, with a corresponding test file.

ViktorT-11 (commented):
Thanks! I think the LIT status check in the `itest` will likely solve the issue with just that change, but it probably makes sense to include extra lnd wallet check for other reasons.

Leaving just some minor feedback below.

terminal.go
```diff
@@ -857,6 +877,49 @@ func (g *LightningTerminal) checkRunning(ctx context.Context,
...
+					"within %v: %w", timeout, lastErr)
+			}
+
+			return fmt.Errorf("lnd's RPC interceptor did not "+
+				"leave WAITING_TO_START within %v", timeout)
+
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
```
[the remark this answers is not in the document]
bitromortac: Fixed. We extracted 'waitForLndRPCReady' into 'lnd_ready.go' and refactored the timeout check to use 'time.NewTimer(timeout)' with a deferred 'Stop()' to prevent resource leaks.

lnd_ready_test.go
```diff
@@ -0,0 +1,116 @@
...
+				},
+			},
+			timeout:     300 * time.Millisecond,
+			expectError: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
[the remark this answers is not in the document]
bitromortac: Fixed. Renamed the test file to 'lnd_ready_test.go' and added 'tc := tc' to capture the loop variable for parallel subtests.

itest/litd_node.go
bitromortac: Good point, have changed it such that we check for errors.

lnd_ready_test.go
bitromortac: Yes, was initially even thinking about deleting the test :). Moved it to another file pair.

bitromortac:
> but it probably makes sense to include extra lnd wallet check for other reasons

I saw that there was another itest flake with that, so I also tried to fix that.

ViktorT-11 (approved):
Thanks for the updates :pray:!

PR #1355, opened 2026-07-16
docs: Add new release notes for `v0.17.1-alpha`

ViktorT-11:
This PR adds a release notes doc for `v0.17.1-alpha`, as we have many open PRs which still currently adds their respective release notes entry for the doc for `v0.17.0-alpha`, despite that release already being published.

We also move 2 entries that were accidentally merged to the doc for `v0.17.0-alpha` despite not being part of that release.


bitromortac (approved):
> We also move 2 entries that were accidentally merged to the doc for v0.17.0-alpha despite not being part of that release.

Ah right, thanks! Perhaps you could also update the contributor list.

PR #1356, opened 2026-07-18
itest: stop litd nodes via litd's StopDaemon

Vandit1604:
This is split out of #1329 at @ViktorT-11's request, so it can be reviewed on its own. #1329 will be rebased on top of this.

### What this does

The itest harness stops a node through litd's own `StopDaemon` when it has a lit connection, and falls back to lnd's `StopDaemon` otherwise. The lit connection (`litConn`) was only ever set up on the no-seed path in `Start`, so seed based nodes (stateless init, seed restarts) always fell back to lnd.

Two changes:

1. `litConn` setup moves into `initLightningClient`, so every node gets one, seed or not.
2. Stateless init nodes write no macaroon files to disk, so there is no lit macaroon to authenticate with. For those, the harness now bakes a lit super macaroon from the admin macaroon it gets back from `Init` / `InitChangePassword`, writes it to the node's directory, and uses that for `litConn`.

This mirrors what a real stateless setup does:

```
lncli create --stateless_init --save_to /tmp/admin.macaroon
litcli --macaroonpath /tmp/admin.macaroon bakesupermacaroon --save_to /tmp/lit.macaroon
```

### Why it matters

On master this is a no-op in behaviour, since stopping lnd also brings litd down either way. It becomes necessary with #1329, where litd deliberately keeps running after lnd stops. Without it, seed based nodes stop lnd only, litd stays up by design, and the harness times out waiting for the process to exit.

### Testing

`terminal stateless init mode` passes on this branch on top of master.

### DIFF
--- a/itest/litd_node.go
+++ b/itest/litd_node.go
@@ -32,6 +32,7 @@ import (
 	"github.com/lightninglabs/loop/looprpc"
 	"github.com/lightninglabs/pool/poolrpc"
 	"github.com/lightninglabs/taproot-assets/taprpc"
+	"github.com/lightningnetwork/lnd/fn/v2"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/invoicesrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/routerrpc"
@@ -45,7 +46,10 @@ import (
 	"github.com/lightningnetwork/lnd/lntest/wait"
 	"github.com/lightningnetwork/lnd/macaroons"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 	"gopkg.in/macaroon.v2"
 )
 
@@ -776,31 +780,9 @@ func (hn *HarnessNode) Start(litdBinary string, litdError chan<- error,
 		return nil
 	}
 
-	err = hn.initLightningClient(conn)
-	if err != nil {
-		return fmt.Errorf("could not init Lightning Client: %w", err)
-	}
-
-	// Also connect to Lit's RPC port for any Litd specific calls.
-	litConn, err := connectLitRPC(
-		context.Background(), hn.Cfg.LitAddr(), hn.Cfg.LitTLSCertPath,
-		hn.Cfg.LitMacPath,
-	)
-	if err != nil {
-		return fmt.Errorf("could not connect to Lit RPC: %w", err)
-	}
-	hn.litConn = litConn
-
-	ctxt, cancel := context.WithTimeout(
-		context.Background(), lntest.DefaultTimeout,
-	)
-	defer cancel()
-	return wait.NoError(func() error {
-		litConn := litrpc.NewProxyClient(hn.litConn)
-
-		_, err = litConn.GetInfo(ctxt, &litrpc.GetInfoRequest{})
-		return err
-	}, lntest.DefaultTimeout)
+	// initLightningClient also connects to Lit's RPC port and sets
+	// hn.litConn so the node can be stopped via litd's own StopDaemon.
+	return hn.initLightningClient(conn, fn.None[[]byte]())
 }
 
 // WaitForLNDWalletReady waits until the wallet state flips from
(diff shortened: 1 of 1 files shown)


litbot-9000:

docs/release-notes/release-notes-0.17.1.md
```diff
@@ -27,6 +27,13 @@
 
 ### Technical and Architectural Updates
 
+* [Stop itest litd nodes via litd's own
```
ViktorT-11: As this PR only effects the tests, there's no need to add a release note entry for this PR :)

itest/litd_node.go
```diff
@@ -1087,6 +1075,60 @@ func (hn *HarnessNode) initClientWhenReady(timeout time.Duration) error {
...
+			RootKeyIdSuffix: 0,
+			ReadOnly:        false,
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	// BakeSuperMacaroon hex encodes the returned macaroon.
+	superMacBytes, err := hex.DecodeString(resp.Macaroon)
```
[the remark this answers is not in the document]
ViktorT-11: The request would have resulted in an error if that was the case, so no need to address this.

itest/litd_node.go
```diff
@@ -1087,6 +1075,60 @@ func (hn *HarnessNode) initClientWhenReady(timeout time.Duration) error {
...
+	if err != nil {
+		return err
+	}
+
+	macPath := filepath.Join(
+		filepath.Dir(hn.Cfg.LitMacPath), "lit-super.macaroon",
+	)
+	if err := os.WriteFile(macPath, superMacBytes, 0644); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
ViktorT-11: This is only in a test environment, so I'd view this as optional to address.

itest/litd_node.go
```diff
@@ -1252,6 +1316,48 @@ func (hn *HarnessNode) initLightningClient(conn *grpc.ClientConn) error {
...
+	// Also connect to Lit's RPC port for any litd specific calls. We do
+	// this here, rather than only on the no-seed path in start, so that
+	// seed based nodes (which reach this method via Init or Unlock) get a
+	// litConn too. Stop relies on it to shut the node down via litd's own
+	// StopDaemon, which is the only way to bring litd down now that its
+	// lifecycle is decoupled from lnd's. For stateless nodes litMacPath
+	// points at a baked super macaroon; otherwise the lit macaroon on disk
+	// is used.
+	macPath := hn.Cfg.LitMacPath
+	if hn.litMacPath != "" {
```
ViktorT-11: Instead of relying on `hn.litMacPath` being `!= ""`, I would recommend using the `fn.Option[string]` type, and default the field to `None` in the constructor, and then set it to `Some` once set. You can then use `hn.litMacPath.IsSome()` here.

ViktorT-11 (commented):
Generally looks great! I think this is very close to be ready to be merged :fire:!

Vandit1604:
Heads up on the red sqlite itest, it is a startup race rather than the review changes.

`bakeLitSuperMacaroon` ran inside `Init`, right after the wallet unlocks but before litd's macaroon service finishes coming up. On the slow SQL tranche the bake lost that race (`macaroon service has not started yet`, then `signature mismatch` once I added a retry). It has been latent since the harness commit and only tipped over now that master picked up the auto-bake-on-startup change in #1324 and the SQL-migration wait.

Fixed by moving the bake into `initLightningClient`, after `waitTillServerStarted`, so litd's macaroon service is up before we bake. Green locally, pushed to rerun CI.

docs/release-notes/release-notes-0.17.1.md
Vandit1604: Removed.

itest/litd_node.go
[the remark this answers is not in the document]
Vandit1604: Agreed, left as is.

itest/litd_node.go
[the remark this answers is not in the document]
Vandit1604: Done, switched to 0600.

itest/litd_node.go
Vandit1604: Done, litMacPath is now fn.Option[string].

itest/litd_node.go
```diff
@@ -384,6 +388,13 @@ type HarnessNode struct {
 	// litConn is the underlying connection to Lit's grpc endpoint.
 	litConn *grpc.ClientConn
 
+	// litMacPath, when set, is the macaroon file used to authenticate
+	// litConn. It is set for stateless init nodes, which have no lit
+	// macaroon on disk: we bake a super macaroon from the in-memory admin
+	// macaroon and point this at it so the node can still be stopped via
+	// litd's own StopDaemon.
+	litMacPath fn.Option[string]
```
ViktorT-11: With your other changes, I actually think this field now brings minimal value to be kept as a struct member of `HarnessNode`.

Instead, I would update `bakeLitSuperMacaroon` to return a `(string, error)`, and return the `macPath` it has generated when successful. You can then remove this field.

 Then you could update the callsite to baking function to:
```
// In stateless init mode no lit macaroon is written to disk, so bake a
// super macaroon from the in-memory admin macaroon. This has to happen
// after the server has started, since LiT only finishes setting up its
// macaroon service around then, and baking any earlier fails macaroon
// verification.
macPath := hn.Cfg.LitMacPath
var bakeErr error
statelessInitAdminMac.WhenSome(func(adminMac []byte) {
	macPath, bakeErr = hn.bakeLitSuperMacaroon(adminMac)
})
if bakeErr != nil {
	return fmt.Errorf("could not bake lit super macaroon: %w",
		bakeErr)
}
```

ViktorT-11 (approved):
Thanks for this, LGTM :pray::fire:! 

Leaving one comment to make the design even cleaner regarding the `litMacPath` field.

itest/litd_node.go
Vandit1604: Done, removed the field and return the path from bakeLitSuperMacaroon.

Vandit1604:
Hi @ViktorT-11 can we merge this if everything looks good?

litbot-9000:
@bitromortac: review reminder

litbot-9000:
@bitromortac: review reminder

Vandit1604:
Hi Viktor, small nudge on this one. It has been approved for a few weeks. If you merge it, I can rebase #1329 on top and get it green. That was our plan when we split this out. No rush if you are busy.

litbot-9000:
@bitromortac: review reminder

litbot-9000:
@bitromortac: review reminder

litbot-9000:
@bitromortac: review reminder

litbot-9000:
@bitromortac: review reminder

Vandit1604:
Hi @bitromortac, when you have a moment, would you mind taking a look at this one? Viktor has approved it, and it only changes `itest/litd_node.go`. #1329 is stacked on it, so a review here would help both move forward. Thanks!

PR #1359, opened 2026-07-21
multi: wait for lnd with a configurable timeout during the SQL migration

ellemouton:
### Problem

A node upgraded to litd 0.17.0 (lnd 0.21.1) started up but every API call
returned `lnd is not ready for: /lnrpc.Lightning/GetInfo`. The litd status
endpoint showed the `lit` sub-server errored with:

> could not create stores: error applying migrations to SQLite store: ...
> error listing macaroon IDs when migrating stores to SQL after 120 attempts:
> rpc error: code = Unknown desc = the RPC server is in the process of
> starting up, but not yet ready to accept calls

### Root cause

The kvdb-to-SQL data migration (`Mig6ProgrammaticMigration`) calls lnd's
`ListMacaroonIDs` RPC to fetch the macaroon root-key IDs needed for the
firewalldb migration. That RPC is served by lnd's main Lightning server,
which the interceptor rejects with `ErrRPCStarting` until lnd reaches its
`rpcActive` state (`rpcperms/interceptor.go`).

The migration polled this RPC a fixed **120 times with a 500ms delay = 60s**.
On a large routing node, lnd took **~76s** from wallet-unlock to `rpcActive`
(16s just to open the main bbolt DB, then building all server subsystems for
a node with ~97k graph channels), so the poll timed out ~16s too early.
The migration failure aborts litd startup; litd stays up but never marks the
LND sub-server ready, so the rpc proxy returns "lnd is not ready for ..." for
every call — even though lnd itself finished starting moments later.

Timeline from the reporter's log:

| Time | Event |
|------|-------|
| 18:01:05 | wallet unlocked → migration starts, poll attempt 1/120 |
| 18:02:05 | migration fails after 120 attempts (60s) |
| 18:02:20 | lnd reaches `rpcActive` (~16s too late) |

Because the migration is transactional with `ResetVersionOnError`, restarting
litd retries it, which is why a restart (warm DB cache → faster lnd) usually
recovers — making this timing/size dependent.

### Fix

Replace the fixed attempt cap with a wait bounded by a new, configurable
`--lndreadytimeout` option, defaulting to a generous **10 minutes** (~8x the
observed ~76s). The loop still aborts early on daemon shutdown (`ctx`
cancellation), and the returned error preserves lnd's real "RPC server is in
the process of starting up" message.

### Notes

- litd's other lnd-readiness waits (`setupBasicLNDClient` /
  `setupFullLNDClient`) actually loop **indefinitely**, bailing only on
  shutdown / lnd-death. I chose a *bounded* (but generous, configurable)
  timeout here deliberately, because this wait happens inside the migration's
  open SQL write transaction — an unbounded wait there is riskier. Happy to
  switch to unbounded if reviewers prefer matching the client-setup pattern.
- The `ListMacaroonIDs` fetch still happens inside the write transaction (as
  before). Moving it out would be a cleaner but larger refactor; left as a
  possible follow-up.

### DIFF
--- a/db/migsets/programmatic_migrations.go
+++ b/db/migsets/programmatic_migrations.go
@@ -21,12 +21,19 @@ import (
 	"github.com/lightningnetwork/lnd/sqldb/v2"
 )
 
+// listMacaroonIDRetryDelay is the delay between successive attempts to reach
+// lnd's ListMacaroonIDs RPC while waiting for lnd to become ready. 500ms keeps
+// the poll responsive (lnd is typically ready within a minute or two) without
+// busy-looping against a not-yet-ready RPC server.
+const listMacaroonIDRetryDelay = 500 * time.Millisecond
+
 // Mig6ProgrammaticMigration generates and returns the programmatic migration
 // entry containing the kvdb to SQL migration for all of litd's database stores.
 func Mig6ProgrammaticMigration(ctx context.Context,
 	basicClient lnrpc.LightningClient, db *sqldb.BaseDB,
 	accountsDir, networkDir string, clock clock.Clock,
-	migVersion uint) migrate.ProgrammaticMigrEntry {
+	migVersion uint,
+	lndReadyTimeout time.Duration) migrate.ProgrammaticMigrEntry {
 
 	mig6queries := sqlcmig6.NewForType(db, db.BackendType)
 	mig6executor := sqldb.NewTransactionExecutor(
@@ -49,6 +56,7 @@ func Mig6ProgrammaticMigration(ctx context.Context,
 				return kvdbToSqlProgrammaticMigration(
 					ctx, basicClient, accountsDir,
 					networkDir, db, clock, q6,
+					lndReadyTimeout,
 				)
 			}, sqldb.NoOpReset,
 		)
@@ -87,7 +95,8 @@ func Mig6ProgrammaticMigration(ctx context.Context,
 
 func kvdbToSqlProgrammaticMigration(ctx context.Context,
 	basicClient lnrpc.LightningClient, accountsDir, networkDir string,
-	_ *sqldb.BaseDB, clock clock.Clock, q *sqlcmig6.Queries) error {
+	_ *sqldb.BaseDB, clock clock.Clock, q *sqlcmig6.Queries,
+	lndReadyTimeout time.Duration) error {
 
 	start := time.Now()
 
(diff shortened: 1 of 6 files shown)


db/migsets/programmatic_migrations.go
```diff
@@ -185,41 +188,59 @@ func kvdbToSqlProgrammaticMigration(ctx context.Context,
...
 		select {
-		case <-ctx.Done():
-			return fmt.Errorf("context canceled while retrying "+
-				"to list macaroon IDs: %w", ctx.Err())
+		case <-waitCtx.Done():
+			return fmt.Errorf("error listing macaroon IDs "+
+				"when migrating stores to SQL after %d "+
+				"attempts (waited up to %v for lnd's RPC "+
+				"server to become ready): %w", attempt,
+				lndReadyTimeout, err)
```
ViktorT-11: nit: I think we should mention in the error that the user can increase the `lndreadytimeout` config option to increase the time litd waits here.

Additionally, we could potentially mention the actual time we waited here, as this error will be returned early if the user shuts down litd, and just mentioning the potential wait time can therefore be a bit confusing.

ViktorT-11 (approved):
Niiice! Thanks for the fix :pray:. Leaving just a tiny nit regarding an error message, but other than that I think this is ready to go :fire:!

db/migsets/programmatic_migrations.go
```diff
@@ -185,41 +188,59 @@ func kvdbToSqlProgrammaticMigration(ctx context.Context,
...
+	var (
+		macaroonIDList *lnrpc.ListMacaroonIDsResponse
+		attempt        int
 	)
+	for {
+		attempt++
 
-	var macaroonIDList *lnrpc.ListMacaroonIDsResponse
-	for i := 1; i <= maxListMacaroonIDAttempts; i++ {
 		macaroonIDList, err = basicClient.ListMacaroonIDs(
```
bitromortac: We should pass `waitCtx` instead of `ctx` to `basicClient.ListMacaroonIDs` so individual RPC calls respect the timeout budget. We can also update the `select` block to check `ctx.Err()` explicitly before `waitCtx.Done()` so that daemon shutdown during retries returns a context cancellation error rather than a readiness timeout error. Would be good to add a unit test for the elapsing timeout or ctx cancellation.

db/migsets/programmatic_migrations.go
```diff
@@ -185,41 +188,59 @@ func kvdbToSqlProgrammaticMigration(ctx context.Context,
...
-	// We'll fetch the macaroonIDList from `lnd` next. Note that since lnd's
-	// RPC servers may not have been fully started yet if the execution of
-	// accounts and session migration were really quick, we poll the request
-	// up to 120 times with a 0.5 second delay between the attempts. This
-	// should be a sufficient amount of time for the wallet to have been
-	// loaded and for the RPC servers to started.
-	const (
-		maxListMacaroonIDAttempts = 120
-		listMacaroonIDRetryDelay  = 500 * time.Millisecond
+	// We'll fetch the macaroonIDList from lnd next. This is a call to
```
bitromortac: nit: could be more succinct. I think the essential info is that we need lnd to be ready for being able to return the macaroon ids.

db/migsets/programmatic_migrations.go
ellemouton: Good points, done. The error now names `--lndreadytimeout` as the knob to raise, and reports the actual elapsed time (`%d attempts over %v`) instead of just the nominal budget — so an early abort on shutdown no longer implies we waited the full timeout. Shutdown is also now reported as a cancellation rather than a readiness timeout (see bitromortac's comment below).

db/migsets/programmatic_migrations.go
ellemouton: All three done. Pulled the poll loop out into `listMacaroonIDsWhenLndReady` so it is testable in isolation: it now passes `waitCtx` to `ListMacaroonIDs` (a single hanging call can no longer outlive the budget), and on `waitCtx.Done()` it checks the parent `ctx` explicitly first, so a shutdown mid-retry surfaces as `context.Canceled` rather than a readiness timeout. Added `TestListMacaroonIDsWhenLndReadyTimeout` and `TestListMacaroonIDsWhenLndReadyCancel` — the test lnd stub takes an optional error now so it can simulate an lnd that never becomes ready.

db/migsets/programmatic_migrations.go
ellemouton: Agreed, trimmed from 17 lines to 6 and led with the essential bit: the firewalldb migration needs lnd's macaroon root key IDs, and that RPC only serves once lnd is "RPC active". The "why the timeout is generous" rationale moved to the helper's doc comment and the retry-delay rationale to a package-level const, so it is next to the thing it explains rather than in the call site.

PR #1368, opened 2026-08-03
app: general ui improvements + sidebar + better mobile support

dstrukt:
### Summary

General UI improvements to the Terminal frontend, in four commits:

- **Sidebar**: pinned to the left edge of the window, so it no longer shifts when navigating to and from Pool. It collapses when the window is resized below 1200px, is expanded by default on desktop, and the page content no longer moves when it is toggled.
- **Home page**: removed the "What's different?" section and its screenshots, consolidated the connection copy into a single sentence, and moved Learn More inline as a text button. The plan is to consolidate the Home and Connect views, as the primary function of both views is to push to Terminal Web, thus simplifying the choice(s) in Lightning Terminal.
- **Mobile**: the Home and auth pages reflow down to phone widths, and the QR code scales to fit the modal instead of being clipped.
- **Theme**: the page background is now `lightningBlack`, ported from the terminal-web repo.

### Screenshots

Just general clean-up and adding some responsive tags to help this function better on mobile, starting with the Home page, will move on to the other views in future PRs.

Screenshot 2026-08-03 at 10 32 02 AM
Screenshot 2026-08-03 at 10 32 07 AM
Screenshot 2026-08-03 at 10 32 17 AM
Screenshot 2026-08-03 at 10 32 24 AM
Screenshot 2026-07-30 at 4 26 08 PM

### Technical Notes

- `Aside` is fixed positioned but had no `left` offset, so it fell back to its static position at the centered container's edge. It and the toggle now render outside of the container with `left: 0`. The toggle has to move as well, since it is absolutely positioned while collapsed and would otherwise still resolve against the container.
- The collapse breakpoint was only checked on startup. It now also runs from a debounced resize handler via `SettingsStore.syncAutoCollapse()`, and only acts when the window crosses the breakpoint so that it does not override the sidebar being toggled by hand.
- `Content` now reserves the sidebar's width in both states. Collapsing on desktop leaves an empty gutter rather than handing the space back to the page, so Loop and History no longer widen when it is collapsed. Pool is unaffected, since it is full width and already ignored the margin.
- `min-width: 1024px` moved off `html/body/#root` and into `Layout`, where it is applied to the pages that are laid out for desktop widths (keyed off a new `appView.responsive`). A viewport meta tag was added, which the app was missing entirely. Pool, Loop and History still scroll horizontally instead of being squeezed. Making them responsive is follow-up work.
- The QR code was a fixed 500px canvas, clipped by the modal below roughly 600px wide. It is now `QRCodeSVG` with `max-width: 100%`, which scales cleanly because the svg carries a `viewBox`.
- `colors.blue` is deliberately unchanged. 30 call sites use it, mostly as a foreground color on light surfaces, and against the darker background it now reads as a raised panel.

### Steps to Test

1. On a window wider than 1440px, navigate between Home and Pool. The sidebar should stay flush against the left edge of the window.
2. Click the hamburger on any page. The page content should not move, horizontally or vertically.
3. Drag the window below 1200px. The sidebar should collapse rather than cover the content, and expand again above 1200px. Reload while wide and confirm it comes back expanded.
4. At a phone width such as 414px, confirm Home and the auth page reflow with no horizontal scrolling, the connect buttons stack, and the hamburger does not overlap the page title. Pool and Loop should still scroll horizontally, unchanged.
5. Open "Connect with QR" at a narrow width and confirm the code scales to fit the modal.
6. Confirm the page background is black, the sidebar border disappears when collapsed, and the nav hover indicator is no longer pink while the active item's white bar is unchanged.

### Related Issues & Pull Requests

Closes #

Depends on:


litbot-9000:
@jamaljsr: review reminder

litbot-9000:
@jamaljsr: review reminder

jamaljsr (approved):
tACK LGTM 👌

Issue #583, opened 2023-07-03
more control of LND account spending rates


AndySchroder:
With LND accounts we can allocate an initial balance to the account and also set an expiration date of the account.

It would be nice to have the following additional features:

- Daily/hourly spending limit amount.
- Max sending transaction size.
- Max number of transaction allowed to be sent per second, hour, and day.

These additional controls can help in the following types of situations:

- Accounts for children or employees where the user is spending someone else's balance and we would like to but some guide rails on what they can do.
- Providing a macaroon to a service provider so that they can automatically request payment for their service on a regular interval.
- Limit machine to machine payments where the machine may be easier to break into and steal the macaroon (think a car like in [Distributed Charge](http://andyschroder.com/DistributedCharge/)).
- Limit machine to machine payments in case the machine gets out of control (software bug) or the user doesn't realize their high product consumption rate or seller's high pricing structure.
- Implementing something like the [Bolt Card](https://www.boltcard.org/).

levmi:
This is exactly the type of feedback that we were hoping to get when launching the MVP version of LND accounts. Great ideas and love the tie to specific use cases. We'll take this feedback into account and start to explore how we could implement these features/functionality. Thanks for the feedback! It's super helpful!

AndySchroder:
> It would be nice to have the following additional features:
> 
>     * Daily/hourly spending limit amount.
> 
>     * Max sending transaction size.
> 
>     * Max number of transaction allowed to be sent per second, hour, and day.


Looks like we now have the "Max sending transaction size." ask in https://github.com/lightninglabs/lightning-terminal/pull/1369 as a global option! We might want to add a per account version of this too though.

PR #1369, opened 2026-08-06
accounts: add configurable account spending controls

### LINKED ISSUE #495: validate whether the sum of all account balances is smaller or equal to the channel balance held in the underlying LND node
In the documentation, it says "The accounts feature does not validate whether the sum of all account balances is smaller or equal to the channel balance held in the underlying LND node."

It seems like every time a new account is created, a check should exist to make sure that the amount being allocated to the account is not greater than the total outbound capacity of the node minus the current sum of the amounts in all the other accounts.

**Additional context**
It's possible this statement that I've quoted is being misinterpreted? Is there some limitation related to channel capacity inbound vs outbound that makes this complicated and I don't realize it?

ViktorT-11:
Fixes #495 & #583.

In this PR the accounts system is enhanced by two a additional config options.

1. A `max-payment-size-msat` option allows the account admin to set the max cap for the total of individual payment amounts + fee.

2. A `check-channel-balance` flag is added, which if set, makes the accounts system reject account balance allocations (new accounts, administrative credits and administrative balance increases) that would push the sum of all account balances above the node's available total local (outbound) channel balance.

### DIFF
--- a/accounts/service.go
+++ b/accounts/service.go
@@ -19,9 +19,21 @@ import (
 )
 
 // Config holds the configuration options for the accounts service.
+//
+//nolint:ll
 type Config struct {
 	// Disable will disable the accounts service if set.
 	Disable bool `long:"disable" description:"disable the accounts service"`
+
+	// MaxPaymentSizeMsat, when greater than zero, is the maximum value (in
+	// millisatoshis) that a single account payment may have. Payments whose
+	// amount exceeds this cap are rejected by the account interceptor. This
+	// provides a guard rail against a compromised or misbehaving account
+	// macaroon draining its balance in a single large payment.
+	//
+	// It defaults to 0, which disables the cap and preserves the historical
+	// behaviour of allowing payments up to the full account balance.
+	MaxPaymentSizeMsat uint64 `long:"max-payment-size-msat" description:"the maximum total amount in millisatoshis, including fees, that a single account payment may debit; 0 (the default value) disables the cap"`
 }
 
 // trackedPayment is a struct that holds all information that identifies a
@@ -56,6 +68,10 @@ type InterceptorService struct {
 
 	routerClient lndclient.RouterClient
 
+	// maxPaymentSize, when greater than zero, is the maximum value that a
+	// single account payment may have. See Config.MaxPaymentSizeMsat.
+	maxPaymentSize lnwire.MilliSatoshi
+
 	mainCtx       context.Context
 	contextCancel fn.Option[context.CancelFunc]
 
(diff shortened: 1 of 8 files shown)


accounts/service.go
```diff
@@ -34,6 +34,22 @@ type Config struct {
...
+	//
+	// It defaults to false to preserve the historical behaviour, where the
+	// operator is trusted to manage over-provisioning themselves; accounts
+	// can legitimately be created before channels are opened or funded.
+	//
+	// NOTE: The total channel balance may still decrease below the already
+	// allocated account balance. This can occur if the node operator
+	// decreases the total channel balance through non-account related
+	// activity.
+	CheckChannelBalance bool `long:"check-channel-balance" description:"reject account balance allocations that would push the sum of all account balances above the node's available local channel balance. Note that The total channel balance can still decrease below the already allocated account balance. This can occur if the node operator decreases the total channel balance through non-account related activity."`
```
ellemouton: nit: `Note that The` -> `Note that the`

ellemouton: nit: perhaps a small note in docs/accounts.md?

accounts/service.go
```diff
@@ -377,6 +417,26 @@ func (s *InterceptorService) UpdateAccount(ctx context.Context,
...
+	// If the balance is being increased, make sure the increase doesn't
+	// over-provision the node's available local channel balance (no-op
+	// unless enabled).
+	if accountBalance >= 0 {
+		current, err := s.store.Account(ctx, accountID)
+		if err != nil {
+			return nil, fmt.Errorf("fetching account: %w", err)
+		}
+
+		newBalanceMsat := int64(accountBalance) * 1000
```
ellemouton: duplicate of line 409 above? can we just use `balance` again rather? 

also, for both could probs use `lnwire.NewMSatFromSatoshis(accountBalance)` instead of manual multiplication

accounts/service.go
```diff
@@ -502,13 +570,59 @@ func (s *InterceptorService) CheckBalance(ctx context.Context, id AccountID,
...
+//
+// NOTE: the service lock MUST be held when calling this method.
+func (s *InterceptorService) checkChannelBalanceReservationUnsafe(
+	ctx context.Context, delta lnwire.MilliSatoshi) error {
+
+	// The check is opt-in. It is also skipped if the service hasn't been
+	// started with a lightning client yet, so that account operations don't
+	// block on an unavailable dependency.
+	if !s.checkChannelBalance || s.lightningClient == nil {
+		return nil
```
ellemouton: if lightning client not available... should return error no?

accounts/service.go
```diff
@@ -502,13 +570,59 @@ func (s *InterceptorService) CheckBalance(ctx context.Context, id AccountID,
...
+	}
+
+	var allocated int64
+	for _, acct := range accounts {
+		allocated += acct.CurrentBalance
+	}
+
+	// The node's spendable (outbound) capacity is the sum of the local
+	// balances across all open channels.
+	chanBalance, err := s.lightningClient.ChannelBalance(ctx)
```
ellemouton: not super ideal that we are doing a network call under a lock. might be worth fetching the chan balance before locking or at the very least commenting acknowledging it

ellemouton (commented):
Nice fixes! couple comments

accounts/service.go
ViktorT-11: Added a section for both of these config options in `docs/accounts.md` in a separate commit :)

accounts/service.go
ViktorT-11: Good call! Updated :)

accounts/service.go
ViktorT-11: Good catch! Added a fix for this in the new function which addresses https://github.com/lightninglabs/lightning-terminal/pull/1369#discussion_r3732208251

accounts/service.go
ViktorT-11: Added a new `fetchChannelBalance` function which is executed outside of the loc to address this issue :)

ViktorT-11:
Thanks for the review @ellemouton :tada:! I've addressed your feedback with the latest push.

ellemouton (approved):
noice!

PR #1371, opened 2026-08-10
terminal, config: fix tapd startup deadlock, require an HTLC interceptor

GeorgeTsagk:
## Description

Fixes a startup deadlock in integrated mode, and closes an HTLC forwarding gap while tapd's interceptor isn't attached.

- litd waited for lnd to be synced to chain before starting the sub-servers. lnd only reports synced once its blockbeat has caught up, and block processing can block on tapd's aux sweeper, which blocks until tapd starts, which only happened after that wait. With a channel being resolved on chain, litd stalled 60 seconds per block and tapd never came up. litd now starts tapd first, then waits for chain sync itself, then starts the remaining sub-servers, so loop, pool and faraday still start against a synced lnd.

- We set lnd's `requireinterceptor` when tapd runs in-process, so lnd fails forwards back rather than forwarding them with no RFQ policy check while tapd's interceptor isn't attached.

Both are gated on lnd *and* tapd being integrated.

### Reviewer notes

The first version skipped the chain sync wait outright, which let faraday, loop and pool start against a catching-up lnd. @bitromortac pointed out the narrower shape, and this version does that instead.

His prototype also gates lnd's sweeper on an explicit aux readiness signal. Once tapd is started before anything waits on the sync, the sweeper waits until tapd is up either way, so it isn't a dependency here and belongs in lnd on its own merits.

The itest is his.

`requireinterceptor` is independent of the startup shape. tapd registers its interception stream in a goroutine, so no start-time signal closes that gap, and it reopens whenever the stream is re-established.

### DIFF
--- a/config.go
+++ b/config.go
@@ -997,6 +997,21 @@ func loadConfigFile(preCfg *Config, interceptor signal.Interceptor) (*Config,
 			cfg.Lnd.ProtocolOptions.CustomMessage, lnwire.MsgError,
 		)
 
+		// tapd's RFQ subsystem enforces the agreed upon quote for asset
+		// HTLCs through lnd's HTLC interceptor. Whenever no interceptor
+		// is attached, lnd forwards HTLCs without any of those checks.
+		// That window is narrow, as tapd's aux traffic shaper blocks
+		// forwards until tapd is ready, but it is real: it exists
+		// whenever tapd has to re-establish its interception stream,
+		// and for the brief moment between tapd signalling readiness
+		// and its interceptor actually registering. We therefore
+		// require an interceptor to be present whenever tapd runs
+		// in-process, which makes lnd fail forwards back instead of
+		// forwarding them unchecked.
+		if cfg.TaprootAssetsMode == ModeIntegrated {
+			cfg.Lnd.RequireInterceptor = true
+		}
+
 		var err error
 		cfg.Lnd, err = lnd.ValidateConfig(
 			*cfg.Lnd, interceptor, fileParser, flagParser,
--- a/docs/release-notes/release-notes-0.17.5.md
+++ b/docs/release-notes/release-notes-0.17.5.md
@@ -1,0 +1,54 @@
+# Release Notes
+
+- [Lightning Terminal](#lightning-terminal)
+    - [Bug Fixes](#bug-fixes)
+    - [Functional Changes/Additions](#functional-changesadditions)
+    - [Technical and Architectural Updates](#technical-and-architectural-updates)
+- [Integrated Binary Updates](#integrated-binary-updates)
+    - [LND](#lnd)
+    - [Loop](#loop)
+    - [Pool](#pool)
+    - [Faraday](#faraday)
+    - [Taproot Assets](#taproot-assets)
+- [Contributors](#contributors-alphabetical-order)
+
+## Lightning Terminal
+
+### Bug Fixes
+
+* [Fix a startup deadlock when tapd runs in integrated
+  mode](https://github.com/lightninglabs/lightning-terminal/pull/1371):
+  litd waited for lnd to be fully synced to chain before starting any of its
+  sub-servers, but lnd only reports itself as synced once its blockbeat has
+  caught up, and block processing can be blocked on tapd's aux sweeper, which
+  only becomes available once tapd is started. With a channel being resolved on
+  chain, litd stalled for 60 seconds per block, and for as long as blocks kept
+  arriving faster than that, tapd never came up at all. Any channel triggers
+  this, not just asset channels. litd now starts tapd first, then waits for
+  lnd's chain sync, then starts the remaining sub-servers, so loop, pool and
+  faraday still start against a synced lnd. We additionally require an HTLC
+  interceptor to be attached whenever tapd runs in-process, so lnd fails
+  forwards back instead of forwarding them without any RFQ policy checks.
+
+### Functional Changes/Additions
+
+### Technical and Architectural Updates
+
+## RPC Updates
+
+## Integrated Binary Updates
+
+### LND
+
+### Loop
+
+### Pool
+
+### Faraday
+
+### Taproot Assets
+
+# Contributors (Alphabetical Order)
+
+* bitromortac
+* George Tsagkarelis
--- a/itest/litd_startup_stall_test.go
+++ b/itest/litd_startup_stall_test.go
@@ -1,0 +1,301 @@
+package itest
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/lightninglabs/lightning-terminal/litrpc"
+	"github.com/lightninglabs/lightning-terminal/subservers"
+	"github.com/lightningnetwork/lnd/lnrpc"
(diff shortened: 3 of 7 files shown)


bitromortac (commented):
Here an (llm) analysis. I tried to see if this can be reproduced with an itest (see here  https://github.com/bitromortac/lightning-terminal/tree/pr-1371-force-close-stall-itest - needs a cleanup) and was able to analyze it a bit. The tldr is that it can be reproduced with an itest and that only removing the sync wait isn't enough, we need to start tapd before the other subservers. It could be that running the lnd/tapd bundle on the tap repo has a similar problem (but haven't looked into that further).

## It reproduces in an itest, and any channel will do

`itest/litd_startup_stall_test.go` reproduces it deterministically: force close a channel, mine past the CSV while litd is down, restart, and measure how long tapd takes to report running.

The channel does **not** have to be an asset channel. lnd asks every configured aux sweeper for an extra sweep output whenever one is present (`fn.MapOptionZ(t.cfg.AuxSweeper, ...)`, `sweep/fee_bumper.go:1659`), and tapd's `DeriveSweepAddr` does no asset-input filtering before it blocks. A vanilla, non-asset force close is therefore enough to reach into tapd and park the caller, which makes the exposure much wider than "nodes with asset channels".

## It is a stall, not a deadlock

The wait is bounded. lnd's blockbeat dispatcher gives up on a stuck consumer after `DefaultProcessBlockTimeout` (60s in v0.21), logs the failure and carries on, so the startup resumes on its own:

```
CHIO Height[472]: Failed to process block: consumer TxPublisher:
                  process block timeout
CHIO Height[472]: Notified all consumers on new block in 1m0.007482933s
```

The blocked consumer is `TxPublisher`, not `UtxoSweeper` — `prepareSweepTx` and its `DeriveSweepAddr` call live in `sweep/fee_bumper.go`. It becomes unbounded only while the chain *backend* is still feeding lnd blocks faster than one per timeout (initial block download, neutrino catching up), which is presumably where the "retry forever" observation came from. Worth restating in the commit message and release note, because someone debugging a slow startup will not recognise the symptom from the word "deadlock".

## Skipping the chain sync wait alone does not fix it

Measured on this branch as submitted, three runs: **1m0.24s / 1m0.01s / 1m0.22s** for tapd to come up. litd itself no longer waits, but `StartIntegratedServers` iterates a Go **map**, so loop, pool or faraday can be started before tapd — and their startup calls `GetInfo`. `getChainSyncInfo` calls `BlockbeatDispatcher.CurrentHeight()`, which is served by the same goroutine that is blocked notifying consumers, so that sub-server inherits the exact wait this PR removed from litd. Whether a given startup pays the 60s depends on map iteration order.

```
LITD  Not waiting for lnd to be synced ...   <- the change works
LITD  Full lnd client connected              <- 1s
LOOPD Protocol version: MuSig2               <- loop starts first
RPCS  [/lnrpc.Lightning/GetInfo] requested   <- and blocks
      ... 60s ...
CHIO  Height[472]: consumer TxPublisher: process block timeout
LOOP  Connected to lnd node 'Stalled'        <- GetInfo returns
TSVR  Version: 0.8.0-alpha                   <- tapd finally starts
```

## So tapd's start has to be separated out

Starting tapd before the sub-servers that query lnd's sync state closes it, and is a complete fix rather than a mitigation: `waitForReady` blocks until tapd is ready, not for a fixed 60s, and nothing in tapd's own startup depends on the blocked dispatcher. Verified against stock tapd v0.8.0:

```
LITD  Full lnd client connected
TSVR  Version: 0.8.0-alpha            <- tapd first, +3ms
TCHN  Starting AuxSweeper
SWPR  Publishing sweep tx ..., num_inputs=1
CHIO  Height[472]: Notified all consumers in 167.872926ms
LOOPD Protocol version: MuSig2         <- loop after tapd
```

**167.87ms instead of 1m0.008s**, the parked sweep broadcasts immediately, and the force close resolves one block later. It also removes the map-order non-determinism and gives loop/pool/faraday a synced lnd again.

One warning for anyone tempted by the obvious upstream fix: making tapd's aux methods **fail** instead of block does remove the stall (402ms), but lnd classifies the aux error as fatal, drops the monitor record and the input, and
the commitment output is then never swept — the sweeper reports `attempt sweeping 0 inputs` forever. That is why the itest drives the force close all the way to resolution; a test that only asserts prompt startup passes against that regression.

GeorgeTsagk:
Thanks for the thorough review @bitromortac. Added your test.

Ready for another round.

litbot-9000:
@viktort-11: review reminder
@bitromortac: review reminder

litbot-9000:
@viktort-11: review reminder
@bitromortac: review reminder

bitromortac (commented):
I have an alternative design here, I think it may be better to make the dependency of the sweeper on the aux components explicit. A downside to this is that it requires an lnd change. Not sure if a ready channel on the aux component was discussed before.

Here the branch https://github.com/bitromortac/lightning-terminal/tree/prototype-aux-ready-gate and here the lnd branch https://github.com/bitromortac/lnd/tree/prototype-aux-ready-gate.

GeorgeTsagk:
Thanks, the sub-server split is better than what I had. Taking it: tapd first, then wait for the sync in litd, then start the rest, so loop, pool and faraday keep coming up against a synced lnd. That was the loose end in my reviewer notes.

I'd leave the sweeper gate out of this PR though. With tapd started first, the sweeper waits until tapd is up either way, so the gate doesn't shorten anything here, and it costs us an lnd dependency. The durability argument is real, a sync wait reintroduced ahead of tapd can't deadlock with the gate in place, so I think it's worth its own lnd PR rather than a blocker for this one.

On requireinterceptor, a readiness signal doesn't cover it. tapd registers the interception stream in a goroutine, so Start returns before lnd gets to SetInterceptor, and the gap reopens on every re-establish regardless. Keeping that commit as is.

Does that split of work sound good to you?

bitromortac:
> Thanks, the sub-server split is better than what I had. Taking it: tapd first, then wait for the sync in litd, then start the rest, so loop, pool and faraday keep coming up against a synced lnd. That was the loose end in my reviewer notes.
> 
> I'd leave the sweeper gate out of this PR though. With tapd started first, the sweeper waits until tapd is up either way, so the gate doesn't shorten anything here, and it costs us an lnd dependency. The durability argument is real, a sync wait reintroduced ahead of tapd can't deadlock with the gate in place, so I think it's worth its own lnd PR rather than a blocker for this one.
> 
> On requireinterceptor, a readiness signal doesn't cover it. tapd registers the interception stream in a goroutine, so Start returns before lnd gets to SetInterceptor, and the gap reopens on every re-establish regardless. Keeping that commit as is.
> 
> Does that split of work sound good to you?

The pro aux ready channel argument was that we can guarantee chain sync for other subdaemons and to explicitly signal that the sweeper is dependent on aux components. My main goal was to not change behavior for having chain sync read for other daemons. Tapd seems fine to be started without it as it's able to heal when the chain gets synced.

litbot-9000:
@georgetsagk, remember to re-request review from reviewers when ready

litbot-9000:
@viktort-11: review reminder
@georgetsagk, remember to re-request review from reviewers when ready

litbot-9000:
@viktort-11: review reminder
@georgetsagk, remember to re-request review from reviewers when ready

PR #1372, opened 2026-08-10
app: update outdated frontend dependencies

jamaljsr:
Updates the outdated packages `yarn audit` reports in `/app`, from 419 findings across 42 packages down to 16 across 9. Supersedes #1339, #1361, #1354, #1332 and #1302, each at or above the version those PRs proposed.

Three commits: production dependencies, the build-time tree, then a CI Node bump.

The CI bump is not cosmetic. The frontend job was pinned to Node 16, and yarn 1 throws `Found incompatible module` rather than warning when a non-optional dependency's `engines` doesn't match, so `yarn install --frozen-lockfile` would have failed outright on the new lockfile (`tar` and `esbuild` want 18+, `serialize-javascript` wants 20+). Node was also only pinned in that one job: `backend`, `proto-compile-check` and `itest` all referenced `matrix.node_version` without declaring it in their matrix, so it resolved to an empty string and `setup-node` fell back to whatever the runner shipped, including for two jobs that run `yarn install --frozen-lockfile`. Both are fixed by a `NODE_VERSION` env var. It replaces the matrix entry rather than feeding it, because the `env` context isn't available inside `strategy.matrix`.

## react-router

The only part that touches app code. `react-router-dom` goes to 6.30.4 for the untrusted-path open redirect, and drags in:

- 6.4 folded `history` into `@remix-run/router`, so it leaves the tree entirely. We imported `createBrowserHistory` and `History` from it directly while relying on it transitively, so the store now imports from `@remix-run/router`.
- `@remix-run/router` uses `get` accessors in its type declarations, which needs TypeScript 4.3+. We were on 4.1.6, so typescript goes to 4.9.5. `skipLibCheck` doesn't help, these are parse errors. TS 4.4 also started typing `catch` variables as `unknown`, so `handleError` now takes `unknown` and narrows before reading `.message`, and the five sites that read it directly narrow first too. That fixes a real case: a non-`Error` reaching `handleError` previously produced `notify(undefined)` and showed a blank alert.
- `react-router-dom` is pinned exactly rather than with a caret. `@remix-run/router` is a direct dependency at an exact version and react-router-dom pins its own copy exactly, so a caret would let a future patch release install a second router and hand `<HistoryRouter>` a history built by a different module instance.
- The new history accepts one active listener, but `RouterStore` and `<HistoryRouter>` each register one, so `RouterStore` now subscribes once and fans updates out. It also needs `v5Compat: true`, otherwise listeners never fire on `push()`. Those two accounted for 245 test failures.

Note this doesn't lower the react-router count. It moves past the one item with a published fix, but 6.4+ and 6.30.2+ flag two more that have no fix in the v6 line, and 6.30.4 is the last v6 release. None of them apply to us either way: our whole surface is `{ Route, Routes }` and `{ unstable_HistoryRouter }`, no `<Link>`, no `useNavigate`, no SSR. I'd still take it, since the undeclared `history` dependency was worth fixing regardless.

## Dev tree

Mostly lockfile-only. `yarn upgrade` doesn't re-resolve transitive entries in yarn 1, so the affected entries were dropped from `yarn.lock` and reinstalled, picking the newest version each parent's range already allows. That covers the four dependabot PRs and about a dozen others, and is why the lockfile diff is large next to the manifest.

Two things worth a look:

- `browserslist` and `caniuse-lite` had to move too. `@babel/core` 7.29 emits `opera_mobile` and the pinned `@babel/helper-compilation-targets@7.17.7` rejects it, which broke `yarn build` while `tsc` and tests stayed green.
- The `svgo` and `resolve-url-loader` resolutions force majors past what their parents pin, so I checked output rather than assuming. The production bundle is byte for byte identical with and without `svgo` (react-scripts sets `svgo: false`, so that plugin never runs), and `resolve-url-loader` 5 leaves the emitted CSS unchanged down to the content hash.

## Still reported

Nine packages, each either without a published fix or held back by a parent that pins an incompatible range. None of them are reachable from the shipped bundle, and I went through each one against how we actually use it:

- `webpack-dev-server` is the one I'd most like to move, and it only applies while `yarn start` is running. wds 5 removed the `onBeforeSetupMiddleware` and `onAfterSetupMiddleware` hooks react-scripts 5.0.1 calls directly, so forcing it breaks `yarn start`. The items that are specific to proxy config don't apply as we have it set up, since our proxy filters on the RPC path prefixes.
- `lodash.pick` is a phantom. reactour declares it but never imports it, so it isn't in the bundle. Deleting the package outright still builds, byte for byte identical. It'll keep showing in `yarn audit` regardless, since that walks the manifest and not the import graph.
- The `protoc-gen-js` chain has nothing published to move to, and only runs on `yarn protos` against the protoc release archive.
- `ip` sits behind `webpack-dev-server > bonjour > multicast-dns > dns-packet` for mDNS discovery. No app code reaches it.
- react-router, as above.

If we want the remaining gap closed, getting off react-scripts clears `webpack-dev-server` and `ip` together. Separate project.

## Testing

`yarn lint`, `yarn tsc`, `yarn test:ci` (432/432; master is 431, plus one new case covering a non-`Error` reaching `handleError`), `yarn build` and `yarn build-storybook` all pass.

Release builds run on the Dockerfiles' Node, not CI's, so I ran `yarn install --frozen-lockfile` and `yarn build` under Node 22 as well: both succeed and the bundle is byte for byte identical to the Node 24 one.

`yarn start` is the one path CI doesn't cover, so it's worth a manual click through given the `RouterStore` change.


ViktorT-11 (approved):
Nice, thanks :pray:! LGTM

Issue #644, opened 2023-09-26
Start retrying/restarting the various sub-servers if they error out. 


ellemouton:
Before #541, if Lit failed to connect to a remote LND instance, then Lit itself would also shutdown. If users are running Lit in a docker container, then this would be fine since docker would continually try to restart LiT until it succeeded. This is useful in cases where a user wants to restart LND. 

The reason for adding the status server is that if LiT does fail to start for some reason that needs to be addressed by the user, then the user would still be able to query the status server to see which sub-system is failing. This is nice but now Lit wont shutdown (and hence wont be restarted by the docker container) until the user specifically requests it to (via `litcli stop`). 

To address this, we should let LiT do this "restarting" instead. So if it fails to start anything, it should retry itself. For example: many of the subservers depend on LND and so will throw a critical error if LND shuts down. So if this happens, LiT should pick it up, continue to try and reconnect to LND and once it has succeeded, it should try restart all the dependent sub-servers.

Vandit1604:
Picking this up. This is the natural follow-up to #1329 (keep litd running after lnd stops), which tears down the lnd-dependent sub-servers and keeps the status endpoint alive but never tries to recover. I'd stack the reconnect work on that branch.

Proposed shape:

- Hook the reconnect into the `lndQuit` handler in `waitForShutdown`, right after `shutdownLndDependentSubServers`. A bounded backoff loop re-dials lnd and, on success, restarts the lnd-dependent sub-servers and flips status back to running. It also refreshes the connection ID (the existing TODO in `terminal.go`).
- Extract the connect + bring-up tail of `start` so both initial startup and reconnect share one code path.
- Scope v1 to **remote mode** only. In integrated mode lnd runs in-process via `lnd.Main` over a bufconn that dies with it, so "reconnect" means relaunching `lnd.Main` in-process — big and fragile, and a container restart is the intended recovery there. Happy to do integrated as a follow-up if you want it.

One open question: after the sub-servers are stopped, is restarting the same loop/pool/tapd/faraday instances supported, or should I rebuild fresh instances on reconnect? Leaning towards rebuilding to avoid restart-after-stop edge cases. Any preference?

I'll add an itest that bounces a remote lnd and asserts the sub-servers return to running.

Vandit1604:
[...]

PR #1373, opened 2026-08-13
subservers: report remote sub-server disconnects at runtime

Vandit1604:
### Change

In remote mode, a sub-server that disconnected after startup was still reported as `running` by the status server, so `litcli status` kept showing it as healthy. Only failures during startup were tracked; a disconnect at runtime was not.

This adds a watcher on each remote sub-server's gRPC connection. When the connection drops (`TransientFailure`/`Shutdown`) the sub-server is marked errored, and when it recovers (`Ready`) it is marked running again, so the status server reflects the real runtime state.

Two smaller things came along with it:
- `Manager.Stop` now also tears down remote sub-servers (it skipped them before), so the watcher and the remote connection are stopped on shutdown. The remote branch in `stop()` was effectively dead before, since remote servers never reached it.
- `status.SetErrored` logged "could not start the X sub-server", which is wrong for a runtime disconnect, so the log is now neutral.

Note: the watcher nudges an idle connection with `conn.Connect()` so a disconnect is still observed when no requests are in flight. That keeps the remote connection warm rather than letting it idle out, which is the trade-off for observability here.

### Scope

This is the first, smallest step towards #644, as suggested by @ViktorT-11: get the status reporting right before adding any reconnect/restart logic. It intentionally does not restart sub-servers or touch the lnd reconnect path.

Known gaps left as follow-ups:
- Integrated `pool`/`faraday` expose no runtime error channel, so their disconnects are still not tracked.
- Actually restarting the lnd-dependent sub-servers after a disconnect.
- A remote server that is already down at connect time is briefly shown as running until the first failure transition (the connection is dialled lazily).

### Testing

`TestWatchRemoteConn` starts a real gRPC backend, stops it, and asserts the disconnect is reported, then restarts it and asserts the recovery is reported.

Part of #644.

### DIFF
--- a/subservers/subserver.go
+++ b/subservers/subserver.go
@@ -1,6 +1,7 @@
 package subservers
 
 import (
+	"context"
 	"fmt"
 	"sync"
 
@@ -8,6 +9,7 @@ import (
 	"github.com/lightningnetwork/lnd/lncfg"
 	"github.com/lightningnetwork/lnd/lnrpc"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 )
 
 const (
@@ -57,8 +59,10 @@ func (s *subServerWrapper) setStarted(started bool) {
 // stop the subServer by closing the connection to it if it is remote or by
 // stopping the integrated process.
 func (s *subServerWrapper) stop() error {
-	// If the sub-server has not yet started, then we can exit early.
-	if !s.started() {
+	// An integrated sub-server that never started has nothing to stop. A
+	// remote sub-server has no started flag, so always tear it down to
+	// close its connection and stop its connection watcher.
+	if !s.Remote() && !s.started() {
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


litbot-9000:
@viktort-11: review reminder

litbot-9000:
@viktort-11: review reminder

litbot-9000:
@viktort-11: review reminder

subservers/subserver.go
```diff
@@ -150,3 +154,72 @@ func (s *subServerWrapper) connectRemote() error {
...
 	return nil
 }
+
+// watchRemoteConn watches the remote sub-server's connection and reports its
+// runtime health through the given callbacks. onError is called when the
+// connection is lost after startup and onRunning when it recovers, so the
+// status server reflects a runtime disconnect instead of keeping the
+// sub-server marked as running. The watcher stops when the sub-server is
+// stopped.
+func (s *subServerWrapper) watchRemoteConn(onError func(error),
```
Cyberguru1, line 169: It might be worth adding an `isWatching` flag to `subServerWrapper` to make `watchRemoteConn` idempotent. Checking this before starting the [goroutine](https://github.com/lightninglabs/lightning-terminal/pull/1373/changes#diff-4c119e8549de2006a66039647051ef88fb4e0cbb6b3d2284e45c740efc1a0fe0R172) lets any extra calls exit early, so we avoid spawning duplicate goroutines and triggering duplicate error callbacks.

subservers/subserver_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	erroredChan := make(chan struct{}, 8)
+	runningChan := make(chan struct{}, 8)
+	ss.watchRemoteConn(
+		func(error) { erroredChan <- struct{}{} },
+		func() { runningChan <- struct{}{} },
+	)
+	defer func() {
+		close(ss.quit)
+		ss.wg.Wait()
```
Cyberguru1: Could we switch this to `require.NoError(t, ss.stop())` so we exercise the full teardown logic and close `remoteConn` properly.

 Just like `Name()`, we will also need to add `Remote()` to `nameOnlySubServer`. When I tested `ss.stop()` locally without it, it panicked with a nil pointer dereference when `stop()` evaluated `s.Remote()`.

Cyberguru1 (changes requested):
Great contribution, @Vandit1604 🙌

I’ve done a first-pass review and left a few comments below.

subservers/subserver.go
Vandit1604: Done.

subservers/subserver_test.go
Vandit1604: Done.

litbot-9000:

Vandit1604:
Hi @Cyberguru1, thanks for the review! `watchRemoteConn` is idempotent now, and the test calls `ss.stop()` so the full teardown runs. Could you take another look whenever it suits you?

PR #1374, opened 2026-08-14
multi: don't fail startup when lnd takes a while to open its databases

ellemouton:
### Problem

Users on litd `v0.17.x` with `lnd v0.21.2` are seeing litd abort startup, with `/v1/status` reporting both the `lit` and `lnd` sub-servers as errored:

```
"lit": {
  "running": false,
  "error": "could not start Lit: lnd RPC not ready: lnd's RPC interceptor did not leave WAITING_TO_START within 15s"
},
"lnd": {
  "running": false,
  "error": "lnd RPC not ready: lnd's RPC interceptor did not leave WAITING_TO_START within 15s"
}
```

...on nodes where lnd is perfectly healthy and simply still starting up.

### Root cause

#1353 added a wait for lnd's RPC interceptor to leave `WAITING_TO_START`, bounded by `defaultConnectTimeout` (a fixed 15s). The premise was that once the `readyChan` signal we wait on just before it fires, lnd is nearly ready. It isn't.

`readyChan` is lnd's `lis.Ready`, closed in `startGrpcListen` as soon as the gRPC listener socket is bound. In `lnd.Main` that happens *before*:

1. leader election (if running with etcd clustering),
2. `implCfg.DatabaseBuilder.BuildDatabase` — the bbolt/channeldb open, channeldb migrations, graph cache population, and on 0.21 the kvdb-to-native-SQL migration,
3. `implCfg.BuildWalletConfig`, which is where `SetWalletNotCreated`/`SetWalletLocked` finally moves lnd out of `WAITING_TO_START`.

So the 15s budget has to cover lnd's entire database open — the step lnd itself warns "might take a few minutes" about. On nodes with a large channel and graph state, or on the first startup after a version bump that migrates them, blowing through 15s is the expected outcome, not a symptom of anything being wrong.

`defaultConnectTimeout` is also a gRPC dial timeout, shared with `lnd_connection.go` and `session_rpcserver.go`, so it was never an appropriate budget for a database migration and can't just be raised in place.

Note that this failure is not fatal to lnd — `start()` returning an error leaves the status server up and `lnd.Main` running — but litd abandons the rest of its startup, so the lnd clients, the accounts service, the session server, the firewall middleware and the integrated sub-servers never start.

### Fix

Bound the wait by the existing `--lndreadytimeout` instead, which already covers the other place we wait on lnd's readiness (the kvdb-to-SQL migration) and defaults to a generous 10 minutes. Its description is widened to cover both uses and to name both error strings, so the flag is greppable from either failure.

Two supporting changes so that a long timeout doesn't trade one bad failure mode for another:

- The wait now also selects on `lndQuit` and the error queue, matching the `waitForSignal` closure just below it. A genuinely dead lnd is surfaced immediately rather than after the full timeout.
- Progress is logged every 30s, so a slow-but-healthy startup is distinguishable from a hang in the logs.

### Testing

- Two new cases in `TestWaitForLndRPCReady` covering the early-abort paths, each with a 1 hour timeout so they fail if the abort doesn't work.
- Existing timeout and poll cases unchanged and still passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config.go
+++ b/config.go
@@ -115,16 +115,17 @@ const (
 
 	defaultFirstLNCConnTimeout = 10 * time.Minute
 
-	// defaultLndReadyTimeout is the default maximum time that litd
-	// waits for lnd's RPC server to become ready. It currently applies
-	// to the one-time kvdb-to-SQL data migration, which must call into
-	// lnd's main RPC server (ListMacaroonIDs); that only becomes
-	// available once lnd reaches its "RPC active" state. On nodes with a
-	// large channel/graph state this can take well over a minute after
-	// the wallet is unlocked, and timing out aborts litd startup
-	// entirely (requiring a manual restart), so we default to a
-	// deliberately generous value that comfortably covers even large
-	// nodes.
+	// defaultLndReadyTimeout is the default maximum time that litd waits
+	// for lnd's RPC server to become ready. It applies both to the startup
+	// wait for lnd's RPC interceptor to leave WAITING_TO_START (which only
+	// happens once lnd has opened, and if necessary migrated, all of its
+	// databases) and to the one-time kvdb-to-SQL data migration, which
+	// must call into lnd's main RPC server (ListMacaroonIDs) and so needs
+	// lnd to have reached its "RPC active" state. On nodes with a large
+	// channel/graph state either can take many minutes, and timing out
+	// aborts litd startup entirely (requiring a manual restart), so we
+	// default to a deliberately generous value that comfortably covers
+	// even large nodes.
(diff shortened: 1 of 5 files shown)


gkrizek:
I tested this out on an affected node and does appear to resolve the issue.

ViktorT-11 (approved):
Thanks a lot for the fix @ellemouton :pray:! LGTM

bitromortac (commented):
Thank you for taking care about this :pray:

PR #1380, opened 2026-08-19
terminal: notify systemd when litd is ready

### LINKED ISSUE #585: add systemd `Type=notify` support
litd doesn't do systemd notify like lnd does (https://github.com/lightningnetwork/lnd/blob/86e7b4e1e02e29117dcebba319b0da98e66a3365/contrib/init/lnd.service#L22) so we don't know when litd is ready to use. With `Type=notify`, `systemctl start litd.service` just hangs.

Be able to use `Type=notify` in systemd service files.

While you are adding this, it would be good to add an example systemd service file so that people know the best way to launch litd.

`Type=simple` works, but we don't have consistency with lnd and we don't know when litd is fully up and running for dependent services to use.

**Additional context**
Would like systemd service files for all lightning labs products to be nearly identical.

I think lnd is implementing this here: https://github.com/lightningnetwork/lnd/blob/86e7b4e1e02e29117dcebba319b0da98e66a3365/signal/signal.go#L25 .

See also, https://manpages.debian.org/bookworm/systemd/systemd.service.5.en.html#OPTIONS .

Vandit1604:
Fixes #585.

### Problem

litd does not tell systemd when it has finished starting. So a systemd unit with `Type=notify` never leaves the "activating" state, and `systemctl start litd.service` just hangs. lnd already supports `Type=notify`, so litd is inconsistent with it.

### Fix

Notify systemd once all sub-servers are up:

- `notifySystemdReady()` sends `READY=1` in `Run()`, only after `start()` succeeds. If litd is not run under systemd (`NOTIFY_SOCKET` unset), the call is a no-op. litd is already up at that point, so a notify failure is logged and does not bring it down.
- `notifySystemdStopping()` sends `STOPPING=1` on shutdown.

This uses `github.com/coreos/go-systemd/v22`, which is already an indirect dependency, so only its `// indirect` marker is dropped. No build tags are needed, because `SdNotify` is a no-op when `NOTIFY_SOCKET` is unset, the same as lnd.

The issue also asks for an example service file, so this adds `contrib/init/litd.service` with `Type=notify`, mirroring lnd's `contrib/init/lnd.service`.

### Testing

`systemd_test.go` binds a datagram socket, points `NOTIFY_SOCKET` at it, calls `notifySystemdReady()`, and asserts `READY=1` is received. A second test checks the no-socket case is a clean no-op. The full `terminal` package test suite passes (26 tests).


systemd_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// TestNotifySystemdReady checks that notifySystemdReady sends the READY=1
+// message to the systemd notification socket when NOTIFY_SOCKET is set.
+func TestNotifySystemdReady(t *testing.T) {
+	// Bind a datagram socket and advertise it as the systemd notify socket.
+	dir, err := os.MkdirTemp("", "notify")
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
```
Cyberguru1, line 19: Use `t.TempDir()` instead, as it handles both the creation and cleanup of the temporary directory automatically.

systemd.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"github.com/coreos/go-systemd/v22/daemon"
+)
+
+// notifySystemdReady notifies systemd that litd has finished starting up and is
+// ready to serve requests, so a unit using Type=notify can leave the
+// "activating" state. If litd is not running under systemd (NOTIFY_SOCKET is
+// unset) the call is a no-op. A failure is only logged: litd is already up, so
+// it keeps running regardless.
+func notifySystemdReady() {
```
Cyberguru1, line 12: I think updating both `notifySystemdReady()` and `notifySystemdStopping()` to return `(bool, error)` would be better so `systemd_test.go` can explicitly assert error and notification status rather than relying on implicit non-panicking execution.

systemd.go
```diff
@@ -0,0 +1,37 @@
...
+		log.Infof("Notified systemd that litd is ready")
+	} else {
+		log.Debugf("Not running under a systemd Type=notify unit; " +
+			"skipped the readiness notification")
+	}
+}
+
+// notifySystemdStopping notifies systemd that litd is shutting down. It is
+// best-effort: any error is logged and ignored, since litd is stopping anyway.
+func notifySystemdStopping() {
```
Cyberguru1, line 32: Could we add unit test coverage for `notifySystemdStopping()`?

Cyberguru1 (changes requested):
Great work @Vandit1604 🙌!

Left a few first-pass feedback below.

Vandit1604:
Done: `t.TempDir()`, both functions now return `(bool, error)` with the tests asserting them, and `notifySystemdStopping` has coverage. Rebased on master too.

litbot-9000:

Vandit1604:
Hi @Cyberguru1, thanks for the suggestions! The tests use `t.TempDir()` now, both notify functions return `(bool, error)` with asserts, and `notifySystemdStopping` has its own test. Could you take another look whenever it suits you?

PR #1382, opened 2026-08-24
firewall: map inbound fees for FeeReport endpoint

bitromortac:
This commit passes through inbound fees in the FeeReport privacy mapping in analogy to the outbound fees.


ViktorT-11 (approved):
LGTM :pray:!

PR #1403, opened 2026-09-17
session: clamp implausible timestamps during KV to SQL migration

### LINKED ISSUE #1402: KV to SQL session migration aborts on session with implausible expiry timestamp
## Summary

On upgrading to v0.17.4 (from an older version that still used the bbolt session store), litd aborts on startup during the KV to SQL session store migration (migration version 6) with:

```
error migrating session store to SQL: migration of non-linked session failed:
unable to unmarshal migrated session: unable to get migrated session from sql store:
sql: Scan error on column index 5, name "expiry":
unsupported Scan, storing driver.Value type string into type *time.Time
```

The migration then reruns and fails identically on every subsequent start, since `Mig6ProgrammaticMigration` sets `ResetVersionOnError: true` and the failed transaction rolls back fully. The node can never start with the sqlite backend until the offending bbolt record is removed by hand.

## Root cause

One of the sessions in the legacy bbolt `session.db` holds a garbage `expiry`. In this case a session written by an early litd version (it also has `created_at` at Unix epoch zero, which several of the user's other oldest sessions share) has an expiry whose deserialized `time.Time` lands at year -154755863931.

The migration path is:

1. `getBBoltSessions` deserializes each session, with `DeserializeSession` turning the stored `uint64` unix seconds back into `time.Time` via `time.Unix` ([session/tlv.go](https://github.com/lightninglabs/lightning-terminal/blob/master/session/tlv.go#L241)).
2. `migrateSingleSessionToSQL` inserts the session, writing `session.Expiry.UTC()` into the `expiry TIMESTAMP NOT NULL` column.
3. The validation step reads the row back with `GetSessionByAlias` and scans it into `Session.Expiry time.Time`.

The failure is in step 3, and it's a modernc.org/sqlite parsing limitation: the driver writes a `time.Time` bound parameter as `t.String()`, and on read it only converts TEXT back to `time.Time` when the text parses against one of its fixed layouts, all of which assume a 4-digit year. A time outside years 0-9999 writes fine, but parses back as a plain string, which then fails the scan into `*time.Time`. Repro against modernc.org/sqlite v1.38.2 (the version pinned by v0.17.x):

```go
db.Exec("CREATE TABLE s (id INTEGER PRIMARY KEY, expiry TIMESTAMP NOT NULL)")
db.Exec("INSERT INTO s (expiry) VALUES (?)", time.Unix(1<<62, 0))
var got time.Time
db.QueryRow("SELECT expiry FROM s").Scan(&got)
// sql: Scan error on column index 0, name "expiry":
// unsupported Scan, storing driver.Value type string into type *time.Time
```

Normal times, the zero time, and the `9999-01-01` "never expire" sentinel all round-trip fine, which is why only nodes with one of these bad records hit this.

Note this isn't a driver version regression: modernc.org/sqlite v1.34.5 (pinned by litd up to v0.16.x) and v1.38.2 have identical `formatTime`/`parseTime` handling.

## Workaround

Delete the offending session record from the legacy bbolt `session.db` (or the whole file, if the old LNC sessions aren't needed). With litd stopped, the migration then completes on next start. We walked a user through this on a production node: 54 sessions, one with year -154755863931, removal of that single record unblocked the migration.

## Suggested fixes

First, the migration should sanitize implausible timestamps rather than aborting and rolling back forever. The subtlety is that `migrateSessionsToSQLAndValidate` deep-equals the read-back session against the original KV session, so clamping only at the insert would trade the Scan error for `ErrMigrationMismatch`. Any normalization needs to apply to the KV session before both the insert and the comparison, in the same manner as the existing `overrideSessionTimeZone`. A sanity check at `DeserializeSession` time (reject or clamp unix seconds outside a plausible range) would catch it even earlier and would also cover `session/sql_store.go`'s normal `GetSession` path, which would fail identically if such a row ever landed in SQL.

Second, and cheap: the read-back error path loses the session ID. The insert path wraps with `unable to migrate session(%v), kvSession.ID` but the validation path at [session/sql_migration.go](https://github.com/lightninglabs/lightning-terminal/blob/master/session/sql_migration.go#L199) omits it, so the log gives no way to identify which record is the problem. Adding the ID to those two `fmt.Errorf` wraps would have made this a one-minute diagnosis instead of requiring a bbolt dump.

Roasbeef:
In this PR, we make the KV to SQL session store migration tolerate legacy bbolt sessions that hold implausible timestamps, fixing #1402.

Sessions written by early litd versions can carry an expiry whose deserialized `time.Time` lands far outside the range of years the SQL drivers can parse, for example year -154755863931 in the wild (reported in #1402, on a production node that could no longer start with the sqlite backend). The failure is a modernc.org/sqlite parsing limitation: the driver writes a `time.Time` bound parameter as `t.String()`, but on read it only converts TEXT back to `time.Time` when the text parses against one of its fixed layouts, all of which assume a four digit year. A time outside years 0-9999 therefore writes fine, comes back as a plain string on the migration's validation read-back, and fails the scan into `*time.Time`, which aborts the whole migration (which then reruns and fails identically on every startup, since `ResetVersionOnError` is set).

## The fix

`sanitizeSessionTime` clamps the session's timestamps into the representable range before both the insert and the deep-equal validation, following the existing `overrideSessionTimeZone` pattern (clamping only at the insert would trade the scan error for `ErrMigrationMismatch`):

- Anything before the Unix epoch is clamped to it. Such a session is long expired either way.
- Anything after year 9999 is clamped to the end of it. Such a session never expires either way.
- The zero value stays untouched, as it marks unset timestamps.

A warning naming the session is logged whenever a value is actually clamped, so operators can see it happened.

We also include the session ID in the read-back validation error wrap, which previously omitted it. In #1402 the log gave no way to tell which of the node's 54 sessions was the problem, which is what turned this into a bbolt-dump exercise.

Only the migration path is touched: deserialization stays faithful, so nothing changes for ordinary KV reads, and new sessions created through `NewSession` can never hold such a value. The accounts migration doesn't need the same treatment, as its TLV stores timestamps as nanoseconds, which bounds the decoded year to around +/- 2262.

## Test

The new migration test case writes a session with an expiry of `time.Unix(-1<<40, 0)` into the bbolt store and asserts the migration completes with the clamped value. Against the pre-fix code it fails with exactly the error from #1402:

```
migration of non-linked session failed: unable to unmarshal migrated session([...]):
unable to get migrated session from sql store:
sql: Scan error on column index 5, name "expiry":
unsupported Scan, storing driver.Value type string into type *time.Time
```

Fixes #1402


litbot-9000:
@viktort-11: review reminder

